PR #13, opened 2022-08-23
Feat/libsolv

tdejager:
First initial integration of libsolv, into rattler, so that this can be used for solving dependencies as an alternative to pubgrub.


crates/rattler/src/libsolv/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+mod ffi;
+mod pool;
+mod queue;
+mod repo;
+mod solvable;
+mod solver;
+mod transaction;
+
+/// Convenience method to convert from a string reference to a CString
```
baszalmstra, line 11:
```suggestion
/// Convenience method that converts a string reference to a CString
```

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+    /// Create the whatprovides on the pool which is needed for solving
+    pub fn create_whatprovides(&mut self) {
+        // Safe because pointer must exist
+        unsafe {
+            ffi::pool_createwhatprovides(self.as_ptr().as_mut());
+        }
+    }
+}
+
+/// Intern string like types
```
baszalmstra, line 90:
```suggestion
/// Interns string like types into a `Pool` returning a `StringId`
```

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+        StringId(ffi::pool_strn2id(
+            pool.as_ptr().as_mut(),
+            c_str.as_ptr(),
+            length.try_into().expect("string too large"),
+            1,
+        ))
+    }
+}
+
+/// Finds a previously interned string or returns `None` if it wasnt found.
```
baszalmstra, line 108:
```suggestion
/// Finds a previously interned string or returns `None` if it wasn't found.
```

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+            pool.as_ptr().as_ptr(),
+            c_str.as_ptr(),
+            length.try_into().expect("string too large"),
+            0,
+        );
+        if id == 0 {
+            None
+        } else {
+            Some(StringId(id))
+        }
```
baszalmstra, line 127:
```suggestion
        (id != 0).then(|| StringId(id))
```

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+/// Interns from Target type to Id
```
baszalmstra, line 131:
```suggestion
/// Interns an instance of `Self` into a [`Pool`] returning a handle (or `Id`) to the actual data. 
/// Interning reduces memory usage by pooling data together which is considered to be equal, sharing 
/// the same `Id`. However, a `Pool` also only releases memory when explicitly asked to do so or on
/// destruction.
```

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+pub trait Intern {
+    type Id;
+
+    /// Interns the type in the [`Pool`]
+    fn intern(&self, pool: &mut PoolRef) -> Self::Id;
+}
+
+pub trait FindInterned: Intern {
```
baszalmstra, line 139:
```suggestion
/// Enables retrieving the `Id` of previously interned instances of `Self` through the `Intern`
/// trait.
pub trait FindInterned: Intern {
```

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+    fn find_interned_id(&self, pool: &PoolRef) -> Option<Self::Id>;
+}
+
+/// Wrapper for the StringId of libsolv
+#[derive(Copy, Clone)]
+pub struct StringId(pub(super) ffi::Id);
+
+impl StringId {
+    /// Resolve to the interned type returns a string reference
+    pub fn resolve<'a>(&self, pool: &'a PoolRef) -> &'a str {
```
baszalmstra, line 150: This could fail if you create two different pools and use the Id of Pool a to retrieve a string from Pool b. 

I think the c_str, in that case, will be `nullptr` so I think this method should return an Option.

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+        // Safe because the new-type wraps the ffi::id and cant be created otherwise
+        unsafe {
+            let c_str = ffi::pool_id2str(pool.as_ptr().as_ptr(), self.0);
+            CStr::from_ptr(c_str).to_str().expect("utf-8 parse error")
+        }
+    }
+}
+
+/// Intern implementation for string reference
```
baszalmstra, line 159: I would remove this comment completely, it doesn't add anything that is not immediately clear from the line below. Also doc-comments for traits are not visible to the user.

```suggestion
```

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+    }
+}
+
+impl<'s> FindInterned for &'s str {
+    fn find_interned_id(&self, pool: &PoolRef) -> Option<Self::Id> {
+        find_intern_str(pool, self)
+    }
+}
+
+/// Intern implementation for owned Strings
```
baszalmstra, line 174: Same goes for this.

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,300 @@
...
+impl<'s> Intern for &'s String {
```
baszalmstra, line 175: Why do you need the lifetime here? Can't you just implement this for `String`? Since the `intern` method already takes a reference to `Self` anyway?

.github/workflows/rust-compile.yml
```diff
@@ -43,6 +43,7 @@ jobs:
       - uses: actions/checkout@v3
         with:
           submodules: recursive
+          lfs: true
```
baszalmstra, line 46:
```suggestion
          # The tests require data that is stored in LFS
          lfs: true
```

crates/rattler/src/libsolv/queue.rs
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    /// Returns the ffi::Queue as a const pointer
+    pub fn as_inner_ptr(&self) -> *const ffi::Queue {
+        &self.queue as *const ffi::Queue
+    }
+}
+
+impl<T: Into<ffi::Id>> Queue<T> {
+    /// Push a single id into the queue
```
baszalmstra, line 58:
```suggestion
    /// Pushes a single id to the back of the queue
```

crates/rattler/src/libsolv/queue.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn push_id(&mut self, id: T) {
+        unsafe {
+            ffi::queue_insert(self.as_inner_mut(), self.queue.count, id.into());
+        }
+    }
+
+    /// Push multiple id's into the queue
```
baszalmstra, line 65: You can only pass a single element here, is the comment correct?

crates/rattler/src/libsolv/repo.rs
```diff
@@ -0,0 +1,118 @@
...
+        unsafe { std::mem::transmute(self) }
+    }
+
+    /// Returns the pool that created this instance
+    pub fn pool(&self) -> &PoolRef {
+        // Safe because a `PoolRef` is a wrapper around `ffi::Pool`
+        unsafe { &*(self.as_ref().pool as *const PoolRef) }
+    }
+
+    /// Add conda json to the repo
```
baszalmstra, line 66:
```suggestion
    /// Reads the content of the file pointed to by `json_path` and adds it to the instance.
```

crates/rattler/src/libsolv/repo.rs
```diff
@@ -0,0 +1,118 @@
...
+            }
+
+            // Libsolv needs this function to be called so we can work with the repo later
+            // TODO: maybe wolf knows more about this function
+            ffi::repo_internalize(self.as_ptr().as_mut());
+        }
+        Ok(())
+    }
+
+    /// Add a solvable to the Repo
```
baszalmstra, line 96:
```suggestion
    /// Adds a `Solvable` to this instance
```

crates/rattler/src/libsolv/repo.rs
```diff
@@ -0,0 +1,118 @@
...
+    pub fn add_solvable(&mut self) -> SolvableId {
```
baszalmstra, line 97: What is the purpose of this function? There doesn't seem to be a way to associate any data with the solvable.

crates/rattler/src/libsolv/solvable.rs
```diff
@@ -0,0 +1,113 @@
...
+            let solvables = pool.as_ref().solvables;
+            // Apparently the solvable is offset by the id from the first solvable
+            let solvable = solvables.offset(self.0 as isize);
+            Solvable(NonNull::new(solvable).expect("solvable cannot be null"))
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct SolvableInfo {
```
baszalmstra, line 37: What is the difference between a `SolvableInfo` and `Solvable`? Couldn't you just add accessor methods for all of the fields inside this struct to `Solvable`?

crates/rattler/src/libsolv/solvable.rs
```diff
@@ -0,0 +1,113 @@
...
+impl Solvable {
+    /// Returns a reference to the Repo that created this instance.
+    pub fn repo(&self) -> &RepoRef {
+        RepoRef::from_ptr(
+            NonNull::new(unsafe { self.0.as_ref() }.repo)
+                .expect("the `repo` field of an ffi::Solvable is null"),
+        )
+    }
+
+    /// Lookup a certain string from the solvable.
```
baszalmstra, line 53:
```suggestion
    /// Looks up a string value associated with this instance with the given `key`.
```

crates/rattler/src/libsolv/solvable.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn lookup_str(&self, string_id: StringId) -> Option<&str> {
+        let str = unsafe { ffi::solvable_lookup_str(self.0.as_ptr(), string_id.into()) };
```
baszalmstra, line 55:
```suggestion
    fn lookup_str(&self, key: StringId) -> Option<&str> {
        let str = unsafe { ffi::solvable_lookup_str(self.0.as_ptr(), key.into()) };
```

crates/rattler/src/libsolv/solvable.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    /// Returns a solvable info from a solvable
+    pub fn solvable_info(&self) -> SolvableInfo {
+        let pool = self.repo().pool();
+        let solvable = self.0.as_ptr();
+
+        let id = StringId(unsafe { (*solvable).name });
+        let version = StringId(unsafe { (*solvable).evr });
+
+        let solvable_build_flavor = "solvable:buildflavor"
```
baszalmstra, line 77: We should probably turn these into constants.

crates/rattler/src/libsolv/solver.rs
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    /// Returns the pool that created this instance
+    pub fn pool(&self) -> &PoolRef {
+        // Safe because a `PoolRef` is a wrapper around `ffi::Pool`
+        unsafe { &*(self.as_ref().pool as *const PoolRef) }
+    }
+
+    /// Creates a string of 'problems' that the solver still has which it encountered while solving
+    /// the matchspecs use this function to print the existing problems to string
```
baszalmstra, line 69:
```suggestion
    /// the matchspecs. Use this function to print the existing problems to string.
```

crates/rattler/src/libsolv/solver.rs
```diff
@@ -0,0 +1,119 @@
...
+                CStr::from_ptr(problem)
+                    .to_str()
+                    .expect("could not parse string")
+            };
+            writeln!(&mut output, " - {}", problem).expect("could not write into string");
+        }
+        output
+    }
+
+    /// Run the libsolv solver that solves the problems, which are probably matchspecs in the pool
```
baszalmstra, line 87:
```suggestion
    /// Solves all the problems in the `queue`, or returns an error if problems remain.
```

crates/rattler/src/libsolv/solver.rs
```diff
@@ -0,0 +1,119 @@
...
+            Ok(())
+        } else {
+            Err(anyhow!(
+                "encountered problems while solving:\n {}",
+                self.solver_problems()
+            ))
+        }
+    }
+
+    /// Create a transaction from the solver
```
baszalmstra, line 105:
```suggestion
    /// Creates a transaction from the solutions found by the solver.
```

crates/rattler/src/libsolv/solver.rs
```diff
@@ -0,0 +1,119 @@
...
+    pub fn create_transaction(&mut self) -> Transaction {
```
baszalmstra, line 106: Should this consume the `Solver`?

crates/rattler/src/libsolv/transaction.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+impl DerefMut for Transaction<'_> {
+    fn deref_mut(&mut self) -> &mut Self::Target {
+        unsafe { self.0 .0.cast().as_mut() }
+    }
+}
+
+#[derive(Debug)]
+pub enum InstallOperation {
```
baszalmstra, line 46: Missing doc-comment

crates/rattler/src/libsolv/transaction.rs
```diff
@@ -0,0 +1,133 @@
...
+            | ffi::SOLVER_TRANSACTION_CHANGED => InstallOperation::Install,
+            ffi::SOLVER_TRANSACTION_REINSTALL => InstallOperation::Reinstall,
+            ffi::SOLVER_TRANSACTION_ERASE => InstallOperation::Remove,
+            ffi::SOLVER_TRANSACTION_IGNORE => InstallOperation::Ignore,
+            _ => InstallOperation::Unknown,
+        }
+    }
+}
+
+pub struct OperationOnSolvable {
```
baszalmstra, line 69: Missing doc-comment

crates/rattler/src/libsolv/transaction.rs
```diff
@@ -0,0 +1,133 @@
...
+        unsafe { std::mem::transmute(self) }
+    }
+
+    /// Returns the pool that owns this instance.
+    pub fn pool(&self) -> &PoolRef {
+        // Safe because a `PoolRef` is a wrapper around `ffi::Pool`
+        unsafe { &*(self.as_ref().pool as *const PoolRef) }
+    }
+
+    /// Return the solvable operations
```
baszalmstra, line 93:
```suggestion
    /// Returns the solvable operations
```

baszalmstra (changes requested):
I think in general the API of how we interact with libsolv can be improved. Creating a queue, adding matchspecs, passing that to the solver, and creating a transaction from the solver seems like a cumbersome API. I think it would be good to provide a much higher level API where you go 

```
matchspecs & installed specs & repos & other (input)
     v
  "solve"
     v
"new set of specs"
```

But that's for a second iteration I think.

crates/rattler/src/libsolv/solver.rs
tdejager: Depends if you ever want to create multiple transactions?

crates/rattler/src/libsolv/pool.rs
tdejager: We've discussed this before and it didn't work, but I do not recall why.

crates/rattler/src/libsolv/pool.rs
tdejager: Removed

crates/rattler/src/libsolv/pool.rs
tdejager: Yeah, so it seems to be kind of strange behavior by libsolv. Maybe you can have a look :)

crates/rattler/src/libsolv/queue.rs
tdejager: Clarified doc

crates/rattler/src/libsolv/repo.rs
tdejager: Removed it

crates/rattler/src/libsolv/solver.rs
baszalmstra: Can you call solve twice and get different transactions? Or do they add up or something?

crates/rattler/src/libsolv/pool.rs
baszalmstra: I tried removing the lifetime and that works fine.

I think previously we used a different approach with traits. But with the current setup no lifetime is required because the `intern` method already takes `self` by reference.

crates/rattler/src/libsolv/pool.rs
```diff
@@ -0,0 +1,307 @@
...
+    pub fn resolve<'a>(&self, pool: &'a PoolRef) -> Option<&'a str> {
+        // Safe because the new-type wraps the ffi::id and cant be created otherwise
+        unsafe {
+            let c_str = ffi::pool_id2str(pool.as_ptr().as_ptr(), self.0);
+            let c_str = CStr::from_ptr(c_str).to_str().expect("utf-8 parse error");
+            if c_str == "<NULL>" {
+                None
+            } else {
+                Some(c_str)
+            }
```
baszalmstra, line 160:
```suggestion
            (c_str != "<NULL>").then_some(c_str)
```

baszalmstra: Such strange behavior, what if you would want to store the string `"<NULL>"`?

Should we maybe assert in the `intern` method that `"<NULL>"` is not a valid internable string?

Also should we turn `"<NULL>"` into a constant?

crates/rattler/src/libsolv/queue.rs
```diff
@@ -0,0 +1,87 @@
...
+impl<T: Into<ffi::Id>> Queue<T> {
+    /// Pushes a single id to the back of the queue
+    pub fn push_id(&mut self, id: T) {
+        unsafe {
+            ffi::queue_insert(self.as_inner_mut(), self.queue.count, id.into());
+        }
+    }
+
+    /// Push and id and flag into the queue
+    pub fn push_id_and_flags(&mut self, id: T, flags: i32) {
```
baszalmstra, line 66: Nitpicky but would it make sense to rename this method to `push_id_with_flags`? The `and` suggests that the flags are also stored in the queue.

crates/rattler/src/libsolv/pool.rs
baszalmstra: Fixed

PR #15, opened 2022-09-26
fix: remove dependencies required for the library

baszalmstra:
Splits the crate in two: a binary and a library. This reduces the number of packages required for the library because we can move those to the binary crate.


baszalmstra:
Clippy errors are fixed in #14

PR #36, opened 2023-01-25
feat: use nvml instead of libcuda to detect available cuda version

baszalmstra:
Based on a discussion in the [Conda issue tracker](https://www.github.com/conda/conda/pull/11667) this implements checking the Cuda driver version via `nvml`.

Also adds a bunch of documentation with regard to what this library is and how to use it.


wolfv:
Great!

PR #37, opened 2023-02-03
misc: add licenses

baszalmstra:
Adds BSD 3-Clause license.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,6 +1,6 @@
 BSD 3-Clause License
 
-Copyright (c) 2023, Prefix GmbH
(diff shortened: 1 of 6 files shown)


LICENSE
```diff
@@ -0,0 +1,28 @@
+BSD 3-Clause License
+
+Copyright (c) 2023, Prefix GmbH
```
wolfv, line 3: If we use prefix, then better to use `prefix.dev GmbH` as that's the actual name :)

wolfv:
Looks good except for the two tiny comments :)

wolfv (approved):

PR #38, opened 2023-02-03
feat: extract from path

baszalmstra:
Functions to extract packages from a path.

### DIFF
--- a/crates/rattler_package_streaming/src/fs.rs
+++ b/crates/rattler_package_streaming/src/fs.rs
@@ -2,7 +2,7 @@ use crate::{ArchiveType, ExtractError};
 use std::fs::File;
 use std::path::Path;
 
-/// Extracts the contents a `.tar.bz2` package archive from the specified remote location.
+/// Extracts the contents a `.tar.bz2` package archive at the specified path to a directory.
 ///
 /// ```rust,no_run
 /// # use std::path::Path;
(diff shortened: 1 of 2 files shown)


wolfv (approved):

wolfv:
looks super nice.

crates/rattler_package_streaming/src/fs.rs
```diff
@@ -0,0 +1,51 @@
+use crate::{ArchiveType, ExtractError};
+use std::fs::File;
+use std::path::Path;
+
+/// Extracts the contents a `.tar.bz2` package archive from the specified remote location.
```
wolfv, line 5: remote location? I interpret the function as taking a `fs::path`?

crates/rattler_package_streaming/src/fs.rs
```diff
@@ -0,0 +1,51 @@
...
+///     Path::new("/tmp"))
+///     .unwrap();
+/// ```
+pub fn extract_conda(archive: &Path, destination: &Path) -> Result<(), ExtractError> {
+    let file = File::open(archive)?;
+    crate::read::extract_conda(file, destination)
+}
+
+/// Extracts the contents a package archive from the specified remote location. The type of package
+/// is determined based on the path of the url.
```
wolfv, line 36: based on the file extension of the path?

crates/rattler_package_streaming/src/fs.rs
baszalmstra: Changed

crates/rattler_package_streaming/src/fs.rs
baszalmstra: Changed

PR #39, opened 2023-02-04
feat: validate package content

baszalmstra:
Functionality to validate the contents of a package archive.


crates/rattler/src/validation.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    #[error("sha256 hash mismatch, expected '{0}' but file on disk is '{1}'")]
+    HashMismatch(String, String),
+}
+
+/// Determine whether the files in the specified directory match what is expected according to the
+/// `info/paths.json` file in the same directory.
+pub fn validate_package_files(package_dir: &Path) -> Result<(), PackageValidationError> {
+    // Read the 'paths.json' file which describes all files that should be present
+    let paths = PathsJson::from_path(&package_dir.join("info/paths.json"))
```
wolfv, line 52: Some older packages do not have the `paths.json` entry, and only have a flat file called `files` that contains a list of all paths that are supposedly there. I think for maximum compatibility we should also implement that as a fallback if `paths.json` isn't availbe. The `files` file also doesn't contain any hashes, so any validation beyond the existence check of the file is hard :)

crates/rattler/src/validation.rs
```diff
@@ -0,0 +1,298 @@
...
+    entry: &PathsEntry,
+    metadata: Metadata,
+) -> Result<(), PackageEntryValidationError> {
+    debug_assert!(entry.path_type == PathType::SoftLink);
+
+    if !metadata.is_symlink() {
+        return Err(PackageEntryValidationError::ExpectedSymlink);
+    }
+
+    // TODO: Validate symlink content
```
wolfv, line 142: I think in mamba we have a bit of an optimization to not recompute the SHA256 of a file that we already checked (e.g. if the symlink is going to a file of the same package).

wolfv:
Looks really good! 

For older packages, we might be able to construct a `paths.json` (with some missing data) from the two files: `has_prefix` and `files` where `has_prefix` is the file that contains all files with prefixes.

E.g. for zlib:

has_prefix:

```txt
/Users/runner/miniforge3/conda-bld/zlib-split_1665761013714/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol text lib/pkgconfig/zlib.pc
```

and `files`:

```txt
include/zconf.h
include/zlib.h
lib/libz.a
lib/libz.dylib
lib/pkgconfig/zlib.pc
```

wolfv:
Here is an example of such a package – 6 years old `zlib`: https://anaconda.org/conda-forge/zlib/1.2.8/download/linux-64/zlib-1.2.8-3.tar.bz2

crates/rattler/src/validation.rs
baszalmstra: Yeah, I dont think we need to check the SHA256 hash of the file it points to because it *must* point to another file in the same package (which is then already checked). But we could also check whether the symlink itself is correct. E.g. `../a` points to the same file as `./../a` but the symlinks themselves are different.

I doubt it really matters though because during linking the symlinked files are symlinked in the target folder again anyway. But for completeness, I left the comment in.

crates/rattler/src/validation.rs
baszalmstra: I just pushed code to also check the deprecated `files` format by converting the `files` and friends to a `paths.json` file. I also added a few tests for this.

crates/rattler/src/validation.rs
```diff
@@ -0,0 +1,361 @@
...
+    debug_assert!(entry.path_type == PathType::SoftLink);
+
+    if !metadata.is_symlink() {
+        return Err(PackageEntryValidationError::ExpectedSymlink);
+    }
+
+    // TODO: Validate symlink content. Dont validate the SHA256 hash of the file because since a
+    // symlink will most likely point to another file added as a hardlink by the package this is
+    // double work. Instead check that the symlink is correct e.g. `../a` points to the same file as
+    // `b/../../a` but they are different.
```
wolfv, line 191: just fyi, symlinks can also point to files from other packages.

crates/rattler/src/validation.rs
```diff
@@ -0,0 +1,361 @@
...
+    };
+
+    // Validate all the entries
+    validate_package_directory_from_paths(package_dir, &paths)
+        .map_err(|(path, err)| PackageValidationError::CorruptedEntry(path, err))?;
+
+    Ok(paths)
+}
+
+/// Determine whether the files in the specified directory match wat is expected according to the
```
wolfv, line 96: small typo (wat -> what).

wolfv (approved):

PR #42, opened 2023-02-06
feat: package cache

baszalmstra:
Provides functionality to cache extracted Conda packages in a directory. The cache object holds an in-memory cache of which valid packages exist. It checks the filesystem if an entry doesn't exist in memory yet. Validation (from #39 ) is used to determine whether the cache is corrupt. If the filesystem also doesn't provide the requested package it is fetched into the cache.

The cache does not provide a method to download and extract package archives but instead relies on a `populate` function. The goal is to decouple "caching" and "download/extract". This makes it easier to test, provide different downloading functionalities, queueing, etc.

Builds on #39 which should be merged first!

### DIFF
--- a/crates/rattler/Cargo.toml
+++ b/crates/rattler/Cargo.toml
@@ -27,7 +27,7 @@ nom = "7.1.0"
 once_cell = "1.8.0"
 pin-project-lite = "0.2.9"
 rattler_conda_types = { version = "0.1.0", path = "../rattler_conda_types" }
-rattler_package_streaming = { version = "0.1.0", path = "../rattler_package_streaming" }
+rattler_package_streaming = { version = "0.1.0", path = "../rattler_package_streaming", features = ["reqwest", "tokio"] }
 regex = "1.5.4"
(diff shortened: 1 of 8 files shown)


baszalmstra:
@wolfv Do you have ideas for edge cases or different test strategies?

crates/rattler/src/validation.rs
```diff
@@ -0,0 +1,361 @@
...
+    };
+
+    // Validate all the entries
+    validate_package_directory_from_paths(package_dir, &paths)
+        .map_err(|(path, err)| PackageValidationError::CorruptedEntry(path, err))?;
+
+    Ok(paths)
+}
+
+/// Determine whether the files in the specified directory match wat is expected according to the
```
wolfv, line 96: small typo (wat -> what).

wolfv (approved):

wolfv:
I think the biggest edge cases I have in mind are all around:

- file locking when writing to the cache from multiple processes (especially when network file systems are involved)
- layered caches (e.g. one in `/usr/cache` that is read-only, and one in `~/cache` that is read-write)
- finding a writable cache

wolfv (approved):

PR #48, opened 2023-02-09
feat: install individual packages

baszalmstra:
This is the first implementation for an API to install packages with Rattler. A test is available to check that installing an explicit lock file for python actually allows you to run Python. There are still loads of features missing but the basic functionality works. The most notable missing features are codesigning for macOS, proper handling of clobbered files, and writing into the `conda-meta` folder. Probably more is absent but this seems like a great start.

**Todo**

- [x] Code signing on Apple ARM64
- [ ] ~~Write `conda-meta` file~~ (will do in another PR)
- [ ] ~~Handle clobbering properly~~ (will do in another PR)


crates/rattler/src/package_cache.rs
```diff
@@ -62,13 +65,49 @@ struct Package {
...
+}
+
+impl PackageInfo {
+    /// Try to convert the specified filename into a [`PackageInfo`].
+    pub fn try_from_filename(filename: &str) -> Option<PackageInfo> {
+        // Filenames in the form of: <name>-<version>-<build>(.ext)
+        let (build_string, version, name) = filename.rsplitn(3, '-').next_tuple()?;
+
+        // Remove the file extension from the build string
+        let build_string = build_string
```
wolfv, line 88: packages can have `.` in names. I think we shoudl always match `.conda` or `.tar.bz2` explicitly.

baszalmstra: But can build strings contain strings? At this point in the code the only the build string part is split.

wolfv (approved):

PR #50, opened 2023-02-14
Polish libsolv ffi wrapper

aochagavia:
This commit makes sure we allow calling libsolv functions from shared references, instead of requiring mutability, because otherwise legitimate use cases of the library (like multiple repos for a single pool) are impossible to express.

### DIFF
--- a/crates/rattler/src/solver/libsolv/repo.rs
+++ b/crates/rattler/src/solver/libsolv/repo.rs
@@ -124,14 +124,8 @@ impl RepoRef {
         unsafe { &*(self.as_ref().pool as *const PoolRef) }
     }
 
-    /// Returns the pool that created this instance
-    pub fn pool_mut(&mut self) -> &mut PoolRef {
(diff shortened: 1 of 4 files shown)


crates/rattler/src/solver/libsolv/repo.rs
```diff
@@ -107,7 +107,7 @@ impl RepoRef {
     }
 
     /// Returns a pointer to the wrapped `ffi::Repo`
-    fn as_ptr(&self) -> NonNull<ffi::Repo> {
+    pub fn as_ptr(&self) -> NonNull<ffi::Repo> {
```
baszalmstra, line 110: Does this still need to be `pub`? Id rather not expose the ffi if we don't need to.

baszalmstra (approved):

crates/rattler/src/solver/libsolv/repo.rs
aochagavia: I _think_ I'll need this later on, but I just removed it from the PR because it is unrelated 😉

PR #51, opened 2023-02-15
feat: support installed (virtual) packages in libsolv

aochagavia:
This PR makes it possible to specify which packages are already installed (including virtual packages), so libsolv can take them into account. I also took the liberty of extracting the `rattler_solve` crate, to have a somewhat lighter dependency than `rattler` for use cases when you are not actually planning to install anything (e.g. a server that resolves an environment and returns information about which dependencies should be installed).

I am not totally comfortable about the way we are using unsafe code in the libsolv wrapper (i.e. it might be possible to cause UB within safe Rust, because we are not abstracting the unsafe code away properly). Instead of overhauling it in one go, I hope to iteratively improve it (this time I added a lifetime parameter to `Solvable`, for instance).

Also, there are quite some TODOs within the libsolv wrapper, basically asking what the meaning is of different functions. At some point we will need to go through them before we consider this to be production-ready.


crates/rattler_solve/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+    #[error("error adding repodata: {0}")]
+    ErrorAddingRepodata(#[source] anyhow::Error),
+
+    /// An error occurred when trying to load information about installed packages to the solver
+    #[error("error adding installed packages: {0}")]
+    ErrorAddingInstalledPackages(#[source] anyhow::Error),
+
+    /// The solver backend returned operations that have no known mapping to [`PackageOperationKind`]
+    #[error("unsupported operations")]
+    UnsupportedOperations,
```
baszalmstra, line 30: This error is not very informative to the user. Which operation could not be mapped for instance?

crates/rattler_solve/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+}
+
+/// Represents known information about a single package that is already installed in the Conda
+/// environment
+#[derive(Debug)]
+pub struct InstalledPackage {
```
baszalmstra, line 36: This can be replaced with a `PrefixRecord` from my branch.

crates/rattler_solve/src/lib.rs
```diff
@@ -0,0 +1,85 @@
...
+    /// The package is being updated
+    Update,
+}
+
+/// Represents a dependency resolution problem, to be solved by one of the backends (currently only
+/// libsolv is supported)
+#[derive(Debug)]
+pub struct SolverProblem<'c> {
+    /// All the available channels (and contents) in order of priority
+    pub channels: Vec<(String, &'c RepoData)>,
```
baszalmstra, line 69: Ideally, we'd replace this with a `Vec<RepoDataRecord>`.

crates/rattler_solve/src/package_operation.rs
```diff
@@ -0,0 +1,37 @@
+/// Represents an operation that needs to be applied on a package, in order to bring the Conda
+/// environment to the desired state
+#[derive(Debug, Clone)]
+pub struct PackageOperation {
+    /// The package to be operated on
+    pub package: PackageIdentifier,
```
baszalmstra, line 6: Could we also return a `RepoDataRecord` here? Same as the input?

crates/rattler_solve/src/libsolv/mod.rs
```diff
@@ -0,0 +1,343 @@
...
+    SOLVER_SOLVABLE_PROVIDES, SOLVER_UPDATE,
+};
+use crate::{PackageOperation, RequestedAction, SolveError, SolverProblem};
+
+/// Convenience method that converts a string reference to a CString
+fn c_string<T: AsRef<str>>(str: T) -> CString {
+    CString::new(str.as_ref()).expect("should be convertable from string")
+}
+
+fn request_to_solvable_flags(action: RequestedAction) -> u32 {
```
baszalmstra, line 27: Could also turn this into a transparent type `SolvableFlags(u32)` and create a `From` implementation.

crates/rattler_solve/src/libsolv/solver.rs
```diff
@@ -84,6 +84,11 @@ impl SolverRef {
         output
     }
 
+    /// Sets a solver flag
+    pub fn set_flag(&self, flag: u32, value: bool) {
```
baszalmstra, line 88: Can we type these? :) `SolverFlag`

crates/rattler_solve/src/lib.rs
aochagavia: The idea is to keep the errors backend-agnostic, so basically this error is saying: _your libsolv backend is buggy, because it is returning operations that you don't know how to interpret_. If we want to provide a bit more context, I can think of the following alternatives:

1. Add logging on the libsolv backend side, so the user gets to see something at stderr about unsupported operations.
2. Modify the `UnsupportedOperations` variant to include a `Vec<String>` which can be populated by the solver backend to provide a user-friendly representation of the operations that could not be translated).

I think I'd prefer 2. Do you agree? Or would you suggest a third approach?

baszalmstra: I agree completely :)

crates/rattler_solve/src/lib.rs
aochagavia: I have been tinkering with this, but am not totally convinced it is the right way to go. As far as I've understood, we need to create one repo for each `repodata.json`. Putting all packages together in a single `Vec<RepoDataRecord>` will require additional bookkeeping downstream (e.g. for each package, check if we have already seen its corresponding repo, and create it if it if necessary). It seems more natural to me to keep them separate in the input. Or should we group all available packages in a single repo?

aochagavia: Note to future readers: we discussed this offline and concluded to use `Vec<Vec<RepoDataRecord>>` (and will probably enhance it in the future to include a hash of the repo data for caching purposes)

crates/rattler_solve/src/libsolv/transaction.rs
```diff
@@ -0,0 +1,202 @@
...
+    let channel = channel_mapping
+        .get(&solvable.repo().id())
+        .map(|c| c.to_owned())
+        .unwrap_or_else(|| "unknown".to_owned());
+
+    RepoDataRecord {
+        // TODO: if we want this we will probably need to add it separately to the solvable
+        file_name: "unknown".to_string(),
+        // TODO: if we want this we will probably need to add it separately to the solvable
+        url: Url::from_str("https://example.com").unwrap(),
```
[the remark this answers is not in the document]
aochagavia: Yeah, I already found it and just added a commit for it

crates/rattler_solve/src/libsolv/transaction.rs
```diff
@@ -0,0 +1,202 @@
...
+fn repodata_record_from_solvable(
+    solvable: Solvable,
+    channel_mapping: &HashMap<RepoId, String>,
+) -> RepoDataRecord {
+    let channel = channel_mapping
+        .get(&solvable.repo().id())
+        .map(|c| c.to_owned())
+        .unwrap_or_else(|| "unknown".to_owned());
+
+    RepoDataRecord {
```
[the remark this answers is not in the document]
aochagavia: I will try this in an upcoming PR

crates/rattler_conda_types/src/lib.rs
```diff
@@ -16,6 +16,7 @@ mod version_spec;
 
 pub mod package;
 pub mod prefix_record;
+pub mod virtual_package;
```
baszalmstra, line 19: I would just name this `generic_virtual_package` make it private and add `pub use generic_virtual_package::GenericVirtualPackage` in this file.

crates/rattler_solve/src/libsolv/transaction.rs
```diff
@@ -0,0 +1,206 @@
...
+        channel,
+        package_record: PackageRecord {
+            name: solvable.name(),
+            version: Version::from_str(&solvable.version()).unwrap(),
+            build_number: solvable.build_number().unwrap(),
+            build: solvable.build_string().unwrap(),
+            subdir: solvable.lookup_str(solvable_mediadir).unwrap().to_string(),
+            md5: solvable.md5(),
+            sha256: solvable.sha256(),
+            // We have the data (see [`Repo::add_repodata_records`]), but I'm not sure if it is necessary to retrieve it
```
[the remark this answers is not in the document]
aochagavia: Thanks for the clarification, this is something I'm working on for an upcoming PR, which will retrieve the original `RepoDataRecord` for each solvable

crates/rattler_conda_types/src/lib.rs
aochagavia: Right, will do

baszalmstra (approved):

baszalmstra:
Lets merge this if CI turns green!

PR #52, opened 2023-02-16
move index to its own file, add helper functions and simple test

wolfv:
Two things I found while writing / adjusting this code:

- to my knowledge, run_exports are defined in a seperate file (`info/run_exports.json`) so I was surprised to see this in the index.json definition. Did you encounter a package where run_exports were added to the index? Otherwise I would suggest we remove that for now?
- The `paths.json` object is called `PathsJson` while the `index.json` object is called `Index`. Should we make them both the same?

Also, I've used `serde_json::from_reader(...)` instead of going via reading a string and reusing the FromStr functions. That means we have two `.map_err(Into::into)`. Let me know if you prefer the `FromStr` way.

### DIFF
--- a/crates/rattler_conda_types/src/package/index.rs
+++ b/crates/rattler_conda_types/src/package/index.rs
@@ -1,6 +1,6 @@
-use std::{collections::HashMap, fs::File, path::Path, str::FromStr};
+use std::{fs::File, io::Read, path::Path, str::FromStr};
 
-use crate::{RunExports, Version};
+use crate::Version;
 use serde::{Deserialize, Serialize};
 use serde_with::{serde_as, skip_serializing_none, DisplayFromStr};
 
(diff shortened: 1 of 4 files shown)


baszalmstra:
> to my knowledge, run_exports are defined in a separate file (info/run_exports.json) so I was surprised to see this in the index.json definition. Did you encounter a package where run_exports were added to the index? Otherwise, I would suggest we remove that for now?

Did not know that! Let's remove them in that case. I think you already did that in another PR.

> The paths.json object is called PathsJson while the index.json object is called Index. Should we make them both the same?

Yeah, I switched that at some point but didn't bother to change the rest of the structs. My reasoning for calling them `..Json` is that they really refer to specific JSON files. That's also why I added the helper functions because it's a bit of a leaky abstraction to have to load them through `serde_json` manually. 

> Also, I've used serde_json::from_reader(...) instead of going via reading a string and reusing the FromStr functions. That means we have two .map_err(Into::into). Let me know if you prefer the FromStr way.

The reason I read to string first is twofold:

* Deserializing from string is orders of magnitude faster than from a reader. [[source]](https://github.com/serde-rs/json/issues/160#issuecomment-253446892). For small files like these, I don't think the extra memory usage of reading it to string directly outways the performance penalty.
* The `FromStr` implementation makes it very easy to test. In your tests currently, you have to unpack a certain package archive to be able to test that it functions correctly. With the `FromStr` (but you can also use the `from_reader` function) you can just paste the original JSON in the test. (Although you might also want to check the integration).

wolfv:
OK, made the changes.

PR #55, opened 2023-02-16
feat: download and cache repodata.json

### LINKED ISSUE #47: Remove dependency on `extendhash`
Some parts of `rattler` use the [extendhash](https://github.com/printfn/extendhash) crate. However, the same functionality is also provided by the `md-5` and `sha2` crates which are also used. Since `md-5` and `sha2` seem more complete we should get rid of `extendhash`.

baszalmstra:
This PR adds the ability to download and cache repodata similar to how Mamba and Conda work. This PR implements the proposal from https://github.com/conda-incubator/ceps/pull/46, and adds support for `zst` and `bz2` compressed files. Using file locks it also supports multiple processes to use the same on-disk cache. JLAP is not supported.

**What I still want to do:**

- [x] I need to add some `spawn_blocking` here and there.
- [x] Ability to track progress (for CLI usage).
- [x] Integrate this function into the already existing `RequestRepoDataBuilder` or just remove that.

Nonetheless, I could already use some feedback! 

Also resolves #47


crates/rattler/src/repo_data/cache/mod.rs
```diff
@@ -0,0 +1,170 @@
...
+            r#"{
+            "cache_control": "public, max-age=1200",
+            "etag": "\"bec332621e00fc4ad87ba185171bcf46\"",
+            "has_zst": {
+                "last_checked": "2023-02-13T14:08:50Z",
+                "value": true
+            },
+            "mod": "Mon, 13 Feb 2023 13:49:56 GMT",
+            "mtime_ns": 1676297333020928000,
+            "size": 156627374,
```
wolfv, line 164: this doesn't test the blake2 hash yet, right?

baszalmstra: No this file comes directly from something installed with Mamba. I wanted to ensure that I can load that cache :)

wolfv: OK, got it!

crates/rattler-bin/src/commands/create.rs
```diff
@@ -31,38 +27,38 @@ pub async fn create(opt: Opt) -> anyhow::Result<()> {
...
+    // let solver_problem = SolverProblem {
+    //     channels: repo_data
+    //         .iter()
+    //         .map(|(channel, repodata)| (channel.base_url().to_string(), repodata))
+    //         .collect(),
+    //     specs,
+    // };
+    //
+    // let result = solver_problem.solve()?;
+    // println!("{:#?}", result);
```
aochagavia, line 61: Don't forget to fix the create command to use the new feature

aochagavia (commented):
Looks good!

crates/rattler_repodata_gateway/src/lib.rs
```diff
@@ -0,0 +1,11 @@
+#![deny(missing_docs)]
+
+//! `rattler_repodata_gateway` is a create that provides functionality to interact with Conda
```
aochagavia, line 3: typo: create -> crate

crates/rattler-bin/src/commands/create.rs
baszalmstra: Im planning on refactoring that completely once we have a more stable solver interface. I didn't think this was used (since it doesn't actually do much), so I simply broke it. 

Are you use this?

aochagavia: I am not using it, but I tried to ensure it keeps working after my solver interface changes. I was thinking if you keep it working here it will be easier to resolve the merge conflicts later in a way that makes sense

aochagavia: (if you think it is easier to just ignore the `create` command for the time being, we can leave it broken and fix it later as well)

wolfv (approved):

PR #56, opened 2023-02-17
feat: add PrefixRecord and RepoDataRecord

baszalmstra:
Adds two additional types that represent a single package.

* `PackageRecord` - Already existed but has been tuned to contain only the data that can be found in a `repodata.json`. I removed the `filename` field because its not actually contained within the object found in the repodata. So it was actually always empty.
* `RepoDataRecord` - A standalone object that contains a `PackageRecord`, the `filename`, the`url`, and `channel`. This can be trivially constructed from a `RepoData` object.
* `PrefixRecord` - Contains a `RepoDataRecord` but also additional fields that are filled in when installing a package. These entries can be found in the `conda-meta` directory of an environment.

This diagram shows the relation of the different records. Each type adds additional information.
```
   ┌───────────────┐       ┌────────────────┐     ┌───────────────┐
   │ PackageRecord │◄──────┤ RepoDataRecord │◄────┤ PrefixRecord  │
   └───────────────┘       └────────────────┘     └───────────────┘
```

Todo:

- [x] Write additional module documentation
- [x] Possibly figure out what some of the fields mean that I don't understand yet.


crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -33,33 +33,49 @@ pub struct ChannelInfo {
...
     pub subdir: String,
-    #[serde(default, rename = "fn")]
-    pub filename: Option<String>,
 
+    /// Optionally a MD5 hash of the package archive
     pub md5: Option<String>,
-    //pub legacy_bz2_md5: Option<String>,
-    //pub legacy_bz2_size: Option<usize>,
+
+    /// Optionally a MD5 hash of the package archive
```
wolfv, line 59: this should say SHA256 in the doc, right?

crates/rattler/src/solver/libsolv/repo.rs
```diff
@@ -173,21 +173,22 @@ impl RepoRef {
...
-            unsafe {
-                ffi::repodata_set_location(
-                    data,
-                    solvable_id.into(),
-                    0,
-                    CString::new(subdir)?.as_ptr(),
-                    CString::new(filename)?.as_ptr(),
-                );
-            }
+            // TODO: The filename is not present in a PackageRecord
```
wolfv, line 176: Although the filename is the key to the record in the `repodata.json`, right? Maybe we could add that in? On the other hand, that information will be in the repodata record, so that is fine. I just wonder if repodata record is more expensive to construct

baszalmstra: Yeah, I wanted to keep the `PackageRecord` "pure" in that it represents exactly what is stored in the `repodata.json`. You can quickly convert the `RepoDataRecord` from this and its original channel because you can move the `PackageRecord` directly in it. However, storing all the URLs and channels etc does have some memory overhead. Maybe we can add some form of interning in the future?

baszalmstra: Also for the solver, it now ingests both the `RepoData` with the `PackageRecords` and `Channel` where it comes from. I think we can replace that with simply passing in `RepoDataRecord` because that then contains all the information.

baszalmstra:
@wolfv Do you know what the `preferred_env` means in a `PackageRecord`? Browsing the conda source code also didn't really help me.

baszalmstra:
I added a function called `RepoData::into_repo_data_records` that converts `RepoData` and a `Channel` to a `Vec<RepoDataRecords>`.

wolfv:
I don't know what the `preferred_env` means and can't remember ever having seen it. 

I could only imagine that this is added e.g. to `conda` to indicate that `conda` should be installed in the `base` environment. But that's only a guess...

crates/rattler_conda_types/src/repo_data_record.rs
```diff
@@ -0,0 +1,26 @@
...
+/// the source of the data (like the url and the channel).
+#[derive(Debug, Deserialize, Serialize, Eq, PartialEq, Ord, PartialOrd, Clone)]
+pub struct RepoDataRecord {
+    /// The data stored in the repodata.json.
+    #[serde(flatten)]
+    pub package_record: PackageRecord,
+
+    /// The filename of the package
+    #[serde(rename = "fn")]
+    pub filename: String,
```
wolfv, line 17: Rust uses `Path::file_name()` - I wonder if we should also use a `_` between `file` and `name`? (same for `file_size`, etc.

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -66,23 +84,50 @@ pub struct PackageRecord {
...
 
-    // pub package_type: ?
+    /// The UNIX Epoch timestamp when this package was created. Note that sometimes this is specified in
+    /// seconds and sometimes in milliseconds.
     pub timestamp: Option<usize>,
-
-    pub date: Option<String>,
-
-    pub size: Option<usize>,
+    //pub preferred_env: Option<String>,
```
wolfv, line 101: maybe add a small comment about where you found those?

wolfv:
Just one small comment, looks awesome 🚀

wolfv (approved):

crates/rattler/src/solver/libsolv/repo.rs
```diff
@@ -173,21 +173,22 @@ impl RepoRef {
...
+            // let subdir = record.subdir.as_str();
+            // unsafe {
+            //     ffi::repodata_set_location(
+            //         data,
+            //         solvable_id.into(),
+            //         0,
+            //         CString::new(subdir)?.as_ptr(),
+            //         CString::new(filename)?.as_ptr(),
+            //     );
+            // }
```
aochagavia, line 191: We should probably get rid of this commented-out code

baszalmstra: I can just remove it. I assumed that in your branch this will work slightly different anyway?

crates/rattler_conda_types/src/repo_data_record.rs
baszalmstra: Good one! Will change!

crates/rattler/src/solver/libsolv/repo.rs
baszalmstra: Or actually. I think the `add_repodata` function should take `RepoDataRecord` as its argument instead of the RepoData directly.

aochagavia: Hmm I don't understand this code enough yet to know what we need here, I just thought you had forgotten to remove the comments 😅

crates/rattler_conda_types/src/repo_data/mod.rs
baszalmstra: I added this comment:

```rust
// Looking at the `PackageRecord` class in the Conda source code a record can also include all
// these fields. However, I have no idea if or how they are used so I left them out.
```

crates/rattler/src/solver/libsolv/repo.rs
baszalmstra: No I think we do need this, (not entirely sure), so I left it for future reference.

PR #57, opened 2023-02-18
add about.json parsing like others

wolfv:
This adds the same kind of parsing that we have for `index.json` and other structures.

Also I am wondering if we should implement a `from_archive(Archive: Read)` function that would work the same as `from_package_directory`, but on an unextracted archive?


baszalmstra:
Nice! Implementing the `archive: Read` function also consumes the entire `Read` object which means you cant read from it again. That would only be useful if you want to only read the about.json from the file or you can recreate the `Read` object (which is the case if you have `&[u8]`). 

On another note you also got me thinking about a trait for these files, its a good idea! Maybe you could try implementing that (in another PR)? Otherwise Ill give it a go this week.

wolfv:
sounds good! I won't have time today, maybe tomorrow. If you want to do it, feel free :) I'll have to read a bit more about traits to be able to implement it. We could do that & rename them all to `XYZJson` in one go.

PR #59, opened 2023-02-19
feat: rustdoc

baszalmstra:
This PR fixes some issue with the rustdocs and adds a CI step to verify that in the future this doesn't happen again.

It also adds another CI stage that generates the rustdocs and pushes it to github pages.


wolfv (approved):
Lgtm

PR #60, opened 2023-02-20
fix: windows build issues

baszalmstra:
I had some issues on windows with symlinks and a stupid LFS thing. This PR fixes both of those.


wolfv (approved):
Are we not testing on Windows in the CI?

baszalmstra:
We are, but in Github CI symlinks on windows are enabled. On normal Windows PCs they arent.

PR #67, opened 2023-02-22
fix: improve code structure, documentation and safety for libsolv wrapper

aochagavia:
This is a follow-up for #51 , cleaning up our libsolv wrapper, properly documenting it and ensuring it is (mostly) safe.

There is still an open issue related to strings. When passing C strings to the FFI, in many places our code will panic if a NUL byte is found in the original Rust string (only in a few cases it will propagate an error). Maybe we should check at the beginning that the repodata is NUL-free, so we can stop worrying about this. Any thoughts?


crates/rattler_solve/src/lib.rs
```diff
@@ -227,16 +227,16 @@ mod test_libsolv {
         assert!(matches!(operations[0].kind, PackageOperationKind::Install));
         let info = &operations[0].package;
 
-        assert_eq!("foo-3.0.2.tar.bz2", info.file_name);
+        assert_eq!("foo-3.0.2.conda", info.file_name);
         assert_eq!(
-            "https://conda.anaconda.org/conda-forge/linux-64/foo-3.0.2.tar.bz2",
+            "https://conda.anaconda.org/conda-forge/linux-64/foo-3.0.2.conda",
```
wolfv, line 232: I didn't pay attention to this before, but this is actually an "invalid" package name. It should be `<NAME>-<VERSION>-<BUILDSTRING>.<EXTENSION>`. The `package_record.build` also looks weierd (with the `_conda`).

crates/rattler_solve/src/libsolv/util.rs
```diff
@@ -0,0 +1,17 @@
+#[derive(Copy, Clone, Ord, PartialEq, PartialOrd, Eq)]
+pub enum PackageExtension {
+    TarBz2,
+    Conda,
+}
+
+/// Given a package filename, extracts the filename and the extension if the extension is a known
+/// package extension.
+pub fn extract_known_filename_extension(filename: &str) -> Option<(&str, PackageExtension)> {
```
wolfv, line 9: We have the `ArchiveType` thingy in `rattler_conda_types` - could we use it here?

wolfv: Either `ArchiveIdentifier`: https://github.com/mamba-org/rattler/blob/main/crates/rattler_conda_types/src/package/archive_identifier.rs or https://github.com/mamba-org/rattler/blob/main/crates/rattler_conda_types/src/package/archive_type.rs

crates/rattler_solve/src/libsolv/wrapper/solvable.rs
```diff
@@ -21,60 +13,27 @@ impl From<SolvableId> for ffi::Id {
...
-        let value = unsafe { ffi::solvable_lookup_num(self.as_ptr().as_ptr(), key.0, u64::MAX) };
-        if value == u64::MAX {
-            None
-        } else {
-            Some(value as usize)
-        }
+/// Gets a number associated to this solvable
+pub fn lookup_num(solvable: *mut ffi::Solvable, key: StringId) -> Option<u64> {
+    let value = unsafe { ffi::solvable_lookup_num(solvable as *mut _, key.0, u64::MAX) };
+    if value == u64::MAX {
```
wolfv, line 34: just FYI i think the `ffi::solvable_lookup_num` returns a `unsigned int` (which should be a 32bit integer). I guess using a 64 bit integer is safer anyways, but wanted to note it :)

wolfv: Sorry I am wrong about this! :) It's a `unsigned long long`, so this is correct.

crates/rattler_solve/src/libsolv/input.rs
```diff
@@ -0,0 +1,258 @@
...
+use crate::libsolv::wrapper::solvable::SolvableId;
+use rattler_conda_types::{GenericVirtualPackage, RepoDataRecord};
+use std::cmp::Ordering;
+use std::collections::HashMap;
+use std::ffi::{CString, NulError};
+
+/// Adds [`RepoDataRecord`] to this instance
+pub fn add_repodata_records(
+    pool: &Pool,
+    repo: &Repo,
```
baszalmstra, line 19: I think its important that the passed in `Repo` was created from the `Pool` right? Not doing so might result in undefined behavior. Would it be possible to get a reference to pool from the repo? Or can you add an assertion to ensure this is the case?

wolfv:
Looks really good!

crates/rattler_solve/src/libsolv/input.rs
```diff
@@ -0,0 +1,258 @@
...
+    repo_datas: &[RepoDataRecord],
+) -> Result<(), NulError> {
+    let data = repo.add_repodata();
```
baszalmstra, line 22: Maybe its more clear if you move the declaration of this variable down closer to where it is used. The function is not very descriptive if you don't know libsolv. A comment would also help.

crates/rattler_solve/src/libsolv/wrapper/pool.rs
```diff
@@ -108,164 +129,96 @@ impl Pool {
             Verbosity::Extreme => 3,
         };
         unsafe {
-            ffi::pool_setdebuglevel(self.0.as_ptr(), verbosity);
+            ffi::pool_setdebuglevel(self.raw_ptr(), verbosity);
         }
     }
 
     /// Set the provided repo to be considered as a source of installed packages
     pub fn set_installed(&self, repo: &Repo) {
```
baszalmstra, line 137: This function can result in undefined behavior if the repo has not been created by the same `Pool`. Maybe you can add an `assert` in here to ensure that Repo belongs to the current Pool?

crates/rattler_solve/src/libsolv/wrapper/pool.rs
```diff
@@ -108,164 +129,96 @@ impl Pool {
...
-
-    // Safe because the function accepts any string
-    unsafe {
-        StringId(ffi::pool_strn2id(
-            pool.0.as_ptr(),
-            c_str.as_ptr(),
-            length.try_into().expect("string too large"),
-            1,
-        ))
+    pub fn intern_matchspec(&self, match_spec: &MatchSpec) -> MatchSpecId {
```
baszalmstra, line 157: Maybe add a comment to this public function?

crates/rattler_solve/src/libsolv/wrapper/pool.rs
```diff
@@ -316,28 +236,49 @@ impl From<MatchSpecId> for Id {
     }
 }
 
+fn match_spec_to_c_string(match_spec: &MatchSpec) -> CString {
```
baszalmstra, line 239: Maybe we can move the conversion from MatchSpec to string into `Display` of `MatchSpec`. Im sure it will be used more.

crates/rattler_solve/src/libsolv/wrapper/repo.rs
```diff
@@ -1,422 +1,103 @@
...
-                            unimplemented!("found a duplicate package")
-                        }
-                    }
-                } else {
-                    package_to_type.insert(filename, (package_type, solvable_id));
-                };
-            } else {
-                tracing::warn!("unknown package extension: {}", &repo_data.file_name);
-            }
+    /// Adds a new repodata to this repo
```
baszalmstra, line 55: Maybe explain a little bit here what repodata is, or at least that its a libsolv datastructure.

crates/rattler_solve/src/libsolv/wrapper/solvable.rs
```diff
@@ -21,60 +13,27 @@ impl From<SolvableId> for ffi::Id {
...
-        self.0
-    }
-
-    /// Looks up a string value associated with this instance with the given `key`.
-    pub fn lookup_str(&self, key: StringId) -> Option<&str> {
-        let str = unsafe { ffi::solvable_lookup_str(self.0.as_ptr(), key.into()) };
-        if str.is_null() {
-            None
+    /// Resolves the id to a pointer to the solvable
+    pub fn resolve_raw(self, pool: &Pool) -> NonNull<ffi::Solvable> {
```
baszalmstra, line 17: Should this function remaning local to the crate because it returns an ffi object? (`pub(super)`)

baszalmstra: Also, explain that this function might panic. Similar to resolving of strings.

crates/rattler_solve/src/libsolv/wrapper/solvable.rs
```diff
@@ -21,60 +13,27 @@ impl From<SolvableId> for ffi::Id {
...
-
-    pub fn get_usize(&self, key: StringId) -> Option<usize> {
-        let value = unsafe { ffi::solvable_lookup_num(self.as_ptr().as_ptr(), key.0, u64::MAX) };
-        if value == u64::MAX {
-            None
-        } else {
-            Some(value as usize)
-        }
+/// Gets a number associated to this solvable
+pub fn lookup_num(solvable: *mut ffi::Solvable, key: StringId) -> Option<u64> {
```
baszalmstra, line 32: Same, should this also be `pub(super)`?

crates/rattler_solve/src/libsolv/wrapper/solver.rs
```diff
@@ -96,17 +89,11 @@ impl SolverRef {
...
     /// Creates a transaction from the solutions found by the solver.
-    pub fn create_transaction(&mut self) -> Transaction {
-        let transaction =
-            NonNull::new(unsafe { ffi::solver_create_transaction(self.as_ptr().as_mut()) })
-                .expect("solver_create_transaction returned a nullptr");
-        Transaction::new(transaction)
-    }
-}
+    pub fn create_transaction(&self) -> Transaction {
+        let transaction = NonNull::new(unsafe { ffi::solver_create_transaction(self.raw_ptr()) })
```
baszalmstra, line 93: I think that this might be a legitimate error. This could happen if `solve` hasn't been called yet. Maybe turn this into a Result?

crates/rattler_solve/src/libsolv/wrapper/transaction.rs
```diff
@@ -1,167 +1,70 @@
...
+                solvable_id.into(),
+                ffi::SOLVER_TRANSACTION_SHOW_ALL as std::os::raw::c_int,
+            )
         }
+    }
 
-        if !unsupported_operations.is_empty() {
-            return Err(unsupported_operations);
-        }
+    /// Returns the second solvable associated to the transaction
```
baszalmstra, line 57: What does this mean? Can you clarify what `obs` means?

baszalmstra (commented):
This looks really good @aochagavia !

baszalmstra:
> There is still an open issue related to strings. When passing C strings to the FFI, in many places our code will panic if a NUL byte is found in the original Rust string (only in a few cases it will propagate an error). Maybe we should check at the beginning that the repodata is NUL-free, so we can stop worrying about this. Any thoughts?

We add all the data ourselves into libsolv right? Could we check when we add it to libsolv if strings contain `Nul`?

wolfv:
A string with `NUL` should be super unlikely, right? Because any C style parser would totally trip over it.

tdejager (changes requested):

crates/rattler_solve/src/libsolv/wrapper/solver.rs
```diff
@@ -1,89 +1,82 @@
...
 
-impl SolverRef {
-    /// Returns a pointer to the wrapped `ffi::Solver`
-    fn as_ptr(&self) -> NonNull<ffi::Solver> {
-        // Safe because a `SolverRef` is a transparent wrapper around `ffi::Solver`
-        unsafe { NonNull::new_unchecked(self as *const Self as *mut Self).cast() }
+    /// Returns a user-friendly representation of a problem
+    ///
+    /// Safety: the caller must ensure the id is valid
+    unsafe fn problem2str(&self, id: ffi::Id) -> &CStr {
```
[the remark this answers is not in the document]
aochagavia: I intentionally tried to use the same names as libsolv when it comes to functions that wrap the ffi, and to use better names when building our own logic on top of them. IMO this makes the most sense, but if there is consensus that we should do things as you say I'll gladly change it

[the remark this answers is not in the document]
aochagavia: @baszalmstra @wolfv, do you have a horse in this race?

crates/rattler_solve/src/libsolv/input.rs
```diff
@@ -0,0 +1,258 @@
...
+
+    // Get all the IDs
+    let solvable_buildflavor_id = pool.find_interned_str(SOLVABLE_BUILDFLAVOR).unwrap();
```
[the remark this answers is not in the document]
aochagavia: Technically, this can never fail, because these strings are internal to libsolv and always present. If there is something wrong there, a panic seems like the only sensible option.

crates/rattler_solve/src/libsolv/util.rs
aochagavia: Good catch! `ArchiveType` is perfect here

crates/rattler_solve/src/lib.rs
aochagavia: Yeah, that is a consequence of my unfamiliarity with the conda ecosystem. I'll update it to use more realistic values

crates/rattler_solve/src/libsolv/input.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    repo.internalize();
+
+    Ok(())
+}
+
+/// When adding packages, we want to make sure that `.conda` packages have preference over `.tar.bz`
+/// packages. For that reason, when adding a solvable we check first if a `.conda` version of the
+/// package has already been added, in which case
```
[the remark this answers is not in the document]
aochagavia: Oops, looks like the rest of the comment got lost there

crates/rattler_solve/src/libsolv/input.rs
```diff
@@ -0,0 +1,258 @@
...
+use crate::libsolv::wrapper::pool::Pool;
+use crate::libsolv::wrapper::repo::Repo;
+use crate::libsolv::wrapper::repodata::Repodata;
+use crate::libsolv::wrapper::solvable::SolvableId;
+use rattler_conda_types::{GenericVirtualPackage, RepoDataRecord};
+use std::cmp::Ordering;
+use std::collections::HashMap;
+use std::ffi::{CString, NulError};
+
+/// Adds [`RepoDataRecord`] to this instance
```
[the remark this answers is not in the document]
aochagavia: Hmm... This seems to be a leftover from the time when `add_repodata_records` was a method on `Repo`. I'll update it.

crates/rattler_solve/src/libsolv/input.rs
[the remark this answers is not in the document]
baszalmstra: Could you add that in a comment or in an `expect`?

crates/rattler_solve/src/libsolv/wrapper/solver.rs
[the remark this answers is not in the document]
baszalmstra: I get both arguments. I have no strong opinion.🤷

crates/rattler_solve/src/libsolv/input.rs
aochagavia: Why do you think that passing a `Repo` that does not belong to the `Pool` will result in UB? At least on our side, when we resolve `SolvableId`s we make sure to avoid UB. Are there some assumptions in the libsolv side that we should take into account too? I haven't checked libsolv's code, and maybe it pays off to be extra careful here anyway, so I'll add the check just in case (even without UB it will result in more predictable errors if anyone ever misuses the function)

crates/rattler_solve/src/libsolv/input.rs
aochagavia: Sure!

crates/rattler_solve/src/libsolv/wrapper/pool.rs
aochagavia: Will do

crates/rattler_solve/src/libsolv/wrapper/repo.rs
aochagavia: Will do

crates/rattler_solve/src/libsolv/wrapper/transaction.rs
aochagavia: I am not entirely sure what obs means here (I took it from mamba). Since this code is being deleted in #75, I suggest we leave it unchanged

crates/rattler_solve/src/libsolv/wrapper/solver.rs
aochagavia: In that case I think we should automatically create the transaction after solving, to remove altogether the possibility of misusing the API. I'll merge this into `Solver::solve`

crates/rattler_solve/src/libsolv/input.rs
baszalmstra: Because a `Repo` references data from the `Pool` from which it was created. The current API allows this:

```rust
let pool_a = Pool::new();
let pool_b = Pool::new();
let repo = pool_a.create_repo("cool");
add_repodata_records(pool_b, repo); // <-- could result in undefined behavior.
```

The repo also contains a pointer to the pool. If we just add an assert that that pointer matches the pass in pool we can be sure that its correct.

crates/rattler_solve/src/libsolv/wrapper/solvable.rs
aochagavia: I intentionally exposed a pointer to `ffi::Solvable` because we need mutable access to it in `input.rs`. The ffi object is difficult to wrap in a safe interface, because it is mostly useful when having a mutable reference to it. By exposing it as a raw pointer, the user takes on the responsibility of ensuring there is at most one reference to `ffi::Solvable` at a given time.

I have spent about 30 minutes trying to create a `Solvable` wrapper that is safer, but am not satisfied with the result. If you think it is truly worth it, I can give it a second try. Otherwise I'd suggest we leave it as is

aochagavia:
Just finished addressing all the comments!

Regarding the NUL issue, I suggest we tackle it in an upcoming PR to keep complexity at bay in this one

baszalmstra (approved):

crates/rattler_solve/src/libsolv/wrapper/solver.rs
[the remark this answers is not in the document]
tdejager: All good!

tdejager:
@aochagavia could you just add the `expect` to the last comment?

aochagavia:
> @aochagavia could you just add the `expect` to the last comment?

You mean replacing the calls to `unwrap` by `expect` in `input::add_repodata_records`? I added a comment at the beginning of them, like Bas suggested, which seems enough to me, but I can switch to `expect` too if you feel that's necessary

PR #68, opened 2023-02-23
Shell activation helpers

wolfv:
I ported some functions over to assist in shell activation (WIP!).

This loads environment variables and activation scripts and places them in an `Activator` struct.

I was thinking we could have another function that would consume one (or two) `Activator` objects and emit the shell code based on them.
Something like:

```
activate_environment(to_activate: Activator, to_deactivate: Option<Deactivator>) {
 /// additional logic to activate the env in the shell
 return shell script...
}
```


crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+            .map(|r| r.unwrap().path())
+            .filter(|path| path.is_file())
+            .collect::<Vec<_>>();
+
+        env_var_files.sort();
+
+        // TODO figure out how to more safely `unwrap`?
+        let env_var_json: Vec<serde_json::Value> = env_var_files
+            .iter()
+            .map(|path| fs::read_to_string(path).unwrap())
```
wolfv, line 135: Should I use `try_for_each` here?

baszalmstra: You can use : (from the top of my head)

```rust
.map(|path| fs::read_to_string(path))
.and_then(|text| serde_json::from_str(&json).map_err(Into::into)) // Into::into to convert serde_json error to std::io::Error.
.collect::<Result<Vec<_>, _>()?
```

crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+    if !path.exists() {
+        return Ok(vec![]);
+    }
+
+    let paths = fs::read_dir(path).unwrap();
+
+    let mut scripts = paths
+        .into_iter()
+        .filter(|r| r.is_ok())
+        .map(|r| r.unwrap().path())
```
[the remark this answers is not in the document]
tdejager: And you can use `filter_map` to do it in one call.

crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+
+        // load json but preserve the order of dicts - for this we use the serde preserve_order feature
+        let state_json: serde_json::Value = serde_json::from_str(&state_json).unwrap();
+        let state_env_vars = state_json["env_vars"].as_object().unwrap();
+
+        for (key, value) in state_env_vars {
+            // TODO log warning?
+            // println!("{}: {} (overwritten)", key, value.as_str().unwrap());
+            env_vars.insert(key.to_uppercase(), value.as_str().unwrap().to_string());
+        }
```
baszalmstra, line 157: You could write this as an iterator.

crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+    pub paths: Vec<PathBuf>,
+
+    /// A list of scripts to run when activating the environment
+    pub activation_scripts: Vec<PathBuf>,
+
+    /// A list of scripts to run when deactivating the environment
+    pub deactivation_scripts: Vec<PathBuf>,
+
+    /// A list of environment variables to set when activating the environment
+    pub env_vars: IndexMap<String, String>,
```
tdejager, line 59: Are you using the  `IndexMap` so you can sort them :)?

crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+///
+/// # Arguments
+///
+/// * `prefix` - The path to prefix the path entries with
+/// * `operating_system` - The operating system that the path entries are for
+///
+/// # Returns
+///
+/// A vector of path entries
+fn prefix_path_entries(
```
baszalmstra, line 172: I dont quite get what this means, does this return where binaries can be stored?

tdejager (changes requested):

crates/rattler_shell_helpers/src/lib.rs
wolfv: Nope, so that they stay sorted. It's not well specified and IMO the conda folks chose the wrong json format (it should've been an ordered list) ...

wolfv: I am using the `preserve_order` feature of serde, btw.

baszalmstra (commented):
This is awesome! Thanks!

Im wondering if it would make sense to decouple the activator from the shell. That way you can create an `Activator` from an environment and then create shell scripts for any shell from that activator instance.

Im also wondering if it would make sense to split the behavior for environment variables and activation scripts into two objects for the sake of separation of concerns.

You use `unwrap` quite a bit. I think its almost always better to use `except` instead because then you can also provide a reason why this should always work. But even better would be to propagate possible errors up the callstack. To that end I also think its better to use `thiserror` instead of `anyhow` because then you provide type safe errors back to the user who can act accordingly. Since any `thiserror` can always be converted to an `anyhow` error it provides more control to the user of your API.

One final note, would it also be possible to construct an Activator from a non-installed environment but simply from `PrefixRecords` and/or cached package directories? Perhaps then we can act on a more immutable state. This could be very useful while installing an environment as the prefix directory is completely in flux during the installation phase which might complicate the discovery of files. WDYT?

wolfv:
Thanks for the comments, will fix them up. I am not sure if decoupling the "Activator" or whatever from the notion of a "shell" makes a lot of sense, since at the end of the day "activation" itself mostly makes sense in a shell or script. For the same reason I am also not very convinced by the idea of creating an Activator from e.g. the conda-meta data (instead of the file system) - if we want to do this "on-the-fly" then we can better write a short bash script that performs the same actions? Also for the `env` variables, we actually do need to read the files to figure them out.

crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+/// * `prefix` - The path to the root of the conda environment
+///
+/// # Returns
+///
+/// A map of environment variables
+///
+/// # Errors
+///
+/// If the `state` file or the `env_vars.d` directory cannot be read, an error is returned.
+fn collect_env_vars(prefix: &Path) -> anyhow::Result<IndexMap<String, String>> {
```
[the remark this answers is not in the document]
wolfv: I want to preserve the order (also used serde_json preserve-order feature in this crate).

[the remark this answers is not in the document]
wolfv: We have our very own now :)

crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,354 @@
...
+    let pkg_env_var_dir = prefix.join("etc/conda/env_vars.d");
+    let mut env_vars = IndexMap::new();
+
+    if pkg_env_var_dir.exists() {
+        let env_var_files = pkg_env_var_dir.read_dir()?;
+        let mut env_var_files = env_var_files
+            .into_iter()
+            .filter(|r| r.is_ok())
+            // is safe because we filtered on `is_ok()`
+            .map(|r| r.unwrap().path())
```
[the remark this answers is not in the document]
wolfv: Done! :)

crates/rattler_shell_helpers/src/lib.rs
```diff
@@ -0,0 +1,418 @@
...
+
+        let state_env_vars = state_json["env_vars"].as_object().ok_or_else(|| {
+            ActivationError::InvalidEnvVarFileStateFile {
+                file: state_file.to_path_buf(),
+            }
+        })?;
+
+        for (key, value) in state_env_vars {
+            if state_env_vars.contains_key(key) {
+                println!(
```
baszalmstra, line 217: Can you use `tracing` instead of `println`? That way the user of the library is able to configure how they want to log things.

baszalmstra (commented):
Aside from one comment the rest of the code looks great!

I am still wondering if it makes sense to separate the Activator from the shell. If the activator collects all scripts for all shells you can generate the required script for any shell later. WDYT?

PR #69, opened 2023-02-23
feat: added reading of conda-lock files

tdejager:
Allows reading of conda-lock files.


crates/rattler_conda_types/src/conda_lock/mod.rs
```diff
@@ -0,0 +1,201 @@
...
+}
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Hash)]
+pub struct PackageName(String);
+#[derive(Serialize, Deserialize, Eq, PartialEq, Hash)]
+/// This is basically a MatchSpec but will never contain the package name
+pub struct VersionConstraint(String);
+
+#[derive(Serialize)]
+pub enum PackageHashes {
```
wolfv, line 77: I am curious why we need this vs. two `Option<String>`?

Also I am wondering if we should deserialize them straight away as the binary representations as properly "typed" checksums or keep them as strings?

wolfv:
pretty cool! :)

crates/rattler_conda_types/src/conda_lock/mod.rs
tdejager: I just like it that you cannot have the case where both are None :)

wolfv:
I am wondering if this should go in conda-types or a seperate crate.

crates/rattler_conda_types/src/conda_lock/mod.rs
tdejager: Yeah so I had the feeling that if both values are none the whole thing should just be excluded from the yaml.

crates/rattler_conda_types/src/conda_lock/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub git_user_name: String,
+    /// Git user.email field of global config
+    pub git_user_email: String,
+    /// sha256 hash of the most recent git commit that modified one of the input files for this lockfile
+    pub git_sha: String,
+}
+
+#[derive(Serialize, Deserialize)]
+/// Represents whether this is a dependency managed by pip or conda
+pub enum Manager {
```
baszalmstra, line 74: You could also use `#[serde(rename_all="lowercase")]`

crates/rattler_conda_types/src/conda_lock/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    #[serde(rename = "conda")]
+    Conda,
+    #[serde(rename = "pip")]
+    Pip,
+}
+
+#[derive(Serialize, Deserialize, Eq, PartialEq, Hash)]
```
baszalmstra, line 81: Why did you use new types here? Especially for package name String looks pretty logical. Maybe you can at least implement Display for it?

There is now also no way to access the string name because the string is not `pub`.

crates/rattler_conda_types/src/conda_lock/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+pub struct PackageName(String);
+#[derive(Serialize, Deserialize, Eq, PartialEq, Hash)]
+/// This is basically a MatchSpec but will never contain the package name
+pub struct VersionConstraint(String);
```
baszalmstra, line 85: Would be nice if you add a `parse` function that turns this into a `MatchSpec`.

crates/rattler_conda_types/src/conda_lock/mod.rs
baszalmstra: Maybe you can add a `serde_with` module to `rattler_digest` to serialize a digests? That would be neet!

crates/rattler_conda_types/src/conda_lock/mod.rs
```diff
@@ -0,0 +1,222 @@
...
+    pub used_env_vars: Vec<String>,
+}
+
+#[cfg(test)]
+mod test {
+    use super::PackageHashes;
+    use crate::conda_lock::CondaLock;
+    use serde_yaml::from_str;
+
+    #[test]
```
baszalmstra, line 181: Maybe you can add an `insta` test for a few lock files here? That would alert us in the future if something changes in the serialization/deserialization code.

crates/rattler_conda_types/src/conda_lock/mod.rs
tdejager: Done

crates/rattler_conda_types/src/conda_lock/mod.rs
tdejager: Removed `PackageName` added Display for `VersionConstraint`

crates/rattler_conda_types/src/conda_lock/mod.rs
tdejager: Will do this in subsequent PR.

tdejager:
Added Digest parsing to `rattler_digest` crate, so that should work now!

crates/rattler_conda_types/src/conda_lock/mod.rs
tdejager: Yes, added it for one lock file. Was already working on another lock file in a different PR.

wolfv:
Wow, that looks pretty slick! :)

PR #71, opened 2023-02-24
Feat/use conda lock

tdejager:
Use the conda lock for installation of a python environment.

Next up: writing lock files.


crates/rattler_conda_types/src/conda_lock/mod.rs
```diff
@@ -239,14 +294,60 @@ mod test {
...
+                .len(),
+            180
+        );
+        assert_eq!(
+            conda_lock
+                .packages_for_platform(Platform::OsxArm64)
+                .collect::<Vec<_>>()
+                .len(),
+            183
+        );
```
baszalmstra, line 351: This is a bit fragile. You could also use `insta` to print all the filenames of the packages.

baszalmstra (approved):
Nice! 🤗  One small comment!

crates/rattler_conda_types/src/conda_lock/mod.rs
tdejager: This is done.

PR #72, opened 2023-02-24
feat: create command

baszalmstra:
This is a work-in-progress attempt at the create command (definitely needs more work). This still needs some big refactors to pull a lot of code out of the rattler binary itself. It does work however!!

rattler create

This PR also does a pretty significant refactor of the Solver API. Instead of providing the solver with installed packages and specifying which operations you want to perform. You provide it with the current state (pinned and locked packages) and the matchspecs that your desired environment should contain. It will then return all the packages that should exist in this environment. The tool itself then takes care of determining which packages to reinstall/install or remove. This is very different from the previous approach but makes it more stateless.

I also changed the libsolv implementation a little bit. This diverges from the mamba implementation. I introduced the `SolverGoal` struct which defines what the solver should do. I added several handy functions that make it a bit more insightful.

I also fixed several bugs. :)

Anyway, more to come!

@aochagavia I can integrate your changes into this!


tdejager:
Great work!! 😁😁

baszalmstra:
I split up the changes into #73, #74, and #75 to make it a bit easier to review. I'll turn this PR into a draft until those PRs are merged. (possibly also #67 )

crates/rattler-bin/src/commands/create.rs
```diff
@@ -367,7 +367,7 @@ async fn install_package_to_environment(
     // Create the conda-meta directory if it doesnt exist yet.
     let target_prefix = target_prefix.to_path_buf();
     match tokio::task::spawn_blocking(move || {
-        let conda_meta_path = target_prefix.join("conda-meta");
+        let conda_meta_path = target_prefix.join("prefix");
```
wolfv, line 371: hmm, I wonder if we should keep that at `conda-meta` to not throw off micromamba or conda tools and others

baszalmstra: woops! thats a mistake!

tdejager (approved):

crates/rattler-bin/src/commands/create.rs
```diff
@@ -11,54 +35,613 @@ pub struct Opt {
...
+    }
+}
+
+/// Completely remove the specified package from the environment.
+async fn remove_package_from_environment(
+    target_prefix: &Path,
+    package: &PrefixRecord,
+) -> anyhow::Result<()> {
+    // TODO: Take into account any clobbered files, they need to be restored.
+    // TODO: Can we also delete empty directories?
```
wolfv, line 402: yes, that's how it's done in conda & mamba.

wolfv (approved):

PR #73, opened 2023-02-26
fix: validate index.json file

baszalmstra:
Updates package cache validation code to also check that a valid `index.json` file exists. The parsed files are also returned so you don't have to reparse them.

Code taken from #72 to split it up a bit.

### DIFF
--- a/crates/rattler/src/validation.rs
+++ b/crates/rattler/src/validation.rs
@@ -10,7 +10,7 @@
 //! `paths.json` file is missing these deprecated files are used instead to reconstruct a
 //! [`PathsJson`] object. See [`PathsJson::from_deprecated_package_directory`] for more information.
 
-use rattler_conda_types::package::{PackageFile, PathType, PathsEntry, PathsJson};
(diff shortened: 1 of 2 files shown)


crates/rattler/src/validation.rs
```diff
@@ -33,6 +33,9 @@ pub enum PackageValidationError {
 
     #[error("the path '{0}' seems to be corrupted")]
     CorruptedEntry(PathBuf, #[source] PackageEntryValidationError),
+
+    #[error("failed to verify index.json")]
+    InvalidIndexJson(#[source] std::io::Error),
```
wolfv, line 38: I am wondering if we should call this `ReadIndexJsonError` (like `ReadPathsJsonError`) – or rename that one to `InvalidPathsJson`?

wolfv:
Looks good to me, one tiny comment

crates/rattler/src/validation.rs
baszalmstra: Yeah good one, I renamed it to `ReadIndexJsonError` and changed the error message as well.

wolfv (approved):

PR #74, opened 2023-02-26
fix: repodata cache was always out of date

baszalmstra:
Fixes an issue where the cache was always out of date because the check was reversed. 🙈 

Code taken from https://github.com/mamba-org/rattler/pull/72 to split it up a bit.

### DIFF
--- a/crates/rattler_repodata_gateway/src/fetch/mod.rs
+++ b/crates/rattler_repodata_gateway/src/fetch/mod.rs
@@ -628,7 +628,7 @@ fn validate_cached_state(cache_path: &Path, subdir_url: &Url) -> ValidatedCacheS
     let cache_state = match RepoDataState::from_path(&cache_state_path) {
         Err(e) if e.kind() == ErrorKind::NotFound => {
             // Ignore, the cache just doesnt exist
-            tracing::info!("repodata cache state is missing. Ignoring cached files...");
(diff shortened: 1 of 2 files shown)


crates/rattler_repodata_gateway/src/fetch/mod.rs
```diff
@@ -722,13 +722,17 @@ fn validate_cached_state(cache_path: &Path, subdir_url: &Url) -> ValidatedCacheS
                 max_age: Some(duration),
                 ..
             }) => {
-                if duration > cache_age {
-                    tracing::info!("Cache is out of date. Assuming out of date...");
+                if cache_age > duration {
+                    tracing::debug!(
```
wolfv, line 726: Is `debug` something that is not compiled into the final binary? I think information about the cache is kinda informative to analyze wether the cache is hit or not, so I'd be OK with keeping it at `info` (if debug is gone at runtime) so that users could access this information by changing the verbosity flag.

wolfv (approved):

crates/rattler_repodata_gateway/src/fetch/mod.rs
baszalmstra: It is my understanding that all these statements are left in. Let me verify.

baszalmstra: Yep they are :)

PR #75, opened 2023-02-26
feat: stateless solver

baszalmstra:
This modifies the solver API to be more "stateless". Instead of specifying what is installed and the solver telling you what to change, the solver now always returns the "desired state". Transforming an installed environment into the desired state is left up to the consumer instead.

I changed the previous behavior (which is similar to Mamba) to make it easier for tools to create "clean" environments where no packages remain from a previous installation. This was previously only possible by determining the exact specs that went into creating an environment (which would imply adding a history). The API should still enable tools to work "incrementally" by modifying the installation behavior instead of the solver behavior.

I also think this refactor will make it easier to integrate different solvers because they do not need to take care of the "transaction" anymore.

 #72 (which already incorporates these changes) contains code to construct a `Transaction` from an installed environment and the result of solving.

@aochagavia This one might be particularly interesting to you! Especially the `SolveGoal`.

Code taken from #72 to split it up a bit.


wolfv:
nice!

crates/rattler_solve/src/lib.rs
```diff
@@ -29,55 +28,53 @@ pub enum SolveError {
...
-    /// All the available packages
+    /// All available packages
     pub available_packages: Vec<Vec<RepoDataRecord>>,
 
-    /// All the packages currently installed
-    pub installed_packages: Vec<PrefixRecord>,
+    /// Records of packages that are previously selected.
+    ///
+    /// If the solver encounters multiple variants of a single package (identified by its name), it
+    /// will sort the records ands select the best possible version. However, if there exists a
```
aochagavia, line 48: typo: _ands_ -> _and_

crates/rattler_solve/src/lib.rs
```diff
@@ -29,55 +28,53 @@ pub enum SolveError {
...
+    /// locked version it will prefer that variant instead. This is useful to reduce the number of
+    /// packages that are updated when installing new packages.
+    ///
+    /// Usually you add the currently installed packages or packages from a lock-file here.
+    pub locked_packages: Vec<RepoDataRecord>,
+
+    /// Records of packages that are previously selected and CANNOT be changed.
+    ///
+    /// If the solver encounters multiple variants of a single package (identified by its name), it
+    /// will sort the records ands select the best possible version. However, if there is a variant
```
aochagavia, line 58: typo: _ands_ -> _and_

crates/rattler_solve/src/libsolv/mod.rs
```diff
@@ -1,25 +1,37 @@
...
+    fn solve(&mut self, problem: SolverProblem) -> Result<Vec<RepoDataRecord>, SolveError> {
         // Construct a default libsolv pool
         let pool = Pool::default();
 
         // Setup proper logging for the pool
         pool.set_debug_callback(|msg, flags| {
             tracing::event!(tracing::Level::DEBUG, flags, "{}", msg);
         });
-        pool.set_debug_level(Verbosity::Low);
+        pool.set_debug_level(Verbosity::Extreme);
```
aochagavia, line 26: Is this something we want to keep or were you using it for debugging and forgot to remove it? I have no preference, just curious

crates/rattler_solve/src/libsolv/mod.rs
```diff
@@ -35,53 +47,67 @@ impl SolverBackend for LibsolvSolver {
...
+        // Lock the currently pinned packages
+        for locked_solvable in pinned_solvables {
+            goal.lock(locked_solvable);
+        }
+
+        // Specify the matchspec requests
+        for spec in problem.specs {
             let id = spec.intern(&pool);
-            queue.push_id_with_flags(id, SolvableFlags::from(request));
+            goal.install(id, false)
```
aochagavia, line 96: I like the `favor`, `lock` and `install` methods!

crates/rattler_solve/src/libsolv/wrapper/solve_goal.rs
```diff
@@ -0,0 +1,106 @@
...
+    SOLVER_SOLVABLE_PROVIDES, SOLVER_UPDATE, SOLVER_WEAK,
+};
+use crate::libsolv::wrapper::pool::MatchSpecId;
+use crate::libsolv::wrapper::solvable::SolvableId;
+use std::os::raw::c_int;
+
+/// Wrapper for libsolv queue type that stores jobs for the solver. This type provides a safe
+/// wrapper around a goal state for libsolv.
+pub struct SolveGoal {
+    queue: ffi::Queue,
```
aochagavia, line 13: Why not use the existing `Queue<T>` wrapper here (i.e. `queue: Queue<SolvableId>`)? I think we are currently duplicating part of that code (which seems extra undesirable because it is unsafe)

crates/rattler_solve/src/libsolv/wrapper/transaction.rs
```diff
@@ -109,46 +108,8 @@ impl TransactionRef {
...
-                            package: repodata_record.clone(),
-                            kind: PackageOperationKind::Install,
-                        });
-                    }
-                    ffi::SOLVER_TRANSACTION_ERASE => {
-                        package_operations.push(PackageOperation {
-                            package: repodata_record.clone(),
-                            kind: PackageOperationKind::Remove,
-                        });
+                        required_packages.push(repodata_record.clone());
```
aochagavia, line 112: I love it when you get the chance to delete branches like this 😉

aochagavia (commented):
Left some comments. Looks like a great step forward!

crates/rattler_solve/src/libsolv/wrapper/repo.rs
```diff
@@ -94,6 +94,7 @@ impl RepoRef {
         let mut package_to_type: HashMap<&str, (PackageExtension, SolvableId)> = HashMap::new();
```
tdejager, line 94: I think you don't need the  type here right?

crates/rattler_solve/src/libsolv/wrapper/repo.rs
```diff
@@ -306,12 +307,15 @@ impl RepoRef {
             } else {
                 tracing::warn!("unknown package extension: {}", &repo_data.file_name);
             }
+
+            // Store the solvable id
+            solvable_ids.push(solvable_id);
         }
 
         // TODO: What does this do?
         unsafe { ffi::repo_internalize(self.as_ptr().as_ptr()) };
```
tdejager, line 316: Do we know what this does already?

tdejager (changes requested):

crates/rattler_solve/src/libsolv/wrapper/solve_goal.rs
baszalmstra: Because the SolverGoal needs to store both `SolvableId` and `MatchSpecId` in the queue. I could make an enum out of that and use `Queue<SolvableOrMatchSpecId>`. WDYT?

crates/rattler_solve/src/libsolv/mod.rs
baszalmstra: Nice catch!

crates/rattler_solve/src/libsolv/wrapper/transaction.rs
```diff
@@ -67,17 +66,17 @@ impl TransactionRef {
...
 
-    /// Returns the package operations derived from the transaction
+    /// Returns which packages should be installed in the environment
     ///
-    /// If the transaction contains libsolv operations that have no mapping to `PackageOperation`,
-    /// an error is returned containing their ids
-    pub fn get_package_operations(
+    /// If the transaction contains libsolv operations that are not "install" an error is returned
+    /// containing their ids.
+    pub fn get_required_packages(
```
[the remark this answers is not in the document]
baszalmstra: Yeah we need a bigger refactor for that anyway. The `SolverProblem` now also consumes all the `RepoData` I think we could also pass the RepoData by reference there. Would you be ok doing that in another PR?

crates/rattler_solve/src/libsolv/wrapper/repo.rs
baszalmstra: Unfortunately, the code doesn't compile without this here.

crates/rattler_solve/src/libsolv/wrapper/transaction.rs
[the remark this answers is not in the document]
tdejager: Sure!

tdejager (approved):

crates/rattler_solve/src/libsolv/wrapper/repo.rs
wolfv: I think that will internalize all the strings into the big libsolv Pool or something like that

PR #77, opened 2023-02-27
add activation script writing

wolfv:


crates/rattler_shell/src/native_path.rs
```diff
@@ -0,0 +1,55 @@
+/// Convert a path to a native path.
```
wolfv, line 1: I'll remove this file for now.

baszalmstra (approved):

PR #78, opened 2023-02-28
Link json

wolfv:
Turns out we were missing `link.json` parsing and are also currently not creating entry points for `noarch`-Python packages. That will come in another PR! :)

### DIFF
--- a/crates/rattler_conda_types/src/package/entry_point.rs
+++ b/crates/rattler_conda_types/src/package/entry_point.rs
@@ -1,0 +1,97 @@
+//! A struct for a single Python entry point.
+
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use std::fmt::Display;
+use std::str::FromStr;
+
+/// A struct for a single Python entry point. An entry point is a command that
+/// runs a function in a Python module. For example, the entry point
+/// `jlpm = jupyterlab.jlpmapp:main` will run the `main` function in the
+/// `jupyterlab.jlpmapp` module when the command `jlpm` is run on the command line.
+/// The main usage for entry points is in `noarch: python` packages.
+///
+/// The entry point is represented as a string in the format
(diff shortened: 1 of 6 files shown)


crates/rattler_conda_types/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 //! `rattler-conda-types` contains data models for types commonly found within the Conda ecosystem.
 //! The library itself doesnt provide any functionality besides parsing the data types.
 
+#[deny(missing_docs)]
```
baszalmstra, line 4: This now only applies to this file. If you want to do this for the whole library use:

```suggestion
#![deny(missing_docs)]
```

crates/rattler_conda_types/src/package/entry_point.rs
```diff
@@ -0,0 +1,97 @@
...
+
+    /// The name of the function to run.
+    pub function: String,
+}
+
+impl FromStr for EntryPoint {
+    type Err = String;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut parts = s.splitn(2, '=');
```
baszalmstra, line 37: You can use `split_once` instead of `splitn`. That will give you a tuple of the left and right side.

crates/rattler_conda_types/src/lib.rs
wolfv: argh, shoot, I was happy that it looked already pretty decent. but actually more work to do that deserves a seperate PR!

crates/rattler_conda_types/src/package/entry_point.rs
wolfv: Cool!, done

baszalmstra (approved):

baszalmstra:
LGTM! Awesome work @wolfv ! 🦖 🦕

PR #82, opened 2023-02-28
matchspec & version docs

wolfv:
I think we should attribute the original conda documentation/files that these were taken from. 

@baszalmstra do you want to check the matchspec parsing? some of the examples didn't work because of a `BracketsError`.

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -9,10 +9,10 @@ mod parse;
 /// comprise a [`crate::PackageRecord`] can be used to compose a [`MatchSpec`].
 ///
 /// [`MatchSpec`] can be composed with keyword arguments, where keys are any of the
-/// attributes of [`PackageRecord`]. Values for keyword arguments are the exact values the
-/// attribute should match against. Many fields can also be matched against non-exact values--by
-/// including wildcard `*` and `>`/`<` ranges--where supported. Any non-specified field is
-/// the equivalent of a full wildcard match.
+/// attributes of [`crate::PackageRecord`]. Values for keyword arguments are the exact
+/// values the attribute should match against. Many fields can also be matched against non-exact
+/// values -- by including wildcard `*` and `>`/`<` ranges--where supported. Any non-specified field
+/// is the equivalent of a full wildcard match.
 ///
 /// MatchSpecs can also be composed using a single positional argument, with optional
 /// keyword arguments. Keyword arguments also override any conflicting information provided in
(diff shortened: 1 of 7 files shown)


wolfv:
I am wondering if we should parse build_number into a `VersionSpec` to make it possible to match on build numbers as well, e.g. `[build_number=">=5"]` makes sense.

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -7,6 +7,104 @@ mod parse;
...
+///
+///   - channel
+///   - subdir
+///   - name
+///   - version
+///   - build
+///
+/// In the future, the namespace field might be added to this list.
+///
+/// Alternatively, an exact spec is given by `*[sha256=01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546b]`.
```
wolfv, line 107: Todo implement `md5` and `sha256`.

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -30,7 +130,17 @@ impl Display for MatchSpec {
...
+            write!(f, "{}", channel)?;
+        }
+
+        if let Some(subdir) = &self.subdir {
+            write!(f, "/{}", subdir)?;
+        }
+
+        if let Some(namespace) = &self.namespace {
+            write!(f, ":{}:", namespace)?;
+        } else if self.channel.is_some() || self.subdir.is_some() {
```
wolfv, line 142: if we only have a subdir, we should do `*/<subdir>`.

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -288,11 +298,14 @@ fn parse(input: &str, channel_config: &ChannelConfig) -> Result<MatchSpec, Parse
...
 
     match_spec.namespace = namespace.map(ToOwned::to_owned).or(match_spec.namespace);
+
     if let Some(channel_str) = channel_str {
-        match_spec.channel = Some(
-            Channel::from_str(channel_str, channel_config)
-                .map_err(ParseMatchSpecError::ParseChannelError)?,
-        )
+        if let Some((channel, subdir)) = channel_str.rsplit_once('/') {
+            match_spec.channel = Some(channel.to_string());
```
wolfv, line 304: Should we check if channel or subdir is `*` and then insert `None`?

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -163,9 +165,17 @@ fn parse_bracket_vec_into_components(
...
+    let mut match_spec = match_spec;
+
+    for elem in bracket {
+        let (key, value) = elem;
+        match key {
+            "version" => match_spec.version = Some(VersionSpec::from_str(value)?),
+            "build" => match_spec.build = Some(value.to_string()),
+            "build_number" => match_spec.build_number = Some(value.parse()?),
+            "fn" => match_spec.file_name = Some(value.to_string()),
+            _ => Err(ParseMatchSpecError::InvalidBracketKey(key.to_owned()))?,
```
wolfv, line 177: we should add `sha256` and `md5` as in the docs. I think `url` might also be a field that `conda` supports.

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -44,15 +45,16 @@ pub enum ParseMatchSpecError {
...
-    pub fn from_str(
-        input: &str,
-        channel_config: &ChannelConfig,
-    ) -> Result<MatchSpec, ParseMatchSpecError> {
-        parse(input, channel_config)
+impl FromStr for MatchSpec {
+    type Err = ParseMatchSpecError;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        parse(s)
```
wolfv, line 57: Should we remove the `parse` function and put everything into `from_str`?

crates/rattler_solve/src/lib.rs
```diff
@@ -422,7 +421,7 @@ mod test_libsolv {
             )
         }
 
-        if pkgs.len() == 0 {
+        if pkgs.is_empty() {
```
wolfv, line 424: clippy fixed that automatically, whoops!

wolfv:
I think another note is that MatchSpecs can match on strings with `*`, e.g. one can do `xtensor*` and it would find `xtensor-blas`, `xtensor-python`, etc.

So I am wondering if we should (similar to `VersionSpec`) have another class for `StringMatcher` (or `StringSpec`), and for build numbers a `NumberSpec` or `NumberMatcher`?

If we go with `...Matcher`, maybe rename `VersionSpec` to `VersionMatcher`.

baszalmstra (approved):

PR #83, opened 2023-02-28
feat: install python entry points

baszalmstra:
Reads the `link.json` file during install and creates Python entry points in parallel with the rest of the files.


crates/rattler/src/install/python.rs
```diff
@@ -68,6 +68,23 @@ impl PythonInfo {
...
     }
 
+    /// Constructs a shebang that will run the rest of the script as Python.
+    pub fn shebang(&self, target_prefix: &str) -> String {
+        let target_path = Path::new(target_prefix).join(self.path());
+        let target_path = target_path.as_os_str().to_string_lossy();
+
+        // Shebangs cannot be larger than 127 characters and executables with spaces are
+        // problematic.
+        if target_path.len() > 127 - 2 {
```
wolfv, line 78: I think we should also check if the `target_path` contains a space here (as the comment indicates :))

crates/rattler/src/install/snapshots/rattler__install__entry_point__test__entry_point_script.snap
```diff
@@ -0,0 +1,15 @@
+---
+source: crates/rattler/src/install/entry_point.rs
+expression: script
+---
+#!/prefix\bin/python3.11
```
wolfv, line 5: there is something strange going on with the `/` vs `\` here? Or is this just an optical illusion? :)

wolfv (approved):
Looks awesome, two small comments! 🚀

crates/rattler/src/install/snapshots/rattler__install__entry_point__test__entry_point_script.snap
baszalmstra: Since I join paths on windows I think that creates the backslash..

wolfv: but should not all path elements be backslash's in that case?

baszalmstra: Yep! Thats also why the tests were failing, I normalized the slashes now :)

crates/rattler/src/install/python.rs
```diff
@@ -71,11 +71,11 @@ impl PythonInfo {
     /// Constructs a shebang that will run the rest of the script as Python.
     pub fn shebang(&self, target_prefix: &str) -> String {
         let target_path = Path::new(target_prefix).join(self.path());
-        let target_path = target_path.as_os_str().to_string_lossy();
+        let target_path = target_path.as_os_str().to_string_lossy().replace('\\', "/");
```
wolfv, line 74: hmm, unsure if that's the right move here. Do we use shebangs only on Unix? 

If we use shebangs on Windows (e.g. with `cygwin`) we should use `cygpath` to convert the path to a unix style path (`/c/Windows/System Profile/ ...`)

baszalmstra: Im not sure it's actually used on Windows. I copied this code from mamba. The launcher reads the shebang to determine the python instance to start I think. I would have to dig deeper though..

baszalmstra: Actually, thinking about it. The `target_prefix` that is passed in should already be the correct prefix (cygwin or not). Since windows works both with `\` or `/` I think this is fine/

PR #85, opened 2023-02-28
feat: allow caching repodata.json as .solv file

aochagavia:
The two first commits contain fixes that were necessary to use rattler to solve conda environments in a standalone application: `ChannelInfo` is used in one of `RepoData`'s public fields, so it makes sense to make it public too; making the `dowload_progress` closure `Send` and `Sync` ensures we can download `repodata.json` in an async context.

The third commit adds functionality to cache repodata.json for libsolv, which results in huge speedups (solve `python` in ~300ms on my machine, instead of >2000ms). I am not totally satisfied with the modifications to the `SolverBackend` trait, but I couldn't find a better way of expressing what I wanted in the current type system (maybe `impl Trait` in traits would help here, but the feature is currently unstable). Also, the conditional compilation side of things is a bit ugly, but as long as it stays properly encapsulated inside the libsolv wrapper I think it is tolerable.


baszalmstra (approved):
The PR looks good to me! 

I think in the future it would be nice if we could also enable this for Windows! But that's for another day.

wolfv:
Ouch, sorry, seems like my conflict resolution has made the tests not pass anymore. Let me know if I should look into it!

PR #86, opened 2023-03-01
Writing conda lock files

tdejager:
Add support from writing from the `conda-lock` datastructures into yaml. This is the first step because it's still bare-bones serialization. We do include the initial part of the `content_hash` calculation though, which was a bit difficult to reverse engineer. 

Currently, I'm only allowing the write of fields I think we will use, we can also extend those later.

We are still missing, but I want to do this in a next PR:

- [ ] Create conda lock from `Vec<RepodataRecord>` so that we can use it in the create for example
- [ ] Add default virtual packages for a canonical lock file.
- [ ] Add virtual packages to the content_hash, and see if we can get similar output for the python `conda-lock` and our version.


wolfv:
It looks like your rebase went haywire somewhere :(

PR #87, opened 2023-03-01
Feat/write conda lock

tdejager:
Add support from writing from the conda-lock datastructures into yaml. This is the first step because it's still bare-bones serialization. We do include the initial part of the content_hash calculation though, which was a bit difficult to reverse engineer.

Currently, I'm only allowing the write of fields I think we will use, we can also extend those later.

We are still missing, but I want to do this in a next PR:

- [ ] Create conda lock from Vec<RepodataRecord> so that we can use it in the create for example
- [ ] Add default virtual packages for a canonical lock file.
- [ ] Add virtual packages to the content_hash, and see if we can get similar output for the python conda-lock and our version.


wolfv:
nice!

crates/rattler_conda_types/src/conda_lock/content_hash.rs
```diff
@@ -0,0 +1,132 @@
...
+    Ok(serde_json::to_string(&content_hash_data)?
+        // Replace these because python encodes with spaces
+        // and serde does not
+        .replace(":", ": ")
+        .replace(",", ", "))
+}
+
+/// Calculate the content hash for a platform and set of matchspecs
+pub fn calculate_content_hash(
+    _platform: &Platform,
```
wolfv, line 96: why the underscore if you use the variable later?

crates/rattler_conda_types/src/conda_lock/content_hash.rs
```diff
@@ -0,0 +1,132 @@
...
+    // TODO: Add default list of virtual packages and then create the content hashing
+
+    // Create the python dict
+    let content_hash_data = ContentHashData {
+        channels: channels.to_vec(),
+        specs,
+    };
+
+    // Create the json
+    Ok(serde_json::to_string(&content_hash_data)?
```
wolfv, line 87: Can we use the PrettyFormatter? https://docs.rs/serde_json/latest/serde_json/ser/struct.PrettyFormatter.html

Replacing `:` and `,` looks "dangerous" :D

wolfv:
looks goooooood!

wolfv:
Bas and I had a quick look and think that it's possible to use an implementation of the PrettyFormatter to get the format right (and it's not so much code to write).

crates/rattler_conda_types/src/conda_lock/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+    pub locked_packages: HashMap<Platform, LockedPackages>,
+
+    /// MatchSpecs input
+    /// This is only used to calculate the content_hash
+    /// for the lock file
+    pub input_specs: Vec<MatchSpec>,
+}
+
+impl LockFileBuilder {
+    pub fn new(
```
baszalmstra, line 34: Missing docs :)

crates/rattler_conda_types/src/conda_lock/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+        channels: impl IntoIterator<Item = Channel>,
```
baszalmstra, line 35: Since you can construct a `Channel` from `String` or `&str` maybe you could do:

```suggestion
        channels: impl IntoIterator<Item = impl Into<Channel>>,
```

then you can also simply pass a `String` directly, like: `["conda-forge"]`

crates/rattler_conda_types/src/conda_lock/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+                .map(|into_channel| into_channel.into())
+                .collect(),
+            platforms: platforms.into_iter().collect(),
+            input_specs: input_spec.into_iter().collect(),
+            ..Default::default()
+        }
+    }
+
+    /// Add locked packages per platform
+    pub fn add_locked_packages(mut self, locked_packages: LockedPackages) -> Self {
```
baszalmstra, line 51: Should this panic if there are already entries for that platform?

crates/rattler_conda_types/src/conda_lock/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+impl LockedPackages {
+    pub fn new(platform: Platform) -> Self {
+        Self {
+            locked_packages: Vec::new(),
+            platform,
+        }
+    }
+
+    /// Add a locked package
+    pub fn add_locked_package(mut self, locked_package: LockedPackage) -> Self {
```
baszalmstra, line 107: I always like it better to use `&mut self` in these kind of methods. You can still do proper chaining in that case but you can also do:

```rust
let mut locked_packages = ...
locked_packages.add_locked_package(...)
```

otherwise you always have to move the result back into the binding.

crates/rattler_conda_types/src/conda_lock/builder.rs
```diff
@@ -0,0 +1,209 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::conda_lock::builder::{LockFileBuilder, LockedPackage, LockedPackages};
+    use crate::conda_lock::{Channel, CondaLock, PackageHashes};
+    use crate::{ChannelConfig, MatchSpec, Platform};
+
+    #[test]
+    fn create_lock_file() {
```
baszalmstra, line 182: This function doesnt actually test anything :)

crates/rattler_conda_types/src/conda_lock/content_hash.rs
tdejager: We will need it in a later PR

tdejager: For now its unused

crates/rattler_conda_types/src/conda_lock/content_hash.rs
tdejager: Let's see!

crates/rattler_conda_types/src/conda_lock/content_hash.rs
baszalmstra: the first line in the function is passed the platform. 🙃

crates/rattler_conda_types/src/conda_lock/builder.rs
tdejager: Yeah I kinda wanted it to function as an identity function

crates/rattler_conda_types/src/conda_lock/builder.rs
tdejager: I get that problem, but I kinda think that goes against the builder pattern anyways.

I guess I like this a bit more because otherwise it's hard to pass in the `LockedPackages` directly into the builder, as it will be a mutable reference. Also e.g Clap also does it in this way.

crates/rattler_conda_types/src/conda_lock/builder.rs
tdejager: Added

crates/rattler_conda_types/src/conda_lock/builder.rs
tdejager: Done. I wanted that initially anyways :)

crates/rattler_conda_types/src/conda_lock/content_hash.rs
tdejager: Right, changed it!

crates/rattler_conda_types/src/conda_lock/content_hash.rs
tdejager: Seems to work @baszalmstra @wolfv! Check it out :)

baszalmstra (approved):

PR #95, opened 2023-03-06
Solve problems

wolfv:
This PR parses solver problems into a vector of enums. We need to post process that list of problems into a graph, then compress it and print it to get similar error messages as with `libmamba` :)

### DIFF
--- a/crates/rattler_solve/src/libsolv/wrapper/solver.rs
+++ b/crates/rattler_solve/src/libsolv/wrapper/solver.rs
@@ -4,9 +4,12 @@ use std::ptr::NonNull;
 
 use crate::libsolv::wrapper::pool::Pool;
 use crate::libsolv::wrapper::solve_goal::SolveGoal;
+use crate::libsolv::wrapper::solve_problem::SolveProblem;
 
 use super::ffi;
 use super::flags::SolverFlag;
+use super::queue::Queue;
+use super::solvable::SolvableId;
 use super::transaction::Transaction;
 
 /// Wrapper for libsolv solver, which is used to drive dependency resolution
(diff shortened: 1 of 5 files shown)


crates/rattler_solve/src/libsolv/wrapper/pool.rs
```diff
@@ -196,6 +196,21 @@ impl Pool {
             (id != 0).then_some(StringId(id))
         }
     }
+
+    /// Returns a list of solvables that match the provided id (id can be created from matchspec)
+    pub fn select_by_dep_id(&self, id: Id) -> Vec<SolvableId> {
```
baszalmstra, line 201: Is this a MatchSpecId? Maybe we can use that type here?

crates/rattler_solve/src/libsolv/wrapper/solver.rs
```diff
@@ -87,6 +150,10 @@ impl Solver<'_> {
             // Safe because we know the `transaction` ptr is valid
             Ok(unsafe { Transaction::new(self, transaction) })
         } else {
+            let problems = self.all_solver_problems();
+            println!("problems: {:#?}", problems);
```
baszalmstra, line 154: Left over `println`

crates/rattler_solve/Cargo.toml
```diff
@@ -13,6 +13,8 @@ tracing = "0.1.29"
 serde = { version = "1.0.130", features = ["derive"] }
 url = "2.2.2"
 hex = "0.4.3"
+graphlib = "0.6.3"
+petgraph = "0.6.3"
```
baszalmstra, line 17: These don't appear to be used yet?

wolfv: removed!

crates/rattler_solve/src/libsolv/wrapper/solver.rs
wolfv: yeah I used it for testing. Shoudl we keep the PR WIP and leave it in, or try to merge it already?

crates/rattler_solve/src/libsolv/wrapper/pool.rs
wolfv: Nope, unfortunately it's something else (a `relid` or `depid` ... but maybe we should make a new type for it on the Rust side).

baszalmstra: Yeah lets :)

crates/rattler_solve/src/libsolv/wrapper/solver.rs
baszalmstra: Wip is fine 👍

wolfv: I removed it for now

crates/rattler_solve/src/libsolv/wrapper/pool.rs
wolfv: I removed this function for the time being and would rather implement it similar to how we do String Id and friends

baszalmstra (approved):

PR #99, opened 2023-03-07
fix: robustly handle NUL in repodata.json

### LINKED ISSUE #98: rattler_solve, NUL characters and robustness
We are currently using `libsolv` as solver backend in `rattler_solve` through FFI calls. When passing the information about the available packages to `libsolv`, parts of the data (such as a package's build string or license) need to be converted from a Rust string to a NUL-terminated string (using the `CString` Rust type). This assumes that the original string does not contain NUL characters, which is not always guaranteed.

In its current state, rattler is unable to solve an environment if the `repodata.json` for a particular channel includes strings with `\u0000`. Adding the following package to `test-data/channels/dummy/linux-64/repodata.json`, for instance, causes all tests related to that file to fail:

```json
"baz-1.0-unix_py36h1af98f8_2\u0000.tar.bz2": {
  "build": "unix_py36h1af98f8_2\u0000",
  "build_number": 1,
  "depends": [
    "__unix"
  ],
  "license": "MIT",
  "license_family": "MIT",
  "md5": "bc13aa58e2092bcb0b97c561373d3905",
  "name": "bar",
  "sha256": "97ec377d2ad83dfef1194b7aa31b0c9076194e10d995a6e696c9d07dd782b14a",
  "size": 414494,
  "subdir": "linux-64",
  "timestamp": 1605110689658,
  "version": "1.2.3"
}
```

I see two alternatives to go on about this:

1. Ignore the issue. Other C-based tools in the conda ecosystem would probably break as well if there are NUL escapes in repodata.json, and maybe we can just trust the maintainers of a channel to take that into consideration.
2. Use a custom method to construct `CString`s that are always valid, by replacing NUL characters with something else (similar to `String::from_utf8_lossy`). I will push a PR for this shortly, so you can see how it would look like.

aochagavia:
Fixes #98

### DIFF
--- a/crates/rattler_solve/Cargo.toml
+++ b/crates/rattler_solve/Cargo.toml
@@ -16,6 +16,7 @@ hex = "0.4.3"
 
 [dev-dependencies]
 insta = { version = "1.16.0", features = ["yaml"] }
+rstest = "0.16.0"
 serde_json = "1.0.68"
 url = "2.2.2"
 
--- a/crates/rattler_solve/src/lib.rs
+++ b/crates/rattler_solve/src/lib.rs
@@ -8,7 +8,6 @@ mod solver_backend;
 
 pub use libsolv::LibsolvBackend;
 pub use solver_backend::SolverBackend;
-use std::ffi::NulError;
 
 use rattler_conda_types::GenericVirtualPackage;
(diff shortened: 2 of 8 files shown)


baszalmstra:
Thanks this looks good!

PR #101, opened 2023-03-08
add some WIP rattler index functions

wolfv:
The sort alphabetically works for the map, but not for the set. I think it works for the map because it converts to a BTree internally. I was thinking it might be more efficient to convert RepoData to a generic so that we can use sorted containers (like a BTree) in the indexing functions?

I don't know yet how to sort the `HashSet` as well. 

There are also some inconsistencies between `IndexJson` and `PackageRecord` that we should take a look at :) Also it looks like `IndexJson` is currently missing `track_features` and `features` information.

### DIFF
--- a/crates/rattler_conda_types/src/lib.rs
+++ b/crates/rattler_conda_types/src/lib.rs
@@ -8,7 +8,7 @@ mod explicit_environment_spec;
 mod match_spec;
 mod no_arch_type;
 mod platform;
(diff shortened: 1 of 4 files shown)


baszalmstra:
Can this be closed @wolfv ?

wolfv:
There is still the indexing from packages in here .. but we can close for the time being :)

PR #109, opened 2023-03-10
feat: enable downloading repodata_from_patches.json

baszalmstra:
Enable downloading of `repodata_from_packages.json` with the `fetch_repo_data` function.

### DIFF
--- a/crates/rattler_repodata_gateway/src/fetch/mod.rs
+++ b/crates/rattler_repodata_gateway/src/fetch/mod.rs
@@ -87,6 +87,39 @@ impl Default for CacheAction {
     }
 }
 
+/// Defines which type of repodata.json file to download. Usually you want to use the
+/// [`Variant::AfterPatches`] variant because that reflects the repodata with any patches applied.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Variant {
+    /// Fetch the `repodata.json` file. This `repodata.json` has repodata patches applied. Packages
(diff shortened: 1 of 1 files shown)


crates/rattler_repodata_gateway/src/fetch/mod.rs
```diff
@@ -87,6 +87,39 @@ impl Default for CacheAction {
     }
 }
 
+/// Defines which type of repodata.json file to download. Usually you want to use the
+/// [`Variant::AfterPatches`] variant because that reflects the repodata with any patches applied.
+#[derive(Copy, Clone, Debug, PartialEq, Eq)]
+pub enum Variant {
+    /// Fetch the `repodata.json` file. This `repodata.json` has repodata patches applied. Packages
+    /// may have also been removed from this file (yanked).
+    AfterPatches,
```
wolfv, line 96: there is also the `current_repodata.json` which is a pruned version. Should we already include that?

wolfv (approved):

crates/rattler_repodata_gateway/src/fetch/mod.rs
baszalmstra: Good one! Ill add it.

baszalmstra: Done in [f75706f](https://github.com/mamba-org/rattler/pull/109/commits/f75706f0acbb6b5cb64b8d4e35d5daaf7457a038)

PR #112, opened 2023-03-12
add package writing functions

wolfv:
This adds some low-level package write functions. Unfortunately the zstd writing doesn't work for some reason (seems to write a partial file for the first file and 0 bytes for the second) and I don't know why :)

### DIFF
--- a/crates/rattler_conda_types/src/package/package_metadata.rs
+++ b/crates/rattler_conda_types/src/package/package_metadata.rs
@@ -1,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+/// This is metadata about the package version that is contained in `.conda` packages only, in the
+/// outer zip archive.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PackageMetadata {
+    /// The version of the conda package format. This is currently always 2.
+    pub conda_pkg_format_version: u64,
+}
--- a/crates/rattler_conda_types/src/package/package_metadata.rs
+++ b/crates/rattler_conda_types/src/package/package_metadata.rs
@@ -7,3 +7,11 @@ pub struct PackageMetadata {
     /// The version of the conda package format. This is currently always 2.
     pub conda_pkg_format_version: u64,
 }
+
+impl Default for PackageMetadata {
+    fn default() -> Self {
+        Self {
+            conda_pkg_format_version: 2,
+        }
+    }
+}
--- a/crates/rattler_package_streaming/src/write.rs
+++ b/crates/rattler_package_streaming/src/write.rs
@@ -50,6 +50,12 @@ pub enum CompressionLevel {
     Numeric(u32),
 }
 
+impl Default for CompressionLevel {
+    fn default() -> Self {
+        CompressionLevel::Default
+    }
+}
+
 impl CompressionLevel {
     fn to_zstd_level(self) -> Result<i32, std::io::Error> {
         match self {
@@ -194,9 +200,7 @@ pub fn write_conda_package<W: Write + Seek>(
         zip::write::FileOptions::default().compression_method(zip::CompressionMethod::Stored);
 
     // write the metadata as first file in the zip archive
-    let package_metadata = PackageMetadata {
-        conda_pkg_format_version: 2,
-    };
+    let package_metadata = PackageMetadata::default();
     let package_metadata = serde_json::to_string(&package_metadata).unwrap();
     outer_archive.start_file("metadata.json", options)?;
     outer_archive.write_all(package_metadata.as_bytes())?;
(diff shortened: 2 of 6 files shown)


baszalmstra:
Nice nice! 

> Unfortunately the zstd writing doesn't work for some reason (seems to write a partial file for the first file and 0 bytes for the second) and I don't know why :)

Thats probably because you need to call [`finish`](https://docs.rs/zstd/latest/zstd/stream/write/struct.Encoder.html#method.finish) on the encoders!

wolfv:
Hmm, I thought that would be done here: https://github.com/mamba-org/rattler/pull/112/files#diff-d86da51555092380381d6f546daa7c7baaef0a1e724295c46945c8c516509a18R59

baszalmstra:
That finishes writing the archive but I dont think it calls finish on the underlying stream because the finish call is not part of the trait.

wolfv:
nice, yes, works better with `into_inner().finish()`!

crates/rattler_conda_types/src/package/package_metadata.rs
```diff
@@ -0,0 +1,9 @@
+use serde::{Deserialize, Serialize};
+
+/// This is metadata about the package version that is contained in `.conda` packages only, in the
+/// outer zip archive.
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PackageMetadata {
```
baszalmstra, line 6: Maybe you can add an implementation of `Default` that constructs this with the value 2?

crates/rattler_package_streaming/src/write.rs
```diff
@@ -0,0 +1,281 @@
...
+use std::path::PathBuf;
+
+use bzip2;
+use rattler_conda_types::package::PackageMetadata;
+use tar;
+use zip;
+
+/// a function that sorts paths into two vectors, one that starts with `info/` and one that does not
+/// both vectors are sorted alphabetically for reproducibility
+fn sort_paths(paths: &[PathBuf], base_path: &Path) -> (Vec<PathBuf>, Vec<PathBuf>) {
```
baszalmstra, line 18: You could also take `impl IntoIterator<Item=impl AsRef<Path>>` as the input type and return two iterators instead of directly converting to vecs.

crates/rattler_package_streaming/src/write.rs
```diff
@@ -0,0 +1,281 @@
...
+        .sorted_by(|a, b| a.cmp(b))
+        .map(|p| p.to_path_buf())
+        .collect::<Vec<PathBuf>>();
+
+    (info_paths, other_paths)
+}
+
+/// Select the compression level to use for the package
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CompressionLevel {
```
baszalmstra, line 42: You can implement `Default` for this struct which returns Default.

crates/rattler_package_streaming/src/write.rs
```diff
@@ -0,0 +1,281 @@
...
+        writer,
+        compression_level.to_bzip2_level()?,
+    ));
+
+    // sort paths alphabetically, and sort paths beginning with `info/` first
+    let (info_paths, other_paths) = sort_paths(paths, base_path);
+    for path in info_paths.iter().chain(other_paths.iter()) {
+        // TODO we need more control over the archive headers here to
+        // set uid, gid to 0.
+        archive.append_path_with_name(base_path.join(path), path)?;
```
baszalmstra, line 137: Does this work with executable permissions and such?

crates/rattler_package_streaming/src/write.rs
```diff
@@ -0,0 +1,281 @@
...
+    use std::{fs::File, path::PathBuf};
+
+    use walkdir::WalkDir;
+
+    #[test]
+    fn test_write_tar_bz2() {
+        // repackage a package found at `path` and compare the contents of the tar.bz2
+        // with the contents of the original package
+        // `path` should be a path to a directory containing a `info/` directory
+        let path = PathBuf::from("/Users/wolfv/micromamba/pkgs/zstandard-0.19.0-py311hdcbfb07_1");
```
baszalmstra, line 240: This shouldnt work in CI right?

crates/rattler_package_streaming/tests/write.rs
```diff
@@ -0,0 +1,223 @@
...
+    }
+}
+
+enum FilterFiles {
+    None,
+    NoInfo,
+    NoLicense,
+}
+
+fn compare_two_tar_archives<T: Read>(
```
baszalmstra, line 54: Could you maybe explain a little bit what is going on here?

crates/rattler_package_streaming/src/write.rs
wolfv: Nope

baszalmstra (approved):

PR #118, opened 2023-03-16
feat: better readme

baszalmstra:
[rendered](https://github.com/baszalmstra/rattler/blob/feat/update_readme/README.md)


wolfv:
Do we need to add some comment about `cmake`?

PR #131, opened 2023-03-24
feat: generate bindings statically

baszalmstra:
This PR adds a new crate "tools" which contains commands to perform code generation commands and the like. There is currently only a single command `gen-libsolv-bindings` which will generate the libsolv bindings from the libsolv source code. The bindings are now part of the source code instead of being generated every time you build the `rattler-solv` crate.

This eliminates the dependency on `bindgen` and with that, the dependency on `clang-sys`  when using the `rattler-solv` crate. This should make it much easier to build the code for targets that don't have a runtime version of clang available (like musl or wasm).

A test is also included that checks whether the bindings are still up to date to ensure we don't use out-of-date bindings by accident.


wolfv (approved):

wolfv:
slick!

PR #132, opened 2023-03-26
always write file type

wolfv:
For some reason, micromamba didn't end up in the binary replacement loop when not explicitly mentioning `binary`. I think we should check wether that is wrong behavior in micromamba, or wether we should (always) write the filetype when binary vs. text or idk.

I also noticed a mix of text & binary and no classification when I checked some other paths.json files and couldn't figure out what the default is from just looking at them now.

### DIFF
--- a/crates/rattler_conda_types/src/package/paths.rs
+++ b/crates/rattler_conda_types/src/package/paths.rs
@@ -166,7 +166,7 @@ pub struct PathsEntry {
     pub path_type: PathType,
 
     /// The type of the file, either binary or text.
-    // #[serde(default, skip_serializing_if = "FileMode::is_binary")]
+    #[serde(default)]
     pub file_mode: FileMode,
 
     /// Optionally the placeholder prefix used in the file. If this value is `None` the prefix is not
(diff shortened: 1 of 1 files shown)


crates/rattler_conda_types/src/package/paths.rs
```diff
@@ -166,7 +166,7 @@ pub struct PathsEntry {
     pub path_type: PathType,
 
     /// The type of the file, either binary or text.
-    #[serde(default, skip_serializing_if = "FileMode::is_binary")]
+    // #[serde(default, skip_serializing_if = "FileMode::is_binary")]
```
baszalmstra, line 169: Should we at least keep the default in, for the cases where it is not mentioned?

wolfv:
yes, good point! We could also spend some more time figuring out how conda-build behaves in this case :)

wolfv:
OK, I checked more deeply. They only write the file mode if there is a `prefix_placeholder`, and otherwise leave it out. So these two fields are coupled for serialization.

baszalmstra:
Ah that makes sense! We can also do that by moving the prefix placeholder to the enum and making it optional

baszalmstra:
This is superseded by #136

PR #136, opened 2023-03-28
fix: couple file_mode and prefix_placeholder

baszalmstra:
@wolfv figured out that the `file_mode` is only set if the `prefix_placeholder` is set. This makes sense since the file_mode refers to how the placeholder should be interpreted in the file.

To make this more explicit I pulled out both fields and added a new type `PrefixPlaceholder` that combines the two. This means they are either both set or not at all. Using `serde(flatten)` the structure of the file remains the same.

Supersedes #132

### DIFF
--- a/crates/rattler_conda_types/src/package/paths.rs
+++ b/crates/rattler_conda_types/src/package/paths.rs
@@ -97,11 +97,10 @@ impl PathsJson {
                     match path_type {
                         Ok(path_type) => Ok(PathsEntry {
                             path_type,
-                            file_mode: prefix
-                                .map(|entry| entry.file_mode)
-                                .unwrap_or(FileMode::Text),
-                            prefix_placeholder: prefix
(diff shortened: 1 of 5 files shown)


wolfv (approved):

crates/rattler_conda_types/src/package/paths.rs
```diff
@@ -155,6 +154,20 @@ impl PathsJson {
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize, Deserialize, Hash)]
+pub struct PrefixPlaceholder {
+    /// The type of the file, either binary or text. Depending on the type of file either text
+    /// replacement is performed or CString replacement.
+    pub file_mode: FileMode,
+
+    /// The placeholder prefix used in the file. This is the path of the prefix when the package
+    /// was build.
+    #[serde(rename = "prefix_placeholder")]
+    pub placeholder: String,
```
wolfv, line 168: This should've been a PathBuf

PR #137, opened 2023-03-30
re-export hash types from rattler_digest

wolfv:


crates/rattler_digest/src/lib.rs
```diff
@@ -49,6 +49,15 @@ pub use digest;
 use digest::{Digest, Output};
 use std::{fs::File, io::Write, path::Path};
 
+pub use md5::Md5;
+pub use sha2::Sha256;
+
+/// A type alias for the output of a SHA256 hash.
+pub type Sha256Array = sha2::digest::Output<Sha256>;
```
baszalmstra, line 56: Maybe we can name this Sha256Hash?

wolfv: Done!

baszalmstra (approved):

PR #151, opened 2023-04-03
return path and script from activation command

wolfv:
I need the computed PATH for stacking of environments. Would be curious what you think about this API @baszalmstra?


baszalmstra:
I prefer to add a result struct indtead of a tuple to add some context to the variables

wolfv:
I changed it. I think in the future we may want to have another intermediate type that just records all the "modifications" going from one env to the next and that can be "realized" into the script.

wolfv:
Also plain deactivation is something that isn't easily possible yet :)

baszalmstra (approved):

PR #157, opened 2023-04-09
add support for local files

wolfv:
Unfortunately I couldn't get both the `file` and reqwest if/else branches to return the same type. I would have liked to factor out the creation of teh "stream" of data into a function, then we could keep all interfaces.


baszalmstra:
You can use https://docs.rs/tokio-util/latest/tokio_util/either/enum.Either.html to return the same type of future or stream from an if else branch!

wolfv:
Thanks, that worked! I had to change the interface to take a URL directly, because the `IntoUrl` errors out when presented with a `file://` scheme... lmk what you think

wolfv:
unfortunately this moves the parsing of URLs one level "higher" – i think that's fine. What do you think?

PR #166, opened 2023-04-13
Add .arch function

wolfv:
Note: 

- I also had a version that wrote out all options
- We could also implement a stricter type for `Arch`

- I think we should also implement a catch-all platform (`Unknown(String)`) to facilitate adding new platforms easily

### DIFF
--- a/crates/rattler_conda_types/src/platform.rs
+++ b/crates/rattler_conda_types/src/platform.rs
@@ -30,6 +30,21 @@ pub enum Platform {
     Emscripten32,
 }
 
+/// Known architectures supported by Conda.
+#[derive(Debug, Clone, Copy, Eq, PartialEq, Hash)]
+pub enum Arch {
+    X86,
+    X86_64,
+    Aarch64,
+    ArmV6l,
+    ArmV7l,
+    Ppc64le,
+    Ppc64,
+    S390X,
+    Riscv32,
+    Riscv64,
+}
+
 impl Platform {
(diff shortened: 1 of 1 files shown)


crates/rattler_conda_types/src/platform.rs
```diff
@@ -225,6 +225,23 @@ impl From<Platform> for &'static str {
     }
 }
 
+impl Platform {
+    /// Return the arch string for the platform
+    /// The arch is usually the part after the `-` of the platform string.
+    /// Only for 32 and 64 bit platforms the arch is `x86` and `x86_64` respectively.
+    pub fn arch(&self) -> String {
```
baszalmstra, line 232: Can we do this without the unwrap? especially if we get “unknown”. Maybe return option?

wolfv: Good point, especially for `noarch` it should be undefined :)

wolfv: Yeah I do think that's good, but it also makes us have the same data in two places which is what I dislike... but I can write it out and we could add a test or something.

wolfv:
Just for the record, I've also looked at implementing a custom platform (that would be `Custom(String)` or so, but didn't yet, because `String` doesn't implement `Copy` and so we have to refactor a little more code to use `.clone()` in some places.

Or is there a copyable string somewhere that we could/should use?

baszalmstra (approved):

PR #171, opened 2023-04-16
Allow setting timestamp when writing archives

wolfv:
I cobbled something together to help build "reproducible" archives. 

This deterministically sets the timestamp of all files in an archive. That worked for a quick test of building a reproducible package :)

However, we could come up with more interesting schemes of modifying the timestamps (e.g. only forcing the timestamp on files that have been written in the past 24 hours or so, and leave those files that were just copied from the source archive as they are).

Some references: https://reproducible-builds.org/docs/timestamps/


baszalmstra (commented):
Can you also add some docs to the function to indicate what the `timestamp` argument means?

wolfv:
thx for the suggestion, done that. I was considering to add a "options" struct, but not convinced it's needed just yet.

baszalmstra:
Yeah if we add more we should do that. But lets merge as is!

PR #173, opened 2023-04-17
feat: include hash in package cache directories

baszalmstra:
This adds the hash of the package archive to the directory name of an extracted package.

Fixes: https://github.com/mamba-org/rattler/issues/165


wolfv:
I think for this one we discussed to use a different layout, right? I've kinda circled back and thought maybe it _is_ better to use the SHA hashes instead of multiple layers because in that case we could deduplicate packages that appear in multiple channels.

One issue that this doesn't solve is that if we do local development / iterations on builds you might end up with multiple e.g. `curl-8.0.1-h1234` packages which would stay in teh cache indefinitely. `conda` / `conda-build` would have just overriden the previous archive file & extracted files.

baszalmstra:
> I've kinda circled back and thought maybe it is better to use the SHA hashes instead of multiple layers because in that case we could deduplicate packages that appear in multiple channels. 

My thought exactly.

> One issue that this doesn't solve ..

Yeah, that's exactly the downside of this approach. Alternatively, we could also implement multiple strategies so that you can choose a different strategy for local development?

Or, different thought.

* We extract packages under their hash name.
* We also create a symbolic link from the package name to the hashed directory.
* If the symbolically linked directory already exists we delete the hashed folder it (previously) pointed to?

But that doesn't work on Windows because symlinks...

We can also store this mapping from "channel/subdir/packagename" to hash in an accompanying sqlite db?

baszalmstra:
But either way we need #176 first. :)

PR #175, opened 2023-04-18
feat: implement String for platform for convenience

ruben-arts:
Added this to make a change in rattler-build easier.


baszalmstra:
I dont think this implementation makes a lotnof sense. Since Platform inplements display, cant you just call to_string directly on Platform where you need to use this?

ruben-arts:
> I dont think this implementation makes a lotnof sense. Since Platform inplements display, cant you just call to_string directly on Platform where you need to use this?

your right that also works. for some reason, it didn't work at first so I started looking for a solution and this was what that resulted in.

PR #176, opened 2023-04-19
feat: compute hashes while extracting

baszalmstra:
This computes the hashes of extracted archives on the fly. I _think_ this is really fast and shouldn't add much overhead (compared to disk and network access) but we can also put it behind flags that we pass in. Or we can separate this concern into different functions. However, I also think its good practice to check the hash of the downloaded file to ensure consistency so 🤷 

Let me know what you think!

Fixes: https://github.com/mamba-org/rattler/issues/41


wolfv:
just to be sure – this computes the hash of the outer archive on the fly, correct? E.g. the hash of the `.tar.bz2` or `.conda` file?

I think that's great :) Do we take any actions when the hash doesn't match? E.g. removing the archive & extracted contents?

baszalmstra:
> just to be sure – this computes the hash of the outer archive on the fly, correct? E.g. the hash of the .tar.bz2 or .conda file?

Yes, that's exactly what it does!

> Do we take any actions when the hash doesn't match? E.g. removing the archive & extracted contents? 

No not yet, that's what the package cache should do. The extraction process itself is unaware of the expected hash.

PR #181, opened 2023-04-26
add tls features to rattler dependencies as well

wolfv:
I think this is needed.

### DIFF
--- a/crates/rattler-bin/Cargo.toml
+++ b/crates/rattler-bin/Cargo.toml
@@ -15,9 +15,9 @@ name = "rattler"
 path = "src/main.rs"
 
 [features]
-default = ['native-tls']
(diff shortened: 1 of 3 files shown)


baszalmstra:
Ah yes! I also think you need to disable default features of rattler_package_streaming in the rattler crate.

wolfv:
I now have to add it to every crate that depends on `rattler_conda_types` because that has a dev-dependency on package_streaming. So I have to add the feature to every crate more or less. 

Or am I missing something?

baszalmstra:
I dont think you need to add the feature to the dev dependencies! The tests can just run with nativetls.

wolfv:
It doesn't compile though. But actually I could force the tests to use rustls-ROS

wolfv:
Hmm, I still see some native-tls dependency coming from `tokio-native-tls` ... maybe we can fix this together on friday :)

PR #191, opened 2023-05-07
add rattler_auth to authenticate requests

wolfv:


wolfv:
@baszalmstra unfortunately I had to make two functions, one for `reqwest::blocking::Client` and one for `reqwest::Client`. Maybe you have an idea on how to improve this?

Apart from that it works like a charm :)

crates/rattler_auth/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+            _ => Err(AuthenticationParseError::InvalidScheme),
+        }
+    }
+}
+
+pub fn store_authentication_entry(host: &str, authentication: &Authentication) -> Result<()> {
+    let entry = Entry::new("rattler_auth", host)?;
+    match authentication {
+        Authentication::BearerToken(token) => {
+            let password = format!("Bearer {}", token);
```
baszalmstra, line 44: Can you implement Display? Since you also implement FromStr it makes sense to be able to convert from String and back.

crates/rattler_auth/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    match password {
+        Ok(password) => Ok(Some(Authentication::from_str(&password).unwrap())),
+        Err(keyring::Error::NoEntry) => Ok(None),
+        Err(e) => Err(e),
+    }
+}
+
+pub fn delete_authentication_entry(host: &str) -> Result<()> {
+    let entry = Entry::new("rattler_auth", host).unwrap();
```
baszalmstra, line 70: Return error instead of unwrap

crates/rattler_auth/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    entry.delete_password()
+}
+
+pub fn authenticated_request<U: IntoUrl>(client: &Client, url: U) -> reqwest::RequestBuilder {
```
baszalmstra, line 74: Maybe take a requestbuilder as input? That way the user is in charge of the entire request construction.

crates/rattler_auth/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+    let url = url.into_url().unwrap();
+    let host = url.host_str().unwrap();
+
+    let authentication = get_authentication(host).unwrap();
```
baszalmstra, line 78: Please return sensible errors instead of unwrap.

crates/rattler_auth/src/lib.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    println!("Getting authenticated request for host: {}", host);
```
baszalmstra, line 80: Please you tracing instead of println

baszalmstra (changes requested):
Nice work!

Some comments:

* authentication is currently a singleton and it cant be turned on or off. I dont like how intransparent, uncontrollable and potentially thread unsafe it is. It would be nice if you pass in an object for authentication.  There you could also specify the key for the keychain for instance. Authentication::from_keychain would be nice to have!

* Please dont use unwrap in library code unless you are sure the code cannot fail. And even then, please add a comment and/or use expect.

crates/rattler_networking/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+    #[error("Could not parse credentials stored for {host}")]
+    ParseCredentialsError {
+        /// The host for which the credentials could not be parsed
+        host: String,
+    },
+}
+
+impl AuthenticationStorage {
+    /// Store the given authentication information for the given host
+    pub fn store(&self, host: &str, authentication: &Authentication) -> keyring::Result<()> {
```
baszalmstra, line 102: Im wondering, should this be mutable?

crates/rattler_networking/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+            }
+            Authentication::CondaToken(token) => {
+                let password = format!("CondaToken {}", token);
+                entry.set_password(&password)
+            }
+        }
+    }
+
+    /// Retrieve the authentication information for the given host
+    pub fn get(&self, host: &str) -> Result<Option<Authentication>, AuthenticationStorageError> {
```
baszalmstra, line 121: I think `try_get` would be better here because it returns an option.

crates/rattler_networking/src/lib.rs
```diff
@@ -0,0 +1,409 @@
...
+
+    /// The authentication storage
+    auth_storage: AuthenticationStorage,
+}
+
+impl Default for AuthenticatedClient {
+    fn default() -> Self {
+        AuthenticatedClient {
+            client: Client::default(),
+            auth_storage: AuthenticationStorage::new("rattler"),
```
baszalmstra, line 183: I think the `Default` implementation should be "unauthenticated". Using "rattler" as a default key means you are accessing a giant global variable that is used by everything that is using rattler which seems like a huge red-flag to me.

baszalmstra (commented):
This looks pretty close to done to me! A note:

* Since you can only really use an `AuthenticatedClient` and there is no `UnauthenticatedClient` maybe we can just call it `Client`?
* I think by `Default` the "client" should be unauthenticated unless you initialize it with credentials. See my other comment.
* It would be very nice if there is also a `From<reqwest::Client>` implementation. That way the library remains super simple to use with just reqwest.
* Random thought: If there is direct conversion from `reqwest::Client` to `rattler_networking::Client` we can use `impl Into<rattler_networking::Client>` as an argument where authentication is required. That way people can keep using the `reqwest::Client` too.

wolfv:
The idea of having a default key is that multiple tools could share the same namespace for credentials (eg rattler-build and px are both logged in to the same sites).

baszalmstra:
I think thats dangerous behavior because if you set certain (invalid) credentials it affects **all** programs that use the **default** behavior. Its worst kind of a global variable you can have..

To me its also not logical that if you would do ‘px login’ it also works with ‘rattler-build publish’ (or whatever). Id much rather have a seperate ‘rattler-build login’.

Ever more so as this is a library, I think we should allow users to opt-in by specifying the same key. It should not be the default behavior.

wolfv:
Hmm but for each channel you want to use with rattler build (not for publishing, just for pulling packages from) you'd need to re-login. A lot of programs use global configurations for credentials (eg netrc files) or global proxy configurations (HTTP_PROXY environment variable etc).

One can also easily opt out of the global sharing by using a different key. 

I think it would be pretty nice if px and rattler build share the credentials (which we can achieve either way).

wolfv:
just for the record, I am working on a keyring-rs compatible fallback storage that saves the authentication to a file in the `~/.rattler_auth.json` location (trying to make that configurable as well).

It's a bit challenging because of multi-threading and file locking but hopefully can get it done soon.

crates/rattler-bin/src/commands/create.rs
```diff
@@ -93,9 +94,19 @@ pub async fn create(opt: Opt) -> anyhow::Result<()> {
...
+        .join("rattler/auth");
+
+    let authentication_storage = AuthenticationStorage::new("rattler_credentials", &auth_dir);
+
+    let download_client = AuthenticatedClient::from_client(download_client, authentication_storage);
     let multi_progress = global_multi_progress();
 
     let repodata_cache_path = cache_dir.join("repodata");
+    println!("Downloading repodata for {} channels", channel_urls.len());
+    print!("Cache folder: {}", repodata_cache_path.display());
```
baszalmstra, line 109: Left over debug statement?

crates/rattler_networking/src/authentication_storage/storage.rs
```diff
@@ -0,0 +1,183 @@
...
+/// A struct that implements storage and access of authentication
+/// information
+#[derive(Clone)]
+pub struct AuthenticationStorage {
+    /// The store_key needs to be unique per program as it is stored
+    /// in a global dictionary in the operating system
+    pub store_key: String,
+
+    /// Fallback JSON location
+    pub fallback_json_location: PathBuf,
```
baszalmstra, line 23: Why not use the `FallbackStorage` object here, instead of recreating it where you need it?

baszalmstra (approved):

baszalmstra (approved):

PR #192, opened 2023-05-18
Add `blake2b` hash to `.state.json` file in repodata cache

travishathaway:
While working on the JLAP implementation, I discovered that the `blake2` hash that `rattle_repodata_gateway` generates isn't compatible with the ones found in the `repodata.jlap` files currently being generated by the server (rattler uses `blake2s` and JLAP uses `blake2b`).

This pull request is an attempt to write `blake2b` to the `*.state.json` file too.  I chose to add a new `blake2b_hash` field to that file while also adding a couple new function to handle the hashing itself plus serialization and de-serialization.

A couple open questions I still have:

1. Should we just switch to have the `.state.json` to use `blake2b` and drop `blake2s`? or will that break things? (looks like `blake2b` is [more performant on 64-bit platforms](https://www.blake2.net/))
2. My `generate_blake2b256_hash` function relies on receiving a `PathBuf` object and then opens the file itself. Would it be better to instead just use the already open file, read the contents from it and have that be the argument?

For my JLAP implementation I also considered just calculating that hash every time, but it seems like it could speed things up by storing it in this `.state.json` file instead.

Comments and suggestions for improvements welcome.


baszalmstra:
I thinks its fine to switch the hash type. I also think rattler_digest contains a few things you implemented here or was that not sufficient.

wolfv (approved):

wolfv:
Thanks!

PR #196, opened 2023-05-25
ignore trailing .0 in version comparison

### LINKED ISSUE #194: Double check that libabseil matchspec matches a package
I thought that we had to change the parsing of matchspecs but after double checking it seems to be correct. However, we had some resolve issues with `libabseil` and constraints of the following sort:

`libabseil 20230125.0 cxx17*` should match `libabseil-20230125.0-cxx17_hb7217d7_0`.

I'll try to add a test for this

wolfv:
Not sure if this is the cleanest way of implementing this behavior, but in order to fix #194 we need to ignore trailing `0` when comparing versions for equality.

### DIFF
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -115,7 +115,7 @@ mod parse;
 /// this problem by appending an underscore to plain version numbers:
 ///
 /// 1.0.1_ < 1.0.1a =>  True   # ensure correct ordering for openssl
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, Eq)]
 pub struct Version {
     /// The epoch of this version. This is an optional number that can be used to indicate a change
     /// in the versioning scheme.
@@ -171,30 +171,9 @@ impl Version {
 
 impl PartialEq<Self> for Version {
     fn eq(&self, other: &Self) -> bool {
-        for ranges in self
-            .version
-            .components
-            .iter()
-            .zip_longest(other.version.components.iter())
-        {
-            match ranges {
-                EitherOrBoth::Both(left, right) => {
-                    if left != right {
-                        return false;
-                    }
-                }
-                EitherOrBoth::Left(el) | EitherOrBoth::Right(el) => {
-                    if el != &NumeralOrOther::Numeral(0) {
-                        return false;
-                    }
-                }
-            }
-        }
-        self.local.components == other.local.components
+        self.version == other.version && self.local == other.local
     }
 }
-
-impl Eq for Version {}
 
 impl Hash for Version {
     fn hash<H: Hasher>(&self, state: &mut H) {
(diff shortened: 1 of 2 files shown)


crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -171,7 +171,26 @@ impl Version {
 
 impl PartialEq<Self> for Version {
     fn eq(&self, other: &Self) -> bool {
```
baszalmstra, line 173: Could you inplement this as the Eq trait of VersionComponent?

baszalmstra (changes requested):
Can this code also be found somewhere in the conda codebase? I followed that pretty closely.

baszalmstra (changes requested):
Can this code also be found somewhere in the conda codebase? I followed that pretty closely.

wolfv:
I can try to find the place in the conda code but this was an issue i observed while debugging resolvelib stuff

wolfv:
We actually did copy the docs for it! :) https://github.com/mamba-org/rattler/blob/001f326c7c3160637c8778c5a1c33f715266434e/crates/rattler_conda_types/src/version/mod.rs#L71-L72

wolfv:
And this is the code in conda: 


```python
    def _eq(self, t1, t2):
        for v1, v2 in zip_longest(t1, t2, fillvalue=[]):
            for c1, c2 in zip_longest(v1, v2, fillvalue=self.fillvalue):
                if c1 != c2:
                    return False
        return True
```

https://github.com/conda/conda/blob/61edfac743a2461a24fafe48b821a99af61b6bd4/conda/models/version.py#L244-L249

wolfv:
Btw. I think we're also missing an compare of the `epoch` of the two versions.

wolfv:
I am confused by the follownig:

```
    /// The epoch of this version. This is an optional number that can be used to indicate a change
    /// in the versioning scheme.
    norm: String,
```

The norm doesn't seem to be the epoch but rather the string representation of the version?

wolfv:
E.g. this is what I see in debug print:

```
Version { norm: "1.2.0", version: [[0], [1], [2], [0]], local: [] }
Version { norm: "1!1.2.3", version: [[1], [1], [2], [3]], local: [] }
```

wolfv:
I attempted what you asked but there I am a bit unsure about a couple things. 

I had to add `#[derive(Eq)]` to the `Version` for some reason.

Also I had to implement a custom hash function. I am now removing all default values (0) from the hash. Clippy was warning that the hash function needs to match the `PartialEq` function.

baszalmstra:
This looks good to me! The warning about the hash is also very correct! Very nice!

The `norm` is a bit of an odd duck. It represents "sorta" the string it was initially created from. This can be important because (as you've seen with this issue) multiple version representations can represent the exact same version. The `norm` sorta enables keeping track of that. However, modifications to the `Version` are also not expressed in the `norm` which is actually just plain wrong. Maybe we should remove it and implement a proper `Display` for `Version`.

baszalmstra (approved):

Issue #81, opened 2023-02-28
Implement the `jlap` format


wolfv:
It would be nice to implement the `jlap` format for partial repodata updates (using a specific sequenece of JSON patch applications).

That makes interactive usage faster because only a subset of the entire data needs to be fetched over the network.

dholth:
This is the predecessor to the jlap forma in rust https://github.com/dholth/jdiff/. It is basically the same except that there is an array of patches instead of a line-based format.

It turns out Python has very fast json support (generating the patches on pypi was faster), and this was put aside.

wolfv:
Just for reference, this was the prototype I made in C++: https://github.com/mamba-org/mamba/pull/2029/files

PR #197, opened 2023-05-29
JLAP support

travishathaway:
Fixes: #81 

This pull request adds support for the [JLAP format](https://github.com/conda-incubator/ceps/pull/20/files). The new `rattler_repodata_gateway::fetch::jlap` module contains a `patch_repo_data` function that allows you to patch arbitrary repodata.json files. To work, this function just needs a copy of `RepoDataState` an HTTP `client`, a `subdir_url` and the path where `repo_data` is.

Here's an example of how it's meant to be used:

```rust
use std::{path::Path};
use reqwest::Client;
use url::Url;

use rattler_repodata_gateway::fetch::jlap::{patch_repo_data, RepoDataState};

#[tokio::main]
pub async fn main() {
    let repodata_url = Url::parse("https://conda.anaconda.org/conda-forge/osx-64/").unwrap();
    let client = Client::new();
    let cache = Path::new("./cache");
    let current_repo_data = cache.join("c93ef9c9.json");

    let repo_data_state: RepoDataState =  serde_json::from_str(r#"{
       "url": "https://conda.anaconda.org/conda-forge/osx-64/repodata.json.zst",
       "etag": "W/\"49aa6d9ea6f3285efe657780a7c8cd58\"",
       "mod": "Tue, 30 May 2023 20:03:48 GMT",
       "cache_control": "public, max-age=30",
       "mtime_ns": 1685509481332236078,
       "size": 38317593,
       "blake2_hash": "580100cb35459305eaaa31feeebacb06aad6422257754226d832e504666fc1c6",
       "has_zst": {
         "value": true,
         "last_checked": "2023-05-21T12:14:21.904003Z"
       },
       "has_bz2": {
         "value": true,
         "last_checked": "2023-05-21T12:14:21.904003Z"
       },
       "has_jlap": {
         "value": true,
         "last_checked": "2023-05-21T12:14:21.903512Z"
       },
       "jlap": {
         "iv": "0000000000000000000000000000000000000000000000000000000000000000",
         "pos": 0,
         "footer": {
           "url": "repodata.json",
           "latest": "580100cb35459305eaaa31feeebacb06aad6422257754226d832e504666fc1c6"
         }
       }
     }"#).unwrap();

    // Patches `current_repo_data` and returns an updated JLAP state object
    let updated_jlap_state = patch_repo_data(
        &client,
        subdir_url,
        repo_data_state,
        &current_repo_data
    ).await.unwrap();
```

This pull request is still a draft, but I welcome any comments or suggestions for improvements already as I believe this is pretty close to being complete.

I still need to figure out how exactly this is going to be integrated into the rest of rattler.


crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    /// version from the server.
+    HashesNotMatchingError,
+}
+
+/// Represents the numerous patches found in a JLAP file which makes up a majority
+/// of the file
+#[derive(Serialize, Deserialize, Debug)]
+pub struct Patch {
+    /// Next hash of `repodata.json` file
+    pub to: String,
```
baszalmstra, line 121: It would be nice if these types are the hash types.

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+
+    /// Patches to apply to current `repodata.json` file
+    pub patch: json_patch::Patch, // [] is a valid, empty patch
+}
+
+/// Represents the metadata for a JLAP file, which is typically found at the very end
+#[derive(Serialize, Deserialize, Debug)]
+pub struct JLAPMetadata {
+    /// URL of the `repodata.json` file
+    pub url: String,
```
baszalmstra, line 134: You can use the `Url` type here.

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    cache_path.join(cache_file_name)
+}
+
+/// Persist a JLAP file to the provided location
+pub async fn cache_jlap_response(
+    jlap_cache_path: &PathBuf,
+    response_text: &str,
+) -> Result<(), tokio::io::Error> {
+    let mut jlap_file = tokio::fs::File::create(&jlap_cache_path).await?;
+    jlap_file.write_all(response_text.as_bytes()).await?;
```
baszalmstra, line 295: You can use https://docs.rs/tokio/latest/tokio/fs/fn.write.html instead.

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+
+    Ok(())
+}
+
+/// Update an existing cached JLAP file
+pub async fn update_jlap_file(jlap_file: &PathBuf, jlap_contents: &str) -> Result<(), JLAPError> {
```
baszalmstra, line 301: Use the borrowed type `&Path` instead of the owned `&PathBuf`

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+/// Determines the byte offset to use for JLAP range requests
+///
+/// This function assumes we already have a locally cached version of the JLAP file
+pub async fn get_jlap_request_range(
+    jlap_cache: &PathBuf,
+) -> Result<Option<String>, tokio::io::Error> {
+    let mut cache_file = tokio::fs::File::open(jlap_cache).await?;
+    let mut contents = String::from("");
+
+    cache_file.read_to_string(&mut contents).await?;
```
baszalmstra, line 351: You can use https://docs.rs/tokio/latest/tokio/fs/fn.read_to_string.html

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+    let mut repo_data = match tokio::fs::File::open(repo_data_path).await {
+        Ok(contents) => contents,
+        Err(error) => return Err(JLAPError::FileSystemError(error)),
+    };
+
+    let mut repo_data_contents = String::new();
+    match repo_data.read_to_string(&mut repo_data_contents).await {
+        Ok(_) => (),
+        Err(error) => return Err(JLAPError::FileSystemError(error)),
+    }
```
baszalmstra, line 439: You can use https://docs.rs/tokio/latest/tokio/fs/fn.read_to_string.html

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,800 @@
...
+/// 2. Applying patches to this repodata file
+/// 3. Saving this repodata file to disk
+/// 4. Generating a new `blake2b` hash
+///
+/// The return value is the string representation of the `blake2b` hash that
+/// can be verified against our `Metadata` object.
+pub async fn apply_jlap_patches(
+    patches: &Vec<&Patch>,
+    repo_data_path: &PathBuf,
+) -> Result<String, JLAPError> {
```
baszalmstra, line 428: Would be nice to return the typed hash.

baszalmstra (commented):
Hey @travishathaway! Thanks for this awesome work! I didn't do a thorough review yet but here are some of my initial thoughts:

I dislike the 'Manager' name because it doesn't really tell me what it does. I'm also not really sure whether an object is really needed. Won't a simple function suffice? Otherwise, maybe  `PatchContext`, `State`, `PatchDriver` might be a better name?

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
travishathaway: I tried this but these aren't actually URLs 🙃. It's just the filename (e.g. `repodata.json` or `current_repodata.json`) 🤷‍♂️

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+
+/// Finds the initial vector in the JLAP response; this is the first line of a full request
+fn get_initial_vector(jlap_response: &str) -> Option<String> {
+    let parts: Vec<&str> = jlap_response.split('\n').collect();
+
+    if !parts.is_empty() {
+        Some(parts[0].to_string())
+    } else {
+        None
+    }
```
baszalmstra, line 381: This avoids an extra allocation:

```suggestion
    jlap_response.split('\n').next().map(Into::into)   
```

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+
+/// Finds the index of the of the most applicable patch to use
+fn find_current_patch_index(patches: &[Patch], hash: Output<Blake2b256>) -> Option<usize> {
+    for (idx, patch) in patches.iter().enumerate() {
+        if hash == patch.from.unwrap_or_default() {
+            return Some(idx);
+        }
+    }
+
+    None
```
baszalmstra, line 370:
```suggestion
    patches.into_iter().position(|patch| patch.from == Some(hash))
```

crates/rattler_repodata_gateway/src/fetch/cache/mod.rs
```diff
@@ -122,7 +151,7 @@ fn duration_to_nanos<S: Serializer>(time: &SystemTime, s: S) -> Result<S::Ok, S:
         .serialize(s)
 }
 
-fn deserialize_blake2_hash<'de, D>(
+pub fn deserialize_blake2_hash<'de, D>(
```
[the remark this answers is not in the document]
baszalmstra: Actually there already is. `rattler_digest` provides `SerializableHash<T>` that you can use for serializating hashes as strings. See https://github.com/mamba-org/rattler/blob/e904e5f48e06beafbd8698c2626f01fa975cfef2/crates/rattler_conda_types/src/prefix_record.rs#L75 for an example. This does require the use of the `serde_as` crate.

baszalmstra (commented):
Apart from the TODO's and some comments I left this looks good to me! Fantastic! 🥳

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
travishathaway: Will do 👍

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
travishathaway: Nice. Didn't know that method existed 👌

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,711 @@
...
+    if length >= JLAP_FOOTER_OFFSET {
+        lines.truncate(length - JLAP_FOOTER_OFFSET);
+        let patches = lines.join("\n");
+        return patches.into_bytes().len();
+    }
+
+    0
+}
+
+fn parse_patch_json(line: &&str) -> Result<Patch, JLAPError> {
```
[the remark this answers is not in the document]
travishathaway: Okay. That's probably the more "idiomatic" thing to do. Will investigate.

crates/rattler_repodata_gateway/src/fetch/cache/mod.rs
```diff
@@ -80,6 +79,36 @@ impl FromStr for RepoDataState {
...
+    pub url: Option<String>,
+
+    /// blake2b hash of the latest `repodata.json` file
+    #[serde(
+        default,
+        skip_serializing_if = "Option::is_none",
+        deserialize_with = "deserialize_blake2_hash",
+        serialize_with = "serialize_blake2_hash"
+    )]
+    pub latest: Option<blake2::digest::Output<Blake2b256>>,
```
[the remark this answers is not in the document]
travishathaway: Technically, yes. I did this so that `JLAPState::default()` would work, so we could generate an "empty" JLAPState. When a JLAP request fails, for example when the range request fails or it can't find the correct starting point for patches to apply, we need a way to reset the cached JLAP values.

Now that I think about it though, I think we could actually just set the `jlap` value in cache to None instead. That would be a better more explicit (i.e. we are purposely store nothing here versus a default `JLAPState`).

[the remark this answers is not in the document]
travishathaway: This is where I could make those changes:

https://github.com/travishathaway/rattler/blob/6e8669cbd35a327cb1781f56f6cf4dd7379a5dad/crates/rattler_repodata_gateway/src/fetch/mod.rs#L376-L380

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,913 @@
...
+                Some(value) => {
+                    iv = Some(blake2b_256_hash_with_key(line.as_bytes(), &value));
+                }
+                None => {
+                    iv = Some(blake2b_256_hash_with_key(
+                        line.as_bytes(),
+                        &self.initialization_vector,
+                    ));
+                }
+            }
```
baszalmstra, line 343:
```suggestion
            iv = Some(blake2b_256_hash_with_key(
                line.as_bytes(),
                iv.as_deref().unwrap_or(&self.initialization_vector),
            ));
            iv_values.push(iv.unwrap());
```

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,913 @@
...
+        if !iv_values.is_empty() {
+            let new_iv = iv_values[iv_values.len() - 1];
+            if new_iv != self.checksum {
+                return Err(JLAPError::ChecksumMismatch);
+            }
+
+            Ok(new_iv)
+        } else {
+            Err(JLAPError::NoPatchesFound)
+        }
```
baszalmstra, line 356: Indexing vectors is a bit unsafe. Although it doesn't result in undefined behavior it does panic if you index incorrectly. To that end I always prefer to use more semantic means:

```suggestion
        if let Some(new_iv) = iv_values.pop() {
            if new_iv != self.checksum {
                return Err(JLAPError::ChecksumMismatch);
            }
            Ok(new_iv)
        } else {
            Err(JLAPError::NoPatchesFound)
        }
```

baszalmstra (commented):
Left a view more comments. Could you also have a look at my previous comments?

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,913 @@
...
+    /// Patches to apply to current `repodata.json` file
+    pub patch: json_patch::Patch, // [] is a valid, empty patch
+}
+
+/// Represents a single JLAP response
+///
+/// All of the data contained in this struct is everything we can determine from the
+/// JLAP response itself.
+#[derive(Debug)]
+pub struct JLAP<'a> {
```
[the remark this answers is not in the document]
travishathaway: Yes, that would make it more clear.

crates/rattler_repodata_gateway/src/lib.rs
```diff
@@ -25,7 +25,7 @@
 //! Below is a basic example that shows how to retrieve and cache the repodata for a conda channel
 //! using the [`fetch::fetch_repo_data`] function:
 //!
-//! ```rust
+//! ```no_run
```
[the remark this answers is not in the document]
travishathaway: The example will fail if run because the `current_repo_data` file does not exist. Not really sure how to pull that off without including a ton of setup code. There are options to hide certain lines if I'm not mistaken. So maybe that would work.

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
travishathaway: Coming from Python, I miss not being able to do:

```python
something = [1 ,2 ,3]
print(something[-1]) # print the last element
```

😅 

Thanks for the suggested improvement.

travishathaway:
I think this pull request is a good to be merged now. The only thing that could be improved is the addition of more testing scenarios. I still need to account for several error states (e.g. bad JLAP response and `RANGE_NOT_SATISFIABLE` error).

I did create quite a few error states for JLAP that we basically ignore for now in `fetch_repo_data` 😅. This isn't a big deal in my opinion and will hopefully help future maintainers debug this code when/if bugs are found. To that end, it also might be nice to add some error logging that could be optionally enabled. Perhaps you all could advise on how best to do that.
 
Some good next steps would be to set up some profiling to see what kind of speed boost this method actually gives us 🐇. I might try to do this by writing a simple rust application that fetches a repodata.json files.

@baszalmstra thanks for all the great suggestions during the review. I learned a lot! 😎

travishathaway:
This pull request still needs one important change! 😅

Right now, in `rattler_repodata_gateway::fetch::fetch_repo_data` it won't actually exit early if we successfully made a JLAP patch request. This is obviously not ideal because it will just go on to retrieve the full repodata again!

I will take a look at this later today. It shouldn't be too hard to fix.

crates/rattler_repodata_gateway/src/lib.rs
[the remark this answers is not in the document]
baszalmstra: Im not sure I understand why this would fail? Its invoking `fetch::fetch_repo_data` with default arguments. Why does that fail all of a sudden?

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
```diff
@@ -0,0 +1,853 @@
...
+//! that conda compatible applications use to query conda packages. For more information about
+//! how this file format works, please read this CEP proposal:
+//!
+//! - <https://github.com/conda-incubator/ceps/pull/20/files>
+//!
+//! ## Example
+//!
+//! The recommended way to use this module is by using the JLAPManager struct. This struct is meant
+//! to act as a kind of "facade" object which orchestrates the underlying operations necessary
+//! to fetch JLAP data used to update our current `repodata.json` file.
```
baszalmstra, line 16: I think this comment is still out of date.

baszalmstra (commented):
Two final comments! But this looks amazing! :)

crates/rattler_repodata_gateway/src/lib.rs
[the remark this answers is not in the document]
travishathaway: I thought this was in reference to the example in the `jlap/mod.rs` file (wasn't looking at the filename 😂).

The reason I changed that example is that the code sample makes an external request for a file that's pretty large. This slows down testing time, so I figured it would be good to just compile and not actually run it.

Do you still think it's a good idea to revert?

crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs
travishathaway: heh, good catch! Will fix.

travishathaway:
I did a little more work on this today. There were a lot of things I was able to uncover by running tests with real repodata. The tests basically consisted of:

- Starting with an repodata old state (e.g. 2 hours old)
- Running the `fetch_repo_data` function with the following command (`cargo run --release --bin rattler create jupyterlab`)
- Inspecting the `.state.json` file to make sure everything was okay; also observe that no JLAP warnings were given.
- Re-run the same command again to make sure that JLAP was able to be run for a second time and produced the desired effects (i.e. do nothing and give no warnings).
- Re-run it once more to observe the same thing.

A couple things came up during this testing. See this commit for the changes that had to be made: https://github.com/mamba-org/rattler/pull/197/commits/4ab1e6e0a7b0749a42fc5e6f0e362ccf7e5eadb1

Feeling better and better 😅. The only warnings I'm still seeing are these:

```
 WARN rattler_repodata_gateway::fetch: BLAKE2 hash of repodata.json does not match cache state. Ignoring cached files...
```

This probably has to do with how we save the file. In conda, this problem isn't dealt with and is why two versions of the blake2 hash are saved:

https://github.com/conda/conda/blob/b8e356e84ca8fab5e64f03f51069334beccd2f53/conda/gateways/repodata/jlap/fetch.py#L427-L431

I still want to see if it's possible to get these files to line up exactly. It would give me better peace of mind when releasing this feature.

travishathaway:
I did a little more work on this recently. I added some more tests to basically test the scenario I mentioned here: https://github.com/mamba-org/rattler/pull/197#issuecomment-1584319874

I also solved the "blake2" mismatch issue. To solve this, you just have to sort the new repo data before saving it. To do this I created my own `OrderedRepoData` struct (https://github.com/travishathaway/rattler/blob/ec0005e45669f1eab5ba2139c30519cdf8d0be40/crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs#L457-L470) I later realized that `rattler_conda_types::RepoData` exists, and I am trying to use it now, but I have been running into issues using it:

```
thread 'fetch::jlap::test::test_patch_repo_data::case_2_patch_repo_data_with_a_partial_jlap_response_and_a_previous_jlap_cache_state' panicked at 'called `Result::unwrap()` on an `Err` value: JSONParse(Error("invalid type: string \"f284fea068c51b1a0eaea3ac58c300c0\", expected a borrowed string", line: 0, column: 0))', crates/rattler_repodata_gateway/src/fetch/jlap/mod.rs:977:10
```

This comes from trying to convert it from a `serde_json::Value` object:

```rust
    let ordered_doc: RepoData = match serde_json::from_value(doc) {
        Ok(value) => value,
        Err(error) => {
            return Err(JLAPError::JSONParse(error));
        }
    };
```

Not sure what's going on here. Is it because the `MD5` field is expected a string reference? It might be worth just keeping the `OrderedRepoData` struct I created if this doesn't work. It also might be a lower cost serialization because we aren't worrying about parsing and verifying of the special fields (e.g. the SHA256 and MD5 fields).

wolfv:
Yes, the `repodata.json` generated by the Repodata struct should be sorted as well (rattler even requires it for some of the sparse parsing logic).

The error reads as if a `&str` should be a `String` or vice versa. I'll have a quick look at the code. Did you push the failing state?

Good to hear that you got rid of the blake2 mismatch!

baszalmstra:
I think the `OrderedRepoData` struct is good enough because it's probably much faster since we don't run any specific deserialization code. 👍

baszalmstra:
Thanks very much for your awesome work @travishathaway !

PR #201, opened 2023-06-02
fix: refactor ci to enable building for different targets

baszalmstra:
Refactors the CI stages to build and test for multiple (cross-platform) targets.


wolfv:
I am a bit hesitant if there is really so much utility in building for all these targets. Mainly because CI time is precious :)

PR #206, opened 2023-06-06
refactor: version tree parsing with nom

### LINKED ISSUE #204: Invalidate `>=3.8<3.9`
What I meant to write was `>=3.8,<3.9` but missing the `,` gives you really weird behavior as it only uses the `>=3.8`.

Conda invalidates this request, as should rattler.

baszalmstra:
Refactors parsing of version trees by replacing the regex parser with nom. Hopefully, we can reuse these parsers in other parts of the library as well.

Fix #204

### DIFF
--- a/crates/rattler_conda_types/src/version_spec/version_tree.rs
+++ b/crates/rattler_conda_types/src/version_spec/version_tree.rs
@@ -357,6 +357,12 @@ mod tests {
             parse_version_epoch::<Err>("1!1.0b2.post345.dev456"),
             Ok(("1.0b2.post345.dev456", 1))
         );
+
+        // Epochs must be integers
+        assert!(
+            parse_version_epoch::<Err>("12.23!1").is_err(),
+            "epochs should only be integers"
+        );
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


crates/rattler_conda_types/src/version_spec/version_tree.rs
```diff
@@ -198,4 +228,158 @@ mod tests {
...
+        );
+        assert_eq!(
+            recognize_constraint::<Err>(">=3.8.*,<3.9"),
+            Ok((",<3.9", ">=3.8.*"))
+        );
+    }
+
+    #[test]
+    fn issue_204() {
+        assert!(VersionTree::try_from(">=3.8<3.9").is_err());
```
wolfv, line 383: Maybe also do a test for a invalid epoch? E.g. `12.23!123.321` shouldn't work (afaik epochs can only be simple integers).

baszalmstra: Done!

wolfv (approved):

PR #218, opened 2023-06-16
feat: topological sort of PackageRecords

baszalmstra:
Adds topological sorting of `PackageRecord`s from [`rattler-server`](https://github.com/mamba-org/rattler-server). This implementation differs slightly in that it allows any `Vec` of something that resembles a `PackageRecord` to be sorted. This includes `RepoDataRecord` and `PrefixRecord`.

### DIFF
--- a/crates/rattler_conda_types/src/repo_data/topological_sort.rs
+++ b/crates/rattler_conda_types/src/repo_data/topological_sort.rs
@@ -4,7 +4,9 @@ use fxhash::{FxHashMap, FxHashSet};
 /// Sorts the packages topologically
 ///
 /// This function is deterministic, meaning that it will return the same result regardless of the
-/// order of `packages` and of the `depends` vector inside the records
+/// order of `packages` and of the `depends` vector inside the records.
+///
+/// Note that this function only works for packages with unique names.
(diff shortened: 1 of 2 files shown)


wolfv (approved):

crates/rattler_conda_types/src/repo_data/topological_sort.rs
```diff
@@ -0,0 +1,1623 @@
...
+    let mut all_packages = packages
+        .into_iter()
+        .map(|p| (p.as_ref().name.clone(), p))
+        .collect();
+
+    get_topological_order(roots, &mut all_packages)
+}
+
+/// Retrieves the names of the packages that form the roots of the graph
+fn get_graph_roots<T: AsRef<PackageRecord>>(records: &[T]) -> Vec<String> {
```
wolfv, line 20: I think this function might fall apart if we have multiple package records with the same name in the vector, right?

Since version 0 might be at the root, but if version 1 adds some dependencies it might not be at the root anymore. 

We might want to clarify that this works only for lists of unique package names.

wolfv (approved):

crates/rattler_conda_types/src/repo_data/topological_sort.rs
baszalmstra: Good one, I added this to the docs.

PR #222, opened 2023-06-20
feat: better parse error message

baszalmstra:
Changed the error message for parsing platforms to also include available values:

```
'foo' is not a known platform.
```

changed to:

```
'foo' is not a known platform. Valid platforms are 'noarch', 'linux-32', 'linux-64', 'linux-aarch64', 'linux-armv6l', 'linux-armv7l', 'linux-ppc64le', 'linux-ppc64', 'linux-s390x', 'linux-riscv32', 'linux-riscv64', 'osx-64', 'osx-arm64', 'win-32', 'win-64', 'win-arm64', 'emscripten-32'
```

### DIFF
--- a/crates/rattler_conda_types/src/platform.rs
+++ b/crates/rattler_conda_types/src/platform.rs
@@ -1,5 +1,7 @@
+use itertools::Itertools;
 use serde::{Deserializer, Serializer};
 use std::cmp::Ordering;
+use std::fmt::Display;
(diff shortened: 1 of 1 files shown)


wolfv:
I still think we should have the option for completely custom `platform` values (they should follow the `{platform}-{arch}` schema though.

wolfv (approved):

PR #223, opened 2023-06-20
zsh also uses `sh` as shell extension

wolfv:
This should fix activation for `zsh` shells (didn't find and execute any activation scripts).


baszalmstra (approved):

baszalmstra:
What happens with powershell though? That will only find `ps1` files right?

wolfv:
afaik powershell isn't able to execute `cmd.exe` (.bat) files and the syntax is wildly different.

PR #224, opened 2023-06-21
add executable to the shell trait and implement it

ruben-arts:
Add `executable` so it can be used to find the command that needs to be run.

### DIFF
--- a/crates/rattler_shell/src/shell/mod.rs
+++ b/crates/rattler_shell/src/shell/mod.rs
@@ -54,6 +54,9 @@ pub trait Shell {
     /// The extension that shell scripts for this interpreter usually use.
     fn extension(&self) -> &str;
 
+    /// The executable that can be called to start this shell.
+    fn executable(&self) -> &str;
+
     /// Constructs a [`Command`] that will execute the specified script by this shell.
     fn create_run_script_command(&self, path: &Path) -> Command;
 }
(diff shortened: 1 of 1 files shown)


crates/rattler_shell/src/shell/mod.rs
```diff
@@ -244,9 +263,12 @@ impl Shell for PowerShell {
     fn extension(&self) -> &str {
         "ps1"
     }
+    fn executable(&self) -> &str {
+        "powershell"
```
wolfv, line 267: pwsh[.exe] is the executable file name of PowerShell [Core] (v6+), the cross-platform edition of PowerShell built on .NET Core / .NET 5+; by contrast, powershell.exe is the executable name of the legacy Windows PowerShell edition (v5.1-), built on the Windows-only .NET Framework (v4.8-)

wolfv: Should we use `pwsh`? cc @baszalmstra

wolfv (approved):

wolfv:
The other option would be that inside the `executable` function we execute `which` and check wether `pwsh` or `powershell` is available on the PATH.

wolfv:
https://github.com/harryfei/which-rs something like this

baszalmstra:
what does micromamba do?

wolfv:
i don't think we have soemthing like thsi for micromamba. We'd use whatever the user supplies (or parent process) for the shell. 

For initialization we iterate through `pwsh` and `powershell` and initialize both.

wolfv (approved):

PR #227, opened 2023-06-21
refactor: version implementation

### LINKED ISSUE #225: `~=2.4,<4` is evaluated wrongly
The MatchSpec currently also matches `3.1` or other versions. In fact, the `<4` should be completely ignored, and the matchspec should match only `2.4.*`.

baszalmstra:
This PR does a complete refactor of the version implementation. The primary goal was to reduce the memory footprint because versions are used in RepoData quite a lot. The secondary goal was to simplify the implementation, I think I achieved this by returning an iterator of the version segments vs accessing multiple arrays to get to the data. The implementation of the `flag` makes it a bit more complex though. If based on a review this is the case I have some ideas.

Just like the previous implementation, this implementation tries to keep most memory on the stack unless the version requires more memory. Most versions are in the form of `1.2.3` so we can take advantage of that. The new implementation takes ~~136~~ 128 bytes of stack memory per Version. The old one was 440. Changes in heap allocations remain the same. More research is needed to figure out how to further reduce this.

I will follow up this PR with another one that refactors the parsing to make it more robust.

Ideally, I also wanted to add some benchmarking but didn't get the time to. Perhaps that's for later!

Oh yeah, this implementation also fixes #255. With the new implementation, it was much simpler to implement dropping a segment for comparison without allocation.

Fix #225

### DIFF
--- a/crates/rattler/src/install/python.rs
+++ b/crates/rattler/src/install/python.rs
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 #[derive(Debug, Clone)]
 pub struct PythonInfo {
     /// The major and minor version
-    pub short_version: (u32, u32),
+    pub short_version: (u64, u64),
 
     /// The relative path to the python executable
     pub path: PathBuf,
--- a/crates/rattler/src/install/python.rs
+++ b/crates/rattler/src/install/python.rs
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 #[derive(Debug, Clone)]
 pub struct PythonInfo {
     /// The major and minor version
-    pub short_version: (usize, usize),
+    pub short_version: (u32, u32),
 
     /// The relative path to the python executable
     pub path: PathBuf,
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -1,11 +1,12 @@
 use std::hash::{Hash, Hasher};
 use std::{
     cmp::Ordering,
+    fmt,
     fmt::{Debug, Display, Formatter},
-    ops::Range,
+    iter,
 };
 
-use itertools::{EitherOrBoth, Itertools};
+use itertools::{Either, EitherOrBoth, Itertools};
 use serde::{Deserialize, Serialize, Serializer};
 use smallvec::SmallVec;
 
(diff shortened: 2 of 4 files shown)


crates/rattler/src/install/python.rs
```diff
@@ -7,7 +7,7 @@ use std::path::{Path, PathBuf};
 #[derive(Debug, Clone)]
 pub struct PythonInfo {
     /// The major and minor version
-    pub short_version: (usize, usize),
+    pub short_version: (u64, u64),
```
wolfv, line 10: I want to live to see Python 18446744073709551615 :D

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -115,306 +116,391 @@ mod parse;
...
-    /// The version of this version. This is the actual version string, including the epoch.
-    version: VersionComponent,
-    /// The local version of this version (everything following a `+`).
-    /// This is an optional string that can be used to indicate a local version of a package.
-    local: VersionComponent,
+    norm: Box<str>,
+
+    /// Individual components of the version.
+    ///
+    /// We store a maximum of 3 components on the stack. If a version consists of more components
```
wolfv, line 128: I think the last sentence of the comment isn't finished?

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -115,306 +116,391 @@ mod parse;
...
+        (self.flags & 0x1) != 0
+    }
+
+    /// Returns true if this version has a local version defined
+    pub fn has_local(&self) -> bool {
+        (self.flags >> 1) > 0
+    }
+
+    /// Returns the index of the first segment that belongs to the local version or `None` if there
+    /// is not local version
```
wolfv, line 153: is _no_ local version

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -634,17 +720,17 @@ mod test {
...
-        assert_eq!(
-            Version::from_str("1.1").unwrap().bump(),
-            Version::from_str("1.2").unwrap()
-        );
-        assert_eq!(
-            Version::from_str("1.1l").unwrap().bump(),
-            Version::from_str("1.2l").unwrap()
-        )
-    }
+    // #[test]
```
wolfv, line 723: Are we ready to re-enable this test? Or do we not need that function anymore?

wolfv: Btw. in `rattler-build` we have some functions like `pin_compatible(otherpkg, min_pin="x.x", max_pin="x.x")` which would resolve to `>=1.4,<1.5`, so being able to "bump" versions is nice.

crates/rattler/src/install/python.rs
baszalmstra: Haha yeah, it's what's stored in a Version because you can have: 202306221907 (year month day hour minute) which is bigger than the largest u32.

wolfv (approved):

PR #230, opened 2023-06-23
expose `sort_topologically` function

wolfv:
I wanted to use this function in `rattler-server` but could not (or did I do something wrong?) :)


baszalmstra:
You should be able to use [`PackageRecord::sort_topologically`](https://docs.rs/rattler_conda_types/0.4.0/rattler_conda_types/struct.PackageRecord.html#method.sort_topologically)

PR #232, opened 2023-06-26
feat: Add path variable modification for the shells

ruben-arts:
Allow for specifying what the path modification will be when activating an environment.


baszalmstra (approved):
Nice! Could you also add a snapshot for powershell?

ruben-arts:
related: https://github.com/prefix-dev/pixi/issues/133

PR #238, opened 2023-06-30
propose a repodata patch function to be able to add `pip` to the `pyt…

wolfv:
…hon` records


baszalmstra (approved):

crates/rattler_conda_types/src/repo_data/topological_sort.rs
```diff
@@ -7,19 +7,48 @@ use fxhash::{FxHashMap, FxHashSet};
...
+            if let Some(cycle) = find_cycles(root, &all_packages, &mut visited, &mut stack) {
+                cycles.push(cycle);
+            }
+        }
+    }
+
+    // print all cycles
+    for cycle in &cycles {
+        tracing::debug!("Found cycle: {:?}", cycle);
+    }
```
baszalmstra, line 35: Is this still useful to leave in?

baszalmstra (approved):
If you add some more docs it look good to me!

wolfv:
We decided that we also want to keep the order of packages when creating the transaction, so I'll have to rework that as well ...

baszalmstra (approved):

PR #241, opened 2023-07-02
Add support for sha256 and md5 field in matchspec

### LINKED ISSUE #207: Add support for sha256 and md5 field in matchspec
Support for matching package records by hash is missing. `MatchSpec` does not include these fields and parsing it should also be added.

This should work:

```
python[sha256=deadbeef, md5=something]
```

0xbe7a:
Closes #207

Currently, `MatchSpec` and `NamelessMatchSpec` share a considerable amount of logic and fields, with the former essentially being the latter plus a name field.

A future PR should look into merging these two classes to reduce redundancy, streamline the codebase and improve maintainability.

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -343,3 +343,52 @@ impl MatchSpec {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use std::str::FromStr;
+
+    use crate::{
+        package::{IndexJson, PackageFile},
+        MatchSpec, PackageRecord,
+    };
+
+    #[test]
+    fn test_digest_match() {
+        let package_dir = tempfile::tempdir().unwrap();
+        let extract_result = rattler_package_streaming::fs::extract(
+            &crate::get_test_data_dir().join("mamba-1.0.0-py38hecfeebb_2.tar.bz2"),
+            package_dir.path(),
+        )
+        .unwrap();
+
+        let index = IndexJson::from_package_directory(package_dir.path()).unwrap();
+        let package_record = PackageRecord::from_index_json(
+            index,
+            None,
+            Some(extract_result.sha256),
+            Some(extract_result.md5),
+        )
+        .unwrap();
+
+        let spec = MatchSpec::from_str("mamba[version==1.0, sha256=aaac4bc9c6916ecc0e33137431645b029ade22190c7144eead61446dcbcc6f97]").unwrap();
+        assert!(!spec.matches(&package_record));
+
+        let spec = MatchSpec::from_str("mamba[version==1.0, sha256=f44c4bc9c6916ecc0e33137431645b029ade22190c7144eead61446dcbcc6f97]").unwrap();
+        assert!(spec.matches(&package_record));
+
+        let spec = MatchSpec::from_str("mamba[version==1.0, md5=aaaa6252c964db3f3e41c7d30d07f6bf]")
+            .unwrap();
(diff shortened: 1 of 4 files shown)


crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -274,6 +338,57 @@ impl MatchSpec {
...
+        .unwrap();
+
+        let index = IndexJson::from_package_directory(package_dir.path()).unwrap();
+        let package_record = PackageRecord::from_index_json(
+            index,
+            None,
+            Some(extract_result.sha256),
+            Some(extract_result.md5),
+        )
+        .unwrap();
```
baszalmstra, line 372: Instead of completely extracting a package and reading the `index.json` to get a single `PackageRecord` I would prefer to construct a fake `PackageRecord` in place. This speeds up the test considerably.

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -146,19 +154,26 @@ impl Display for MatchSpec {
...
         }
 
-        match &self.build {
-            Some(build) => write!(f, " {build}")?,
-            None => (),
+        if let Some(build) = &self.build {
+            write!(f, " {}", build)?;
+        }
+
+        if let Some(md5) = &self.md5 {
```
baszalmstra, line 171: This doesn't really produce a proper matchspec that can be parsed again. It would need the brackets I think?

baszalmstra (changes requested):
Thanks for this PR!

I agree that we should merge the two representations. I'm already working on a PR that refactors both types, reduces the memory footprint, and increases performance! 👍

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -186,12 +201,29 @@ impl MatchSpec {
             }
         }
 
+        if let Some(md5_spec) = self.md5.as_ref() {
+            if let Some(md5_record) = record.md5.as_ref() {
+                if md5_spec != md5_record {
+                    return false;
+                }
+            }
+        }
```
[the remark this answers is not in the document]
0xbe7a: we now only match records that have a hash and the hash is equal

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -186,12 +206,33 @@ impl MatchSpec {
             }
         }
 
+        if let Some(md5_spec) = self.md5.as_ref() {
+            if !record
+                .md5
+                .as_ref()
+                .map_or(false, |md5_record| md5_spec == md5_record)
```
baszalmstra, line 213: You can use something along the lines of Some(md5_spec) == record.md5.as_ref() to simplify the code a bit.

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -226,23 +273,55 @@ impl NamelessMatchSpec {
...
         true
     }
 }
 
 impl Display for NamelessMatchSpec {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
-        match &self.version {
-            Some(version) => write!(f, "{version}")?,
-            None => write!(f, "*")?,
+        if let Some(version) = self.version.as_ref() {
```
baszalmstra, line 302: Can you turn this back into a match too?

baszalmstra (commented):
2 small comments otherwise looks good to me!

baszalmstra (approved):

baszalmstra:
@0xbe7a thanks a lot!!

PR #242, opened 2023-07-02
Use apple-codesign only on macOS

pavelzw:


pavelzw:
Unfortunately, I couldn't get the `ppc64le` and `aarch64` builds to work due to some zlib linking errors when building `rattler_solve` and `libsolv` :(

baszalmstra:
One of our design goals is to be able to create fully functioning environments from one machine to another (or at least make it as easy as possible). This includes from Linux to MacOS. Therefore removing code signing from other platforms is not something Im willing to remove. 

What did you run in to with zlib? Is that something we can fix instead?

baszalmstra:
We are also working on replacing libsolv (see #236 ) maybe that will also help in the future!

pavelzw:
> We are also working on replacing libsolv

Yes, this will already help a lot :)

Although, when I tried to build it for ppc64le in conda-forge, ring didn't build which is a dependency for apple-codesign. We probably need to investigate if we get ring out of the apple codesign dependencies.

pavelzw:
> What did you run in to with zlib?

This will solve itself when we have libsolv-rs :)

baszalmstra:
Im going to close this PR unless someone has a significant need for this. We prefer to fix the problem in `apple-codesign` or `ring`.

PR #243, opened 2023-07-03
Add libsolv_rs

aochagavia:


crates/libsolv_rs/src/solver/mod.rs
```diff
@@ -0,0 +1,1399 @@
...
+    fn set_propagate_learn(
+        &mut self,
+        mut level: u32,
+        solvable: SolvableId,
+        required_by: SolvableId,
+        clause_id: ClauseId,
+    ) -> Result<u32, Problem> {
+        level += 1;
+
+        println!(
```
wolfv, line 400: Should we move to `tracing::info!` here? Or `debug!`.

crates/libsolv_rs/src/solver/mod.rs
```diff
@@ -0,0 +1,1399 @@
...
+                    Decision::new(literal.solvable_id, decision, learned_clause_id),
+                    level,
+                )
+                .expect("bug: solvable was already decided!");
+            println!(
+                "=== Propagate after learn: {} = {decision}",
+                self.pool
+                    .resolve_solvable_inner(literal.solvable_id)
+                    .display()
+            );
```
wolfv, line 487: Just as a future idea - this would probably the place to check for a time-budget, right?

wolfv (approved):

crates/libsolv_rs/src/solver/mod.rs
aochagavia: You mean giving up after a given amount of time? In my eyes the best place for that would be `Solver::resolve_dependencies`, which is using `Solver::set_propagate_learn`

crates/rattler-bin/src/commands/create.rs
```diff
@@ -208,8 +211,13 @@ pub async fn create(opt: Opt) -> anyhow::Result<()> {
 
     // Next, use a solver to solve this specific problem. This provides us with all the operations
     // we need to apply to our environment to bring it up to date.
+    let use_libsolv_rs = opt.use_experimental_libsolv_rs;
     let required_packages = wrap_in_progress("solving", move || {
-        rattler_solve::LibsolvBackend.solve(solver_task)
+        if use_libsolv_rs {
+            rattler_solve::LibsolvRsBackend.solve(solver_task)
```
baszalmstra, line 217: Doesnt the `LibSolvRepoData` also have to be specific?

crates/rattler_solve/Cargo.toml
```diff
@@ -23,8 +23,10 @@ url = "2.4.0"
 hex = "0.4.3"
 tempfile = "3.6.0"
 cfg-if = "1.0.0"
+libsolv_rs = { version = "0.1.0", path = "../libsolv_rs" }
```
baszalmstra, line 26: Ideally we put both implementations behind feature flags.

crates/rattler_solve/src/lib.rs
```diff
@@ -46,6 +48,7 @@ impl fmt::Display for SolveError {
 
 /// Represents a dependency resolution task, to be solved by one of the backends (currently only
 /// libsolv is supported)
+#[cfg_attr(test, derive(Clone))]
```
baszalmstra, line 51: This seems like a hack. Why should it only be cloneable for tests?

aochagavia: Oh, it's just that it's not "necessary" for it to be cloneable in the normal case, so I thought I'd be smart and only generate the clone impl for the case in which we are actually using it (when contributing to other Rust crates, sometimes maintainers prefer to avoid adding `Clone` impls for types in the public API unless it's truly necessary).

If you prefer I can change it to `derive(Clone)`

crates/libsolv_rs/src/solver/mod.rs
aochagavia: I'm postponing that because it's a fairly mechanical change and I want to make sure all the truly important stuff is done first

crates/libsolv_rs/src/solver/mod.rs
wolfv: Yeah, like setting a limit of 1 sec, or 10'000 iterations ...

PR #259, opened 2023-07-12
remove apple-codesign crate

wolfv:


crates/rattler/src/install/link.rs
```diff
@@ -179,8 +193,12 @@ pub fn link_file(
 
             // If the binary changed it requires resigning.
             if content_changed {
-                let signer = UnifiedSigner::new(SigningSettings::default());
-                signer.sign_path_in_place(&destination_path)?;
+                match codesign(&destination_path) {
+                    Ok(_) => {}
+                    Err(e) => {
+                        tracing::warn!("Failed to sign binary: {}", e);
```
baszalmstra, line 199: I think we should return the error here instead of warning. Otherwise we leave the user with a non-functioning environment.

wolfv: Yeah, but then we can't try to execute this on Linux

baszalmstra: Yeah thats fine right? Because the environment is not valid anyway? Otherwise I would suggest we add an option to disable signing or ignore signing errors. But by default I think this is an error that should bubble up.

wolfv: Or we should not try to sign packages when we don't find the binary?

baszalmstra: Maybe I dont get it, but that would still silently result in an environment that doesnt work on regular arm64 macs right?

wolfv: Hypothetically, if you were to cross-compile from Linux to osx-arm64 you don't need to have signed binaries (shared libs also need signing). And you wouldn't try to execute the code.

baszalmstra: But in that (edge) case I would prefer to set an option to not sign the binaries at all.

baszalmstra: If you use this on an arm64 mac and for some reason it fails the user should be notified that it failed.

baszalmstra (approved):

baszalmstra:
@pavelzw So in the end we still ended up removing the apple-codesign binary. 

- The build was utterly failing on the crate.
- The maintainer doesn't seem very open to fixing our issue.
- Cross-creating environments for Mac with ad-hoc certificates will still not work on another mac (defeating its purpose completely)

We ended up using the codesign binary on Mac.

The upside is that maybe now the other targets you were trying will work?

pavelzw:
So ppc64le still has hope 🤪 I'll get right to it once this is merged 😎

PR #261, opened 2023-07-13
Add newlines to run_script commands

maresb:
@baszalmstra, is this what you had in mind with https://github.com/prefix-dev/pixi/pull/210#issuecomment-1634178983? I don't have time to take this any further, so if not feel free to just close this and that out.


baszalmstra:
Almost I meant this:  #262

PR #263, opened 2023-07-13
feat: allow for additional activation scripts

ruben-arts:
Allow the activator to receive additional activation scripts from the user

### DIFF
--- a/crates/rattler_shell/src/activation.rs
+++ b/crates/rattler_shell/src/activation.rs
@@ -409,12 +409,11 @@ impl<T: Shell + Clone> Activator<T> {
                 .map_err(ActivationError::FailedToWriteActivationScript)?;
         }
 
-
         if let Some(additional_activation_scripts) = &variables.additional_activation_scripts {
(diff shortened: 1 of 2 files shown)


crates/rattler_shell/src/activation.rs
```diff
@@ -40,6 +40,9 @@ pub struct ActivationVariables {
 
     /// The type of behaviour of what should happen with the defined paths.
     pub path_modification_behaviour: PathModificationBehaviour,
+
+    /// Paths to additional activation scripts
+    pub additional_activation_scripts: Option<Vec<PathBuf>>,
```
baszalmstra, line 45: can just be Vec right?

wolfv:
I am wondering if this is better vs. cloning the "activator" struct and adding the additional scripts along the way, like:

```
Activator<Bash> full_act = {
  activation_scripts: a.activation_scripts.chain(extra_activation_scripts),
  ...a
};
```

Then we don't need to add extra fields. We might also have to handle extra-deactivation scripts otherwise.

PR #264, opened 2023-07-14
fix: fish format variable

ruben-arts:
Fix how we set variables in fish script


wolfv (approved):

ruben-arts:
> A little `insta` test for these could be good

Good point, added it and tested the result in a fish shell locally

wolfv:
Did you somehow forget to push the test?

wolfv (approved):

Issue #266, opened 2023-07-23
libsolv-rs is choosing slightly older libgfortran5 (on macos-arm64)


wolfv:
When installing `numpy` on macOS ARM64, the new libsolv implementation picks `libgfortran5-12.2.0-h0eea778_32`.

Conda and mamba choose `libgfortran5-12.3.0-ha3a6a3e_0`.

I wonder if that has to do with `libgfortran-5.0.0-12_3_0_hd922786_0` vs. `libgfortran-5.0.0-12_2_0_hd922786_32`. If we compare / sort by build number vs. build string we might pick the older one (because it has a much higer build number).

wolfv:
I think what happens is that first `libgfortran-5.0.0-12_2_0_hXXX_32` is picked (also with micromamba).

With micromamba, the following happens:

- libgfortran-5.0.0 depends on `libgfortran5 >=11.3`, and then libgfortran5-12.3.0 is selected (which has a constraint on `libgfortran *_0`). Since we selected `_32` before, it backtracks, and selects the appropriate `12_3_0_hXXX_0` version. However, the rattler solver picks `12.2` immediately and doesn't do another round of backtracking.

wolfv:
A test-case that shows this behavior was added in #267

baszalmstra:
This issue has become quite stale, especially since we use resolvo most of the time. Would you mind if I close this?

PR #267, opened 2023-07-23
Test propagation (issue #266)

wolfv:
I added a test for the eager propagation issue. 

Not sure how to resolve this properly yet!

### DIFF
--- a/crates/rattler_libsolv_rs/Cargo.toml
+++ b/crates/rattler_libsolv_rs/Cargo.toml
@@ -18,3 +18,4 @@ tracing = "0.1.37"
 
 [dev-dependencies]
 insta = "1.30.0"
+tracing-test = "0.2.4"
--- a/crates/rattler_libsolv_rs/src/solver/mod.rs
+++ b/crates/rattler_libsolv_rs/src/solver/mod.rs
@@ -909,6 +909,7 @@ mod test {
     use crate::id::RepoId;
     use rattler_conda_types::{PackageRecord, Version};
     use std::str::FromStr;
+    use tracing_test::traced_test;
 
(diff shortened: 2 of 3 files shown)


baszalmstra:
Do you mind if I close this one @wolfv ?

baszalmstra:
bump @wolfv

wolfv:
Should we try to add this test to `resolvo` now? I think it's still a valid concern. 

Basically the problem is that the algorithm:

- first selects `libgfortran 5 BN 32`
- then selects `libgfortran5 12.3.0`, but the constraint does not work with the selected build number of `libgfortran`
- backtracks, and selects older `libgfortran 5 BN 0` and then 
- selects latest `libgfortran5 12.3.0`

While it (should?) have done:

- first select `libgfortran 5 BN 32`
- figure out that constraint on `libgfortran5 12.3.0` does not work
- instead select `libgfortran5 12.2.0` where constraint works

wolfv:
Hmm, actually this might also be the "inverse" problem where the solver backtracks too early :D 


```
libgfortran = 5.0.0 32
    libgfortran5>= 11.3
libgfortran = 5.0.0 0
    libgfortran5>=11.3

libgfortran5 = 12.3.0
    -> libgfortran = 5.0.0 0
libgfortran5 = 12.2.0
    -> libgfortran = 5.0.0 32
```

We first select 

`libgfortran = 5.0.0 32`, then find out that the latest `libgfortran5 = 12.3.0` is not compatible, backtrack, select an older `libgfortran` (instead of trying an older `libgfortran5`).

wolfv:
I think it's something I might like to come back to. But we can also close it and not delete the branch

baszalmstra:
We will take a look at this later. Closing for now.

PR #271, opened 2023-07-23
add initial nushell support

wolfv:


baszalmstra:
@wolfv I know you were still working on this. Is this now ready for review?

wolfv:
I think it's ready for review, but I haven't had a chance to actually test this yet :)

wolfv:
I've just tested this a bit with `pixi` & `nushell` (and fixed a bug with `let-env` vs. `$env.`).

There is an issue with `nushell` not executing the `source` command in the PTY right away but I think we might have to fix in pixi.

PR #273, opened 2023-07-27
fix: mmap permission denied

baszalmstra:
This PR is an attempt to fix the issue described in https://github.com/prefix-dev/pixi/issues/234. 

After some investigation, it appears that specifically on macOS it is not allowed to memory map a file that was mounted into an x86 based docker image. This PR adds more diagnostic errors and falls back to reading the entire file to memory if memory mapping fails. In the future it might be nice to stream the file in and do the replacement on the fly.


baszalmstra:
Although this is not strictly required I think it cleans up the code, so Ill merge anyway.

PR #279, opened 2023-08-21
Initial version of rattler python bindings

baszalmstra:
This is the initial work of creating Python bindings for rattler. 

The bindings are implemented using a double indirection approach that I learned from polars. Lets take `Version` as an example:

- Rust: PyVersion (wraps `rattler_conda_types::Version` and exposes it to python using PyO3)
- Python: Version (wraps the exposed `PyVersion` and exposes all functionalities with a pythonic API. Adds types and docstrings)

TODO: 

- [x] Set up better CI (lint, test, build).
- [x] Set up linter (ruff)
- [ ] Determine the initial functionality and implement it. (I think matchspec and version)


py-rattler/src/version.rs
```diff
@@ -0,0 +1,72 @@
...
+    pub fn __init__(version: &str) -> pyo3::PyResult<Self> {
+        Ok(Version::from_str(version)
+            .map(Into::into)
+            .map_err(PyRattlerError::from)?)
+    }
+
+    /// Returns a string representation of the version.
+    pub fn as_str(&self) -> String {
+        format!("{}", self.inner)
+    }
```
tdejager, line 31: Should there also be `__repr__` function?

tdejager (approved):

py-rattler/src/version.rs
wolfv: Tbere is one in the `version.py` file, right?

PR #280, opened 2023-08-21
add retry behavior for package cache downloads

baszalmstra:
Adds retry behavior for extracting packages. I'm not entirely sure how I can test this properly though...

This was reported in pixi: https://github.com/prefix-dev/pixi/issues/273 . I will try to test-drive it with pixi.

### DIFF
--- a/crates/rattler-bin/src/commands/create.rs
+++ b/crates/rattler-bin/src/commands/create.rs
@@ -401,6 +401,7 @@ async fn execute_operation(
                     &install_record.package_record,
                     install_record.url.clone(),
                     download_client.clone(),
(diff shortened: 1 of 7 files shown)


wolfv:
This looks very elegant. I am wondering if we want to stick that into a lower layer (e.g. in the `AuthenticatedClient`, so that every network request can benefit from retrying?

We can also refactor that later to fix the immediate bug.

wolfv (approved):

wolfv:
Bas explained that we can't stick it in a lower level because the streaming needs to be restarted as well. Merging!

PR #281, opened 2023-08-21
fix: local version parsing in constraint

baszalmstra:
Fixes: https://github.com/prefix-dev/pixi/issues/278

The issue was that our Constraint parser didn't account for local versions (e.g. `1.2.3+abc`)


wolfv:
awesome!

PR #285, opened 2023-08-22
feat: normalize package names where applicable

baszalmstra:
This PR fixes an issue where package names of conda are always lowercase but they don't have to be lowercase in specs. This PR introduces a new type `PackageName` that holds both the original source package name as well as the normalized (lowercased) package name. The type is used throughout the rest of the code base which makes this a slightly breaking change.

This should fix the issue reported [here](https://github.com/prefix-dev/pixi/issues/264).


crates/rattler_conda_types/src/package_name.rs
```diff
@@ -0,0 +1,129 @@
...
+}
+
+impl From<String> for PackageName {
+    fn from(value: String) -> Self {
+        Arc::<str>::from(value).into()
+    }
+}
+
+impl From<Arc<str>> for PackageName {
+    fn from(value: Arc<str>) -> Self {
```
wolfv, line 47: We could also potentially validate that all chars are ascii here.

wolfv (approved):

wolfv:
Not really sure if we need to keep the original form or not. And my second question would be that if we do this for all MatchSpecs, are we going to loose a lot of performance (in repodata we can assume that they are already lowercase, I guess).

baszalmstra:
Yeah I was also wondering if it is useful to keep the source representation. I was also thinking that it might be good to use `PackageName` in `PackageRecord` as well. So that we use the same type where we semantically want to refer to a package name. @wolfv WDYT? We can also have a conversion method where we don't check if the name contains uppercase characters and we just assume its valid.

The overhead of `PackageName` is really small but it does increase memory size slightly a little bit.

twrightsman:
@baszalmstra Thanks for working on this! I am also unsure if it is useful within rattler to keep the source representation. As a user, the behavior I'd like to see is `pixi` not normalizing the name in `pixi.toml` whether it was added manually or through things like `pixi add`. So if rattler doesn't track the source representation, then I suppose `pixi` would have to know how to map unnormalized names provided by the user to anything normalized returned by rattler? Just my two cents :smile:

baszalmstra:
I changed this PR quite a bit. `PackageName` is now used everywhere where we are semantically talking about a package name. I also added validation that a `PackageName` only contains valid characters. My gut says the overhead of the implementation is very low but I didnt test it.

tdejager (approved):

PR #286, opened 2023-08-23
Expose ParseMatchSpecError in rattler_conda_types

Wackyator:
This is currently being exposed to be utilised in py-rattler.


wolfv:
I'll merge this as the CI failure is fixed on main already.

PR #287, opened 2023-08-23
make rattler-package-streaming compile with wasm

wolfv:


crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -66,13 +66,27 @@ impl Channel {
...
+                Self {
+                    platforms,
+                    base_url: url,
+                    name: Some(channel.to_owned()),
+                }
+            }
+
+            #[cfg(target_arch = "wasm32")]
+            {
+                let url = Url::try_from("https://anaconda.org").unwrap();
```
baszalmstra, line 84: I don't like this :P

Is one of the functions above not supported?

wolfv:
- Should we move the blocking stuff into a seperate module so that we can exclude it more easily from the compilation?
- make blocking a default feature?

I think then it looks kinda decent!

crates/rattler_conda_types/src/platform.rs
```diff
@@ -146,7 +146,9 @@ impl Platform {
             target_os = "emscripten",
             windows
         )))]
-        compile_error!("unsupported target os");
+        {
+            return Platform::Linux64;
```
baszalmstra, line 150: Maybe add a comment to _why_ this is linux64.

crates/rattler_networking/Cargo.toml
```diff
@@ -13,20 +13,24 @@ readme.workspace = true
 [features]
 native-tls = ['reqwest/native-tls']
 rustls-tls = ['reqwest/rustls-tls']
+blocking = ['reqwest/blocking']
```
baszalmstra, line 16: Maybe you can add this to the `default-features`

crates/rattler_package_streaming/Cargo.toml
```diff
@@ -24,16 +24,16 @@ tokio = { version = "1", optional = true }
...
 
 [features]
 default = ['native-tls']
 tokio = ["dep:tokio", "bzip2/tokio", "tokio/fs", "tokio-util/io", "tokio-util/io-util", "reqwest?/stream", "futures-util"]
-native-tls = ['reqwest?/native-tls']
-rustls-tls = ['reqwest?/rustls-tls']
-reqwest = ["reqwest/blocking"]
+native-tls = ['rattler_networking/native-tls']
+rustls-tls = ['rattler_networking/rustls-tls']
+blocking = ["rattler_networking/blocking"]
```
baszalmstra, line 36: Can you also add this to the default feature set?

crates/rattler_package_streaming/src/reqwest/mod.rs
```diff
@@ -3,11 +3,17 @@
 #[cfg(feature = "tokio")]
 pub mod tokio;
 
+#[cfg(feature = "blocking")]
```
baszalmstra, line 6: Perhaps it would be better to move all the blocking things to a module to remove the code duplication of the blocking feature cfg?

crates/rattler_conda_types/src/channel/mod.rs
baszalmstra: If thats the case, maybe we should just return an error here.

wolfv: The `Url::from_directory_path` doesn't exist on WASM (probably because there is no filesystem.

I am throwing an error instead.

baszalmstra (approved):

PR #288, opened 2023-08-23
feat: support Debian CUDA detection

twrightsman:
Uses paths from [Debian](https://packages.debian.org/bullseye/amd64/libcuda1/filelist). Created to fix https://github.com/prefix-dev/pixi/issues/294. As noted in the issue, simply building from source without this patch results in CUDA correctly being detected, but using the GitHub releases Linux binary doesn't work.

### DIFF
--- a/crates/rattler_virtual_packages/src/cuda.rs
+++ b/crates/rattler_virtual_packages/src/cuda.rs
@@ -113,8 +113,8 @@ fn nvml_library_paths() -> &'static [&'static str] {
         "/usr/lib/x86_64-linux-gnu/libnvidia-ml.so",
         "/usr/lib/wsl/lib/libnvidia-ml.so.1", // WSL
         "/usr/lib/wsl/lib/libnvidia-ml.so",
-        "/usr/lib/x86_64-linux-gnu/nvidia/current/libcuda.so.1", // Debian
-        "/usr/lib/x86_64-linux-gnu/nvidia/current/libcuda.so",
+        "/usr/lib/x86_64-linux-gnu/libcuda.so.1", // Debian
+        "/usr/lib/x86_64-linux-gnu/libcuda.so",
     ];
     #[cfg(windows)]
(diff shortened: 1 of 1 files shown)


wolfv:
@twrightsman thanks! Looks good to me! I wonder if the `current` subdirectory is necessary?

Does this file also exist for you? 

`/usr/lib/x86_64-linux-gnu/libcuda.so.1`

I'll update the branch so that the tests pass (you'll have to pull)

twrightsman:
Thanks for taking a look!

Yes I think you're correct here, `/usr/lib/x86_64-linux-gnu/libcuda.so` and `/usr/lib/x86_64-linux-gnu/libcuda.so.1` exist and are symlinks to places in `/etc/alternatives`. Looks like these would be the better paths since they would be correctly updated through the Debian [Alternatives](https://wiki.debian.org/DebianAlternatives) system.

I'll go ahead and pull down your updates and make those changes.

twrightsman:
Pushed; let me know if I should do any rearranging or changing of commit messages to match the project's conventions.

wolfv:
We can fix the message when squash-merging :) Awesome & thanks for your contribution!

baszalmstra:
I dont quite understand how this fixes the issue. You said that the musl based pixi install doesnt detect cuda correctly but building from source (not building with musl I assume) does work. It sounds to me like this is more of a musl issue.

Did you try building pixi as a musl binary with these changes?

wolfv:
Sorry, I do have more questions after looking a bit closer. 

Do you have this package installed? https://packages.debian.org/sid/libnvidia-ml1
And do you not have the `libnvidia-ml.so` file, at all? Because it seems like `libcuda` and `libnvidia-ml` would be quite different shared libraries.

twrightsman:
@baszalmstra Do you happen to have instructions on how to build with musl? I am happy to test further. I believe Debian defaults to glibc. We can also move this discussion to the issue if you want.

PR #293, opened 2023-08-25
fix: Version::is_dev returning false for dev version (fix #289)

### LINKED ISSUE #289: `Version::is_dev` returns false for dev version
```rust 
let v = Version::from_str("1.2.1dev2");
println!("{}", v.unwrap().is_dev());
```

This should return true, but instead returns false.

Wackyator:


wolfv (approved):
Wow, the fix was much easier than I anticipated. Nice find!

Wackyator:
Yeah, I didnt think it would be this easy either

baszalmstra (approved):

baszalmstra:
Nice work @Wackyator !

PR #294, opened 2023-08-25
fix: workaround for `PIP_REQUIRE_VIRTUALENV` env variable

tusharsadhwani:
If `PIP_REQUIRE_VIRTUALENV` is set to `true` in the user's environment, `maturin develop` fails because the `pip` subcommand doesn't recognize the env as a venv.

This works around that.

### DIFF
--- a/py-rattler/pixi.toml
+++ b/py-rattler/pixi.toml
@@ -7,7 +7,7 @@ channels = ["conda-forge"]
 platforms = ["win-64", "linux-64", "osx-64", "osx-arm64"]
 
 [tasks]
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

wolfv:
Did the `--isolated` option not work? It might be even better?

tusharsadhwani:
No, `--isolated` didn't work for me, but you can try to reproduce it to be doubly sure.

Just `export PIP_REQUIRE_VIRTUALENV=true` in your shell to reproduce the error.

PR #295, opened 2023-08-27
fix: ensure consistent sorting of locked packages

baszalmstra:
This PR replaces some of the HashMaps in the conda-lock data structures with `IndexMap` (keeps order of insertion) and `BTreeMap` (sorts by key). This ensures that we get a consistent output in pixi.

I used `IndexMap` in the places where I thought the order of the source elements was important. For instance, the dependencies specified in the repodata.

I used `BTreeMap` where I think the order doesn't matter but it should be consistent.


wolfv:
Hmm, I am unsure because the IndexMap in the conda-lock part is going to be filled from the `repodata` right? And do we use (want to use) an IndexMap there? I would argue that sorting the dependencies array alphabetically would be fine (and probably also what is done by conda-build in the repodata).

tdejager (approved):

PR #299, opened 2023-08-28
Updates to `NamelessMatchSpec` to allow deserializing

travishathaway:
While working on this issue:

- https://github.com/prefix-dev/pixi/issues/76

I realized that I was going to need to be able to deserialize the `NamelessMatchSpec` struct from a mapping. I believe the easiest way to allow this is by adding `Deserialize` to `derive` in the correct places.

On `NamelessMatchSpec` I also chose to use `DisplayFromStr` for `version` and `build` because I believe that's the easiest way to do it. I would be happy to change it if that's not right though.

Below is my simplified use case:

```rust
use std::fs;
use std::path::PathBuf;

use indexmap::IndexMap;
use clap::{Parser};
use serde::Deserialize;
use serde_with::{serde_as, DisplayFromStr, PickFirst};
use rattler_conda_types::NamelessMatchSpec;

#[derive(Parser)]
#[command(name = "tt - TOML test")]
#[command(author = "Travis Hathaway")]
#[command(version = "0.1.0")]
#[command(about = "Tests with TOML parsing")]
struct Cli {
    /// TOML File
    toml_file: PathBuf,
}

/// Describes the contents of a test project manifest.
#[serde_as]
#[derive(Debug, Clone, Deserialize)]
pub struct MyManifest {
    #[serde_as(as = "IndexMap<_, PickFirst<(_, DisplayFromStr)>>")]
    dependencies: IndexMap<String, NamelessMatchSpec>
}


fn main() -> Result<(), Box<dyn std::error::Error>>{
    let cli = Cli::parse();

    println!("{:?}\n", cli.toml_file);

    if cli.toml_file.exists() {
        let contents = fs::read_to_string(cli.toml_file)?;
        let manifest = toml_edit::de::from_str::<MyManifest>(&contents)?;

        println!("Dependencies:");
        println!("{:?}", manifest.dependencies);

    } else {
        println!("File does not exist");
    }

    Ok(())
}
```

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -1,7 +1,7 @@
 use crate::{PackageName, PackageRecord, VersionSpec};
 use rattler_digest::{serde::SerializableHash, Md5Hash, Sha256Hash};
-use serde::Serialize;
-use serde_with::{serde_as, skip_serializing_none};
+use serde::{Deserialize, Serialize};
+use serde_with::{serde_as, skip_serializing_none, DisplayFromStr};
 use std::fmt::{Debug, Display, Formatter};
 use std::hash::Hash;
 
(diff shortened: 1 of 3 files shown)


baszalmstra (approved):

baszalmstra:
Thanks @travishathaway!

PR #304, opened 2023-08-30
Add VersionWithSource type to py-rattler

Wackyator:


py-rattler/rattler/version/with_source.py
```diff
@@ -0,0 +1,447 @@
...
+class VersionWithSource:
+    """
+    Holds a version and the string it was created from. This is useful if
+    you want to retain the original string the version was created from.
+    This might be useful in cases where you have multiple strings that
+    are represented by the same [`Version`] but you still want to be able to
+    distinguish them.
+
+    The string `1.0` and `1.01` represent the same version. When you print
+    the parsed version though it will come out as `1.0`. You loose the
```
wolfv, line 19: this can't be right (1.0 and 1.01 should not both print as `1.0`, right?)

wolfv: I think `1.1` and `1.01` are the same though.

wolfv:
I think if we add a `VersionBase` class in Python we can get away with implementing the forwarding functions for a few things only once. 

E.g. 

```py
class VersionBase:
    @property
    def epoch(self) -> Optional[str]:
        """
        Gets the epoch of the version or `None` if the epoch was not defined.

        Examples
        --------
        >>> v = Version('2!1.0')
        >>> v.epoch
        2
        """
        return self._version.epoch()

    @property
    def has_local(self) -> bool:
        """
        Returns true if this version has a local segment defined.
        The local part of a version is the part behind the (optional) `+`.

        Examples
        --------
        >>> v = Version('1.0+3.2-alpha0')
        >>> v.has_local
        True
        >>> v2 = Version('1.0')
        >>> v2.has_local
        False
        """
        return self._version.has_local()
```

If both, `Version` and `VersionWithSource` call the inner type `_inner` you can align the two, and inherit with both from `VersionBase`. That should remove some (trivial) duplication.

py-rattler/rattler/version/with_source.py
Wackyator: yes that's a mistake

Wackyator: It's also incorrect in the rattler docs as well, need to fix that

wolfv: Great!

py-rattler/rattler/version/with_source.py
```diff
@@ -0,0 +1,447 @@
...
+        >>> v = VersionWithSource('2!1.0')
+        >>> v.epoch
+        2
+        >>> v2 = VersionWithSource('2!1.0', v.version)
+        >>> v2.epoch
+        2
+        """
+        return self._inner.epoch()
+
+    def bump(self) -> VersionWithSource:
```
baszalmstra, line 74: I think this should return a Version instead. Since you modify the original version anyway it doesnt make sense to keep the source.

baszalmstra:
Instead of having a VersionBase, wouldnt it be possible to have a Version and a VersionWithSource that derives from Version and adds a source string? Thats essentially what it is.

py-rattler/src/version/with_source.rs
```diff
@@ -0,0 +1,59 @@
...
+    fn __hash__(&self) -> u64 {
+        let mut hasher = DefaultHasher::new();
+        self.inner.hash(&mut hasher);
+        hasher.finish()
+    }
+
+    /// Performs comparison between this version and another.
+    pub fn __richcmp__(&self, other: &Self, op: CompareOp) -> bool {
+        op.matches(self.inner.cmp(&other.inner))
+    }
```
baszalmstra, line 58: These are not used right?

baszalmstra (commented):
I like this approach but it raises the question: do we even need the `PyVersionWithSource`?

py-rattler/src/version/with_source.rs
Wackyator: These aren't really needed at the moment, some rattler types require VersionWithSource but I think they could simply be constructed when required?

py-rattler/rattler/version/with_source.py
```diff
@@ -0,0 +1,90 @@
...
+    are represented by the same [`Version`] but you still want to be able to
+    distinguish them.
+
+    The string `1.1` and `1.01` represent the same version. When you print
+    the parsed version though it will come out as `1.1`. You loose the
+    original representation. This class stores the original source string,
+    which can be accessed by `source` property.
+    """
+
+    def __init__(self, source: str, version: Optional[Version] = None):
```
baszalmstra, line 22: I think this signature should be:

```suggestion
    def __init__(self, version: Union[str, Version]):
```

baszalmstra (approved):

PR #305, opened 2023-08-30
Build spec module

YeungOnion:
I think this is a little bigger than it needed to be, but I felt like I'd necessarily need to reimplement a slightly smaller version of everything but the tree in `version_spec`, so I wrote with some of those thoughts in mind. 

---

As I write this I realize that I haven't pulled in a bit and didn't see the work on breaking out the VersionOperator into its parts. So a lot of what I was thinking to do is started up there. I'll move this to draft so I can submit this with some of this using what's existing in version_spec.

Would it be fine to change `version_spec::parse::constraint_parse` to support type T, then call it for `VersionConstraint = Any | Comparison | StrictComparison | Exact` and also call it for `BuildConstraint = Comparison | Exact`? 

Or does that not make sense in the context of build specs?


baszalmstra:
Starting to look good already! Might indeed be a little bit over-engineered but that's fine. :)

@YeungOnion If you can make it generic that would be great yeah!

crates/rattler_conda_types/src/version_spec/mod.rs
```diff
@@ -332,11 +268,8 @@ mod tests {
             Ok(VersionSpec::Group(
                 LogicalOperator::And,
                 vec![
-                    VersionSpec::Range(
-                        RangeOperator::GreaterEquals,
-                        Version::from_str("1.2.3").unwrap()
-                    ),
-                    VersionSpec::Range(RangeOperator::Less, Version::from_str("2.0.0").unwrap()),
+                    VersionSpec::OrdRange(OrdOperator::Ge, Version::from_str("1.2.3").unwrap()),
```
YeungOnion, line 271: looks like fmt not working here

YeungOnion:
Think in my overzealousness and irregularity on working on this, I missed [this](https://github.com/mamba-org/rattler/pull/296) and I'm not sure if I deprecated any feature there. I don't see a diff where I move `StrictVersion` to `Version`, but I'm not sure if `StrictVersion` is a different type or it was a way to specify that this `Version` would have one of the `StrictRangeOperator`s, if so, I'd like to hear about the background for that decision as it seems duplicative to me, but I don't have the context @tdejager when they added those other operators and StrictVersion

tdejager:
Hi @YeungOnion this was added to support the case where a Version like `1.0.0` != `1.0`. Normally, trailing zero's on a version are considered the same as a version without trailing zeros. This if fine normally. Only for the case for **startsWith** and **compatible** it is not. Because in that case you want to start the range for a specific version.

The range `1.0.*`  is unequal to the range `1.*`. And because `1.0.0` and `1.0` were considered the same version this caused problems when using `Hash` for example because these values `tensorflow 2.0.*` and `tensorflow 2.*` would hash to the same values. This is no longer the case with the `StrictVersion` type.

While in most cases like an equality of a single version `v1.0` == `v1.0.0` the versions can be considered the same. Also for a greater than, like `> 1.0` you do not care about the trailing zeros either so these all hash to the same values, and are equal. E.g `>= 1.0` == `>= 1.0.0`, because they both equal to the same range

YeungOnion:
I think this gives a good chunk of the context. I still don't understand why the operator alone isn't enough, I do see that there's a different `impl PartialEq` for it that uses the number of components as well as the segment comparison.

Is it not used explicitly as part of the spec? I see that [it's not used in the version constraint enum](https://github.com/mamba-org/rattler/blob/main/crates/rattler_conda_types/src/version_spec/constraint.rs#L19) and I may be misunderstand where a user would specify that a spec is for a `StrictVersion`

YeungOnion:
Separate meta topic: I think my PR got far too big.

Looking for feedback or tips for not doing things like this in the future. As you can see, I've not the practice or discipline of incremental progress yet.

YeungOnion:
I'll start a few new PR's as this one modified more than it needed to at once. I also cut back on some unneeded generality.

Issue #311, opened 2023-09-05
Python Bindings v0.1.0 (WIP)


baszalmstra:
With #279 we have the first implementation of python bindings for rattler.  🎉 

For v0.1.0 of the bindings we would like to be able to achieve the following goal:

🥅 _You can **solve** and **install** an environment using the Python bindings._

We want to be able to do this to make it easier for users coming from Python to start using rattler as their base library for everything conda instead of having to rely on `conda` or `mamba`.

To that end, the following code needs to be worked on:

* You can fetch repodata from a remote or local channel.
   * The main entry point is: `fetch_repo_data`.
   * We have to determine how we will manage memory. Since RepoData takes up a significant amount of memory we want to reduce copying, duplication, etc. The solver also takes `PackageRecord`s by reference how do we deal with that?
   * We will have to investigate how to handle `async`
   * We need bindings for the following types:
      * [ ] `AuthenticatedClient`
      * [ ] `Platform`
      * [ ] `RepodataRecord`/`PackageRecord` (we can skip implementing all the fields for now?? maybe it would be nice to be able to access some)
   * `fetch_repo_data` returns an object that uses a scope to drop a lock. How do we deal with that?
   * Out of scope for now is:
      * the new `Gateway` implementation unless this lands pretty soon (would simplify the whole API a lot). Might be worth putting some effort in?
      * Download callback is not a requirement but would be nice to have.
* Pass `RepodataRecord` or `PackageRecord` to a solver function to solve the environment.
  *   @baszalmstra Needs some refining here ...
* Install an environment from a list of `RepodataRecord`s. 
  *   @baszalmstra Needs more refining here ..

wolfv:
I am not sure I understand why we need RepodataRecord and friends. We could also decide to hide all that behind the API, right? And just have a "Repodata" object that holds the deserialized repodata?

baszalmstra:
Yeah I like the idea of a higher level api. But would the repodata hold all records or a selection?

baszalmstra:
We are closing this in favor of

- #356 
- #355 
- #354 
- #357 
- #353

PR #315, opened 2023-09-07
feat: Add AuthenticatedClient to py-rattler

Wackyator:
Adds the AuthenticatedClient to py-rattler from #311

### DIFF
--- a/py-rattler/rattler/networking/authenticated_client.py
+++ b/py-rattler/rattler/networking/authenticated_client.py
@@ -1,0 +1,38 @@
+from __future__ import annotations
+from typing import Self
+from rattler.rattler import PyAuthenticatedClient
+
+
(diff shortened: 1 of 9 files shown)


py-rattler/rattler/networking/authenticated_client.py
```diff
@@ -0,0 +1,38 @@
...
+        """
+        authenticated_client = cls.__new__(cls)
+        authenticated_client._client = client
+        return authenticated_client
+
+    def __str__(self) -> str:
+        """
+        Returns the string representation of the AuthenticatedClient
+        """
+        return ""
```
baszalmstra, line 27: Maybe we can just forgo this implementation if it just returns nothing.

baszalmstra (approved):

PR #323, opened 2023-09-08
fix: add N generic everywhere

baszalmstra:
Adds the generic package name all over the place where this wasn't the case yet.

### DIFF
--- a/crates/rattler_libsolv_rs/src/lib.rs
+++ b/crates/rattler_libsolv_rs/src/lib.rs
@@ -32,7 +32,7 @@ pub use transaction::Transaction;
 
 pub use mapping::Mapping;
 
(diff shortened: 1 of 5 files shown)


crates/rattler_libsolv_rs/src/lib.rs
```diff
@@ -32,6 +32,10 @@ pub use transaction::Transaction;
 
 pub use mapping::Mapping;
 
+/// Blanked trait implementation for something that we consider a package name.
```
tdejager, line 35:
```suggestion
/// Blanket trait implementation for something that we consider a package name.
```

tdejager (changes requested):

PR #327, opened 2023-09-11
feat: allow disabling jlap

baszalmstra:
This adds the `jlap_enabled` flag to the `FetchRepoDataOptions` struct. This allows you to completely disable `JLAP`. By default `JLAP` is enabled. It also adds a bunch of new logging statements so we can trace a little better what is happening when we have future JLAP issues.


crates/rattler_repodata_gateway/src/fetch/mod.rs
```diff
@@ -139,14 +139,27 @@ impl Variant {
...
+#[derive(Clone)]
 pub struct FetchRepoDataOptions {
     /// How to use the cache. By default it will cache and reuse downloaded repodata.json (if the
     /// server allows it).
     pub cache_action: CacheAction,
 
     /// Determines which variant to download. See [`Variant`] for more information.
     pub variant: Variant,
+
+    /// True if repodata can be fetch incrementally using JLAP.
```
tdejager, line 151:
```suggestion
    /// When enabled repodata can be fetched incrementally using JLAP
```

tdejager (approved):

PR #329, opened 2023-09-11
drop support for repodata.json.bz2

### LINKED ISSUE #308: Drop support for repodata.json.bz2
Conda hasn't fetched `repodata.json.bz2` for quite some years now, although conda-index supports it as a precaution. It would make sense to drop support for the format entirely.

`anaconda.org/channel` supports `.zst` now, as well as defaults and conda-forge. Unknown whether `.bz2` with its slow unpack speed is faster than on-the-fly `Content-Encoding: gzip`.

dholth:
As explained in #308 `repodata.json.bz2` should be obsolete. Remove from rattler.

Fix #308

Does this make unwanted API changes to rattler?

### DIFF
--- a/crates/rattler_repodata_gateway/src/fetch/cache/mod.rs
+++ b/crates/rattler_repodata_gateway/src/fetch/cache/mod.rs
@@ -43,9 +43,6 @@ pub struct RepoDataState {
     /// Whether or not zst is available for the subdirectory
     pub has_zst: Option<Expiring<bool>>,
 
-    /// Whether a bz2 compressed version is available for the subdirectory
-    pub has_bz2: Option<Expiring<bool>>,
-
     /// Whether or not JLAP is available for the subdirectory
     pub has_jlap: Option<Expiring<bool>>,
 
--- a/crates/rattler_repodata_gateway/src/fetch/mod.rs
+++ b/crates/rattler_repodata_gateway/src/fetch/mod.rs
@@ -990,11 +990,6 @@ mod test {
                 tokio::io::copy(&mut input, &mut encoder).await?;
                 encoder.shutdown().await?;
             }
-            Encoding::Bz2 => {
-                let mut encoder = async_compression::tokio::write::BzEncoder::new(file);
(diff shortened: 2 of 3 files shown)


baszalmstra (approved):
Looks good to me! Thanks @dholth

baszalmstra:
Ah it looks like CI is still failing. Would you be able to fix that?

wolfv:
I am wondering if we should merge this yet as some other index-providers (artifactory / quetz / ...) could potentially still benefit from `bz2`.
Maybe @xhochy has an opinion?

xhochy:
At the moment we don't seem to use the bzip2 feature. There are some weird constellations in the wild though where it would have been useful:

1. Artifactory currently doesn’t work correctly with zstd compressed repodata. Here we simply fallback to the gzip compressed default.
2. Some virus scanning proxies don't compress the repodata.json endpoint. Here we can use the zstd one instead.

Luckily we don't have the combination of Artifactory that goes through a corporate proxy. There bzip2 would still be helpful.

dholth:
A while ago we noticed that uncompressed `repodata.json` transfers were faster than on-the-fly-compressed Content-Encoding: gzip `repodata.json` transfers for >30Mbps? links. `bz2` also decompresses slowly, and on-the-fly `bz2` is a bad idea unlike on-the-fly zstd which is probably faster than Content-Encoding: gzip. So the `bz2` format has a small niche these days. It looks like conda dropped it around [2016-06-23](https://github.com/conda/conda/blob/c5057099b55ea46312fd1fe66a265cb953f22888/CHANGELOG.md?plain=1#L3967C22-L3967C22)

baszalmstra:
Mmm, is there a big downside in keeping the bz2 option in there? The only problem I see is that we are performing an extra HEAD request and that we prefer it over the uncompressed version. Maybe we could change that instead of completely dropping it?

baszalmstra:
Given that `bzip2` still seems to have some merit in some corporate environments and since it's also not doing any harm I'm going to close this PR (and the corresponding issue) for now.

PR #331, opened 2023-09-12
test: added job to check for lfs links

tdejager:
Added job to test if everything stayts an LFS pointer.


baszalmstra (approved):

wolfv:
We do wanna get rid of LFS though, no? It cost me 25 bucks per month :(

PR #344, opened 2023-09-20
Matchable trait

YeungOnion:
Adds trait for implementing `matches` method to compare types that are some kind of `Spec` to the type which represents a single instance of the matchable Spec can describe.

## Question:
Should this be associated type (as implemented) or a generic? I think the difference has to do if the trait will be implemented for generic types or only needed for concrete types.

---
remaining TODOs before moving from draft
impl for VersionSpec


YeungOnion:
There is one very significant effect here regarding this change. I don't know how best to expose the `MatchSpec` method for `matches` without requiring other crates to `use rattler_conda_types::Matcher` and I don't know if this is necessary.

I'd like some of your input on what should be usable with this crate as a dependency based on the re-exports you already use.

YeungOnion:
think I boofed this branch, redoing

PR #345, opened 2023-09-20
`Matchable` trait, mentioned in #340

YeungOnion:
Adds trait for implementing `matches` method to compare types that are some kind of `Spec` to the type which represents a single instance of the matchable Spec can describe, this was mentioned in #340 

## Question:
- Should this be associated type (as implemented) or a generic? I think the difference has to do if the trait will be implemented for generic types or only needed for concrete types.
- How should this be exported? I don't think all uses of the crate should need to import `Matchable` to call `MatchSpec::matches`.

---
remaining TODOs before moving from draft
impl for VersionSpec

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/matcher.rs
+++ b/crates/rattler_conda_types/src/match_spec/matcher.rs
@@ -5,6 +5,10 @@ use std::{
     str::FromStr,
 };
 
+pub trait Matcher<Element: ?Sized> {
+    fn matches(&self, other: &Element) -> bool;
+}
+
 /// Match a given string either by exact match, glob or regex
 #[derive(Debug, Clone)]
 pub enum StringMatcher {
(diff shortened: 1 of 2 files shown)


baszalmstra:
> Should this be associated type (as implemented) or a generic? I think the difference has to do if the trait will be implemented for generic types or only needed for concrete types.

I would go for the associated type. In our use case we have specs that match a certain other type. 

> How should this be exported? I don't think all uses of the crate should need to import Matchable to call MatchSpec::matches.

I see your point with regard to this. But how would you work around that though? We could not expose the trait and implement a function on the `..Spec` type as well but then we can just as well not have the `Matchable` trait. 

Do we really want to abstract "matching" something for the benefit of having an abstraction? In the rest of the code, we will most likely not use this abstraction either .. 🤔

YeungOnion:
I think if we can't see a way to reuse it, then assume YAGNI. Thanks!

PR #347, opened 2023-09-21
Fix/auth fallback

ruben-arts:
The [authentication](https://github.com/prefix-dev/pixi/issues/334) issue was a result of the `~` not being recognized transformed in an actual home directory. This fixes that. 

This also creates a set of defaults, so `pixi` can use the same function as rattler.

### DIFF
--- a/crates/rattler_networking/src/lib.rs
+++ b/crates/rattler_networking/src/lib.rs
@@ -20,11 +20,21 @@ pub struct AuthenticatedClient {
     auth_storage: AuthenticationStorage,
 }
 
+/// Default location for fallback authentication storage.
+pub fn get_default_auth_store_location() -> PathBuf {
+    dirs::home_dir().unwrap().join(".rattler")
+}
+
+/// Setup default authentication storage
+pub fn default_authentication_storage() -> AuthenticationStorage {
+    AuthenticationStorage::new("rattler", &get_default_auth_store_location())
+}
+
 impl Default for AuthenticatedClient {
     fn default() -> Self {
         AuthenticatedClient {
             client: Client::default(),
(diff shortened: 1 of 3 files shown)


crates/rattler_networking/src/lib.rs
```diff
@@ -20,11 +20,21 @@ pub struct AuthenticatedClient {
     auth_storage: AuthenticationStorage,
 }
 
+/// Default location for fallback authentication storage.
+pub fn get_default_auth_store_location() -> PathBuf {
+    dirs::home_dir().unwrap().join(".rattler")
```
ruben-arts, line 25: Clean the unwrap

ruben-arts: Done

crates/rattler_networking/src/lib.rs
```diff
@@ -20,12 +22,24 @@ pub struct AuthenticatedClient {
...
-impl Default for AuthenticatedClient {
+/// Returns the default auth storage directory used by rattler.
+/// Would be placed in $HOME/.rattler, except when there is no home then it will be put in '/rattler/'
+pub fn default_auth_store_fallback_directory() -> &'static Path {
+    static FALLBACK_AUTH_DIR: OnceLock<PathBuf> = OnceLock::new();
+    FALLBACK_AUTH_DIR.get_or_init(|| {
+        dirs::home_dir()
+            .map(|home| home.join(".rattler/"))
+            .unwrap_or_else(|| {
+                tracing::warn!("Home directory could not be found, using '/rattler'");
```
baszalmstra, line 33:
```suggestion
                tracing::warn!("using '/rattler' to store fallback authentication credentials because the home directory could not be found");
```

baszalmstra (approved):

PR #359, opened 2023-09-28
Use SparseRepoData in fetch_repo_data

Wackyator:
This would allow more functionality as well as speed up the loading of repodata a little.


py-rattler/src/repo_data/sparse.rs
```diff
@@ -0,0 +1,102 @@
...
+}
+
+#[pymethods]
+impl PySparseRepoData {
+    #[new]
+    pub fn new(
+        channel: PyChannel,
+        subdir: String,
+        path: PathBuf,
+        patch_func: Option<&PyAny>,
```
baszalmstra, line 33: Maybe remove this as your not using it?

baszalmstra (approved):

PR #360, opened 2023-09-28
Nushell fixes

### LINKED ISSUE #306: Rename `behaviour` to `behavior` and always prefer US spelling
We currently have `PathModificationBehaviour` as a enum in the Rust code. 
IMO we should be consistent with US spelling (behavior).

wolfv:
There is one more fix I need to do.

Closes #306

### DIFF
--- a/crates/rattler_shell/src/shell/mod.rs
+++ b/crates/rattler_shell/src/shell/mod.rs
@@ -1,6 +1,6 @@
 //! This module contains the [`Shell`] trait and implementations for various shells.
 
-use crate::activation::PathModificationBehaviour;
+use crate::activation::PathModificationBehavior;
 use enum_dispatch::enum_dispatch;
 use itertools::Itertools;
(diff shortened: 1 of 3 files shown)


crates/rattler_shell/src/shell/mod.rs
```diff
@@ -422,12 +422,22 @@ impl Shell for Fish {
...
 pub struct NuShell;
 
 impl Shell for NuShell {
     fn set_env_var(&self, f: &mut impl Write, env_var: &str, value: &str) -> std::fmt::Result {
+        // escape backslashes for Windows (make them double backslashes)
+        let value = if cfg!(windows) {
+            escape_backslashes(value)
+        } else {
+            value.to_string()
+        };
```
baszalmstra, line 440: Dont you always want to escape these? Its not windows specific but more how nushell interprets a string right?

baszalmstra (approved):

PR #365, opened 2023-10-04
misc: simplify code and change warning to debug

### LINKED ISSUE #363: Reduce warning messages when using fallback auth storage
Example of the issue in https://github.com/prefix-dev/rattler-build/issues/178.

This is because the `get` function is called multiple times and every time it gives this warning. Which isn't even that interesting to the user. So maybe it should be a debug message.

ruben-arts:
closes #363 

This reduces the way the message used to scream at you, while it is totally normal to not have a keyring on your system. 

Still not the best as it might be better to give this warning only once but this should help.

### DIFF
--- a/crates/rattler_networking/src/authentication_storage/storage.rs
+++ b/crates/rattler_networking/src/authentication_storage/storage.rs
@@ -111,9 +111,9 @@ impl AuthenticationStorage {
                     e,
                     self.fallback_storage.path.display()
                 );
-                match self.fallback_storage.get_password(host)?{
-                    None => { return Ok(None) }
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

wolfv:
Hmm, ideally I think we would show it as a warning once because we end up storing the passwords in clear text which is not great.

ruben-arts:
@wolfv How would you approach that for this function? We could do a check on generation of the AuthenticationStorage ?

wolfv:
Agreed, it should be a warning when storing the token, not when retrieving it. :)

PR #375, opened 2023-10-11
Fix/xonsh extension

ruben-arts:
This now also updates how xonsh' `run_script` function works as it can source both bash and xsh scripts it should use a different `source` command depending on the extention of the file.


crates/rattler_shell/src/activation.rs
```diff
@@ -109,8 +109,14 @@ fn collect_scripts<T: Shell>(path: &Path, shell_type: &T) -> Result<Vec<PathBuf>
         .filter_map(|r| r.ok())
         .map(|r| r.path())
         .filter(|path| {
-            path.is_file()
-                && path.extension().and_then(OsStr::to_str) == Some(shell_type.extension())
+            let ext = path.extension().and_then(OsStr::to_str);
```
baszalmstra, line 112: This seems like a big hack. The shell type is already generic, why not add a function to determine if a certain file can be run be a shell?

ruben-arts: Put it in a function in the Shell trait

baszalmstra: Perfecto!

baszalmstra (approved):

PR #378, opened 2023-10-12
refactor: split lock and add pip

baszalmstra:
Splits the lockfile implementation into its own crate and adds support for pip packages. I assume this will need more changes but for an initial version it works. :)


crates/rattler_lock/src/conda.rs
```diff
@@ -0,0 +1,215 @@
...
+pub struct CondaLockedDependency {
+    /// What are its own dependencies mapping name to version constraint
+    #[serde(default)]
+    #[serde_as(deserialize_as = "crate::utils::serde::MatchSpecMapOrVec")]
+    pub dependencies: Vec<String>,
+    /// URL to find it at
+    pub url: Url,
+    /// Hashes of the package
+    pub hash: PackageHashes,
+    /// ???
```
wolfv, line 26: Do we know more about this field?

baszalmstra: No, I copied this from the previous implementation

PR #385, opened 2023-10-19
feat: add strict channel priority option

ruben-arts:
Skips options for a package in other channels if it is already found in channels that are defined first.

### DIFF
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -122,6 +122,7 @@ impl SparseRepoData {
         repo_data: impl IntoIterator<Item = &'a SparseRepoData>,
         package_names: impl IntoIterator<Item = PackageName>,
         patch_function: Option<fn(&mut PackageRecord)>,
+        strict_channel_priority: bool,
     ) -> io::Result<Vec<Vec<RepoDataRecord>>> {
         let repo_data: Vec<_> = repo_data.into_iter().collect();
 
@@ -136,8 +137,13 @@ impl SparseRepoData {
 
         // Iterate over the list of packages that still need to be processed.
         while let Some(next_package) = pending.pop_front() {
+            let mut found_in_channel = None;
             for (i, repo_data) in repo_data.iter().enumerate() {
                 let repo_data_packages = repo_data.inner.borrow_repo_data();
+                if strict_channel_priority && repo_data_packages.info == found_in_channel {
+                    continue;
+                }
+
                 let base_url = repo_data_packages
                     .info
                     .as_ref()
@@ -161,6 +167,10 @@ impl SparseRepoData {
                     patch_function,
                 )?;
                 records.append(&mut conda_records);
+
+                if strict_channel_priority && !records.is_empty() {
+                    found_in_channel = repo_data_packages.info.clone();
+                }
 
                 // Iterate over all packages to find recursive dependencies.
                 for record in records.iter() {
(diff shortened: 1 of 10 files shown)


crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -136,7 +137,17 @@ impl SparseRepoData {
...
+            let mut found_in_channel = None;
             for (i, repo_data) in repo_data.iter().enumerate() {
+                // If package was found in other channel, skip this repodata
+                if strict_channel_priority {
+                    if let Some(url) = found_in_channel.clone() {
+                        if repo_data.channel.base_url != url {
+                            continue;
+                        }
+                    }
+                }
```
baszalmstra, line 149: This can be done much shorter:

```suggestion
                if found_in_channel.map_or(false, |c| c != &repo_data.channel.base_url) {
                    continue;
                }
```

You dont need to check the `strict_channel_priority` because the `found_in_channel` is only set if that is true.

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -162,6 +173,10 @@ impl SparseRepoData {
                 )?;
                 records.append(&mut conda_records);
 
+                if strict_channel_priority && !records.is_empty() {
+                    found_in_channel = Some(repo_data.channel.base_url.clone());
+                }
```
baszalmstra, line 178: Just store a reference to forgo a clone:

```suggestion
                if strict_channel_priority && !records.is_empty() {
                    found_in_channel = Some(&repo_data.channel.base_url);
                }
```

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -430,9 +458,32 @@ mod test {
...
+
+        let sparse_data = load_sparse(["pytorch-cpu"], false).await;
+        let total_records = sparse_data
+            .iter()
+            .filter(|repo| {
+                repo.into_iter()
+                    .any(|package| package.file_name.contains("pytorch-cpu"))
+            })
+            .count();
+        assert_eq!(total_records, 2);
```
baszalmstra, line 481: I think the text could be made more explicit like this:

```suggestion
        // If we load pytorch-cpy from all channels (non-strict) we expect records from both
        // conda-forge and pthe pytorch channels.
        let sparse_data = load_sparse(["pytorch-cpu"], false).await;
        let channels = sparse_data
            .into_iter()
            .flatten()
            .filter(|record| record.package_record.name.as_normalized() == "pytorch-cpu")
            .map(|record| record.channel)
            .unique()
            .collect_vec();
        assert_eq!(
            channels,
            vec![
                String::from("https://conda.anaconda.org/conda-forge/"),
                String::from("https://conda.anaconda.org/pytorch/")
            ]
        );

        // If we load pytorch-cpy from strict channels we expect records only from the first channel
        // that contains the package. In this case we expect records only from conda-forge.
        let strict_sparse_data = load_sparse(["pytorch-cpu"], true).await;
        let channels = strict_sparse_data
            .into_iter()
            .flatten()
            .filter(|record| record.package_record.name.as_normalized() == "pytorch-cpu")
            .map(|record| record.channel)
            .unique()
            .collect_vec();
        assert_eq!(
            channels,
            vec![String::from("https://conda.anaconda.org/conda-forge/")]
        );
```

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -122,6 +122,7 @@ impl SparseRepoData {
         repo_data: impl IntoIterator<Item = &'a SparseRepoData>,
         package_names: impl IntoIterator<Item = PackageName>,
         patch_function: Option<fn(&mut PackageRecord)>,
+        strict_channel_priority: bool,
```
baszalmstra, line 125: It would also be good to add some documentation to this function on what `strict_channel_priority` means

crates/rattler_repodata_gateway/src/sparse/mod.rs
ruben-arts: That is much better!

crates/rattler_repodata_gateway/src/sparse/mod.rs
ruben-arts: Nice!

crates/rattler_repodata_gateway/src/sparse/mod.rs
ruben-arts: Fair, added it

baszalmstra (approved):
LGTM!

PR #388, opened 2023-10-30
make archive type `serialize` and `deserialize`

wolfv:
This would be useful to serialize the rattler-build outputs.

### DIFF
--- a/crates/rattler_conda_types/src/package/archive_type.rs
+++ b/crates/rattler_conda_types/src/package/archive_type.rs
@@ -1,7 +1,8 @@
+use serde::{Deserialize, Serialize};
 use std::path::Path;
 
 /// Describes the type of package archive. This can be derived from the file extension of a package.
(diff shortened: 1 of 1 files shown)


baszalmstra:
Maybe it helps to use: https://serde.rs/variant-attrs.html#rename_all ?

wolfv:
@baszalmstra would you rather have them serialized as lowercase, or as `tar.bz2` / `conda`? I don't have a strong preference. 
I more or less need this in rattler-build, but we could also skip serializing this option in rattler-build.

baszalmstra:
> @baszalmstra would you rather have them serialized as lowercase, or as `tar.bz2` / `conda`? I don't have a strong preference. I more or less need this in rattler-build, but we could also skip serializing this option in rattler-build.

I have no strong preference, but I think if you serialize this to yaml the most beautiful would be `snake_case` perhaps?

baszalmstra:
This seems to already be integrated in #392

PR #389, opened 2023-10-31
feat: add forward compatibility error

baszalmstra:
Adds an error when parsing a lockfile with a version higher than the currently supported version. 

Should make https://github.com/prefix-dev/pixi/issues/424 more clear.


ruben-arts (approved):
Danke

PR #391, opened 2023-11-01
Match the `channel` in the solve

ruben-arts:
@baszalmstra what do you think of something like this?

If do an unstrict channel priority this works with solving for channel specific packges.

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -279,7 +279,7 @@ pub struct NamelessMatchSpec {
 
 impl NamelessMatchSpec {
     /// Match a MatchSpec against a PackageRecord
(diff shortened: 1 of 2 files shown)


baszalmstra:
I dont think this is great because it changes the API. Maybe it would make more sense to add a match to the `RepoDataRecord` instead? But either way, you dont know in which form the channel is in the record (could be a url, could be a name). I think it would be better to add this to the solver so users also get better feedback on why a variant wasnt selected.

wolfv:
@baszalmstra I was discussing with Ruben about this yesterday. My approach idea was that we would filter the selected repodatas afterwards for strict priority / channel specificity.

Once we have support in resolvo to add "disallow_rules" we can turn the filter into adding a rule for each of the disallowed packages.

Maybe you can help explain how & where we could add such a filter.

ruben-arts:
Closed in favor of #394

PR #394, opened 2023-11-07
Feature/add channel prio and specific to solver info

ruben-arts:
- [x] tests 
- [x] Ability to specify a channel in the matchspec.
- [x] Only pick a record with a similar name from the first channel.


wolfv:
Looks good to me actually!

PR #396, opened 2023-11-08
Start implementing TUF for signed repodata 

wolfv:
This adds some simple validation functions, compatible with `conda-content-trust`.
It is just a POC.

There are some nice serde adaptors that we could use to decode to right types straight away. No signing is implemented yet.


baszalmstra:
Going to close this for now. We will revisit this later.

PR #398, opened 2023-11-08
fix: environment activation for windows

ruben-arts:
Closes https://github.com/prefix-dev/pixi/issues/437

The activation runs a this `openssl_activate.bat` when building the environment:
```batch
@echo off
if "%SSL_CERT_FILE%"=="" (
    set SSL_CERT_FILE=%CONDA_PREFIX%\Library\ssl\cacert.pem
    set __CONDA_OPENSLL_CERT_FILE_SET="1"
)
```

Which results in the following result when we call the `env` in cmdexe:
```
__CONDA_OPENSLL_CERT_FILE_SET="1"
```
This seems to be the only shell that does it that way. But our parser just keeps those `"` around which breaks because we add them again in our powershell `fn set_env_var` 

This PR strips the `"`.


baszalmstra (approved):
Very nice find!

PR #401, opened 2023-11-10
refactor: make the channel in the matchspec struct an actual channel

ruben-arts:
This should really help make installation more correct. 

I got the issue that `conda-forge::python` and `conda::python` got the same result after installation. 
So I changed the type of the channel to be a channel from the start. 

WDYT?

### DIFF
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 use std::path::{Component, Path, PathBuf};
 use std::str::FromStr;
 
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Serialize, Deserializer};
 use smallvec::SmallVec;
 use thiserror::Error;
 use url::Url;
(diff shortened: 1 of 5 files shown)


baszalmstra (commented):
If you turn it into an Arc<Channel> this looks good to me!

ruben-arts:
> If you turn it into an Arc this looks good to me!

Done, please check if that is correct.

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -38,7 +38,7 @@ impl Default for ChannelConfig {
 }
 
 /// `Channel`s are the primary source of package information.
-#[derive(Debug, Clone, Serialize, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Serialize, Deserialize, Eq, PartialEq, Hash)]
```
baszalmstra, line 41: This is no longer required right? Since you have a custom deserializer?

ruben-arts: Nup

ruben-arts: Removed

PR #406, opened 2023-11-16
refactor: Make various Record types interoperable 

Wackyator:
Adds:
  - Interoperability between the Record types
  - Additional missing fields in record types

TODO:
  [] Ensure docs
  [] Ensure tests
  [] Replace new `PyRecord` in places which previously used the individual `Py{x}Record` types


py-rattler/rattler/package/package_name.py
```diff
@@ -71,6 +71,15 @@ def normalized(self) -> str:
         """
         return self._name.normalized
 
+    def __eq__(self, other: object) -> bool:
+        """
+        Returns True if this instance represents the same PackageName as `other`.
```
baszalmstra, line 76: Maybe add a doctest so we dont break this in the future? :)

baszalmstra: A comparison with a string would also be nice to have! E.g. 

```python
PackageName("foo") == "foo"
```

Not required for this PR

py-rattler/rattler/prefix/prefix_record.py
```diff
@@ -3,25 +3,19 @@
...
-class PrefixRecord:
-    def __init__(self, source: str) -> None:
-        if not isinstance(source, str):
-            raise TypeError(
-                "PrefixRecord constructor received unsupported type "
-                f" {type(source).__name__!r} for the `source` parameter"
-            )
-        self._record = PyPrefixRecord(source)
+class PrefixRecord(RepoDataRecord):
+    _record: PyRecord
```
baszalmstra, line 12: Does `PackageRecord` also specify this? Since the type is always `PyRecord` there is no need to duplicate that here (or in `RepoDataRecord`).

py-rattler/rattler/prefix/prefix_record.py
```diff
@@ -3,25 +3,19 @@
...
 
     @classmethod
-    def _from_py_prefix_record(cls, py_prefix_record: PyPrefixRecord) -> PrefixRecord:
+    def _from_py_record(cls, py_record: PyRecord) -> PrefixRecord:
```
baszalmstra, line 15: I assume this exact method also exists in `PackageRecord`? I dont think you need to duplicate that here in that case.

But it would also be nice to assert that the `PyRecord` here is indeed a `PrefixRecord`.

py-rattler/rattler/repo_data/package_record.py
```diff
@@ -10,19 +14,220 @@ class PackageRecord:
...
-        cls, py_package_record: PyPackageRecord
+    @staticmethod
+    def from_index_json(
+        path: os.PathLike[str],
+        size: Optional[int] = None,
+        sha256: Optional[str] = None,
+        md5: Optional[str] = None,
     ) -> PackageRecord:
+        """
+        Builds a PackageRecord from a IndexJson.
```
baszalmstra, line 27:
```suggestion
        Builds a PackageRecord from an `index.json` file. These can be found in an extracted conda package archive.
```

py-rattler/rattler/repo_data/package_record.py
```diff
@@ -10,19 +14,220 @@ class PackageRecord:
...
+        regardless of the order of records and of the depends vector inside the records.
+        Note that this function only works for packages with unique names.
+
+        Examples
+        --------
+        ```python
+        >>> from os import listdir
+        >>> from os.path import isfile, join
+        >>> records = [
+        ...     PackageRecord.from_index_json(join("../test-data/conda-meta/", f))
```
baszalmstra, line 58: These `conda-meta` files are not actually `index.json` files. Would it be better to simply load them as PrefixRecords? 

```suggestion
        ...     PrefixRecord.from_path(join("../test-data/conda-meta/", f))
```

py-rattler/rattler/repo_data/package_record.py
```diff
@@ -10,19 +14,220 @@ class PackageRecord:
...
+        """
+        return [
+            PackageRecord._from_py_record(p)
+            for p in PyRecord.sort_topologically(records)
+        ]
+
+    @classmethod
+    def _from_py_record(cls, py_record: PyRecord) -> PackageRecord:
         """
         Construct Rattler PackageRecord from FFI PyPackageRecord object.
```
baszalmstra, line 76: Comment mismatches.

py-rattler/rattler/repo_data/package_record.py
```diff
@@ -10,19 +14,220 @@ class PackageRecord:
...
+
+    @property
+    def constrains(self) -> List[str]:
+        """
+        Additional constraints on packages.
+        constrains are different from depends in that packages
+        specified in depends must be installed next to this package,
+        whereas packages specified in constrains are not required to
+        be installed, but if they are installed they must follow
+        these constraints.
```
baszalmstra, line 105:
```suggestion
        Constrains are different from depends in that packages
        specified in depends must be installed next to this package,
        whereas packages specified in constrains are not required to
        be installed, but if they are installed they must follow
        these constraints.
```

py-rattler/rattler/repo_data/record.py
```diff
@@ -94,7 +73,7 @@ def file_name(self) -> str:
         return self._record.file_name
 
     @classmethod
-    def _from_py_record(cls, py_record: PyRepoDataRecord) -> RepoDataRecord:
+    def _from_py_record(cls, py_record: PyRecord) -> RepoDataRecord:
         """
         Construct Rattler RepoDataRecord from FFI PyRepoDataRecord object.
```
baszalmstra, line 78: Comment mismatch

py-rattler/src/error.rs
```diff
@@ -99,3 +104,4 @@ create_exception!(exceptions, IoException, PyException);
 create_exception!(exceptions, SolverException, PyException);
 create_exception!(exceptions, TransactionException, PyException);
 create_exception!(exceptions, LinkException, PyException);
+create_exception!(exceptions, ConvertSubdirException, PyException);
```
baszalmstra, line 107: `ConvertSubdirException` is missing both from the bindings as well as in the `error.py` file.

py-rattler/src/package_name.rs
```diff
@@ -54,4 +59,16 @@ impl PyPackageName {
     pub fn normalized(&self) -> String {
         self.inner.as_normalized().into()
     }
+
+    /// Compute the hash of the name.
+    fn __hash__(&self) -> u64 {
```
baszalmstra, line 64: Hash implementation in python seems to be missing

py-rattler/src/prefix_record.rs
```diff
@@ -1,138 +1,6 @@
...
-        self.inner.requested_spec.clone()
-    }
-
-    pub fn as_str(&self) -> String {
-        format!("{:?}", self.inner)
-    }
-}
+use pyo3::{pyclass, pymethods};
+use rattler_conda_types::prefix_record::PrefixPaths;
+use std::path::PathBuf;
```
baszalmstra, line 3: Can this file be removed?

py-rattler/src/record.rs
```diff
@@ -0,0 +1,406 @@
...
+
+#[derive(Clone)]
+pub enum RecordInner {
+    Prefix(PrefixRecord),
+    RepoData(RepoDataRecord),
+    Package(PackageRecord),
+}
+
+impl PyRecord {
+    pub fn as_package_record(&self) -> PyResult<&PackageRecord> {
```
baszalmstra, line 33: I think you can remove the `PyResult` from this function since it always succeeds.

py-rattler/src/record.rs
```diff
@@ -0,0 +1,406 @@
...
+
+use crate::{
+    error::PyRattlerError, package_name::PyPackageName, prefix_record::PyPrefixPaths,
+    version::PyVersion,
+};
+
+#[pyclass]
+#[repr(transparent)]
+#[derive(Clone)]
+pub struct PyRecord {
```
baszalmstra, line 21: Can you add some comments here that explain the structure of this type? With the enum etc. And why we do it this way?

baszalmstra: Please also explain why there is an `inner`.

py-rattler/src/record.rs
```diff
@@ -0,0 +1,406 @@
...
+        }
+    }
+}
+
+impl From<PyRecord> for PrefixRecord {
+    fn from(value: PyRecord) -> Self {
+        match value.inner {
+            RecordInner::Prefix(r) => r,
+            _ => panic!("invalid conversion tried!"),
+        }
```
baszalmstra, line 281: I guess this should be `TryFrom` instead of `From` given that it can fail.

py-rattler/src/record.rs
```diff
@@ -0,0 +1,406 @@
...
+
+impl From<RepoDataRecord> for PyRecord {
+    fn from(value: RepoDataRecord) -> Self {
+        Self {
+            inner: RecordInner::RepoData(value),
+        }
+    }
+}
+
+impl From<PyRecord> for RepoDataRecord {
```
baszalmstra, line 310: Should this also be `TryFrom`?

py-rattler/rattler/repo_data/package_record.py
Wackyator: Yeah this would work too, lemme do this instead

py-rattler/src/record.rs
Wackyator: It's true, but it makes `as_{x}_record` API a little inconsistent.

py-rattler/src/prefix_record.rs
Wackyator: Yeah, I'll rename it to `prefix_paths` since that's the only thing here and it's still being utilised

py-rattler/src/record.rs
Wackyator: Will change this thing.

baszalmstra: You could also rename the other functions into try_as_.. since they return a result?

baszalmstra (approved):
Did you still want to add doctests to all the getters of record?

py-rattler/rattler/repo_data/package_record.py
```diff
@@ -10,19 +14,414 @@ class PackageRecord:
...
+        ]
+
+    @classmethod
+    def _from_py_record(cls, py_record: PyRecord) -> PackageRecord:
+        """
+        Construct Rattler PackageRecord from FFI PyRecord object.
+        """
+
+        # quick sanity check
+        assert isinstance(py_record, PyRecord)
```
baszalmstra, line 82: Mm I meant more something along the lines of:

```py
assert py_record.is_package_record()
assert py_record.is_repodata_record()
assert py_record.is_prefix_record()
```

py-rattler/rattler/package/package_name.py
```diff
@@ -71,9 +71,67 @@ def normalized(self) -> str:
...
+    def __hash__(self) -> int:
+        """
+        Computes the hash of this instance.
+
+        Examples
+        --------
+        ```python
+        >>> hash(PackageName("test-abc")) == hash(PackageName("test-abc"))
+        True
+        >>> hash(PackageName("test-abc")) == hash(PackageName("abc-test"))
```
wolfv, line 83: If this wraps the rattler Package name, I think one of the key features is that it's case-insensitive.

So maybe checking that `hash(PackageName("test-abc")) == hash(PackageName("test-ABC"))`   would be a better check (same for equality).

baszalmstra (approved):

PR #407, opened 2023-11-16
fix: redact tokens from urls in errors

baszalmstra:
This PR adds redaction of tokens in URLs used by errors. I may have missed a few locations, but I think I got them all..


wolfv (approved):
Wow, amazing!

PR #415, opened 2023-11-24
refactor: rename pip to pypi in lock

baszalmstra:
This renames `pip` to `pypi` in lock files. `pip` is actually the installer and we don't really use that. When we talk about pip packages we usually mean pypi packages.

Since this is a forward-incompatible change I bumped the lock-file version to 3.

### DIFF
--- a/crates/rattler_lock/src/lib.rs
+++ b/crates/rattler_lock/src/lib.rs
@@ -165,11 +165,11 @@ impl LockedDependency {
     }
 
     /// Returns a reference to the internal [`PypiLockedDependency`] if this instance represents
-    /// a pip package.
-    pub fn as_pip(&self) -> Option<&PypiLockedDependency> {
+    /// a pypi package.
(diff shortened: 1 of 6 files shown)


crates/rattler_lock/src/lib.rs
```diff
@@ -160,16 +160,16 @@ impl LockedDependency {
...
         match &self.kind {
             LockedDependencyKind::Conda(conda) => Some(conda),
-            LockedDependencyKind::Pip(_) => None,
+            LockedDependencyKind::Pypi(_) => None,
         }
     }
 
-    /// Returns a reference to the internal [`PipLockedDependency`] if this instance represents
+    /// Returns a reference to the internal [`PypiLockedDependency`] if this instance represents
     /// a pip package.
```
ruben-arts, line 168: Should these `as_pip` and `is_pip` also change?

baszalmstra: Yes! Just did that :)

PR #416, opened 2023-11-24
fix: always write paths normalized

baszalmstra:
Make sure that when we write (relative) file paths, we normalize them.


wolfv (approved):
love it

PR #421, opened 2023-11-28
feat: download to disk before extracting

baszalmstra:
This changes the behavior of downloading packages. Decompressing and extracting the stream turns out to often be slower than downloading the file. This causes the download speed to decrease not utilizing the full bandwidth of the system.

This PR changes the behavior to first download the file to memory (or to disk if the file is too large) and only then extract the whole archive. In the future, we could possibly make this even more awesome by using a FIFO file queue so we can buffer the download to disk while still extracting at the same time.

I didn't do formal tests but on my machine, the downloads "feel" much faster.

### DIFF
--- a/crates/rattler_package_streaming/Cargo.toml
+++ b/crates/rattler_package_streaming/Cargo.toml
@@ -17,16 +17,17 @@ futures-util = { version = "0.3.28", optional = true }
 itertools = "0.11.0"
 rattler_conda_types = { version = "0.13.0", path = "../rattler_conda_types" }
 rattler_digest = { version = "0.13.0", path = "../rattler_digest" }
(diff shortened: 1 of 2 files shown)


wolfv:
LGTM 👍

ruben-arts:
I did some tests with installing pixi. 
I installed pixi with this version of rattler and compared it with the release v0.8.0 
| try | This rattler       | Old  rattler     |
|-----|-----------|-----------|
| 1   | **1m16,815s** | 1m22,916  |
| 2   | 1m12,818s | **1m3,479s**  |
| 3   | 1m6,628s  | **0m57,770s** |

wolfv (approved):

baszalmstra:
Since the benefit doesn't seem to be huge. I'll close this for now.

PR #423, opened 2023-11-29
improve lockfile version mismatch error

ruben-arts:
This might help the user
Would close https://github.com/prefix-dev/pixi/issues/504


baszalmstra:
Mm Im not really sure about this. It seems very specific at the wrong place. Maybe it would make more sense to add this behavior in pixi itself? 

Ideally we would propagate an error up the tree. Let me try something.

PR #427, opened 2023-12-04
fix: use filelock for authentication fallback storage 

baszalmstra:
The tests for the authentication storage randomly failed. I think this was caused by multiple threads writing to the fallback storage at the same time. This was only guarded by a mutex but multiple objects referring to the same file could exists. I replaced the mutex with a filelock to ensure that the multiple threads (and processes) are synchronized.

By request I also changed the logging behavior to ensure that the warning about fallback storage is only printed once (per file).

### DIFF
--- a/crates/rattler_networking/src/authentication_storage/fallback_storage.rs
+++ b/crates/rattler_networking/src/authentication_storage/fallback_storage.rs
@@ -1,8 +1,8 @@
 //! Fallback storage for passwords.
-use std::{
-    path::PathBuf,
-    sync::{Arc, Mutex},
-};
+use fslock::LockFile;
+use once_cell::sync::Lazy;
+use std::collections::{HashMap, HashSet};
+use std::{path::PathBuf, sync::Mutex};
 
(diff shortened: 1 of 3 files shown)


crates/rattler_networking/src/authentication_storage/fallback_storage.rs
```diff
@@ -30,44 +31,68 @@ pub enum FallbackStorageError {
...
     /// does not exist
-    fn read_json(&self) -> Result<std::collections::HashMap<String, String>, FallbackStorageError> {
+    fn read_json(&self) -> Result<HashMap<String, String>, FallbackStorageError> {
         if !self.path.exists() {
-            tracing::warn!(
-                "Can't find path for fallback storage on {}",
-                self.path.to_string_lossy()
-            );
-            return Ok(std::collections::HashMap::new());
+            static WARN_GUARD: Lazy<Mutex<HashSet<PathBuf>>> =
```
baszalmstra, line 86: @wolfv this should ensure that the warning is only printed once per filename for the current process.

PR #435, opened 2023-12-12
add option to force usage of fallback_auth_store

0xbe7a:
When running within a CI, it is sometimes easier not to rely on the native authentication toolchain, but to use a simple credentials file instead (see https://github.com/prefix-dev/rattler-build/issues/334). This PR adds an option to force the use of this fallback file and disable the use of the system toolchain.

### DIFF
--- a/crates/rattler_networking/src/authentication_storage/backends/keyring.rs
+++ b/crates/rattler_networking/src/authentication_storage/backends/keyring.rs
@@ -1,0 +1,96 @@
+//! Backend to store credentials in the operating system's keyring
+
+use std::str::FromStr;
+use keyring::Entry;
+use anyhow::Result;
+
+use crate::{authentication_storage::StorageBackend, Authentication};
+
+#[derive(Clone, Debug)]
+/// A storage backend that stores credentials in the operating system's keyring
+pub struct KeyringAuthenticationStorage {
+    /// The store_key needs to be unique per program as it is stored
+    /// in a global dictionary in the operating system
+    pub store_key: String,
+}
+
+impl KeyringAuthenticationStorage {
+    /// Create a new authentication storage with the given store key
+    pub fn from_key(store_key: &str) -> Self {
+        Self {
+            store_key: store_key.to_string(),
+        }
+    }
+}
+
+/// An error that can occur when accessing the authentication storage
+#[derive(thiserror::Error, Debug)]
+pub enum KeyringAuthenticationStorageError {
+    /// An error occurred when accessing the authentication storage
+    #[error("Could not retrieve credentials from authentication storage: {0}")]
+    StorageError(#[from] keyring::Error),
+
+    /// An error occurred when serializing the credentials
+    #[error("Could not serialize credentials {0}")]
+    SerializeCredentialsError(#[from] serde_json::Error),
+
+    /// An error occurred when parsing the credentials
+    #[error("Could not parse credentials stored for {host}")]
+    ParseCredentialsError {
+        /// The host for which the credentials could not be parsed
+        host: String,
+    },
+}
+
+impl Default for KeyringAuthenticationStorage {
+    fn default() -> Self {
+        Self::from_key("rattler")
+    }
+}
+
+impl StorageBackend for KeyringAuthenticationStorage {
+    fn store(
+        &self,
+        host: &str,
+        authentication: &Authentication,
+    ) -> Result<()> {
+        let password = serde_json::to_string(authentication)?;
+        let entry = Entry::new(&self.store_key, host)?;
+
+        entry.set_password(&password)?;
+
+        Ok(())
+    }
+
+    fn get(&self, host: &str) -> Result<Option<Authentication>> {
+        let entry = Entry::new(&self.store_key, host)?;
+        let password = entry.get_password();
+
+        let p_string = match password {
+            Ok(password) => password,
+            Err(keyring::Error::NoEntry) => {
+                return Ok(None);
+            }
+            Err(_) => return Ok(None),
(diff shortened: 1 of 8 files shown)


wolfv:
I think this is "OK" for me. Not a super-fan of the option :)

I am wondering if we should rather create another authentication storage that can be supplied instead of the current one. The new one would rely only on the fallback storage and uses plain JSON as database, and then we would have a Rust trait for an authentication storage?

0xbe7a:
Yeah, that sounds a bit more elegant.

crates/rattler_networking/src/authentication_storage/storage.rs
```diff
@@ -140,6 +162,10 @@ impl AuthenticationStorage {
             cache.remove(host);
         }
 
+        if self.force_fallback_storage {
```
wolfv, line 165: At the end we do also delete the password from any fallback storage. So maybe this is not needed here?

wolfv (changes requested):

0xbe7a:
I still need to test this with rattler-build

wolfv:
Nice work! Looks good to me. If we can get rid of anyhow it's perfect :)

0xbe7a:
I adapted and tested the rattler-build PR with our real-world example. Works great! I also changed the file location of the `.json` file, as the format changed slightly and we now store valid JSON.

0xbe7a:
Would you prefer to return a `Box<dyn Error + Send + Sync>`instead of an anyhow `Result`? Since I wanted to leave the option for users to bring your own backend, you can't really type the error type.

wolfv:
right. let's try to ask @baszalmstra what he thinks is best.

crates/rattler_networking/src/authentication_storage/backends/keyring.rs
```diff
@@ -0,0 +1,91 @@
...
+        Ok(())
+    }
+
+    fn get(&self, host: &str) -> Result<Option<Authentication>> {
+        let entry = Entry::new(&self.store_key, host)?;
+        let password = entry.get_password();
+
+        let p_string = match password {
+            Ok(password) => password,
+            Err(keyring::Error::NoEntry) => {
```
baszalmstra, line 67: Both err statements seem to donthe same.

crates/rattler_networking/src/authentication_storage/storage.rs
```diff
@@ -1,155 +1,100 @@
...
+/// This struct implements storage and access of authentication
+/// information backed by multiple storage backends
+/// (e.g. keyring and file storage)
+/// Credentials are stored and retrieved from the backends in the
+/// order they are added to the storage
 pub struct AuthenticationStorage {
-    /// The store_key needs to be unique per program as it is stored
-    /// in a global dictionary in the operating system
-    pub store_key: String,
+    backends: Vec<Arc<Box<dyn StorageBackend + Send + Sync>>>,
```
baszalmstra, line 24: Arc Box seems superfluous. Maybe just Arc is enough?

crates/rattler_networking/src/authentication_storage/storage.rs
```diff
@@ -190,4 +135,20 @@ impl AuthenticationStorage {
...
+            cache.insert(host.to_string(), None);
+        }
+
+        for backend in &self.backends {
+            if let Err(e) = backend.delete(host) {
+                tracing::warn!("Error deleting credentials from backend: {}", e);
+            }
+        }
+
+        Ok(())
```
baszalmstra, line 152: Shouldnt this return an error if it failed?

baszalmstra (commented):
Thanks for this PR, almost looks good to me. I left some small comments.

And I would just use anyhow Result.

crates/rattler_networking/src/authentication_storage/storage.rs
0xbe7a: We want Arc for `Clone` and need to Box to store the `dyn Trait`

crates/rattler_networking/src/authentication_storage/storage.rs
0xbe7a: What does "fail" mean here? An attempt is made to delete the credentials in _all_ known backends, regardless of whether the previous backend returned an error.

crates/rattler_networking/src/authentication_storage/storage.rs
baszalmstra: but cant you use the arc for dyn too?

crates/rattler_networking/src/authentication_storage/storage.rs
baszalmstra: Well each individual backend might fail but this function will still report success. I think if any of them fail the function should return an error too.

crates/rattler_networking/src/authentication_storage/storage.rs
0xbe7a: You're right, I didn't know that. Thanks!

crates/rattler_networking/src/authentication_storage/storage.rs
0xbe7a: But then the fallback behavior differs from `store/get`, where errors are only logged and the next backend is used. What do you think about returning an error if all backends fail, which would be consistent with the other methods?

baszalmstra: Yeah that sounds good! Maybe that should also be the behavior for the other methods?

baszalmstra (approved):

wolfv (approved):

PR #436, opened 2023-12-12
Move index functionality to rattler and create python bindings

BenjaminLowry:
Moves the functionality for indexing a local conda channel from [rattler-build](https://github.com/prefix-dev/rattler-build/blob/main/src/index.rs) to a new crate in rattler and create python bindings so that you can index dependencies via py-rattler as faster alternative to using something like conda index. This could happen after solving and downloading dependencies from the python bindings, for example.

This would replace the logic in rattler-build and thus I will create a follow up PR to remove this logic and instead use this new crate in rattler-build.


BenjaminLowry:
@wolfv @baszalmstra Hey! This is my first attempt at doing what we discussed of moving the index functionality to rattler. There are some higher level design decisions that happy to talk about further like I created a new crate for the index logic, and the interface I created for the python function may not be what you expected, so let me know what you think. I didn't change any of the indexing logic.

I tried running most of the CI tasks manually but still might need to fix some things, could I get approved for running CI?

py-rattler/Cargo.toml
```diff
@@ -48,7 +49,7 @@ tokio = { version = "1.32" }
 thiserror = "1.0.44"
 url = "2.4.1"
 
-openssl = { version = "0.10", optional = true }
+openssl = { version = "0.10", optional = true, features = ["vendored"] }
```
baszalmstra, line 52: We have features in this crate that control this behavior. See rustls-tls, nativetls and openssl-vendored.

py-rattler/rattler/index/index.py
```diff
@@ -0,0 +1,31 @@
...
+    Will always index the "noarch" subdirectory, and thus this subdirectory should always be present, because
+    conda channels at a minimum must include this subdirectory.
+
+    Arguments:
+        channel_directory: A `os.PathLike[str]` that is the directory containing subdirectories
+                           of dependencies to index.
+        target_platform(optional): A `Platform` to index dependencies for
+
+    Returns:
+        True iff indexing was successful
```
baszalmstra, line 26:
```suggestion
        True if indexing was successful
```

baszalmstra: I think this boolean might not be necessary if the only alternative to false is an exception.

crates/rattler_index/src/lib.rs
```diff
@@ -0,0 +1,195 @@
+//! Indexing of packages in a output folder to create up to date repodata.json files
```
baszalmstra, line 1: Please add deny(missing_docs) attribute.

crates/rattler_index/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+homepage.workspace = true
+repository.workspace = true
+license.workspace = true
+readme.workspace = true
+
+[dependencies]
+fs-err = "2.11.0"
+rattler_conda_types = { version = "0.14.0", path = "../rattler_conda_types" }
+rattler_digest = { version = "0.14.0", path = "../rattler_digest" }
+rattler_package_streaming = { version = "0.14.0", path = "../rattler_package_streaming" }
```
baszalmstra, line 17: Please disable default features by default (for rattler crates at least).

py-rattler/src/index.rs
```diff
@@ -0,0 +1,16 @@
+use pyo3::{pyfunction, PyResult};
+use rattler_conda_types::Platform;
+use rattler_index::index;
+
+use std::path::PathBuf;
+
+use crate::{error::PyRattlerError, platform::PyPlatform};
+
+#[pyfunction]
+pub fn py_index(channel_directory: PathBuf, target_platform: Option<PyPlatform>) -> PyResult<bool> {
```
baszalmstra, line 10: You might be able to release the GIL here which makes it easier to run this function on a thread.

baszalmstra (commented):
Looks good! I left some superficial comments.

py-rattler/src/index.rs
BenjaminLowry: Okay yeah, I added allow_threads, lmk if that's not what you meant

py-rattler/rattler/index/index.py
BenjaminLowry: Yeah that makes sense, and honestly I think boolean could be bad too because hypothetically returning false gives us absolutely no information and thus shouldn't be a valid result. So changed to no boolean so either finishes or throws an exception

py-rattler/Cargo.toml
BenjaminLowry: So I added this just as an attempt to fix the issues I'm seeing in the CI related to the openssl-sys crate. Do you have any idea what's going on? I ran this successfully on a linux machine I have access to, and when I uninstalled libssl-dev then I got a similar error message, but not sure how my PR could have caused this. Possibly the latest version of openssl-sys changed? Curious if you've seen this before

wolfv:
@BenjaminLowry to get rid of the `Unverified` badge, maybe you want to use the same email address in your git config that you are using on Github? :)

Thanks for doing this, by the way, I am excited to get rid of this in rattler-build!

BenjaminLowry:
@wolfv Yeah absolutely! I'm excited to start using it too. Fixed the verification

py-rattler/Cargo.toml
baszalmstra: There are two libraries that provide TLS for our networking operations (on linux). `openssl` (through `native-tls`) and `rustls`. We default to use `native-tls` (so `openssl` on linux) but allow using `rustls`. We use `rustls` for the single executable binaries of `pixi` but use `openssl` when building for conda-forge.

The problem is that if one rattler crate depends on another rattler crate you need to propagate the proper tls features otherwise the "default" features "leak" through the compilation.

In CI we add the proper features. When we build the wheels for py-rattler we [add the `vendored-openssl` feature](https://github.com/mamba-org/rattler/blob/66cad536862c28bae09f80efd8fe29442822194b/.github/workflows/release-python.yml#L95). 

So, in essence, you can remove this change and use features to get the desired behavior for different targets.

crates/rattler_index/src/lib.rs
```diff
@@ -0,0 +1,197 @@
...
+
+        let mut repodata = RepoData {
+            info: Some(ChannelInfo {
+                subdir: platform.clone(),
+                base_url: None,
+            }),
+            packages: Default::default(),
+            conda_packages: Default::default(),
+            removed: Default::default(),
+            version: Some(1),
```
baszalmstra, line 162:
```suggestion
            version: Some(2),
```

This can be `2`. We implemented https://github.com/conda-incubator/ceps/pull/61 which bumps the version to 2. I agree that this would be better implemented into the writing itself. ;)

baszalmstra (approved):
I left two small comment but otherwise looks good to me!

BenjaminLowry:
@baszalmstra Thanks, addressed!

baszalmstra:
Thanks! If you can remove the vendored feature from openssl then Ill merge!

baszalmstra:
Ah that is already the case! athanks!

PR #437, opened 2023-12-15
fix: consistent clobbering & removal of `__pycache__`

wolfv:
This attempts to fix two issues:

- remove empty `__pycache__` folders for `noarch` packages (otherwise the package is still "detected" in the `site-packages` folder).
- do not attempt to link individual files more than once per installation. This happens when there are "clobbering" files. Currently a random file will win (due to async execution). 

However, there is still two remaining issues with clobbering:

- We should deterministically decide on a "winner". I think it should be by topologically sorting the packages and take the last (top) package as a winner.
- When removing packages we also need to take care of the (un-)clobbering. This is a bit more tricky because we might not even have the clobbering package in the package cache anymore. However, I think a workaround could be to still link the files into the environment but with the name of the source package appended (e.g. `crates2.json` would become `crates2.json-clobber-from-mypackage`. For unclobbering we could just rename the file.


wolfv:
cc @Wackyator we could also re-use the `unlink_package` in the Python bindings! :)

crates/rattler/src/install/mod.rs
```diff
@@ -248,6 +251,21 @@ pub async fn link_package(
...
+        for entry in paths_json.paths.iter() {
+            let path = entry.relative_path.clone();
+
+            if let Entry::Vacant(e) = paths_registry.entry(path.clone()) {
+                e.insert(index_json.name.as_normalized().to_string());
+            } else {
+                clobber_paths.insert(path);
+            }
+        }
+    }
```
baszalmstra, line 267: I think it makes sense to move this clobber registry into its own type and have a function on that type that does this. It splits up and compartmentalizes the logic a little better. 

I also feel like the overhead of cloning the package name hundreds of times (once for every path) is pretty resource heavy. If you create an object for the registry you could also have a lookup from index to name. You can then store indices in the map from path to package name instead.

Would also be nice if you could create the registry from an existing set of `PrefixRecords` to ensure you don't overwrite existing files maybe.

crates/rattler/Cargo.toml
```diff
@@ -25,11 +25,13 @@ dirs = "5.0.1"
 futures = "0.3.28"
 fxhash = "0.2.1"
 hex = "0.4.3"
+indexmap = "2.1.0"
 itertools = "0.11.0"
 memchr = "2.6.4"
 memmap2 = "0.7.1"
 nom = "7.1.3"
 once_cell = "1.18.0"
+parking_lot = "0.12.1"
```
baszalmstra, line 34: I dont think you're actually using this.

baszalmstra (commented):
Nice! I like the simplicity!

crates/rattler/src/install/mod.rs
wolfv: yep, gonna take care of this next week!

Wackyator:
looks awesome, will add this to the bindings :)

crates/rattler/src/install/driver.rs
```diff
@@ -130,6 +140,31 @@ impl InstallDriver {
...
+    pub fn clobber_registry(&self) -> MutexGuard<ClobberRegistry> {
+        self.clobber_registry.lock().unwrap()
+    }
+
+    /// Call this after all packages have been installed to perform any post processing that is
+    /// required.
+    ///
+    /// This function will select a winner among multiple packages that might write to a single package
+    /// and will also execute any `post-link.sh/bat` scripts
+    pub fn post_process(
```
baszalmstra, line 155: Would it make sense to give this a more descriptive name, like `unclobber` or something similar?

baszalmstra (commented):
I like this a lot @wolfv ! Lets get it merged asap!

wolfv:
OK, now I just need to write some tests for the unlinking parts and then I think we're done with this one :) ... the easiest would be to resolve some actual packages, install, then unlink them. But I don't think we have a test harness for that yet.

wolfv:
@baszalmstra when you get around to review this, please make sure to really carefully review the `unlink` parts because if that goes south, we could delete user data which we definitely would not want!

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -0,0 +1,648 @@
...
+        } else {
+            self.package_names
+                .push(PackageName::try_from(name).unwrap());
+            self.package_names.len() - 1
+        };
+
+        for entry in paths_json.paths.iter() {
+            let path = entry.relative_path.clone();
+
+            if let Entry::Vacant(e) = self.paths_registry.entry(path.clone()) {
```
baszalmstra, line 128: I think it would be better to use `get` here instead of `entry`. So you dont have to clone the `path`. Only if no entry exists yet.

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -0,0 +1,648 @@
...
+                e.insert(name_idx);
+            } else {
+                // rename to
+                let mut new_path = path.clone();
+                new_path.set_file_name(Self::clobber_name(&path, &self.package_names[name_idx]));
```
baszalmstra, line 133: Do you need the `new_path.set_file_name(` part here? Seems duplicate as you also use clone and `set_file_name` in the `clobber_name` function.

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -0,0 +1,648 @@
...
+
+                self.clobbers
+                    .entry(path.clone())
+                    .or_insert_with(|| vec![*self.paths_registry.get(&path).unwrap()])
```
baszalmstra, line 137: The lookup seems duplicate here since you already index `self.path_registry` above.

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -0,0 +1,648 @@
...
+        new_path.set_file_name(format!(
+            "{}{}",
+            file_name.to_string_lossy(),
+            clobber_template(package_name),
+        ));
+        new_path
+    }
+
+    /// Register the paths of a package when linking the package.
+    pub fn register_paths(
```
baszalmstra, line 105: Would be nice to describe a little more when this function is called (before actually linking files) and what the returned value is.

crates/rattler/src/install/driver.rs
wolfv: idk. post_process is more generic so we could potentially do some other things as well in this step ...

crates/rattler/src/install/clobber_registry.rs
wolfv: :+1:

wolfv:
@baszalmstra I took care of your comments. Two things we might still want to do:

- test "deeper nested" clobbers (`bla/bli/clobber.txt`).
- add something about clobbers to `conda-meta` files.

crates/rattler/src/install/driver.rs
```diff
@@ -69,18 +74,23 @@ impl InstallDriver {
...
                     }}
                 }
             }
         });
+
+        let clobber_registry = if let Some(prefix_records) = prefix_records {
+            ClobberRegistry::from_prefix_records(prefix_records)
+        } else {
+            ClobberRegistry::default()
+        };
```
baszalmstra, line 88:
```suggestion
        let clobber_registry = prefix_records.map(ClobberRegistry::from_prefix_records).unwrap_or_default();
```

PR #441, opened 2023-12-15
Default value for conda_packages in repodata.json

BenjaminLowry:
I am using rattler in a context that has to deal with _really_ old repodata.json files that don't have the "packages.conda" field and thus they fail with this error message:

```
called `Result::unwrap()` on an `Err` value: Custom { kind: InvalidData, error: Error("missing field `packages.conda`", line: 205, column: 1) }
```

I recognize that packages.conda has been created by conda build/index for years now, but this means rattler isn't backwards compatible with the (very) old state of repodatas and it is not trivial to upgrade the repodatas that I am dealing with. Without this, I will be potentially blocked from using rattler in important contexts. Rattler would still produce this field as an empty map by default, and it would default to an empty map when it's missing from a repodata.json that it deserializes. 

Thoughts? Also let me know if I'm missing other changes necessary in order to do this.

### DIFF
--- a/crates/rattler_conda_types/src/repo_data/mod.rs
+++ b/crates/rattler_conda_types/src/repo_data/mod.rs
@@ -38,7 +38,7 @@ pub struct RepoData {
 
     /// The conda packages contained in the repodata.json file (under a different key for
     /// backwards compatibility with previous conda versions)
-    #[serde(rename = "packages.conda", serialize_with = "sort_map_alphabetically")]
+    #[serde(default, rename = "packages.conda", serialize_with = "sort_map_alphabetically")]
     pub conda_packages: FxHashMap<String, PackageRecord>,
 
     /// removed packages (files are still accessible, but they are not installable like regular packages)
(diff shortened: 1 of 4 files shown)


baszalmstra (approved):
Makes sense to me! Thanks!

baszalmstra:
If you can fix the formatting issue, Ill merge it!

BenjaminLowry:
Thanks for the lightning fast review!

baszalmstra:
Oh you also need to change this in the sparse repodata reading. in sparse/mod.rs from the top of my mind.

BenjaminLowry:
@baszalmstra Okay added there as well, I didn't add a similar test as what exists in that file because that recursive repodata requires a channel url, unless I could give it a dummy one?

BenjaminLowry:
And I think windows flaked?

baszalmstra:
I think you can pass it a file path as well, but dont worry about it. The change is small enough. Ill merge when CI turns green!

PR #442, opened 2023-12-16
fix wildcard expansion for stored credentials of domains

0xbe7a:
While testing the `auth-file` feature for `rattler-build` (https://github.com/prefix-dev/rattler-build/pull/413) I noticed that the wildcard extension for domains does not work correctly:

`https://repo.prefix.dev` was matched with `repo.prefix.dev` and `*.repo.prefix.dev`, but not with `*.prefix.dev`. On closer inspection, it looks like the function never worked properly: See https://github.com/mamba-org/rattler/pull/252

This PR fixes the wildcard handling and adds a unit test for it.

### DIFF
--- a/crates/rattler_networking/src/authentication_storage/storage.rs
+++ b/crates/rattler_networking/src/authentication_storage/storage.rs
@@ -119,7 +119,7 @@ impl AuthenticationStorage {
                         Some(domain) => domain,
                         None => return Ok((url, None)),
                     };
-                    
+
                     let mut splits = domain.split('.').collect::<Vec<&str>>();
 
                     while !splits.is_empty() {
@@ -133,7 +133,7 @@ impl AuthenticationStorage {
                             Ok(None) => {
                                 splits.remove(0);
                             }
-                            _ => {},
+                            _ => {}
                         }
                     }
 
--- a/crates/rattler_networking/src/authentication_storage/storage.rs
+++ b/crates/rattler_networking/src/authentication_storage/storage.rs
@@ -109,49 +109,42 @@ impl AuthenticationStorage {
         url: U,
     ) -> Result<(Url, Option<Authentication>), reqwest::Error> {
         let url = url.into_url()?;
-        if let Some(host) = url.host_str() {
-            let credentials = self.get(host);
+        let host = match url.host_str() {
+            Some(host) => host,
+            None => return Ok((url, None)),
+        };
 
-            let credentials = match credentials {
-                Ok(None) => {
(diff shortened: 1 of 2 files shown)


crates/rattler_networking/src/authentication_storage/storage.rs
```diff
@@ -115,10 +115,29 @@ impl AuthenticationStorage {
             let credentials = match credentials {
                 Ok(None) => {
                     // Check for credentials under e.g. `*.prefix.dev`
-                    let mut parts = host.rsplitn(2, '.').collect::<Vec<&str>>();
-                    parts.reverse();
-                    let wildcard_host = format!("*.{}", parts.join("."));
-                    self.get(&wildcard_host)
+                    let domain = match url.domain() {
```
baszalmstra, line 118: Perfect usecase for let Some(domain) = ... else { return ... } syntax!

(Sorry im on a phone)

crates/rattler_networking/src/authentication_storage/storage.rs
```diff
@@ -115,10 +115,29 @@ impl AuthenticationStorage {
...
+                        Some(domain) => domain,
+                        None => return Ok((url, None)),
+                    };
+
+                    let mut splits = domain.split('.').collect::<Vec<&str>>();
+
+                    while !splits.is_empty() {
```
baszalmstra, line 125: Instead of allocating a Vec you could also use something like


while let Some((_, rest)) = domain.split_once('.') {
  

  domain = rest
}

baszalmstra (commented):
Thanks! I left some small notes but the idea is great!

crates/rattler_networking/src/authentication_storage/storage.rs
0xbe7a: I wanted the early return instead of nesting everything and i think the match syntax is cleaner for this

baszalmstra: I think you misunderstand, I mean this syntax: https://doc.rust-lang.org/beta/rust-by-example/flow_control/let_else.html. 

Its relatively new syntax similar to a match statement but easier to read in cases like this. It also allows an early return.

0xbe7a: Wow, that's some nifty syntax, thanks!

baszalmstra (approved):

baszalmstra:
Thanks @0xbe7a !

PR #446, opened 2023-12-18
Allow * (wildcard) in package_name for pixi search

Johnwillliam:
Allow * (wildcard) in package_name for pixi search.

Still unsure what kind of impact this could have on other places, anyone that could help with this?

https://github.com/prefix-dev/pixi/issues/378

### DIFF
--- a/crates/rattler_conda_types/src/package_name.rs
+++ b/crates/rattler_conda_types/src/package_name.rs
@@ -49,7 +49,7 @@ impl PackageName {
 #[derive(Clone, Debug, Error)]
 pub enum InvalidPackageNameError {
     /// The package name contains illegal characters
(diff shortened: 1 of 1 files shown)


baszalmstra:
I dont understand how this helps search? I dont think* is a legal character in a package name so im not inclined to merge this.

Johnwillliam:
I added a possibility to use a wildcard in the search in pixi, wasnt sure how to implement this without allowing “*”. If you check the pr i linked maybe that makes more sense?

baszalmstra:
Would it not be possible to use the string representation of a package name and match on that?

Johnwillliam:
Ahh you're right, I didn't see it correctly. I will fix it later in pixi!

Johnwillliam:
Ahh you're right, I didn't see it correctly. I will fix it later in pixi!

PR #452, opened 2023-12-19
feat: refactor `Version.bump()` to accept bumping `major/minor/patch/last`

hadim:
See https://github.com/mamba-org/rattler/issues/449 for context.

- [x] add new enum `VersionBumpType`
- [x] refactor `Version.bump()` to `Version.bump(bump_type: VersionBumpType)`
- [x] add unit tests
- [x] adapt the python binding

This PR is breaking backward compat but since rust does not support overloading or default args, I could not find a more creative way to keep backward compat. That being said, it might not be that much of a problem. If it is, please let me know, and I can always keep `bump()` as it is and create a new function.


hadim:
I would also need some help/guidance for the python binding.

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -239,6 +242,31 @@ impl Version {
             flags = flags.with_has_epoch(true);
         }
 
+        // Sanity check whether the version has enough segments for this bump type.
+        let segment_count = self.segment_count();
+        match bump_type {
```
baszalmstra, line 247: It would be better to return an error instead of panic.

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -1347,4 +1388,97 @@ mod test {
             Version::from_str("3!4.5a.6b").unwrap()
         );
     }
+
+    #[test]
+    fn bump_dev() {
+        let version = Version::from_str("5!1.alpha+3.4").unwrap();
```
baszalmstra, line 1394: This is not testing anything is it?

baszalmstra:
> I would also need some help/guidance for the python binding.

@Wackyator ?

crates/rattler_conda_types/src/version/mod.rs
hadim: Dev artifact. Will remove.

crates/rattler_conda_types/src/version/bump.rs
```diff
@@ -0,0 +1,26 @@
...
+    Minor,
+    /// Bump the patch version number.
+    Patch,
+    /// Bump the last  version number.
+    Last,
+}
+
+/// VersionBumpError is used to specify the type of error that occurred when bumping a version.
+#[derive(Debug, Clone)]
+pub struct VersionBumpError {
```
baszalmstra, line 17: Please use thiserror for error types.

hadim: I refactored `VersionBumpError`. let me know if it's good enough.

Wackyator:
Hey @hadim, this looks great, appreciate the PR! I can help you with the Python bindings if you have any kind of issues.

hadim:
> Hey @hadim, this looks great, appreciate the PR! I can help you with the Python bindings if you have any kind of issues.

I am not sure how to 1) propagate the Result and the Error 2) integrate the new enum VersionBumpType

Wackyator:
> > Hey @hadim, this looks great, appreciate the PR! I can help you with the Python bindings if you have any kind of issues.
> 
> I am not sure how to 1) propagate the Result and the Error 2) integrate the new enum VersionBumpType

For propagating the Result and Error, you can return the [PyResult](https://docs.rs/pyo3/latest/pyo3/type.PyResult.html) type from functions & extend the [`PyRattlerError`](https://github.com/mamba-org/rattler/blob/25ea070edde53c8097222e6d2d1cecff3e36037d/py-rattler/src/error.rs#L18) and create the exception for it.
After that you can simply do a `PyRattlerError::from` the errors.
You can see an example of [here](https://github.com/mamba-org/rattler/blob/main/py-rattler/src/index.rs)

For the `VersionBumpType` I'd say introduce seperate functions for them instead of making an enum on the python side, that seems more pythonic.

hadim:
@Wackyator @baszalmstra: LGTM - ready for another round of review. Thanks!

crates/rattler_conda_types/src/version/bump.rs
```diff
@@ -0,0 +1,31 @@
...
+#[derive(Clone)]
+pub enum VersionBumpType {
+    /// Bump the major version number.
+    Major,
+    /// Bump the minor version number.
+    Minor,
+    /// Bump the patch version number.
+    Patch,
+    /// Bump the last  version number.
+    Last,
```
wolfv, line 13: Can you add a `Segment(uint)` type? So that we can explicitly select the segment to bump?

We might also consider that bumping non-existent segments should be possible. Usually, all non-existent segments are implicit `.0`, so bumping the `patch` of a `1.0` could also result in `1.0.1`.

py-rattler/src/version/mod.rs
```diff
@@ -132,10 +132,35 @@ impl PyVersion {
         }
     }
 
-    /// Returns a new version where the last numerical segment of this version has been bumped.
-    pub fn bump(&self) -> Self {
-        Self {
-            inner: self.inner.bump(),
+    /// Returns a new version where the major segment of this version has been bumped.
+    pub fn bump_major(&self) -> PyResult<Self> {
```
wolfv, line 136: If we had that integer option we could pass an int from Python (with -1 being the last element, for example), and bump like that. 
Then in Python we could also define an enum for the different named segments and allow both, passing an enum or an int.

wolfv:
Really nice work. I had some simple remarks. Let me know what you think!

crates/rattler_conda_types/src/version/bump.rs
hadim: > Can you add a Segment(uint) type? So that we can explicitly select the segment to bump?

Good idea. Will do.

hadim: > We might also consider that bumping non-existent segments should be possible. Usually, all non-existent segments are implicit .0, so bumping the patch of a 1.0 could also result in 1.0.1.

Don't you prefer to instead fail, so it also builds incentive for the user to explicitly use `x.x.x` instead of `x` or `x.x`?

Unless you feel strongly here, I would prefer to let it for another PR (I can open a ticket if you want).

py-rattler/src/version/mod.rs
hadim: I am fine with Python API.

What we have so far is:

- `bump_major()`
- `bump_minor()`
- `bump_patch()`
- `bump_last()`

I can add the below as well:

- `bump_segment(int)`
- `bump(enum)`

Does that sound good to you?

crates/rattler_conda_types/src/version/bump.rs
hadim: Added support for `VersionBumpType::Segment(i32)` in 9283770893b34b895694b1b28652ffbb618db55f (rust side only).

py-rattler/src/version/mod.rs
hadim: Python side done at https://github.com/mamba-org/rattler/pull/452/commits/4ac1858337f0039d950dcda5408caea378abe917

I didn't implement `bump(enum)` since it seems a bit overkill at this point, but let me know if you still want it.

py-rattler/src/version/mod.rs
```diff
@@ -132,10 +132,43 @@ impl PyVersion {
...
+        match self.inner.bump(VersionBumpType::Major) {
+            Ok(v) => Ok(Self { inner: v }),
+            Err(e) => Err(PyRattlerError::from(e).into()),
+        }
+    }
+
+    /// Returns a new version where the minor segment of this version has been bumped.
+    pub fn bump_minor(&self) -> PyResult<Self> {
```
Wackyator, line 144: Same as above

py-rattler/src/version/mod.rs
```diff
@@ -132,10 +132,43 @@ impl PyVersion {
...
+        match self.inner.bump(VersionBumpType::Minor) {
+            Ok(v) => Ok(Self { inner: v }),
+            Err(e) => Err(PyRattlerError::from(e).into()),
+        }
+    }
+
+    /// Returns a new version where the patch segment of this version has been bumped.
+    pub fn bump_patch(&self) -> PyResult<Self> {
```
Wackyator, line 152: Same as above

py-rattler/src/version/mod.rs
```diff
@@ -132,10 +132,43 @@ impl PyVersion {
...
+        match self.inner.bump(VersionBumpType::Patch) {
+            Ok(v) => Ok(Self { inner: v }),
+            Err(e) => Err(PyRattlerError::from(e).into()),
+        }
+    }
+
+    /// Returns a new version where the last segment of this version has been bumped.
+    pub fn bump_last(&self) -> PyResult<Self> {
```
Wackyator, line 160: Same as above

py-rattler/src/version/mod.rs
```diff
@@ -132,10 +132,43 @@ impl PyVersion {
...
+        match self.inner.bump(VersionBumpType::Last) {
+            Ok(v) => Ok(Self { inner: v }),
+            Err(e) => Err(PyRattlerError::from(e).into()),
+        }
+    }
+
+    /// Returns a new version where the given segment of this version has been bumped.
+    pub fn bump_segment(&self, index: i32) -> PyResult<Self> {
+        match self.inner.bump(VersionBumpType::Segment(index)) {
```
Wackyator, line 169: Same as above

py-rattler/tests/unit/test_version.py
```diff
@@ -17,3 +18,26 @@ def test_version_dash_normalisation() -> None:
 
     with pytest.raises(Exception):
         Version("1-.0dev+3.4-")
+
+
+def test_bump() -> None:
```
Wackyator, line 23: These tests seem a little redundant as you already have them as part of the doctests.

Wackyator (changes requested):
Good work, just a couple of nits. :)

py-rattler/src/version/mod.rs
```diff
@@ -132,10 +132,43 @@ impl PyVersion {
         }
     }
 
-    /// Returns a new version where the last numerical segment of this version has been bumped.
-    pub fn bump(&self) -> Self {
-        Self {
-            inner: self.inner.bump(),
+    /// Returns a new version where the major segment of this version has been bumped.
+    pub fn bump_major(&self) -> PyResult<Self> {
+        match self.inner.bump(VersionBumpType::Major) {
```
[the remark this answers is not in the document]
hadim: Thanks. It seems best indeed!

py-rattler/tests/unit/test_version.py
hadim: Ok. Should I remove all of them?

Wackyator: You can keep the exception tests but I dont think the rest are required.

hadim:
All good @Wackyator on my side.

---

Out of curiosity: are you using VSCode and rust-analyzer by any chance? I'm asking because I do and noticed that any `.rs` files located within `py-rattler/` are failing to do any kind of linting/problem-detection within the VSCode editor, while all the `.rs` files located into `crates/` works just fine. I tried playing with `members = ["crates/*", "py-rattler/*"]` in `Cargo.toml` but no success.

_Sorry, I know it's not really related to that PR, but didnt want to open a new issue for this._

Wackyator:
> Out of curiosity: are you using VSCode and rust-analyzer by any chance? I'm asking because I do and noticed that any `.rs` files located within `py-rattler/` are failing to do any kind of linting/problem-detection within the VSCode editor, while all the `.rs` files located into `crates/` works just fine. I tried playing with `members = ["crates/*", "py-rattler/*"]` in `Cargo.toml` but no success.
> 
> _Sorry, I know it's not really related to that PR, but didnt want to open a new issue for this._

I'm indeed using vscode and rust-analyzer for py-rattler, I usually open `rattler/py-rattler/` in a separate vscode instance & it works fine.

Wackyator (approved):

Wackyator:
This looks good to me, others can point out anything I may have missed.

hadim:
> I'm indeed using vscode and rust-analyzer for py-rattler, I usually open rattler/py-rattler/ in a separate vscode instance & it works fine.

Ok, thanks. I was trying to avoid that and after going into the rabbit hole I found that:

```json
{
    "rust-analyzer.linkedProjects": [
        "./py-rattler/Cargo.toml"
    ]
}
```

into `.vscode/settings.json` solve the issue.

py-rattler/tests/unit/test_version.py
```diff
@@ -17,3 +18,17 @@ def test_version_dash_normalisation() -> None:
...
+        Version("1").bump_minor()
+
+    with pytest.raises(VersionBumpError):
+        Version("1.5").bump_patch()
+
+    with pytest.raises(VersionBumpError):
+        Version("1.5").bump_segment(-5)
+
+    with pytest.raises(VersionBumpError):
+        Version("1.5").bump_segment(5)
```
baszalmstra, line 34: Could you move these to the doctest of `bump_segment`? @Wackyator this is possible right?

baszalmstra (commented):
I left one final comment with regards to the unit test but other than that this looks good to me!

py-rattler/tests/unit/test_version.py
hadim: Should be good now.

baszalmstra (approved):

PR #461, opened 2024-01-02
feat: better detection of hardlinks and fallback to copy

baszalmstra:
This PR adds:

- Slightly better error messages when weird io errors happen during linking
- Falling back to copying when hardlinking or symlinking fails
- An improved method to determine if hardlinking is possible.


wolfv:
Should we already try to use CoW when available? :)

baszalmstra:
@wolfv Yeah I was planning on adding that after this PR 😀

wolfv (approved):

PR #463, opened 2024-01-02
fix: reflink files to destination if supported

baszalmstra:
This adds the ability to reflink files from the cache. 

@wolfv Since I dont have a machine capable of doing reflinks, would you be able to test this on your system? Simply installing an environment with the `create` command with this branch and main should give you a clue.


wolfv:
I tested locally, looks good!

wolfv:
We see the `clonefileat` in the flamegraph :)

Screenshot 2024-01-08 at 13 45 10

PR #469, opened 2024-01-05
feat: improve zstd compression

### LINKED ISSUE #133: Use zstd multithreading for compression / package writing
We can enable the `zstdmt` feature on the zstd crate and use multithreading as described here: 

https://docs.rs/zstd/latest/zstd/stream/write/struct.Encoder.html#method.multithread

For the package writing functions here: https://github.com/mamba-org/rattler/blob/main/crates/rattler_package_streaming/src/write.rs

orhun:
This PR improves the zstd compression by:

- enabling multithreading for compression (closes #133)
- sets the size of the archive which affects compression effectiveness
  - see [`set_pledged_src_size`](https://docs.rs/zstd/latest/zstd/stream/write/struct.Encoder.html#method.set_pledged_src_size)
  - see <https://github.com/conda/conda-package-streaming/issues/57>

### DIFF
--- a/crates/rattler_package_streaming/src/write.rs
+++ b/crates/rattler_package_streaming/src/write.rs
@@ -1,5 +1,4 @@
 //! Functionality for writing conda packages
-use std::env;
 use std::fs::{self, File};
 use std::io::{self, Seek, Write};
 use std::path::{Path, PathBuf};
@@ -136,13 +135,12 @@ pub fn write_tar_bz2_package<W: Write>(
 }
 
 /// Write the contents of a list of paths to a tar zst archive
-///
-/// The number of threads can be configured via `ZSTD_NUMTHREADS` environment variable.
 fn write_zst_archive<W: Write>(
     writer: W,
     base_path: &Path,
     paths: impl Iterator<Item = PathBuf>,
     compression_level: CompressionLevel,
+    num_threads: Option<u32>,
     timestamp: Option<&chrono::DateTime<chrono::Utc>>,
 ) -> Result<(), std::io::Error> {
     // Create a temporary tar file
@@ -158,12 +156,7 @@ fn write_zst_archive<W: Write>(
     let mut tar_file = File::open(&tar_path)?;
     let compression_level = compression_level.to_zstd_level()?;
     let mut zst_encoder = zstd::Encoder::new(writer, compression_level)?;
-    zst_encoder.multithread(
-        env::var("ZSTD_NUMTHREADS")
-            .ok()
-            .and_then(|v| v.parse().ok())
-            .unwrap_or_else(|| num_cpus::get() as u32),
-    )?;
+    zst_encoder.multithread(num_threads.unwrap_or_else(|| num_cpus::get() as u32))?;
     zst_encoder.set_pledged_src_size(tar_file.metadata().map(|v| v.len()).ok())?;
     zst_encoder.include_contentsize(true)?;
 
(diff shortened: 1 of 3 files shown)


wolfv (approved):

crates/rattler_package_streaming/src/write.rs
```diff
@@ -195,19 +208,23 @@ pub fn write_conda_package<W: Write + Seek>(
 
     let (info_paths, other_paths) = sort_paths(paths, base_path);
 
-    outer_archive.start_file(format!("pkg-{out_name}.tar.zst"), options)?;
+    let archive_path = format!("pkg-{out_name}.tar.zst");
```
wolfv, line 211: maybe call it just `.tar` or/and use `tempfile` here?

wolfv:
Looks good to me! Excellent improvement :)

crates/rattler_package_streaming/src/write.rs
```diff
@@ -137,22 +137,35 @@ pub fn write_tar_bz2_package<W: Write>(
...
         append_path_to_archive(&mut archive, base_path, &path, timestamp)?;
     }
+    archive.finish()?;
 
-    archive.into_inner()?.finish()?;
+    // Compress it as tar.zst
+    let mut tar_file = File::open(&tar_path)?;
+    let compression_level = compression_level.to_zstd_level()?;
+    let mut zst_encoder = zstd::Encoder::new(writer, compression_level)?;
+    zst_encoder.multithread(num_cpus::get() as u32)?;
```
wolfv, line 159: I think this should be configurable (default to this number, but users can choose lower number if they want).

wolfv:
cc @dholth does this reflect what you are doing in `conda-package-handling`?

wolfv:
Actually, ChatGPT says that the `zstd` encoding is not deterministic if a differnet number of threads is used. For that reason alone we should make it configurable (and store it in the rattler-build output) so that we can recreate the exact packages :)

wolfv:
Hmm, there are more opinions on this topic: 

- https://github.com/facebook/zstd/issues/999#issuecomment-474125238
- https://github.com/facebook/zstd/issues/2079#issuecomment-612195299

baszalmstra (approved):

dholth:
I should increase the number of threads. I think the comment is wrong about the downsides. https://github.com/conda/conda-package-handling/blob/main/src/conda_package_handling/conda_fmt.py#L24-L27

crates/rattler_package_streaming/src/write.rs
orhun: done in 221873d3d27b7b67349a4860c6e0ac2955a0d6eb

crates/rattler_package_streaming/src/write.rs
orhun: Added `ZSTD_NUMTHREADS` variable in 6ae2197c5c7f677b0632dd7e6379d796b3084679

The inspiration came from: https://github.com/facebook/zstd/issues/2241

I added a doc comment to the function but it is private thus not very visible. I think we can mention this somewhere, where would be a good place? I thought about putting it under a new section as "Environment Variables" in README.md where we list the other variables as well (I think there is only `CONDA_PREFIX` for now).

wolfv:
Not sure that the env var is the right approach for rattler (maybe for `rattler-build`). We shouldl make it configurable via struct or argument (like the compression level).

orhun:
I added an argument to the `write_conda_package` in 5aaf0903ef875b93cbe6a0daf9778459cc973537

PR #472, opened 2024-01-09
Add `read_package_file` function

wolfv:
This function allows us to extract a single `PackageFile` from a given archive.

### DIFF
--- a/crates/rattler_repodata_gateway/src/utils/simple_channel_server.rs
+++ b/crates/rattler_repodata_gateway/src/utils/simple_channel_server.rs
@@ -45,7 +45,8 @@ impl SimpleChannelServer {
 
         // Spawn the server. Let go of the JoinHandle, we can use the graceful shutdown trigger to
         // stop the server.
-        let _ = tokio::spawn(server);
+        let future = tokio::spawn(server);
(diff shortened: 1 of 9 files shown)


baszalmstra (approved):

crates/rattler_repodata_gateway/src/utils/simple_channel_server.rs
```diff
@@ -45,7 +45,8 @@ impl SimpleChannelServer {
 
         // Spawn the server. Let go of the JoinHandle, we can use the graceful shutdown trigger to
         // stop the server.
-        let _ = tokio::spawn(server);
+        let future = tokio::spawn(server);
```
baszalmstra, line 48: I think this should just be _future instead of dropping.

wolfv: clippy was suggesting this.

baszalmstra: Yeah its a bug. Assignment to _ immediately drops the value. Which might be counterintuitive. But we do want to keep the value around (I think).

But whatever, the tests still succeed so nvm!

PR #475, opened 2024-01-11
fix: reduce tracing level for reflink

baszalmstra:
When reflinking is not supported by the filesystem users would be spammed with error logs. I've reduced their level to `debug` now.

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -375,13 +375,13 @@ fn reflink_to_destination(
             }
             Err(e) => {
                 return if allow_hard_links {
-                    tracing::error!(
+                    tracing::debug!(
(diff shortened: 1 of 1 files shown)


wolfv:
Are these removed in production builds? I wonder about performance. Additionally, maybe it would be enough to log this once?

baszalmstra:
Yeah I think it would be better to check these things up front but I created this PR now to unstuck a pixi 0.11.2 release. Ill look at a proper fix later.

baszalmstra:
I think the overhead of skipped log statements is very little but Im not certain.

wolfv:
Supposedly we could use `#[cfg(debug_assertions)]` to remove the statements in release builds. I'll merge this for now :)

PR #482, opened 2024-01-16
Move async http range reader crate into rattler-networking

vlad-ivanov-name:
I preserved history -- let me know if I should squash it (or it can be squashed upon merge)


baszalmstra:
Now that I reread this code, Im wondering if it would make more sense to keep it in the library but allow passing arbitrary headers to requests? I think there is merit in keeping this functionality in its own library. Alternatively if the burden of multiple PRs to multiple repos is too high we could also just move the library itself into this repository but keep it as a seperate crate.

I apologize for steering you in the wrong direction initially, but WDYT?

baszalmstra:
Yeah we do use this crate internally in a few locations as well. I also know of some other uses in the wild. I think having this functionality (async range request streaming) as a separate crate serves more people than us. That is why we split it into its own crate in the first place anyway.

Personally I feel moving the source of the crate into this repository but as a separate crate provides us with the ability to create a single PR in the rattler repository but without tying the functionality to `rattler_networking`. I think that would be a good middle-ground.

baszalmstra:
You mean moving of the `AuthenticatedClient`? I think thats fine in this PR.

baszalmstra (approved):

PR #484, opened 2024-01-18
refactor: lock-file v4

### LINKED ISSUE #465: Consider removing unused fields (like `content_hash`) from lock files
I suggest we remove `content_hash` and some other metadata fields from our implementation, since we are able to verify that a lock-file satisfies certain requirements without all this information anyway.

At the moment `rattler_lock` is slowly diverging from conda lock. At this point the two formats are no longer compatible anyway. I expect that in the future we'll diverge further anyway.

See https://github.com/conda/conda-lock/issues/432.

baszalmstra:
This completely refactors the lock-file implementation in `rattler-lock` significantly changing the format. Old lock-files can still be read by the current implementation.

The goal of the refactor is to allow multiple environments to be stored in the lock-file as well as significantly simplifying the format itself. 

See the top-level crate documentation for more information about the design and considerations.

> [!NOTE]
> This significantly deviates from our previous implementation loosely based on conda-lock. 
> The code is still able to parse `conda-lock.yml` files! Its just not able to output a compatible conda-lock file.
> Note that this has already been the case for a long time but the differences didn't used to be so big.

> [!IMPORTANT]  
> This PR is still in previous because I want to first test out the API in Pixi. Nonetheless, reviews are already welcome! 
> 
> I also realize this is a huge change, Id be happy to walk anyone through the code on discord!

Fix #465


crates/rattler_lock/src/pypi.rs
```diff
@@ -1,30 +1,35 @@
...
+    pub requires_python: Option<VersionSpecifiers>,
 
     /// The URL that points to where the artifact can be downloaded from.
     pub url: Url,
 
     /// Hashes of the file pointed to by `url`.
+    #[serde(flatten)]
     pub hash: Option<PackageHashes>,
 
     /// ???
```
ruben-arts, line 35: Is this still needed?

ruben-arts (commented):
I did a quick read, and the only thing I found strange so far is a line that you **didn't** changed :wink: 

Awesome work

crates/rattler_lock/src/pypi.rs
baszalmstra: Im still not entirely sure. But let's just remove it until we actually use it. :)

crates/rattler_lock/src/pypi.rs
```diff
@@ -1,35 +1,79 @@
...
     /// The URL that points to where the artifact can be downloaded from.
     pub url: Url,
 
     /// Hashes of the file pointed to by `url`.
+    #[serde(flatten)]
     pub hash: Option<PackageHashes>,
 
-    /// ???
-    pub source: Option<Url>,
+    /// A list of dependencies on other packages that the wheel listed.
```
ruben-arts, line 28: Does this also count for the sdists?

ruben-arts (commented):
Really exited for this!! Left a few questions

ruben-arts (commented):
Really exited for this!! Left a few questions

crates/rattler_lock/src/pypi.rs
baszalmstra: Yes, metadata is gathered from sdists in the same way as wheels.

crates/rattler_lock/src/conda.rs
```diff
@@ -1,145 +1,116 @@
...
+        };
 
-    /// Experimental: Defines that the package is an alias for a package from another ecosystem.
-    #[serde(default, skip_serializing_if = "Vec::is_empty")]
-    pub purls: Vec<PackageUrl>,
+        Self {
+            package_record: value.package_record,
+            url: value.url,
+            file_name,
+            // TODO: This is not entirely correct. It should be derived from the `channel` field in
```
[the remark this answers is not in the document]
baszalmstra: Yes, so the `channel` field should be `Some` if it can't be properly derived from the URL of the package. But the channel in `RepoDataRecord` at this point is some undefined String. I left it like this until we get around to fix the `channel` field in `RepoDataRecord`.

crates/rattler_lock/src/lib.rs
```diff
@@ -1,51 +1,151 @@
...
-//! It is modeled on the definitions found at: [conda-lock models](https://github.com/conda/conda-lock/blob/main/conda_lock/lockfile/models.py)
-//! Most names were kept the same as in the models file. So you can refer to those exactly.
-//! However, some types were added to enforce a bit more type safety.
-use ::serde::{Deserialize, Serialize};
-use indexmap::IndexMap;
-use rattler_conda_types::{MatchSpec, PackageName};
-use rattler_conda_types::{NoArchType, Platform, RepoDataRecord};
-use serde_with::serde_as;
-use std::{collections::BTreeMap, io::Read, path::Path, str::FromStr};
+#![deny(missing_docs, dead_code)]
```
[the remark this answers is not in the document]
baszalmstra: Its just a lint to ensure dead-code is not introduced. Not now, and not in the future.

crates/rattler_lock/src/pypi.rs
baszalmstra: I removed the `"wheel"` part of the documentation.

crates/rattler_lock/src/lib.rs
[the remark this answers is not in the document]
ruben-arts: Ah I miss read the `!` for acting as an opposite of this.

PR #488, opened 2024-01-22
Convert authenticated client to reqwest middleware

vlad-ivanov-name:
While initially I wanted to create a number of traits for reqwest types, as it turns out, there are better solutions.

`reqwest_middleware` provides a flexible way to insert middleware that is activated either when the request is being built or just before it's sent. `AuthenticatedClient` is converted to a middleware type that handles auth.

This enables authentication with auth in rattler itself as well as in `async_http_range_reader` through the same interface.

One consequence is removal of the blocking client and features, as blocking comms are not supported in the middleware crate. I would argue that in most places where blocking communication is preferred -- e. g. simple programs performing a limited number of network requests -- it's often enough to just spawn a tokio runtime to wait on the future, because compared to network latency the cost will be negligible. Applications doing intensive network i/o are either async already anyway or likely should be.

This is not yet ready for merge -- I'm not sure what to do with Python binding. Technically we could replicate this type using the middleware client, but it's a bit weird that it's removed from Rust API but stays in Python API.

I will update the docs once there's a view on how to proceed.


wolfv:
I've also made a fix for the shlex problem here: https://github.com/mamba-org/rattler/pull/490

wolfv:
Actually, @baszalmstra is fixing it in #490 so hopefully the tests will soon turn green over here :)

baszalmstra (commented):
This looks very good to me! 

I think its perfectly fine to get rid of the blocking client. We hardly use it except for maybe in some tests.

For the Python bindings, I think we'll just have to add a type that wraps a ClientWithMiddleware. Could just be the AuthenticatedClient that we already have. It indeed diverges from the Rust API slightly but as long as there are no bindings for ClientWithMiddleware I think that is fine.

baszalmstra (approved):

baszalmstra:
LGTM! Is this now ready to merge?

baszalmstra:
Done! Thanks!

PR #492, opened 2024-01-24
chore: try to upgrade most dependencies

wolfv:
- The Zip Reader now reads 3 ranges instead of two, but I didn't investigate further why that would happen ... kinda weird!
- I didn't manage to update `axum`. They removed the `axum::Server` and point to `hyper` instead. However, if I see correctly, the server in hyper is also more complicated. There is a more simple `axum::serve` that I used e.g. in `rattler-server` but idk how to deal with the graceful shutdown etc. that we are using.

### DIFF
--- a/crates/rattler_package_streaming/tests/write.rs
+++ b/crates/rattler_package_streaming/tests/write.rs
@@ -101,7 +101,8 @@ fn compare_two_tar_archives<T: Read>(
         }
         FilterFiles::None => {}
     }
-
+    println!("Map 1: {:?}", map1);
+    println!("Map 2: {:?}", map2);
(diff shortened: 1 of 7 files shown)


crates/rattler_package_streaming/tests/write.rs
```diff
@@ -101,7 +101,8 @@ fn compare_two_tar_archives<T: Read>(
         }
         FilterFiles::None => {}
     }
-
+    println!("Map 1: {:?}", map1);
+    println!("Map 2: {:?}", map2);
```
baszalmstra, line 105: Should these stay?

baszalmstra (approved):

PR #498, opened 2024-01-29
feat: new crate for fetching run exports and etc

swarnimarun:
Summary:
- The said PR generalizes fetch operation for conda data json data files.
- And implements specializations for repo_data.json and run_exports.json.
- The APIs have been largely maintained other than minor changes for better naming of types.
- The import types are untouched or have aliases for API compatibility with previous implementation.
- Field names adjusted should be easy to fix in any package, pixi is using a much older version of the repos hence I haven't been able to update the API usage for it.

Key changes:
- Several types are now generalised types with specialisations. This should not affect any external APIs but still relevant if types are being used in unsafe context.
- All fields with direct reference to `repo_data` have been renamed other than in cases where specialised `impl`s have been introduced.
Eg, `CacheResult { repo_data_json_path, ... }` is now `CacheResult { path, ... }`.
- All specialised APIs are now in files/modules named as such, with unit tests ported for them.
- The tests introduce the entire no-arch and linux-64 files for `run_exports.json` from conda-forge.

Caveats:
- JLAP support is largely very specific to repo_data but is left in the generalised code for the sake of simpler APIs. And has been untouched by the PR.
	- As such JLAP code is not usable for anything other than `repo_data` and should be disabled in options as default for all other types. Eg, `crates/rattler_repodata_gateway/src/fetch/run_exports.rs`

Motivation: Support for handling run_exports.json as per [CEP-12](https://github.com/conda-incubator/ceps/blob/main/cep-12.md)


baszalmstra:
This has been open for a while with little progress. Im also not sure if this is the best approach. Im going to close it for now. Feel free to reopen if you want to continue working on this.

PR #499, opened 2024-01-29
add timeout parameter and SolverOptions to return early

wolfv:
This could be very useful for the bots and also for `rattler-server`.

### DIFF
--- a/crates/rattler_solve/src/resolvo/mod.rs
+++ b/crates/rattler_solve/src/resolvo/mod.rs
@@ -1,6 +1,6 @@
 //! Provides an solver implementation based on the [`resolvo`] crate.
 
-use crate::{IntoRepoData, SolveError, SolverRepoData, SolverTask};
+use crate::{IntoRepoData, SolveError, SolverOptions, SolverRepoData, SolverTask};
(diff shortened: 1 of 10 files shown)


crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -378,6 +388,15 @@ impl<'a> DependencyProvider<SolverMatchSpec<'a>> for CondaDependencyProvider<'a>
 
         Dependencies::Known(dependencies)
     }
+
+    fn should_cancel_with_value(&self) -> Option<Box<dyn std::any::Any>> {
```
nichmor, line 392: sorry for being slow - why the return type is not `Option<Box<dyn CancelReason>>`  but` dyn Any`?

nichmor (approved):
Looks super nice to me!

crates/rattler_solve/src/resolvo/mod.rs
wolfv: It's defined like that in the trait (in `resolvo`) so I think we need to honor that contract and return an `Box<Any>`

PR #510, opened 2024-02-05
feat: add py-rattler-lock 

Wackyator:


baszalmstra:
Looks good! Im just missing tests!

Wackyator:
Let me just add them

py-rattler/rattler/lock/environment.py
```diff
@@ -0,0 +1,217 @@
...
+                (
+                    PypiPackageData._from_py_pypi_package_data(pkg_data),
+                    PypiPackageEnvironmentData._from_py_pypi_env_data(env_data),
+                )
+                for (pkg_data, env_data) in pypi_tup
+            ]
+            for (platform, pypi_tup) in self._env.pypi_packages().items()
+        }
+
+    def conda_repodata_records(self) -> Dict[str, List[RepoDataRecord]]:
```
baszalmstra, line 127: This should also be `Platform`.

py-rattler/rattler/lock/lock_file.py
```diff
@@ -0,0 +1,149 @@
...
+    Lock-files can store information for multiple platforms and for multiple environments.
+    """
+
+    _lock_file: PyLockFile
+
+    def __init__(
+        self,
+        channels: Optional[List[LockFileChannelConfig]] = None,
+        conda_packages: Optional[List[CondaPackageConfig]] = None,
+        pypi_packages: Optional[List[PypiPackageConfig]] = None,
```
baszalmstra, line 26: The construction of this object seems illogical to me. I would expect it to take a 

`Dict[str, Environment]`

you can construct a new `Environment` using `Dict[Platform, List[Union[CondaPackage, Tuple[PyPIPackageData, PypiPackageEnvironmentData]]`.

Internally (in rust) you construct a new `Arc<LockFile>` for this data. 

In the constructor of `LockFile` you combine this data again.

py-rattler/rattler/lock/package.py
```diff
@@ -0,0 +1,172 @@
+from __future__ import annotations
+from typing import Optional, Tuple
+from rattler.lock.pypi import PypiPackageData, PypiPackageEnvironmentData
+
+from rattler.rattler import PyLockPackage
+from rattler.repo_data.record import RepoDataRecord
+
+
+class LockPackage:
```
baszalmstra, line 9: Perhaps `LockedPackage` is a better name?

py-rattler/rattler/lock/package.py
```diff
@@ -0,0 +1,172 @@
...
+    _package: PyLockPackage
+
+    def satisfies(self, spec: str) -> bool:
+        return self._package.satisfies(spec)
```
baszalmstra, line 13: I think this method is ambiguous because the spec is either a Pypi requirement or a Conda matchspec. Maybe it makes more sense to have this method on `RepoDataRecord` and `PyPiPackageData`?

baszalmstra (commented):
I think the way you construct a lock file could really be improved. What would you like the API from python to be? Maybe you can come close to that?

We could also merge this PR without the construction part and create a seperate PR for that?

py-rattler/rattler/lock/lock_file.py
Wackyator: Removed this constructor for now, will raise a separate PR for it.

py-rattler/Cargo.toml
```diff
@@ -53,6 +48,7 @@ thiserror = "1.0.44"
 url = "2.4.1"
 
 openssl = { version = "0.10", optional = true }
+pep508_rs = "0.3.0"
```
baszalmstra, line 51: pep508_rs also exports python bindings I think, we could consider using those!

baszalmstra (approved):

py-rattler/Cargo.toml
Wackyator: Oh, let me have a look.

PR #511, opened 2024-02-05
fix: keep in mind python: noarch packages in clobber calculations

wolfv:
I hit a bug where the `unclobber` function tried to run `fs::rename` on non-existing files (because these files came from `noarch: python` packages where the final path got renamed from `site-packages/bla.py` to `lib/pythonX.X/site-packages/bla.py`).

This fix is relatively ugly (because we compute the "final" name for the file multiple times and entry points are also not properly handled yet.

Maybe we should discuss real quick how we want to handle entry points (entry points don't need to strictly be linked into the env since we can recreate them when we need to).

### DIFF
--- a/crates/rattler/src/install/clobber_registry.rs
+++ b/crates/rattler/src/install/clobber_registry.rs
@@ -2,12 +2,15 @@
 
 use std::{
     collections::{HashMap, HashSet},
-    fs,
     path::{Path, PathBuf},
 };
 
-use rattler_conda_types::{package::PathsJson, PackageName, PrefixRecord};
+use rattler_conda_types::{
+    package::{IndexJson, PathsEntry},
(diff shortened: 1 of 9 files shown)


crates/rattler/src/install/clobber_registry.rs
```diff
@@ -116,25 +120,51 @@ impl ClobberRegistry {
...
+        //                     let file = python_info.bin_dir.join(&entry_point.command);
+        //                     res.push(file);
+        //                 }
+        //             }
+        //         }
+        //     }
+        //     res
+        // } else {
+        //     Vec::new()
+        // };
```
baszalmstra, line 148: Can this be removed?

baszalmstra (approved):

PR #529, opened 2024-02-15
Fix: Add indexed packages to the relevant section in the repodata

johnhany97:
## Before this PR

The produced repodata.json from `rattler_index` would put all the packages, regardless of the file format, into the `packages.conda` section of the repodata.json

## After this PR

TarBz2 packages now go to the `packages` section, while Conda packages still go to `packages.conda`. I've added a test case to validate the desired behaviour.


baszalmstra (approved):
Awesome thanks!

PR #420, opened 2023-11-27
add options to disable zstd and bz2

0xbe7a:
We want to try `rattler` and `rattler-build` within the client infrastructure. This client uses JFrog Artifactory to proxy conda repositories. The deployed version has a bug where the `repodata.json.zst` is cached indefinitely. An option to disable support for bz2 and zstd would allow us to use rattler with this version.

PR #530, opened 2024-02-15
Add repodata config to channel

mariusvniekerk:
This builds on some of the idea of #420 

When fetching from certain channels it would be handy to disable some of the repodata methods that are known to be expensive / non-functional.

This is particularly true for artifactory where the HEAD request for the .bz2 repodata will eager compress the contents and not cache it. :(

Idea is that in downstream tools like pixi a user can define their channel and add some of these options in 

```
channels = [
   { base_url = "https://somehost/channel", repodata_options = { bz2_enabled = False }}
]
```

### DIFF
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -5,7 +5,6 @@ use std::str::FromStr;
 
 use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
-use std::convert::identity;
 use thiserror::Error;
 use url::Url;
 
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -7,6 +7,7 @@ use serde::{Deserialize, Serialize};
 use smallvec::SmallVec;
 use thiserror::Error;
 use url::Url;
+use std::convert::identity;
 
 use super::{ParsePlatformError, Platform};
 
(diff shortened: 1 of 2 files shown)


crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -27,13 +27,49 @@ pub struct ChannelConfig {
...
     pub channel_alias: Url,
+
+    /// An optional configuration for the repodata of the channel.
+    ///
+    /// This *SHOULD* be used to configure how metadata is fetched from the channel.
+    ///
+    pub repodata_options: Option<ChannelRepodataConfig>,
+}
+
+fn _serde_is_true(value: &bool) -> bool {
```
baszalmstra, line 38: Prefixing with _ would mean its unused. I assume you want to make it private to the module, if you dont add "pub" that should be fine.

baszalmstra (commented):
Im a bit on the fence if adding this to channel makes sense responsibility wise. It makes sense when fetching repodata but its used in many more places were it doesnt make sense.

Let me mull it over a little.

baszalmstra:
I have been thinking about this and I think we need to alter the design a little bit. I think we need to split this functionality into two: `Channel` and `FetchChannel` (or whatever, naming is hard). The `Channel` only contains information about a url. The `FetchChannel` struct contains a `Channel` and extra metata information needed for fetching the channel's repodata. You should be able to construct a `FetchChannel` directly from a `Channel` with default values. And `fetch_repodata` function should accept a `FetchChannel`. 

In pixi we can then change some of the logic to work with `FetchChannel` instead of `Channel`.

baszalmstra:
With the new Gateway API, it is now possible to set configuration per channel. 

Im closing this, feel free to reopen if you think this PR still adds more value.

PR #541, opened 2024-02-26
feat: add constructor

Wackyator:
Adds constructors for:
- LockFile 
- Environment 
	
Also adds missing docs to website


py-rattler/src/lock/mod.rs
```diff
@@ -91,6 +120,40 @@ impl From<PyEnvironment> for Environment {
...
+                    record.try_as_repodata_record()?.clone().into(),
+                );
+            }
+        }
+
+        Ok(lock
+            .finish()
+            .environment(&name)
+            .ok_or(Into::<PyErr>::into(PyRattlerError::RequirementError(
+                "Temp Placeholder Error".to_owned(),
```
baszalmstra, line 152: This seems like something that should be changed ;)

baszalmstra (commented):
One small change, but otherwise look great!

py-rattler/src/lock/mod.rs
Wackyator: Oh no, I forgot this one, thanks alot.

baszalmstra (approved):

PR #545, opened 2024-02-27
fix: correct condition to downweigh track-feature packages

wolfv:
We use track features to de-prioritize packages.
Sometimes, the solver is forced to select a given package, e.g. when selecting `numpy-XY-<hash>` it might have a dependency on `pypy`. Numpy variants themselves are not down-prioritized, though.

If we detect that a package forces only dependencies that have track features (such as the case for the numpy variant that only works together with `pypy`) we also "inherit" the down-weighing to `numpy` (otherwise the solver would be forced to use pypy later on).

We already had the code for this but it was missing a `!` – this PR corrects that.

### DIFF
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -458,7 +458,7 @@ mod test {
             .map(std::vec::Vec::len)
             .sum::<usize>();
 
-        assert_eq!(total_records, 21731);
+        assert_eq!(total_records, 21732);
     }
 
     #[tokio::test]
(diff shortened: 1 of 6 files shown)


wolfv:
So everything worked fine if there is only a single package with the track feature. I had to change the repodata to contain 2 builds of 

```json
    "python_abi-3.9-2_cp39.tar.bz2": {
      "build": "2_cp39",
      "build_number": 2,
      "constrains": [
        "pypy <0a0"
      ],
      "depends": [
        "python 3.9.*"
      ],
      "license": "BSD-3-Clause",
      "license_family": "BSD",
      "md5": "39adde4247484de2bb4000122fdcf665",
      "name": "python_abi",
      "sha256": "67231829ea0101fee30c68f788fdba40a11bbee8fdac556daaab5832bd27bf3d",
      "size": 4017,
      "subdir": "linux-64",
      "timestamp": 1624472010037,
      "version": "3.9"
    },
```

And that way I could reproduce the issue + make sure the fix works by adding snapshots.

PR #552, opened 2024-03-01
feat!: optional strict parsing of matchspec and versionspec

baszalmstra:
Adds a mode to all `from_str` functions for `MatchSpec`, `NamelessMatchSpec`, `VersionSpec`, and `Constraint`.

Strict mode disallows some ambigous syntax like:

* `>=1.*` should just be `>=1`
* `*.*` should just be `*`

All parsing functions accept a `ParseStrictness` that can either be `Strict` or `Lenient`. `Lenient` mode should be used when parsing repodata etc, but `Strict` mode can be used to validate new user input.


wolfv:
Looks good to me! 

I wonder if we should expose `parse()` and `parse_strict()` vs. the enum but that's my only comment.

wolfv (approved):

baszalmstra:
> I wonder if we should expose `parse()` and `parse_strict()` vs. the enum but that's my only comment.

`">=1.2.3".parse()` will still work, it will use the lenient parser (because the types also implement `FromStr`). If you want strict parsing you have to call `VersionSpec::from_str("1.2.3", Strict)`.

PR #560, opened 2024-03-08
feat: high level repodata access

### LINKED ISSUE #520: Noisy error from rattler_repodata_gateway::fetch
When fetching repodata it's "OK" if the target architecture is missing - the only required file is the `noarch/repodata.json` basically. 

However, `fetch_repodata` returns a `NotFound` error for non-noarch files. The problem is that this is logged through the `instrument` attribute on the `fetch_repo_data` function and pops up in the logs (e.g. in `rattler-build`).

I am wondering if we could either disable logging an error for this particular error or if we should return a `None` instead.

baszalmstra:
This PR introduces the concept of a `Gateway` into `rattler_repodata_gateway`. 

Fixes #520


crates/rattler_repodata_gateway/src/gateway/sharded_subdir/index.rs
```diff
@@ -0,0 +1,352 @@
...
+                            ))
+                        }
+                    },
+                    |_| Ok(None),
+                )
+                .right_future()
+        };
+
+        // Parse the bytes
+        let parse_fut = tokio_rayon::spawn(move || rmp_serde::from_slice(&decoded_bytes))
```
wolfv, line 86: I have read a little about rayon vs tokio - is it worth it? From what I read it seems most useful in CPU intense workloads while we are presumably mostly bound by IO.

https://www.reddit.com/r/rust/comments/xec77k/rayon_or_tokio_for_heavy_filesystem_io_workloads/

crates/rattler_repodata_gateway/src/gateway/builder.rs
```diff
@@ -0,0 +1,101 @@
...
+
+    /// Set the channel configuration to use for fetching repodata.
+    #[must_use]
+    pub fn with_channel_config(mut self, channel_config: ChannelConfig) -> Self {
+        self.set_channel_config(channel_config);
+        self
+    }
+
+    /// Sets the channel configuration to use for fetching repodata.
+    pub fn set_channel_config(&mut self, channel_config: ChannelConfig) -> &mut Self {
```
wolfv, line 45: Do we need both versions (with and set)?

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+        // Get the token
+        let token = self.token_client.get_token(reporter).await?;
+
+        // Download the shard
+        let shard_url = token
+            .shard_base_url
+            .as_ref()
+            .unwrap_or(&self.shard_base_url)
+            .join(&format!("shards/{shard:x}.msgpack.zst"))
+            .expect("invalid shard url");
```
wolfv, line 129: maybe we should get rid of the expects. i am on the fence :)

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+
+        let mut shard_request = self
+            .client
+            .get(shard_url.clone())
+            .header(CACHE_CONTROL, HeaderValue::from_static("no-store"))
+            .build()
+            .expect("failed to build shard request");
```
wolfv, line 136: here, too.

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/token.rs
```diff
@@ -0,0 +1,163 @@
...
+    pub token: Option<String>,
+    issued_at: Option<DateTime<Utc>>,
+    expires_in: Option<u64>,
+    pub shard_base_url: Option<Url>,
+}
+
+impl Token {
+    /// Returns true if the token is still considered to be valid.
+    pub fn is_fresh(&self) -> bool {
+        if let (Some(issued_at), Some(expires_in)) = (&self.issued_at, self.expires_in) {
```
wolfv, line 145: did we add thsi to the CEP?

Should we initialize `issued_at` with the current time if the token doesn't contain it?

wolfv (approved):

crates/rattler_lock/Cargo.toml
```diff
@@ -27,7 +28,7 @@ serde_repr = { workspace = true }
 thiserror = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 purl = { workspace = true, features = ["serde"] }
-percent-encoding = {  workspace = true }
+rayon = { workspace = true }
```
tdejager, line 31:
```suggestion
```

crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
```diff
@@ -0,0 +1,67 @@
...
+        })
+    }
+}
+
+#[async_trait::async_trait]
+impl SubdirClient for LocalSubdirClient {
+    async fn fetch_package_records(
+        &self,
+        name: &PackageName,
+        _reporter: Option<&dyn Reporter>,
```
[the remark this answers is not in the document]
tdejager: Indeed?

crates/rattler_repodata_gateway/src/utils/body.rs
```diff
@@ -0,0 +1,71 @@
+use bytes::Bytes;
```
tdejager, line 1: Could you maybe explain what this module and its traits do?

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/index.rs
wolfv: Removed rayon, seems fine!

crates/rattler_lock/Cargo.toml
wolfv: Done

crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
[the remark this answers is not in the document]
baszalmstra: We don't have any report callback and this process is usually extremely fast (e.g. 1 or 2 milliseconds).

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/token.rs
baszalmstra: We did add it to the CEP yes: https://github.com/baszalmstra/ceps/blob/sharded-repodata/cep-16.md#authentication

Ill change the issued_at to the current time!

crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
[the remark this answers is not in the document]
wolfv: It's fast for you, but for someone with a spinning disk... :) Maybe we could add a "loading records" and "finished loading records" or something like this?

PR #561, opened 2024-03-08
fix: better handle utf8 paths in CmdExe and Powershell

baszalmstra:
Adds a function to force a shell to handle utf8 encoding. An implementation is provided for cmdexe through the use of the `chcp 65001` command and something similar for Powershell.

This should fix the issue encountered here https://github.com/prefix-dev/pixi/issues/930


baszalmstra:
@wolfv Please take a closer look if you think this is the right approach.

wolfv (approved):

PR #572, opened 2024-03-19
feat: allow passing pypi paths

baszalmstra:
Allow passing paths as pypi package locations instead of just urls.


wolfv:
`pixi run fmt` should fix the remaining issue :)

wolfv (approved):

baszalmstra:
Just testing this in pixi before I merge it :)

PR #574, opened 2024-03-20
fix: run post-link scripts

wolfv:
TODO: 
- also run pre-unlink scripts

fixes:

- https://github.com/prefix-dev/pixi/issues/404
- https://github.com/prefix-dev/rattler-build/issues/185


baszalmstra:
👍

One of the primary reasons that I am very much against executing scripts during installation is that it is a massive security risk. We are basically allowing arbitrary code execution at install time. 

Bun recently introduced the concept that a user has to explicitly allow running installation scripts for certain packages. Would you be open to adding a similar approach here right off the bat? I know this complicates the integration into rattler-build and pixi but I think its very important to protect users.

wolfv:
From our discussion:

- I totally agree that post-link / pre-unlink scripts are ugly and should ideally not exist
- On the same note, activation scripts are "just as bad" (unfortunately!)
- Lastly we always have to trust packages

To make things more secure in the longer run I advocate for:

- sandboxing using cgroups, and other means
- removing activation or post-link scripts wherever possible and substituting for simpler "environment variable" setters (actually already possible but underused in conda-forge)

I will make it configurable wether to run the installation scripts, and we can also emit a warningi or something like that/

crates/rattler/src/install/link_script.rs
```diff
@@ -0,0 +1,251 @@
...
+                &env,
+            ) {
+                Ok(o) if o.status.success() => {}
+                Ok(o) => {
+                    tracing::warn!("Error running post-link script. Status: {:?}", o.status);
+                    tracing::warn!("  stdout: {}", String::from_utf8_lossy(&o.stdout));
+                    tracing::warn!("  stderr: {}", String::from_utf8_lossy(&o.stderr));
+                }
+                Err(e) => {
+                    tracing::error!("Error running post-link script: {:?}", e);
```
baszalmstra, line 117: Shouldnt this result in an error?

crates/rattler/src/install/link_script.rs
```diff
@@ -0,0 +1,251 @@
...
+                }
+            }
+
+            let message_file = target_prefix.join(".messages.txt");
+            if message_file.exists() {
+                let message = std::fs::read_to_string(&message_file)?;
+                tracing::info!(
```
baszalmstra, line 124: Would it make sense to return these, or store them in the driver, so a user can decide what to do with them?

crates/rattler/src/install/link_script.rs
wolfv: That would crash the remaining installation ... not sure if we want that (and I also don't think that's what Conda does).

baszalmstra (approved):

PR #581, opened 2024-03-25
feat: editable pypi packages

baszalmstra:
Adds the `editable` field to pypi packages to indicate they are installed in editable mode.


tdejager:
LGTM

PR #589, opened 2024-04-02
fix: do not do cstring replacement on windows

wolfv:
There is only one form of binary replacement on Windows, and that's for "pyzzer" entrypoints. I don't think they are widely used though.

I think this might fix the issue described in https://github.com/prefix-dev/pixi/issues/1096

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -481,6 +481,8 @@ pub fn copy_and_replace_placeholders(
             )?;
         }
         FileMode::Binary => {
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

baszalmstra:
Maybe it would be good to add a comment in the code why this is disabled for windows. Otherwise LGTM.

wolfv:
I just looked for `pyzzer`: https://pypi.org/project/pyzzer/

I don't think it's very relevant (last release 2013). So IMO we can just ignore that.

baszalmstra (approved):

PR #594, opened 2024-04-05
fix: replace long shebangs with `/usr/bin/env`

wolfv:
Note: I should test the whole replace function as well

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -529,13 +529,11 @@ fn replace_long_shebang(shebang: &str, platform: &Platform) -> String {
     } else {
         assert!(shebang.starts_with("#!"));
         if let Some(captures) = SHEBANG_REGEX.captures(shebang) {
-            let shebang_path = Path::new(&captures[2]);
-            tracing::debug!("New shebang path {}", shebang_path.display());
-            format!(
-                "#!/usr/bin/env {}{}",
-                shebang_path.file_name().unwrap().to_str().unwrap(),
-                &captures[3]
-            )
+            let shebang_path = &captures[2];
+            let filename = shebang_path
+                .rsplit_once('/')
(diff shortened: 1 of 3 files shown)


wolfv:
reference:

https://github.com/mamba-org/mamba/blob/b5195d96b062c841fcf6ead36a611b1cf4707359/libmamba/src/core/link.cpp#L101-L123

https://github.com/conda/conda/blob/c57fbdebed072ced5e03fabac13edcf3b0f444c9/conda/core/portability.py#L239-L240

https://github.com/conda/conda/blob/c57fbdebed072ced5e03fabac13edcf3b0f444c9/conda/core/portability.py#L67-L68

crates/rattler/src/install/link.rs
```diff
@@ -497,6 +498,45 @@ pub fn copy_and_replace_placeholders(
     }
     Ok(())
 }
+use once_cell::sync::Lazy;
+
+static SHEBANG_REGEX: Lazy<Regex> = Lazy::new(|| {
```
baszalmstra, line 503: We really dont need a regex for this I think..

wolfv: The main complication is escaped white spaces in the path ...

crates/rattler/src/install/link.rs
```diff
@@ -364,18 +366,10 @@ fn reflink_to_destination(
             Err(e) if e.kind() == ErrorKind::Unsupported && !allow_hard_links => {
                 return copy_to_destination(source_path, destination_path)
             }
-            Err(e) => {
+            Err(_) => {
                 return if allow_hard_links {
-                    tracing::debug!(
```
wolfv, line 369: removed these as they are super noisy in CI.

baszalmstra (approved):

PR #598, opened 2024-04-08
feat: Add channel priority to solve task and expose to python solve

BenjaminLowry:
There are use cases where disabling strict channel priority is important and so should have functionality in rattler such that it can be disabled. So I added it as a field of the SolveTask and exposed it as a new parameter (that existed before but was deleted) in py-rattler.


crates/rattler_solve/src/libsolv_c/mod.rs
```diff
@@ -96,6 +96,12 @@ impl super::SolverImpl for Solver {
                 "timeout".to_string()
             ]));
         }
+        // Could ignore the strict_channel_priority flag here or do something like below maybe?
```
BenjaminLowry, line 99: This part I'm not so sure what we'd want the behavior to be, it seems to me that this libsolv code doesn't enforce strict channel priority? If so, we can either ignore the option all together to throw if strict priority is true since that's "not supported"?

BenjaminLowry:
@ruben-arts here is my PR that we discussed earlier today, let me know what you think

baszalmstra (commented):
I believe conda has more options, @wolfv is that the case? In that case an enum would be nicer.

BenjaminLowry:
@baszalmstra It does look like conda has three options: strict, flexible and disabled https://conda.io/projects/conda/en/latest/user-guide/tasks/manage-channels.html#strict-channel-priority and they default to flexible. Rattler via resolvo is defaulting to strict. I can add an enum but maybe with only two values, strict and disabled for now? And flexible could be added later if needed?

And I see there is a [flag](https://github.com/mamba-org/mamba/blob/80f1e84d49810f0198629da28cd059d457b4c2e6/micromamba/src/server.cpp#L140) for strict priority for libsolv that micromamba uses but I'm a bit ignorant to how this would be translated to this code.

baszalmstra:
Sounds good! Maybe wolf can help with the libsolv side.

BenjaminLowry:
@baszalmstra @wolfv Sorry for the repeated pings, this is just a fairly important change for our workflows.

If we can't resolve why libsolv isn't behaving expectedly with strict channel priority, I can just throw if someone specifies strict when using libsolv? And strict handling could be added later

crates/rattler_solve/src/lib.rs
```diff
@@ -70,6 +70,22 @@ impl fmt::Display for SolveError {
...
 }
 
+/// Represents the channel priority option to use during solves.
+#[derive(Clone, Copy, PartialEq, Eq)]
+pub enum ChannelPriority {
+    /// The channel that the package is first found in will be used as the only channel
+    /// for that package.
+    Strict,
+
+    /// Conda also has "Flexible" as an option, where packages present in multiple channels
```
wolfv, line 80: maybe we shoudl use `//` so that these comments do not show up in teh documentation for a feature we don't implement.

crates/rattler_solve/src/lib.rs
```diff
@@ -70,6 +70,22 @@ impl fmt::Display for SolveError {
...
+    /// are only taken from lower-priority channels when this prevents unsatisfiable environment
+    /// errors, but this would need implementation in the solvers.
+    /// Flexible,
+
+    /// Packages can be retrieved from any channel as package version takes precedence.
+    Disabled,
```
wolfv, line 86: Shoudl we call this `AllEqual` or `NoPriority/NoPriorities`? Just thinking of something that makes it really clear.

crates/rattler_solve/src/libsolv_c/mod.rs
wolfv: Yep. Probably we need to set the `priority` on the repository, though? I'll take a look.

wolfv (commented):
Just rethinking the naming. Let me take a quick look at libsolv!

crates/rattler_solve/src/lib.rs
BenjaminLowry: Yeah sounds good, I only did that cuz with `//` it formats the comments right against the disabled comment, but likely more confusing for it to be in the documentation

wolfv:
@BenjaminLowry I made a PR against your branch here: https://github.com/BenjaminLowry/rattler/pull/1 It shows how we could set the `priority` for libsolv. 

Libsolv has a `priority` and a `subpriority` integer value. The higher priority should win. We would also need a map or count down the channels and I haven't yet investigated how to do that with the structures at hand.

crates/rattler_solve/src/lib.rs
wolfv: Yeah, totally. I am also OK with keeping the naming. Just sometimes I think we have the opportunity to come up with new & improved terminology :)

BenjaminLowry:
@wolfv Thanks for the PR, I added logic to assign repo priority values where it's determined by the order of the repodatas passed in where the channel with the first repodata is given the highest priority and it descends from there. This is pre-computed so we don't have to do some arbitrary highest priority value and so if you have 5 channels the first will be given 5 and the last will be given 1 and 0 is still available for lowest priority use.

This is now behaving as expected for libsolv with strict and disabled priority, does this implementation seem reasonable? Let me know if I'm not following specific rust-isms as well

baszalmstra (approved):
LGTM!

BenjaminLowry:
@baszalmstra I'm assuming you also want @wolfv to give a +1? Or is this good to go

baszalmstra:
Yeah since @wolfv has a bunch of open comments I was waiting for his approval as well. 😄 

Wolf if you are reading this, can you merge if you are happy with this?

crates/rattler_solve/src/libsolv_c/mod.rs
```diff
@@ -116,15 +155,22 @@ impl super::SolverImpl for Solver {
...
                 continue;
             }
-
             let channel_name = &repodata.records[0].channel;
 
             // We dont want to drop the Repo, its stored in the pool anyway.
-            let repo = ManuallyDrop::new(Repo::new(&pool, channel_name));
+            // let repo = ManuallyDrop::new(Repo::new(&pool, channel_name));
+            let priority: i32 = if task.channel_priority == ChannelPriority::Strict {
+                // TODO we need to fill in the correct values here, derived from some channel -> priority map
```
wolfv, line 167: These comments are outdated, right?

crates/rattler_solve/src/libsolv_c/mod.rs
```diff
@@ -116,15 +155,22 @@ impl super::SolverImpl for Solver {
...
+        for repodata in repodatas.iter() {
             if repodata.records.is_empty() {
                 continue;
             }
-
             let channel_name = &repodata.records[0].channel;
 
             // We dont want to drop the Repo, its stored in the pool anyway.
-            let repo = ManuallyDrop::new(Repo::new(&pool, channel_name));
+            // let repo = ManuallyDrop::new(Repo::new(&pool, channel_name));
```
wolfv, line 165: Keep or remove?

wolfv (approved):

BenjaminLowry:
@baszalmstra Thanks for already working on a release! I saw that py-rattler was bumped to 0.4.0 yesterday, will we be able to ship this change in a new release of py-rattler soon?

baszalmstra:
Sure! I can make another release in the next few days.

baszalmstra:
py-rattler 0.5.0 is on pypi which includes your changes.

PR #602, opened 2024-04-16
feat: make root dir configurable in channel config

wolfv:
This would be useful for making sure that channels are resolved relative to the `pixi.toml` file (and not the current working directory). It also makes the call to current work dir the responsibility of the user and they can catch errors (e.g. the current wrok dir is not available in wasm32 and maybe some other embedded systems or so).

I am not sure about the name though. Happy about suggestions :)

### DIFF
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -11,6 +11,8 @@ use url::Url;
 
 use super::{ParsePlatformError, Platform};
 
+const DEFAULT_CHANNEL_ALIAS: &str = "https://conda.anaconda.org";
+
 /// The `ChannelConfig` describes properties that are required to resolve "simple" channel names to
 /// channel URLs.
 ///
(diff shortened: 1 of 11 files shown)


crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -28,6 +28,10 @@ pub struct ChannelConfig {
     ///
     /// The default value is: <https://conda.anaconda.org>
     pub channel_alias: Url,
+
+    /// For local channels, the root directory from which to resolve relative paths.
+    /// Most of the time you would initialize this with the current working directory.
+    pub root_dir: PathBuf,
```
baszalmstra, line 34: Maybe we can make this Option to make it easier to construct this type. If its None then we use the current directory and otherwise we use the explicitly set one. WDYT?

wolfv: Using current directory is kinda annoying because it can fail though. That's why I liked to insert the type from outside.

baszalmstra: Makes sense but now the Default impl can fail without a way to recover. Maybe we should temove that too in favor of a failable associated function?

PR #604, opened 2024-04-16
fix: unicode activation issues

wolfv:


crates/rattler_shell/src/activation.rs
```diff
@@ -13,7 +13,7 @@ use crate::shell::Shell;
 use indexmap::IndexMap;
 use rattler_conda_types::Platform;
 
-const ENV_START_SEPERATOR: &str = "<=== RATTLER ENV START ===>";
```
wolfv, line 16: This was doing some weird things - it was executing the `env` executable and completely failing. Although that might also be related to the unicode issues in the first place.

With the new one we don't need any special escaping though.

crates/rattler_shell/src/activation.rs
```diff
@@ -433,7 +433,16 @@ impl<T: Shell + Clone> Activator<T> {
         let activation_script_path = activation_script_dir
             .path()
             .join(format!("activation.{}", self.shell_type.extension()));
-        fs::write(&activation_script_path, &activation_detection_script)?;
+
+        // Use CRLF line endings on Windows batch scripts
+        if self.shell_type.extension() == "bat" {
+            activation_detection_script = activation_detection_script.replace('\n', "\r\n");
```
wolfv, line 439: For some reason this is totally necessary.

crates/rattler_shell/src/activation.rs
```diff
@@ -747,44 +760,46 @@ mod tests {
         env_diff.remove("Path");
         env_diff.remove("PATH");
 
-        insta::assert_yaml_snapshot!(shell.executable(), env_diff);
+        insta::assert_yaml_snapshot!("after_activation", env_diff);
```
wolfv, line 763: they should all result in the same env vars so we can use a single snapshot.

crates/rattler_shell/src/activation.rs
baszalmstra: Ideally we make this part of the Shell type instead of checking the extension of the Shell type.

baszalmstra: Like Shell::line_endings or something.

crates/rattler_shell/src/shell/mod.rs
```diff
@@ -421,7 +422,17 @@ impl Shell for PowerShell {
     }
 
     fn executable(&self) -> &str {
-        self.executable_path.as_deref().unwrap_or("pwsh")
+        if let Some(executable_path) = self.executable_path.as_deref() {
+            executable_path
+        } else {
+            // check if `powershell` is available, if not, use `pwsh`
```
baszalmstra, line 428: Can we do this as part of the construction?

wolfv: I am doing this now in `default`.

wolfv:
The only thing I am thinking about is wether we should restore the previous code page after activation is done. Might be nice for the user

PR #605, opened 2024-04-17
feat: Add GCS support for rattler auth

clement-chaneching:
Hello,

### Context
We would like to switch to pixi for our package manager.
We were thinking of having our private packages hosted on a private conda channel on Google Cloud Storage Buckets.
This change will allow pixi to automatically find the packages if we are logged in in our CLI

### Changes:
Change the auth to async for gcloud.
Use the google_cloud_auth module to login ( No need to pass any argument)
Added a test to validate the application_default_credentials path.

### Example pixi.toml:

```
channels = ["conda-forge","https://storage.googleapis.com/<bucket_name>]

[dependencies]
<private_module>= { version = "*", channel = "https://storage.googleapis.com/<bucket_name>" }

```
### Results
If not logged in:
image
After running `gcloud auth application-default login`:
image


crates/rattler_networking/Cargo.toml
```diff
@@ -38,6 +38,7 @@ task-local-extensions = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
 url = { workspace = true }
+google-cloud-auth = "0.13.2"
```
baszalmstra, line 41: Could you add this as a workspace dependency? 

Google-cloud-auth also by default adds the default-tls features. Could you make the tls implementation configurable using the features of the crate?

baszalmstra (changes requested):
Thanks! This looks almost good to me!

clement-chaneching:
Need to fix test & linting

clement-chaneching:
Hello!
Thanks for your reactivity.
I am not sure what is failing in the CI since I didn't modify the tests and I can't find any logs in the failed step. 
Do you have any ideas on this?

baszalmstra:
I also cant see any logs. I restarted the job.

crates/rattler_networking/src/authentication_middleware.rs
```diff
@@ -72,8 +74,39 @@ impl AuthenticationMiddleware {
...
     }
 
+    ///Authenticate with Google Cloud
+    async fn authenticate_with_google_cloud(
+        mut req: reqwest::Request,
+    ) -> reqwest_middleware::Result<reqwest::Request> {
+        let audience = "https://storage.googleapis.com/";
+        let scopes = [
+            "https://www.googleapis.com/auth/cloud-platform",
+            "https://www.googleapis.com/auth/devstorage.full_control",
```
wolfv, line 84: Is this the lowest permission? I don't think we need full control all the time, right? Maybe read-only is possible?

crates/rattler_networking/src/authentication_middleware.rs
```diff
@@ -4,6 +4,8 @@ use crate::{Authentication, AuthenticationStorage};
 use async_trait::async_trait;
 use base64::prelude::BASE64_STANDARD;
 use base64::Engine;
+use google_cloud_auth::project::create_token_source;
```
wolfv, line 7: do we need to make these `use` statements so that they are only enabled when the `google-cloud-auth` feature is enabled?

crates/rattler_networking/src/authentication_middleware.rs
```diff
@@ -104,6 +137,8 @@ impl AuthenticationMiddleware {
                 }
                 Authentication::CondaToken(_) => Ok(req),
             }
+        } else if req.url().host_str() == Some("storage.googleapis.com") {
```
wolfv, line 140: For the `oci` registry support we came up with a new "scheme". I also know that conda has support for an `s3://` scheme. 

Maybe we could also decide that channels that are on Google Cloud should indicate that with a `gcs://storage.googleapis.com/bla`? 

Potentially, people could also have public buckets there (and no need for authentication?).

clement-chaneching:
Thanks for these comments!
We now specify gcs://<bucketname> instead of the storage.googelapis endpoint which is indeed more natural for gcs users:
image

I have added the google-cloud-auth package as a feature so the CI is now failing.
Is it the right way to do it?

wolfv:
Hi @clement-chaneching - this is starting to look good! Thanks for taking care of my comments. How annoyed would you be with one more idea? 
I think ideally, this would live in its own file and be a seperate middleware. Then it can be easily enabled (or disabled) by anyone. You already have the functionality done, so that's all fine. Basically you would just take your code, and put it in it's own `GCSMiddleware` struct. 

WDYT?

Otherwise happy to merge and make the changes later :)

wolfv:
@clement-chaneching the "conditional" inclusion will also be easier when it's in it's own file. Because we can just only include & expose it when the feature is enabled.

clement-chaneching:
Yep it does make sense.
However im not so familiar with rust so not sure how to do it.
I have started this PR to separate my file in a different middleware
https://github.com/clement-chaneching/rattler/pull/1

But not sure how to test it.
I would guess it s on this [pixi file](https://github.com/prefix-dev/pixi/blob/main/src/utils/reqwest.rs) but I cant make it work.

Could you help me on this? :)

clement-chaneching:
Great thanks it's working!

clement-chaneching:
@wolfv Thanks for your comments and your help!
I can test properly in pixi using:
```
#[cfg(feature = "google-cloud-auth")]
use rattler_networking::GCSMiddleware;
And
    #[cfg(feature = "google-cloud-auth")]
    {client_builder = client_builder.with(GCSMiddleware);}
```

Is there anything I need to modify to validate the CI?

wolfv (approved):
Let's merge this! :)

PR #609, opened 2024-04-18
New trampoline

wolfv:
This fixes a bug I encountered with white spaces in the path to the prefix on Windows, and gives us win-arm64 support. I briefly tried to build the trampoline for win-32 as well, but didn't work out because I am missing `dlltool.exe`.

This does not create 2 files anymore, but just a single .exe file. For some reason, the script is stored as a zip encoded file (not sure if I would have done it that way, but whatever). IMO it could be just the text file with offsets stored somewhere. 

If we decide to use it we should add some more attribution :)


crates/rattler-bin/src/commands/create.rs
```diff
@@ -60,12 +60,23 @@ pub struct Opt {
...
     let channel_config = ChannelConfig::default();
-    let target_prefix = env::current_dir()?.join(".prefix");
-
+    let target_prefix = opt.target_prefix.unwrap_or_else(|| {
+        env::current_dir()
+            .expect("could not find current dir")
+            .join(".prefix")
+    });
+    // First create the target prefix if it doesn't exist yet.
+    fs::create_dir_all(&target_prefix).expect("could not create target prefix");
```
baszalmstra, line 76: I think this should result in a user error instead of an expect.

The canonicalize could be an expect because we already created the directory.

crates/rattler/src/install/entry_point.rs
```diff
@@ -39,21 +46,8 @@ pub fn create_windows_python_entry_point(
...
-        .join(format!("{}-script.py", &entry_point.command));
-
-    // Write the contents of the launcher script to disk
-    let script_path = target_dir.join(&relative_path_script_py);
-    std::fs::create_dir_all(
-        script_path
-            .parent()
-            .expect("since we joined with target_dir there must be a parent"),
-    )?;
+) -> Result<[PathsEntry; 1], std::io::Error> {
```
baszalmstra, line 49: This doesnt have to be an array anymore.

crates/rattler/src/install/entry_point.rs
```diff
@@ -62,34 +56,50 @@ pub fn create_windows_python_entry_point(
...
+    launcher.extend_from_slice(python_path.as_bytes());
+    launcher.extend_from_slice(
+        &u32::try_from(python_path.as_bytes().len())
+            .expect("File Path to be smaller than 4GB")
+            .to_le_bytes(),
+    );
+    launcher.extend_from_slice(&LAUNCHER_MAGIC_NUMBER);
+
+    let target_location = target_dir.join(&relative_path_script_exe);
+    fs::create_dir_all(target_location.parent().unwrap())?;
```
baszalmstra, line 91: Unwrap.

baszalmstra (commented):
If this works as expected then this looks good to me!

baszalmstra:
Did you build the binaries yourself? I am a little hesitant to include binaries without a proper reference to the source. I have no way to verify that this binary isnt mallicious.

baszalmstra:
Shall we close this in favor of https://github.com/mamba-org/rattler/pull/639 ?

wolfv:
yeah, let's close this one :)

Issue #613, opened 2024-04-19
Why is paths_data in a PrefixRecord not a list of PathsEntry objects?


iamthebot:
Trying to use py-rattler in a package that implements custom serialization of conda environments.

One thing I noticed is that while `PathsJSON` parses this information out from the original package archive (eg; from paths.json), the JSON files in conda-meta in an environment prefix (which if I understand correctly map to `PrefixRecord`) only capture the path location in `paths_data` as `os.Path`-like objects. This means we lose out on crucial information like `path_type` (eg; softlink vs. hardlink), `prefix_placeholder`, sha256/sha256_in_prefix, etc.

Does it seem reasonable to make this available in the PrefixRecord object? This would unblock us using significantly more py-rattler functionality for our use case.

baszalmstra:
Yeah that sounds good! The Rust side does expose these fields properly.

I would happily accept a PR!

iamthebot:
Yay! Working on that now. @baszalmstra I'm guessing for backwards compatibility this should go in a new property? Or would it be ok to just change `paths_data` to contain `PathsEntry` objects instead?

baszalmstra:
I would prefer to keep the API as much the same between Rust and Python as possible. So yeah just replace the type. If I remember correctly the old type is os.PathLike which is a protocol. Maybe you could implement that protocol for the new type too? That would keep the API break to a minimum.

iamthebot:
@baszalmstra yeah, let me read through the code and see what seems cleanest here. Will throw up a PR later today or monday.

PR #614, opened 2024-04-19
feat: add support for extracting prefix placeholder data to PathsEntry

iamthebot:
This is a prerequisite for #613. Currently, records in conda-meta contain prefix placeholder metadata. This isn't currently being extracted in `PathsEntry`.

This PR extracts the placeholder information so `PathsEntry` now contains exactly what's in the records in conda-meta.

### DIFF
--- a/crates/rattler_conda_types/src/prefix_record.rs
+++ b/crates/rattler_conda_types/src/prefix_record.rs
@@ -1,8 +1,8 @@
 //! Defines the `[PrefixRecord]` struct.
 
+use crate::package::FileMode;
 use crate::repo_data_record::RepoDataRecord;
 use crate::PackageRecord;
-use crate::package::FileMode;
 use rattler_digest::serde::SerializableHash;
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
@@ -44,7 +44,6 @@ impl From<Vec<PathsEntry>> for PrefixPaths {
         }
     }
 }
-
 
 /// Information about a single file installed for a package.
 ///
--- a/crates/rattler_conda_types/src/prefix_record.rs
+++ b/crates/rattler_conda_types/src/prefix_record.rs
@@ -2,6 +2,7 @@
 
 use crate::repo_data_record::RepoDataRecord;
 use crate::PackageRecord;
+use crate::package::FileMode;
 use rattler_digest::serde::SerializableHash;
 use serde::{Deserialize, Serialize};
 use serde_repr::{Deserialize_repr, Serialize_repr};
@@ -44,6 +45,7 @@ impl From<Vec<PathsEntry>> for PrefixPaths {
     }
 }
 
+
 /// Information about a single file installed for a package.
 ///
 /// This struct is similar to the [`crate::package::PathsEntry`] struct. The difference is that this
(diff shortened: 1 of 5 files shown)


crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -85,6 +87,14 @@ pub struct PathsEntry {
...
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub size_in_bytes: Option<u64>,
+
+    /// The file mode of the entry. This is used in conjunction with a prefix_placeholder
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub file_mode: Option<FileMode>,
+
+    /// The original sentinel value used for prefix-replacement from the package
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub prefix_placeholder: Option<String>,
```
baszalmstra, line 97: Is it still useful to have this? This will have been replaced once the files have been installed.

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -85,6 +87,14 @@ pub struct PathsEntry {
     /// The size of the file in bytes
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub size_in_bytes: Option<u64>,
+
+    /// The file mode of the entry. This is used in conjunction with a prefix_placeholder
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub file_mode: Option<FileMode>,
```
baszalmstra, line 93: this adds a number of bytes to the conda-meta output for every path, would it make sense to only serialize this if it is binary?

crates/rattler_conda_types/src/prefix_record.rs
iamthebot: Yes it's extremely useful. If we need to pack the environment the original placeholder and file modes are needed to make the files relocatable if the package cache is cleared. Otherwise we need to fetch all the packages into the package cache (which can be slow and may not even work in the future).

crates/rattler_conda_types/src/prefix_record.rs
iamthebot: This should actually be none for files as long as they don't contain a placeholder originally. So it wouldn't be serialized in the case it's none, right?

crates/rattler_conda_types/src/prefix_record.rs
baszalmstra: I guess what I dont understand is why the original placeholder is needed? This has already been replaced with a predictable value, namely the prefix that the files are in? It seems redundant to store the original placeholder in that case.

iamthebot: Suppose you have a binary file with a placeholder like `placeholder_placeholder_placeholder_placeholder_placeholder`. This has a given length, say, 150 bytes.

During the installation/linking process all instances of that placeholder are replaced with the installation path. This can be followed by a further path component and then the rest is padded with null characters up to the length of the original placeholder's length.

To be able to safely reverse this process, we need to know the original placeholder's length. I suppose a more efficient alternative would be to store the original placeholder length. But then we'd be deviating from what's actually in the contents of these JSON files. It also means that you wouldn't be able to use the original SHA256 of the file to verify making the file relocatable was done predictably (which requires knowing the original placeholder).

baszalmstra: 👍

baszalmstra:
I think you just need to run `cargo fmt`

baszalmstra:
Do you also need a release? Or are you working from git?

PR #616, opened 2024-04-22
fix: issue 615

### LINKED ISSUE #615: UrlOrPath - store absolute Windows URLs with `file:///`?
I am looking at some issues we encountered with absolute paths on Windows. 

If I use the following in a pixi.toml file:

```toml
[pypi-dependencies]
minimal-project = { path = "C:/Users/wolfv/Programs/pixi/examples/pypi-source-deps/minimal-project" }
```

Then it works once, but upon rechecking fails because it can't find the corresponding package. In the lockfile we find:

```
      - pypi: C:/Users/wolfv/Programs/pixi/examples/pypi-source-deps/minimal-project
```

This can be deserialized as a URL, so that's where the failure comes from. We make the distinction for relative paths (because they cannot be stored as `file:///` URL. I am wondering if we should deserialize a `file:///C:/...` URL to Path and fix the problem that way.

Or we don't allow absolute paths.

baszalmstra:
Fixed #615

### DIFF
--- a/crates/rattler_lock/src/lib.rs
+++ b/crates/rattler_lock/src/lib.rs
@@ -602,6 +602,16 @@ mod test {
         insta::assert_yaml_snapshot!(file_name, conda_lock);
     }
 
+    /// Absolute paths on Windows are not properly parsed.
+    /// See: <https://github.com/mamba-org/rattler/issues/615>
+    #[test]
+    fn test_issue_615() {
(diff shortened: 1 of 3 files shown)


wolfv:
I will double check with Ruben because I think it's also about nested file requirements. For example, some package might have a requires-dist that points to a specific absolute path and our validation was potentially failing there.

PR #618, opened 2024-04-22
fix: compare `UrlOrPath`

baszalmstra:
This is a more general fix for https://github.com/prefix-dev/pixi/pull/1224 .

The basic idea is that you can "canonicalize" a `UrlOrPath`. If the stored Url represents an absolute path, it is converted into a path instead. This is not completely trivial because the `Url::to_file_path` method behaves differently depending on the architecture. I had to reimplement this method.

When checking the equality of two `UrlOrPath`s they are first canonicalized so that comparing a `file:///...` and a `/..` url and path works properly. 

`UrlOrPath`s are not canonicalized before serialization, this ensures that loading a lock file and saving it again will result in the same lock file. However, when _creating_ a lock file I think it makes sense to canonicalize any instance you create.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -92,6 +92,7 @@ ouroboros = "0.18.3"
 pathdiff = "0.2.1"
 pep440_rs = { version = "0.5.0" }
 pep508_rs = { version = "0.4.2" }
+percent-encoding = "2.3.1"
 pin-project-lite = "0.2.14"
 plist = "1"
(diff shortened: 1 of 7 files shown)


ruben-arts (approved):
This seems to work for me it fixes the same issue on my rerun example. If you tested it on windows I totally fine with it!

tdejager:
I agree @ruben-arts, will merge this now :)

PR #620, opened 2024-04-23
Expose paths_data as PathEntry in py-rattler

### LINKED ISSUE #613: Why is paths_data in a PrefixRecord not a list of PathsEntry objects?

iamthebot:
Closes #613 and modifies the API in PrefixRecord so that paths_data exposes PathsEntry objects. This allows access to useful data like the original placeholder, sha256, file mode, etc.

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -282,6 +282,7 @@ pub fn link_file(
             .map_err(LinkFileError::FailedToOpenDestinationFile)?;
         metadata.len()
     };
+
     let prefix_placeholder: Option<String> = path_json_entry
         .prefix_placeholder
         .as_ref()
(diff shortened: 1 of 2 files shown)


baszalmstra:
Looks like you are still missing the python classes for the newly introduced types. And would you be able to ensure that PathsEntry inplements the os.Pathlike protocol?

baszalmstra:
Would be nice if the whole thing also implemented the protocol. Then we keep the API intact.

baszalmstra (approved):

PR #624, opened 2024-04-25
feat: create SparseRepoData from byte slices

### LINKED ISSUE #619: Feature request: create SparseRepoData from memory
### Motivation

I have an AWS Lambda function that solves conda environments using rattler. It fetches cached repodata.json files from S3 and loads them into the solver. Using `SparseRepoData` gives us excellent performance, except for the fact that we need to extract the repodata to disk first and memmap it back to memory (resulting in ~1 second of overhead). For simple scenarios, this accounts for _half_ of the time spent solving! Therefore I'm looking into a way of avoiding the disk + memmap roundtrip.

Below is a screenshot from our traces, showing the overhead in red when solving an environment for the single matchspec `python`:

image

### Possible API design

We need a constructor that takes the repodata file's bytes directly instead of a `Path`. In the sketch below I'm using `&[u8]` instead of `Vec<u8>`, because that makes it unnecessary to use self-referential borrows (I think).

```rust
impl SparseRepoData {
    pub fn from_bytes(
        channel: Channel,
        subdir: impl Into<String>,
        bytes: &[u8],
        patch_function: Option<fn(&mut PackageRecord)>,
    ) -> Result<Self, io::Error> {
        todo!()
    }
}
```
### Open questions

1. Do you think this addition to rattler is worth it? Or would you rather have me implement it in a separate codebase?
2. I don't see how my in-memory approach would be compatible with SparseRepoData. From a cursory investigation it looks like I'd need to introduce a separate type (maybe `InMemorySparseRepoData`), but I think I could reuse `LazyRepoData`.

aochagavia:
~This is a breaking change, because it introduces a lifetime parameter for `SparseRepoData`. If that's undesirable, maybe we could create an inner type (e.g. `SparseRepoData2`) wrapped by two outer types (e.g. `MemmappedSparseRepoData` and `ByteSliceSparseRepoData`).~ We ended up using `Bytes` with a self-referential borrow to get rid of the lifetime entirely, so now the change is no longer breaking 🎉.

I also took the liberty to switch from stable to unstable sorting, which has better performance characteristics. If you'd rather leave things as they are now, I can drop the commit.

Fixes #619

### DIFF
--- a/crates/rattler_repodata_gateway/Cargo.toml
+++ b/crates/rattler_repodata_gateway/Cargo.toml
@@ -13,6 +13,7 @@ readme.workspace = true
 [dependencies]
 async-compression = { workspace = true, features = ["gzip", "tokio", "bzip2", "zstd"] }
 blake2 = { workspace = true }
+bytes = { workspae = true, optional = true }
 cache_control = { workspace = true }
 chrono = { workspace = true, features = ["std", "serde", "alloc", "clock"] }
 humansize = { workspace = true }
@@ -61,4 +62,4 @@ tracing-test = { workspace = true }
 default = ['native-tls']
 native-tls = ['reqwest/native-tls']
 rustls-tls = ['reqwest/rustls-tls']
-sparse = ["rattler_conda_types", "memmap2", "ouroboros", "superslice", "itertools", "serde_json/raw_value"]
+sparse = ["rattler_conda_types", "memmap2", "ouroboros", "superslice", "itertools", "serde_json/raw_value", "bytes"]
--- a/crates/rattler_repodata_gateway/Cargo.toml
+++ b/crates/rattler_repodata_gateway/Cargo.toml
@@ -13,7 +13,7 @@ readme.workspace = true
 [dependencies]
 async-compression = { workspace = true, features = ["gzip", "tokio", "bzip2", "zstd"] }
 blake2 = { workspace = true }
-bytes = { workspae = true, optional = true }
+bytes = { workspace = true, optional = true }
 cache_control = { workspace = true }
 chrono = { workspace = true, features = ["std", "serde", "alloc", "clock"] }
 humansize = { workspace = true }
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -406,7 +406,7 @@ fn deserialize_filename_and_raw_record<'d, D: Deserializer<'d>>(
     //
     // Since (in most cases) the repodata is already ordered by filename which does closely resemble
     // ordering by package name this sort operation will most likely be very fast.
-    entries.sort_by(|(a, _), (b, _)| a.package.cmp(b.package));
+    entries.sort_unstable_by(|(a, _), (b, _)| a.package.cmp(b.package));
 
     Ok(entries)
 }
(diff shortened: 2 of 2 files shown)


baszalmstra:
> I also took the liberty to switch from stable to unstable sorting, which has better performance characteristics. If you'd rather leave things as they are now, I can drop the commit.

Did you profile this with a real-world scenario? Most repodata.json is already sorted but just on a slightly different key therefor I didn't think this would matter much.

Im also a little hesitant to add randomness to the repodata.json. I don't think the order of the records in the file matters much but I am not completely sure.

> This is a breaking change, because it introduces a lifetime parameter for SparseRepoData.

Look at the changes, can we use a `'static` lifetime for the memory mapped data, which would still allow use to pass it around in `Arc<SparseRepoData<'static>>`s?

aochagavia:
> Look at the changes, can we use a 'static lifetime for the memory mapped data, which would still allow use to pass it around in Arc<SparseRepoData<'static>>s?

That's indeed possible (in fact, I had to manually specify the `'static` lifetime somewhere and it worked). On the other hand, it might be more user friendly to keep `SparseRepoData` as it is now and introduce an additional `ByteSliceRepoData<'a>`. Both structs would internally use the same code, and users would be spared the pain of updating and having to specify lifetimes everywhere (which might even be confusing for people). What do you think?

> Did you profile this with a real-world scenario? Most repodata.json is already sorted but just on a slightly different key therefor I didn't think this would matter much.

I haven't profiled it. It's just that the stable version allocates half the size of self, according to [the docs](https://doc.rust-lang.org/stable/std/primitive.slice.html#method.sort), so it feels "more correct". The docs even recommend using `sort_unstable` if you can get away with it (unless you really need a stable sort, that is). Also good to know, the sort is deterministic (it uses a constant random seed).

I don't think it _really_ matters in practice, so let me know if you'd rather have me drop the commit 😉

aochagavia:
Rebased and figured out how to make the change non-breaking. This is now ready for review!

crates/rattler_repodata_gateway/Cargo.toml
```diff
@@ -13,6 +13,7 @@ readme.workspace = true
 [dependencies]
 async-compression = { workspace = true, features = ["gzip", "tokio", "bzip2", "zstd"] }
 blake2 = { workspace = true }
+bytes = "1.6.0"
```
baszalmstra, line 16: Since `sparse` is a feature, we should only enable this when the `sparse` feature is used.

baszalmstra (changes requested):
One tiny little comment but otherwise LGTM

crates/rattler_repodata_gateway/Cargo.toml
baszalmstra: Oh and also please add it as a workspace dependency

aochagavia: Done!

crates/rattler_repodata_gateway/Cargo.toml
```diff
@@ -13,6 +13,7 @@ readme.workspace = true
 [dependencies]
 async-compression = { workspace = true, features = ["gzip", "tokio", "bzip2", "zstd"] }
 blake2 = { workspace = true }
+bytes = { workspae = true, optional = true }
```
baszalmstra, line 16:
```suggestion
bytes = { workspace = true, optional = true }
```

aochagavia: Thanks! Weird that the CI didn't catch it

baszalmstra (approved):

PR #639, opened 2024-05-02
add cli-32 and cli-64 from conda project

wolfv:
I think we should still document more where these come from. I'll try to find more info from @jezdez and co.


baszalmstra (approved):

jaimergp:
Those launchers are copied from [conda-build](https://github.com/conda/conda-build/tree/main/conda_build/launcher_sources).

crates/rattler/src/install/entry_point.rs
```diff
@@ -12,8 +12,8 @@ use std::{fs::File, io, io::Write, path::Path};
 /// Get the bytes of the windows launcher executable.
 pub fn get_windows_launcher(platform: &Platform) -> &'static [u8] {
     match platform {
-        Platform::Win32 => unimplemented!("32 bit windows is not supported for entry points"),
-        Platform::Win64 => include_bytes!("../../resources/launcher64.exe"),
+        Platform::Win32 => include_bytes!("../../resources/cli-32.exe"),
```
ruben-arts, line 15: Shouldn't these only be included in their respective targets?

wolfv: No, because you might want to build a win-32 env on win-64, or a win-64 env on linux (and send it to a cloud machine, use pixi-pack, ...)

ruben-arts: Ah oke

wolfv:
> Those launchers are copied from [conda-build](https://github.com/conda/conda-build/tree/main/conda_build/launcher_sources).

This was just an observation or something actionable? :)

Unfortunately, the knowledge on how to _build_ these launchers seems to have been "lost" :D 

I had high hopes that Anaconda could build the win-arm64 launcher (cc @jezdez).

jezdez:
@wolfv this _is_ a good idea IMO, for the https://github.com/conda-incubator/installer team, I'm tapping @marcoesters and @Callek to help move the building of the binaries into an own repo, so we can start using them and build them correctly for new platforms

baszalmstra:
Should this PR be updated with the latest launchers @wolfv ?

wolfv:
Should we try the `zig` built ones @baszalmstra? :)

baszalmstra:
That would have my preference for sure!

jaimergp:
They are not codesigned, if that makes a difference.

baszalmstra:
So the status quo is that we do not have signed binaries. Our binaries are currently very hard to trace back.

I opened in issue in the launchers repo to make this process a bit more streamlined: https://github.com/conda/conda-launchers/issues/11

wolfv:
Yeah we can close this one for now. Ideally, the launchers upload a package to prefix.dev from which we can extract the binaries, or upload binaries to github releases from where we can take them.

PR #641, opened 2024-05-03
fix: sha computation of symlinks was failing sometimes

wolfv:
Sometimes linking was failing because we tried to compute the SHA hash of a symlink-file. When computing the hash, we follow the symlink. But since we are symlinking everything at once, and in parallel, the symlinked file is sometimes not yet there.

In that case, the SHA hash computation was failing. 

With this PR, we change it so that we always store `000...000` for symlinks.
Conda and mamba 'wait' for the files to be ready and then compute (or find) the hash. We could also do that, but on the other hand the "sha" hash of a symlink doesn't make that much sense anyways. 

Thoughts?

ref. https://github.com/prefix-dev/pixi/issues/1309

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -4,8 +4,8 @@ use memmap2::Mmap;
 use once_cell::sync::Lazy;
 use rattler_conda_types::package::{FileMode, PathType, PathsEntry, PrefixPlaceholder};
 use rattler_conda_types::Platform;
-use rattler_digest::HashingWriter;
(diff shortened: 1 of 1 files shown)


wolfv:
Note: before, if we had the paths.json, then we would use the SHA hash that is stored in there (which is also not correct if the symlink target file had a prefix replacement).

baszalmstra (approved):
This is not the final solution because the hashes are now incorrect for symlinks (they also already were). We are planning to add a proper solution where we first process all non-symlinks and process the symlinks after.

PR #643, opened 2024-05-06
fix: use the output of `readlink` as hash for softlinks

wolfv:


wolfv:
Also fixes the `rattler create` when we already have a prefix.

baszalmstra (approved):

Issue #648, opened 2024-05-09
[py-rattler] AttributeError: 'NoneType' object has no attribute 'text' when a PathsEntry has no filemode


iamthebot:
This is semi related to #621 (in the sense that `FileMode` is exposed as a series of boolean properties currently in Python vs. an enum).

Looks like the current implementation will throw an `AttributeError` is a path data entry doesn't have a `FileMode` defined (which will be the case if there's no prefix placeholder) because the `_inner` object won't be defined on `FileMode`.

The most direct fix for this is to just return `False` for the `text` and `binary` properties in this case. At least until we address #621

iamthebot:
I wrote some of this code so I'll push a PR up for this tonight/tomorrow morning @baszalmstra

iamthebot:
This was closed by #651

PR #651, opened 2024-05-10
fix: Address (#648) Missing FileMode causes an AttributeError in py-rattler and fix missing properties

iamthebot:
This contains a fix for #648 by defaulting `_inner` to `None` which prevents an `AttributeError` if trying to get which `FileMode` we have in the case where the `FileMode` isn't defined. It further adds an `unknown` property to check for this case specifically in Python.

I've also added a few property decorators that were missing in that file (eg; for `sha256`).

Note a few files were also slightly reformatted when running ruff so I've added those to spare folks the trouble later (unit tests/conftest).


wolfv:
OK, I double checked and now I do think that this is the correct behavior! Sorry for the noise.

baszalmstra (commented):
LGTM! Thanks!

PR #654, opened 2024-05-13
feat: exclude repodata records based on timestamp

baszalmstra:
Allows setting an `exclude_newer` field on `SolverTask` to exclude any record uploaded after a certain date. This is useful to make repodata from remote source more deterministic.


crates/rattler_solve/Cargo.toml
```diff
@@ -37,7 +37,7 @@ tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
 url = { workspace = true }
 
 [features]
-default = ["libsolv_c"]
+default = ["resolvo"]
```
baszalmstra, line 40: Sneaky little change!

PR #664, opened 2024-05-16
refactor: introducing the installer

baszalmstra:
This PR introduces the `Installer` which handles executing transactions and more. The installer can be observed with the `Reporter` trait. A default implementation is provided which uses indicatif.

### DIFF
--- a/crates/rattler/src/install/installer/mod.rs
+++ b/crates/rattler/src/install/installer/mod.rs
@@ -11,7 +11,10 @@ use std::{
 pub use error::InstallerError;
 use futures::{stream::FuturesUnordered, FutureExt, StreamExt, TryFutureExt};
 #[cfg(feature = "indicatif")]
-pub use indicatif::{DefaultProgressFormatter, IndicatifReporter, IndicatifReporterBuilder, ProgressFormatter};
+pub use indicatif::{
+    DefaultProgressFormatter, IndicatifReporter, IndicatifReporterBuilder, Placement,
+    ProgressFormatter,
+};
 use rattler_conda_types::{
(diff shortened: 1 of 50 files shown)


baszalmstra:
Feel free to review, but please don't merge yet. I wanna test this out in pixi first.

baszalmstra:
With the Python API it is now easier than ever to create a conda environment from scratch:

```python
# Solve an environment with Python 3.9 for the current platform
records = await solve(channels=["conda-forge"], specs=["python=3.9"])

# Link the environment in the directory `my-env`.
await link(records, target_prefix="my-env")

# That's it! You now have a functioning environment with Python 3.9 in the "my-env" directory.
```

baszalmstra:
Renamed `link` to `install`

crates/rattler/src/install/installer/mod.rs
```diff
@@ -0,0 +1,556 @@
...
+
+        Ok(InstallationResult {
+            transaction,
+            pre_process_result,
+            post_process_result,
+        })
+    }
+}
+
+async fn link_package(
```
wolfv, line 448: should we make this (and the next function) pub? Could have some use cases.

baszalmstra: We could also add this behavior to the `link_package` function that is called inside this function. Maybe that makes more sense since its already public. WDYT?

baszalmstra:
I tested it in pixi as well: https://github.com/prefix-dev/pixi/pull/1438 

@wolfv feel free to merge if you have no new objections!

wolfv (approved):

PR #666, opened 2024-05-21
fix: result grouped by subdir instead of channel

baszalmstra:
Fixes an issue with duplicate records in the solver.


nichmor:
small idea-  can we change the error type so the next time it is propagate to pixi and we could output it?

baszalmstra:
Yeah that could work as well! Rhe error originates from passing duplicate filenames to the solver that arent actually unique. This is a user error so we could also return an error.

PR #669, opened 2024-05-22
feat: always store purls as a key in lock file

nichmor:
related to : https://github.com/prefix-dev/pixi/pull/1425

In this PR we will always store purls in lock file so missing purl will be just an empty vector.


crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -132,8 +132,8 @@ pub struct PackageRecord {
 
     /// Package identifiers of packages that are equivalent to this package but from other
     /// ecosystems.
```
baszalmstra, line 134: Maybe add a comment explaining, the difference between `None` and `Some(Vec::new())`

baszalmstra (changes requested):

crates/rattler_conda_types/src/repo_data/mod.rs
nichmor: done!

crates/rattler_lock/src/parse/v3.rs
```diff
@@ -170,7 +170,7 @@ pub fn parse_v3_or_lower(
                             timestamp: value.timestamp,
                             track_features: value.track_features,
                             version: value.version,
-                            purls: value.purls,
+                            purls: Some(value.purls),
```
baszalmstra, line 173: I think this should be `None` if `value.purls` is empty. Basically the snapshots shouldn't really chance because the current state of affairs is that if the purls are not present they are unknown.

nichmor: changed

crates/rattler_lock/src/parse/v3.rs
```diff
@@ -170,7 +170,10 @@ pub fn parse_v3_or_lower(
                             timestamp: value.timestamp,
                             track_features: value.track_features,
                             version: value.version,
-                            purls: value.purls,
+                            purls: match value.purls.is_empty() {
+                                true => None,
+                                false => Some(value.purls),
+                            },
```
baszalmstra, line 176:
```suggestion
                            purls: value.purls.is_empty().then_some(value.purls),
```

nichmor: done!

baszalmstra (approved):

PR #677, opened 2024-05-24
feat: bump characters in versions as well (9d -> 10a) and add `alpha`

wolfv:
As part of the work for compatible `pin` expressions in `conda-build` we might also have to bump characters, as well as adding `0a0` segments when appropriate. 

I am currently working out the details, but this is one step :)


wolfv:
superseded by a number of PRs that were merged :)

PR #686, opened 2024-05-27
fix: VersionSpec starts with, with trailing zeros

### LINKED ISSUE #685: Cannot solve cmdstan=2.33.1=hc24a2e5_0 (linux) from osx-arm64
Hello, I'm trying to emulate `CONDA_OVERRIDE_GLIBC=2.17 conda install -n test cmdstan=2.33.1=hc24a2e5_0 --subdir linux-64` with rattler and the solver doesn't seem to be able to resolve. I'm not too sure what's going on here.

```python
>>> import asyncio
>>> import rattler
>>> virtual_packages = [
...     rattler.GenericVirtualPackage(rattler.PackageName("__glibc"), rattler.Version("2.17"), "0"),
...     rattler.GenericVirtualPackage(rattler.PackageName("__archspec"), rattler.Version("1"), "x86_64"),
...     rattler.GenericVirtualPackage(rattler.PackageName("__unix"), rattler.Version("0"), "0"),
...     rattler.GenericVirtualPackage(rattler.PackageName("__linux"), rattler.Version("0"), "0"),
... ]
>>> asyncio.run(rattler.solve([rattler.Channel("main")], ["cmdstan=2.33.1=hc24a2e5_0"], platforms=["linux-64", "noarch"], virtual_packages=virtual_packages))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/jcmorin/miniconda/envs/test/lib/python3.12/asyncio/runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "/Users/jcmorin/miniconda/envs/test/lib/python3.12/asyncio/runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/jcmorin/miniconda/envs/test/lib/python3.12/asyncio/base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "/Users/jcmorin/miniconda/envs/test/lib/python3.12/site-packages/rattler/solver/solver.py", line 81, in solve
    for solved_package in await py_solve(
                          ^^^^^^^^^^^^^^^
exceptions.SolverException: Cannot solve the request because of: cmdstan ==2.33.1 hc24a2e5_0 cannot be installed because there are no viable options:
└─ cmdstan 2.33.1 would require
   └─ gxx_linux-64 11.2.0.*, which cannot be installed because there are no viable options:
      └─ gxx_linux-64 11.2.0 would require
         └─ binutils_linux-64 ==2.38.0 hc2dff05_0, which cannot be installed because there are no viable options:
            └─ binutils_linux-64 2.38.0 would require
               └─ binutils_impl_linux-64 2.38.0.*, for which no candidates were found.

```
Am I doing something wrong or am I missing something?

Thanks!

baszalmstra:
This fixes an issue where if we check if a version a starts with b, we should ignore any trailing zero segments in b. This is because a version `1` is the same as `1.0` and `1.0.0.0.0`. So in that sense `1` starts with `1.0` as well. 

Im personally not extremely happy with this because I think its complex to understand, but this seems to be the way its implemented in conda and mamba.

Fixes #685

### DIFF
--- a/crates/rattler_conda_types/src/version_spec/mod.rs
+++ b/crates/rattler_conda_types/src/version_spec/mod.rs
@@ -331,6 +331,7 @@ impl VersionSpec {
 #[cfg(test)]
 mod tests {
     use assert_matches::assert_matches;
+    use rstest::rstest;
 
     use crate::version_spec::parse::ParseConstraintError;
     use crate::version_spec::{
@@ -426,6 +427,17 @@ mod tests {
     #[test]
     fn issue_204() {
         assert!(VersionSpec::from_str(">=3.8<3.9", ParseStrictness::Strict).is_err());
+    }
+
+    #[rstest]
+    #[case("2.38.*", true)]
+    #[case("2.38.0.*", true)]
+    #[case("2.38.0.1*", false)]
+    #[case("2.38.0a.*", false)]
+    fn issue_685(#[case] spec: &str, #[case] starts_with: bool) {
+        let spec = VersionSpec::from_str(spec, ParseStrictness::Strict).unwrap();
+        let version = &Version::from_str("2.38").unwrap();
+        assert_eq!(spec.matches(version), starts_with);
     }
 
     #[test]
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -583,7 +583,15 @@ fn segments_starts_with<
         let (left, right) = match ranges {
             EitherOrBoth::Both(left, right) => (left, right),
             EitherOrBoth::Left(_) => return true,
-            EitherOrBoth::Right(_) => return false,
+            EitherOrBoth::Right(segment) => {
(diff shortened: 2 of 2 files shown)


crates/rattler_conda_types/src/version_spec/mod.rs
```diff
@@ -428,6 +429,17 @@ mod tests {
         assert!(VersionSpec::from_str(">=3.8<3.9", ParseStrictness::Strict).is_err());
     }
 
+    #[rstest]
+    #[case("2.38.*", true)]
+    #[case("2.38.*", true)]
```
wolfv, line 434: This should be a different case right?

baszalmstra: Thanks, its getting way to late

PR #689, opened 2024-05-28
feat: add a function to extend version with `0s`

wolfv:
This function will extend a given version to a certain length. 

If the version is already longer than the specified length, it will just return it.

For example:

```
Extending to length of 3:
1 -> 1.0.0
1.2 -> 1.2.0

1!1.2 -> 1!1.2.0
1!1.2+3.4 -> 1!1.2.0+3.4
```

This is in preparation of making the version bumping extend the version.

### DIFF
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -579,9 +579,10 @@ impl Version {
     }
 
     /// Extend the version to the specified length by adding default components (0s).
-    pub fn extend_to_length(&self, length: usize) -> Result<Version, VersionExtendError> {
+    /// If the version is already longer than the specified length it is returned as is.
+    pub fn extend_to_length(&self, length: usize) -> Result<Cow<'_, Version>, VersionExtendError> {
         if self.segment_count() >= length {
-            return Ok(self.clone());
+            return Ok(Cow::Borrowed(self));
         }
 
         // copy everything up to local version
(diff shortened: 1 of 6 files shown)


crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -567,6 +574,52 @@ impl Version {
             Cow::Borrowed(self)
         }
     }
+
+    /// Extend the version to the specified length by adding default components (0s).
+    pub fn extend_to_length(&self, length: usize) -> Result<Version, VersionExtendError> {
```
baszalmstra, line 579: Would be nice to return a `Cow` that is borrowed if the length is already long enough.

```suggestion
    pub fn extend_to_length(&self, length: usize) -> Result<Cow<'_, Version>, VersionExtendError> {
```

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -567,6 +574,52 @@ impl Version {
...
+        // copy everything up to local version
+        let mut segments = self.segments[..self.segment_count()].to_vec();
+        let components_end = segments.iter().map(|s| s.len() as usize).sum::<usize>()
+            + usize::from(self.has_epoch());
+        let mut components = self.components.clone()[..components_end].to_vec();
+
+        // unwrap is OK here because these should be fine to construct
+        let segment = Segment::new(1).unwrap().with_separator(Some('.')).unwrap();
+
+        if length > self.segment_count() {
```
baszalmstra, line 593: This `if` statement is not really needed right?

crates/rattler_conda_types/src/version/mod.rs
wolfv: Done!

crates/rattler_conda_types/src/version/mod.rs
wolfv: removed.

baszalmstra (approved):

PR #704, opened 2024-05-30
feat: parse url and path as matchspec

### LINKED ISSUE #640: Fix panic in matchspec parser
The matchspec parser contains a call to `unimplemented!` (see [here](https://github.com/mamba-org/rattler/blob/e1eea9acf47cb57ee339a49f2fd794de2d602d45/crates/rattler_conda_types/src/match_spec/parse.rs#L390)). For a program that accepts arbitrary matchspecs, that means specific matchspecs could cause the program to unexpectedly crash. I think it'd make sense to extend `ParseMatchSpecError ` with an additional variant for unimplemented bits, to avoid panics.

Would you welcome a PR for this?

ruben-arts:
Fixes #640


crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -370,6 +375,26 @@ impl NamelessMatchSpec {
     }
 }
 
+fn parse_url(input: &str) -> Result<Url, ParseMatchSpecError> {
+    match Url::parse(input) {
+        Ok(url) => Ok(url),
+        Err(_) => {
+            #[cfg(not(target_arch = "wasm32"))]
+            {
+                match PathBuf::from_str(input) {
```
baszalmstra, line 384: This is platform dependent. Therefore it's better to use the `file_url` crate.

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -380,22 +405,25 @@ fn matchspec_parser(
...
-            #[cfg(target_arch = "wasm32")]
-            Err(_) => return Err(ParseMatchSpecError::InvalidPackagePathOrUrl),
-            #[cfg(not(target_arch = "wasm32"))]
-            Err(_) => match PathBuf::from_str(input) {
-                Ok(path) => Url::from_file_path(path)
-                    .map_err(|_err| ParseMatchSpecError::InvalidPackagePathOrUrl)?,
-                Err(_) => return Err(ParseMatchSpecError::InvalidPackagePathOrUrl),
-            },
-        };
+        let url = PathBuf::from_str(input)
```
baszalmstra, line 410: You can't use `PathBuf::from_str` because it is platform-dependent. E.g. `C:\Windows\bla` is parsed differently on Unix and Windows.

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -820,4 +855,53 @@ mod tests {
...
+            "C:/Users/user/conda-bld/linux-64/foo-1.0-py27_0.tar.bz2",
+            Strict,
+        )
+        .unwrap();
+        assert_eq!(
+            spec.url,
+            Some(
+                Url::parse("file:C:/Users/user/conda-bld/linux-64/foo-1.0-py27_0.tar.bz2").unwrap()
+            )
+        );
```
baszalmstra, line 886:
```suggestion
        let spec = MatchSpec::from_str(
            "C:\\Users\\user\\conda-bld\\linux-64\\foo-1.0-py27_0.tar.bz2",
            Strict,
        )
        .unwrap();
        assert_eq!(
            spec.url,
            Some(
                Url::parse("file://C:/Users/user/conda-bld/linux-64/foo-1.0-py27_0.tar.bz2").unwrap()
            )
        );
```

Windows paths are backslashes. I added that here. :)

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -380,22 +405,25 @@ fn matchspec_parser(
...
+
+        let identifier = ArchiveIdentifier::try_from_url(&url)
+            .ok_or(ParseMatchSpecError::InvalidPackagePathOrUrl)?;
+        let file_name = identifier.to_file_name();
 
-        // TODO: Implementing package file specs
-        unimplemented!()
+        return Ok(MatchSpec {
+            url: Some(url),
+            file_name: Some(file_name),
```
baszalmstra, line 424: I would completely remove everything to do with the file_name. The user only specified a URL, we don't need to add more restrictions into the matchspec itself because it no longer matches the user's intent.

baszalmstra (changes requested):
Looks good! 

Maybe you can also add a number of test cases that _fail_ to parse. To also test the non-happy path.

crates/rattler_conda_types/src/match_spec/parse.rs
ruben-arts: The problem is that I can't use file_url to start with as it will break if it isn't an URL. So I start with checking for a path, which succeeds for both and then map the file_url::file-path_to_url if it is a path. otherwise it will try to parse as an normal url.

crates/rattler_conda_types/src/match_spec/parse.rs
ruben-arts: Fixed, using file_url now

crates/rattler_conda_types/src/match_spec/parse.rs
ruben-arts: Resolved!

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -757,11 +764,17 @@ mod tests {
         // Please keep this list sorted.
         let specs = [
             "blas *.* mkl",
+            "C:/Users/user/conda-bld/linux-64/foo-1.0-py27_0.tar.bz2",
```
baszalmstra, line 767:
```suggestion
            "C:\\Users\\user\\conda-bld\\linux-64\\foo-1.0-py27_0.tar.bz2",
```

baszalmstra (commented):
One small nitpick, otherwise LGTM!

baszalmstra (approved):

PR #706, opened 2024-05-30
feat: keep full timestamp precision and return datetime.datetime

### LINKED ISSUE #700: Timestamp in RepoDataRecord is in seconds instead of milliseconds
Hello, I noticed that the timestamp attached to a RepoDataRecord in py-rattler seems to be in seconds instead of being in milliseconds.

From what I understand, this seems to be caused by https://github.com/mamba-org/rattler/blob/a91ca46e49a75e8065d0578760626bfeab3782c4/py-rattler/src/record.rs#L212

According to https://docs.rs/chrono/latest/chrono/struct.DateTime.html#method.timestamp, the timestamp method returns seconds.

The thing is that in repodata.json, the timestamps are in milliseconds, not seconds so we loose precision.

Is this intentional?

wolfv:
fixes #700 

It's a breaking change though!

### DIFF
--- a/py-rattler/rattler/package/index_json.py
+++ b/py-rattler/rattler/package/index_json.py
@@ -297,12 +297,12 @@ class IndexJson:
         ...     "../test-data/with-symlinks/python-3.10.6-h2c4edbf_0_cpython.tar.bz2"
         ... )
         >>> idx_json.timestamp
-        datetime.datetime(2022, 8, 22, 22, 39, 2, 467000)
+        datetime.datetime(2022, 8, 22, 20, 39, 2, 467000, tzinfo=datetime.timezone.utc)
         >>>
         ```
         """
         if timestamp := self._inner.timestamp:
-            return datetime.datetime.fromtimestamp(timestamp / 1000.0)
+            return datetime.datetime.fromtimestamp(timestamp / 1000.0, tz=datetime.timezone.utc)
 
         return None
 
--- a/py-rattler/rattler/package/index_json.py
+++ b/py-rattler/rattler/package/index_json.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 import os
+import datetime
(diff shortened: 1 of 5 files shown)


py-rattler/rattler/package/index_json.py
```diff
@@ -296,12 +297,12 @@ def timestamp(self) -> Optional[int]:
...
         ... )
         >>> idx_json.timestamp
-        1661200742
+        datetime.datetime(2022, 8, 22, 22, 39, 2, 467000)
         >>>
         ```
         """
         if timestamp := self._inner.timestamp:
-            return timestamp
+            return datetime.datetime.fromtimestamp(timestamp / 1000.0)
```
JeanChristopheMorinPerso, line 305: I'm wondering if we could potentially do something like
```python
datetime.datetime(1970, 1, 1, 0, 0, 0, 0, datetime.timezone.utc) + datetime.timedelta(milliseconds=timestamp)
```
to keep the precision. Also note that I think we should return an aware datetime instead of a naive datetime.

wolfv: I was under the impression that this timestamp: `        datetime.datetime(2022, 8, 22, 22, 39, 2, 467000)` contains milliseconds. But also happy to change it.

JeanChristopheMorinPerso: Ah, I see that you are keeping it as a float, so yeah then that works. In this case, you can use `datetime.datetime.fromtimestamp()` and you can pass a timezone
```python
datetime.datetime.fromtimestamp(timestamp / 1000.0, tz=datetime.timezone.utc)
```

wolfv: will add the `utc` parameter.

baszalmstra (approved):

PR #711, opened 2024-06-02
fix: use total duration of timedelta objects

beckermr:
This PR fixes a bug where the py-rattler solver timeout did not use the entire duration of the timedelta object. 

edit: With this fix, it should work. The hack I was using to get around this bug doesn't work and so the solver thought I passed a timeout of zero.

### DIFF
--- a/py-rattler/rattler/solver/solver.py
+++ b/py-rattler/rattler/solver/solver.py
@@ -92,7 +92,7 @@ async def solve(
             pinned_packages=[package._record for package in pinned_packages or []],
             virtual_packages=[v_package._generic_virtual_package for v_package in virtual_packages or []],
             channel_priority=channel_priority.value,
-            timeout=timeout.microseconds if timeout else None,
+            timeout=timeout.total_seconds() * 10e6 if timeout else None,
             exclude_newer_timestamp_ms=int(exclude_newer.replace(tzinfo=datetime.timezone.utc).timestamp() * 1000)
             if exclude_newer
             else None,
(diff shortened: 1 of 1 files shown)


beckermr:
@wolfv This one is ready for review. There is still a bug somewhere else, but we can merge this fix right away.

wolfv:
Hmm, strange - I just checked the code, and we do in fact convert in Rust to a

```rust
timeout: timeout.map(std::time::Duration::from_micros),
```

So not sure what would have been wrong with the code before.

beckermr:
IDK what that code snippet means. 

However, this is a pure-python fix. The time timedelta object holds different amounts of time like this

```
In [23]: td = datetime.timedelta(seconds=100)

In [24]: td.microseconds
Out[24]: 0

In [25]: td.total_seconds()
Out[25]: 100.0

In [26]: td = datetime.timedelta(seconds=100, microseconds=10)

In [27]: td.microseconds
Out[27]: 10

In [28]: td.total_seconds()
Out[28]: 100.00001
```

So the `.microseconds` attribute only holds the microseconds part of the timedelta. It does not have the total timedelta in microseconds. For the total duration, you need to call `.total_seconds()`.

wolfv:
Yep, I also double checked. Your conversion has one `0` too many though. The Python docs suggest to use teh following:

```py
duration / datetime.timedelta(microseconds=1)
```

Do you want to push the fix or should I?

beckermr:
🤦‍♂️ 

will fix.

wolfv:
I'll merge it right away! Thanks for spotting.

beckermr:
Thank you for the review!

beckermr:
FYI the tests are not running here. I didn't know if you noticed or not, so please ignore if this is noise. :)

wolfv:
Yup, thanks for pointing that out :) merged! Hopefully we can get a fix tomorrow for the constraints and then we'll make a new release.

PR #732, opened 2024-06-07
fix: issue 717

### LINKED ISSUE #717: `InvalidBracket` for matchspec
Parsing of invalid matchspec `ray[default,data] >=2.9.0,<3.0.0` fails with 

```
called `Result::unwrap()` on an `Err` value: InvalidBracket
```

Even though the matchspec is invalid, we should not fail with a panic!

See https://github.com/prefix-dev/pixi/issues/1469

baszalmstra:
Fixes #717 
Fixes https://github.com/prefix-dev/pixi/issues/1469

This fixes parsing of `ray[default,data] >=2.9.0,<3.0.0` to fail with `InvalidPackageName` instead of `InvalidBracket`.

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/parse.rs
+++ b/crates/rattler_conda_types/src/match_spec/parse.rs
@@ -20,10 +20,9 @@ use super::{
     matcher::{StringMatcher, StringMatcherParseError},
     MatchSpec,
 };
-use crate::utils::url::parse_scheme;
 use crate::{
     build_spec::{BuildNumberSpec, ParseBuildNumberSpecError},
-    utils::path::is_path,
+    utils::{path::is_path, url::parse_scheme},
     version_spec::{
         is_start_of_version_constraint,
         version_tree::{recognize_constraint, recognize_version},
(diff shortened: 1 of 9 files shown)


baszalmstra:
I also modified the solver to return that the dependencies are unavailable if it contains invalid matchspecs. If no other solution is found this will result in an error like:

```
└─ issue_717 * cannot be installed because there are no viable options:
   └─ issue_717 2.1 is excluded because the constrains 'ray[default,data] >=2.9.0,<3.0.0' failed to parse: 'ray[default,data]' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or .
```

wolfv:
the error message is beautiful, for sure, but I am wondering if we shouldn't just ignore that particular constraint?

baszalmstra:
For constraints it doesnt really matter. Conda simply swallows this matchspec and then constrains any package called "ray[default,data]" which will never exist so it will not constrain anything. In that case I would be fine to just ignore it. However for requirements I think we should keep this way because most likely the spec is there for a reason.

But in either way I prefer to be explicit.

PR #745, opened 2024-06-11
fix: allow empty json repodata

wolfv:
I didn't benchmark this - but I don't think it should have a huge impact. 

LMK what you think, Mr. Bas!

### DIFF
--- a/crates/rattler_conda_types/src/repo_data/mod.rs
+++ b/crates/rattler_conda_types/src/repo_data/mod.rs
@@ -37,7 +37,7 @@ pub struct RepoData {
     pub info: Option<ChannelInfo>,
 
     /// The tar.bz2 packages contained in the repodata.json file
-    #[serde(serialize_with = "sort_map_alphabetically")]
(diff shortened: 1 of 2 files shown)


crates/rattler_shell/src/activation.rs
```diff
@@ -261,7 +261,7 @@ fn collect_env_vars(prefix: &Path) -> Result<IndexMap<String, String>, Activatio
 /// # Returns
 ///
 /// A vector of path entries
-fn prefix_path_entries(prefix: &Path, platform: &Platform) -> Vec<PathBuf> {
+pub fn prefix_path_entries(prefix: &Path, platform: &Platform) -> Vec<PathBuf> {
```
baszalmstra: This change seems unrelated

```suggestion
fn prefix_path_entries(prefix: &Path, platform: &Platform) -> Vec<PathBuf> {
```

wolfv: yep, it's gone!

PR #750, opened 2024-06-14
fix: Support version components with leading zeros

schmelczer:
Some conda packages have versions with components which look like numbers but have leading zeros. For instance:
- [font-ttf-inconsolata](https://prefix.dev/channels/conda-forge/packages/font-ttf-inconsolata) has version `3.000`
- [libre2-11](https://prefix.dev/channels/conda-forge/packages/libre2-11) has version `2023.09.01`

When parsing the above versions with `Version`, we would get `3.0` and `2023.9.1` respectively. This is incorrect as we should aim not to lose any information in the process of parsing a version string.

We also can't just treat version components with leading zeros strings because lexicographic comparison would be unintuitive for them. I believe that we should treat these components in a special way: as numbers with leading zeros. Hence, I introduce a new component type, `NumeralWithLeadingZeros`, which sorts just like `Numeral` but can be used to recover the original version string from it.

I also fixed an unrelated typo `seperators` -> `separators`.

### DIFF
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -763,7 +763,10 @@ impl Component {
 
     /// Checks whether a component is [`Component::Numeral`]
     pub fn is_numeric(&self) -> bool {
-        matches!(self, Component::Numeral(_))
+        matches!(
+            self,
+            Component::Numeral(_) | Component::NumeralWithLeadingZeros(_)
+        )
     }
 
     /// Checks whether the component is a zero.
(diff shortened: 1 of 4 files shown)


baszalmstra:
Why not use VersionWithSource which was designed specifically for this purpose?

baszalmstra:
Since this also increases the size of each component by 8 bytes I would much rather prefer using VersionWithSource over this implementation since then we only pay for these versions when it is important to disambiguate between them.

schmelczer:
I totally missed that, thank you very much for the pointer @baszalmstra!

PR #752, opened 2024-06-17
feat: return pybytes for sha256 and md5 everywhere and use md5 hash for legacy bz2 md5

wolfv:
cc @jaimergp


wolfv:
@baszalmstra do you know wether the raw option vs. the cow<string> makes a big difference in terms of memory? I think in 99% of cases the `legacy_bz2_md5` is going to be empty.

baszalmstra (approved):
In case of stack allocated types it dont think it really matters like in this case.

PR #756, opened 2024-06-20
fix: Fix GatewayQuery.query to filter records based on provided specs

### LINKED ISSUE #755: Gateway.query in Python returns all versions
Hello, I was trying out the `Gateway.query` method and I noticed that it seems to return all versions of a package even if I use a spec like `_go_select[md5=73a9dbf6acbed92b5e45e336f0372bcf]`.

Example:

```python
>>> import rattler
>>> gateway = rattler.Gateway()
>>> import asyncio
>>> asyncio.run(gateway.query(["https://repo.anaconda.com/pkgs/main"], ["linux-64", "noarch"], ["_go_select[md5=73a9dbf6acbed92b5e45e336f0372bcf]"]))

[[RepoDataRecord(url="https://repo.anaconda.com/pkgs/main/linux-64/_go_select-2.1.0-precompiled.tar.bz2"), RepoDataRecord(url="https://repo.anaconda.com/pkgs/main/linux-64/_go_select-2.2.0-nocgo.tar.bz2"), RepoDataRecord(url="https://repo.anaconda.com/pkgs/main/linux-64/_go_select-2.3.0-cgo.tar.bz2"), RepoDataRecord(url="https://repo.anaconda.com/pkgs/main/linux-64/_go_select-2.1.0-precompiled.conda"), RepoDataRecord(url="https://repo.anaconda.com/pkgs/main/linux-64/_go_select-2.2.0-nocgo.conda"), RepoDataRecord(url="https://repo.anaconda.com/pkgs/main/linux-64/_go_select-2.3.0-cgo.conda")], []]
```
Shouldn't it return a single record in this case or at best 2 (the .tar.bz2 and the .conda)?

I also tried with a matchspec like `_go_select=2.3.0=cgo` and I get the same result.

JeanChristopheMorinPerso:
Fixes #755.

The `query` method of `GatewayQuery` was not filtering records based on the provided specs. It would only filter records to be fetched when doing a recursive query.

This PR fixes that.

Note that I'm learning rust and this might not be the right way to fix this. Please let me know if I can improve this! I'd like to learn.

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/query.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/query.rs
@@ -250,7 +250,7 @@ impl GatewayQuery {
 
                         for record in records.iter() {
                             if !request_specs.iter().any(|spec| spec.matches(record)) {
-                                // Do not recurse into records that do not match to root spec.
+                                // Do not return records that do not match to root spec.
                                 continue;
                             }
                             result.len += 1;
--- a/crates/rattler_repodata_gateway/src/gateway/query.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/query.rs
@@ -249,7 +249,7 @@ impl GatewayQuery {
                         let result = &mut result[result_idx];
 
                         for record in records.iter() {
-                            if !request_specs.iter().any(|spec| spec.matches(record)) {
+                            if !self.recursive && !request_specs.iter().any(|spec| spec.matches(record)) {
                                 // Do not return records that do not match to root spec.
                                 continue;
                             }
--- a/crates/rattler_repodata_gateway/src/gateway/query.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/query.rs
@@ -247,8 +247,15 @@ impl GatewayQuery {
                     // Add the records to the result
                     if records.len() > 0 {
                         let result = &mut result[result_idx];
-                        result.len += records.len();
-                        result.shards.push(records);
+
+                        for record in records.iter() {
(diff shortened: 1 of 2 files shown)


JeanChristopheMorinPerso:
It looks like I need to include records passed by direct URL.

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -247,8 +247,15 @@ impl GatewayQuery {
                     // Add the records to the result
                     if records.len() > 0 {
                         let result = &mut result[result_idx];
-                        result.len += records.len();
-                        result.shards.push(records);
+
+                        for record in records.iter() {
+                            if !self.recursive && !request_specs.iter().any(|spec| spec.matches(record)) {
```
JeanChristopheMorinPerso, line 252: `!self.recursive` is a hack... I don't know how we should apply filters when doing recursive queries. Does anyone have an idea of what we should do?

JeanChristopheMorinPerso: I guess what I would need to do is to only filter "top level" packages when in recursive mode and ignore specs that are direct URL specs (since direct URLs are already correctly filtered?

baszalmstra: Yes! But that will already happen correctly because the specs added for the dependencies match all records.

Regardless I think this is better because now we also don't pay for the matches check if the call is not recursive anyway.

baszalmstra (approved):

PR #757, opened 2024-06-24
feat: improve error message when parsing file name

wolfv:
We could also be more lenient and assume a build number / build string of `0` if we only find a single `-`...

Haven't really seen that in the wild though.


baszalmstra:
Maybe also include the string?

PR #760, opened 2024-07-02
fix: skip over implicit `0` components when copying

wolfv:
We copied the implicit components twice:

- once as flag on the segment
- and once as a materialized component from the SegmentIter

This should fix the double-zeros :)


crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -336,6 +336,8 @@ impl Version {
     /// will return the version: `3a.4`.
     pub fn with_segments(&self, segments: impl RangeBounds<usize>) -> Option<Version> {
         // Determine the actual bounds to use
+        // println!("{:?}", self.segments);
```
baszalmstra, line 339: Oopsie! 😅

Remove?

baszalmstra (approved):

PR #761, opened 2024-07-02
fix: lock file stability issues with PyPI types

Hofer-Julian:
Fixes https://github.com/prefix-dev/pixi/issues/1535


Hofer-Julian:
While browsing the code, I found this function.
If I understand it correctly, conda packages should always come before PyPI ones.
Looking at the produced lockfile this doesn't seem to be the case. Maybe more is amiss here.

https://github.com/mamba-org/rattler/blob/76867ed9b45fee40687164f92d5edb29aad41119/crates/rattler_lock/src/parse/serialize.rs#L103-L136

baszalmstra:
> While browsing the code, I found this function. If I understand it correctly, conda packages should always come before PyPI ones. Looking at the produced lockfile this doesn't seem to be the case. Maybe more is amiss here.
> 
> https://github.com/mamba-org/rattler/blob/76867ed9b45fee40687164f92d5edb29aad41119/crates/rattler_lock/src/parse/serialize.rs#L103-L136

That part is to sort the references to the packages found in the environments section. Not the package metadata at the end of the file.

Hofer-Julian:
> > While browsing the code, I found this function. If I understand it correctly, conda packages should always come before PyPI ones. Looking at the produced lockfile this doesn't seem to be the case. Maybe more is amiss here.
> > https://github.com/mamba-org/rattler/blob/76867ed9b45fee40687164f92d5edb29aad41119/crates/rattler_lock/src/parse/serialize.rs#L103-L136
> 
> That part is to sort the references to the packages found in the environments section. Not the package metadata at the end of the file.

I see, thanks for the clarification @baszalmstra!

crates/rattler_lock/src/pypi.rs
```diff
@@ -59,6 +59,7 @@ impl Ord for PypiPackageData {
         self.name
             .cmp(&other.name)
             .then_with(|| self.version.cmp(&other.version))
```
wolfv, line 61: can you also add the hash as a last resort?
Or maybe the platform or something like that, if we have that?

baszalmstra (approved):

baszalmstra:
Nice @Hofer-Julian !

crates/rattler_lock/src/pypi.rs
```diff
@@ -60,14 +60,9 @@ impl Ord for PypiPackageData {
...
-            .then_with(|| {
-                match (
-                    serde_json::to_string(&self.hash),
-                    serde_json::to_string(&other.hash),
-                ) {
-                    (Ok(self_hash), Ok(other_hash)) => self_hash.cmp(&other_hash),
-                    _ => Ordering::Equal,
-                }
+            .then_with(|| match (&self.hash, &other.hash) {
+                (Some(self_hash), Some(other_hash)) => self_hash.cmp(other_hash),
```
wolfv, line 64: Is this necessary? Does Option not have a cmp (with None)?

Because now you are missing `Some, None`, and `None, Some` :)

wolfv (changes requested):

wolfv (approved):

wolfv:
Nice fix!

PR #762, opened 2024-07-03
fix: run clippy on all targets

Hofer-Julian:
I've noticed that vscode showed clippy warnings.
Turned out we don't run clippy on all targets


wolfv (approved):

wolfv:
fantastic!

PR #770, opened 2024-07-08
fix: allow version following package in strict mode

baszalmstra:
I thought it was quite annoying that you cant write `python>=3.9` in strict mode. I changed the strict parser to allow package names to be followed by a version operator without a space.  `python=3.9=*` is still invalid.

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/snapshots/rattler_conda_types__match_spec__parse__tests__test_from_string_Strict.snap
+++ b/crates/rattler_conda_types/src/match_spec/snapshots/rattler_conda_types__match_spec__parse__tests__test_from_string_Strict.snap
@@ -1,6 +1,6 @@
 ---
 source: crates/rattler_conda_types/src/match_spec/parse.rs
-assertion_line: 917
+assertion_line: 915
 expression: evaluated
 ---
 /home/user/conda-bld/linux-64/foo-1.0-py27_0.tar.bz2:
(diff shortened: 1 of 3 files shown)


crates/rattler_conda_types/src/match_spec/snapshots/rattler_conda_types__match_spec__parse__tests__test_from_string_Strict.snap
```diff
@@ -28,9 +29,9 @@ blas *.* mkl:
     base_url: "https://conda.anaconda.org/conda-forge/"
     name: conda-forge
 foo=1.0=py27_0:
-  error: "'foo=1.0=py27_0' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or ."
+  error: "The build string '=py27_0' is not valid it can only contain alphanumeric characters and underscores"
```
Hofer-Julian, line 32:
```suggestion
  error: "The build string '=py27_0' is not valid, it can only contain alphanumeric characters and underscores"
```

crates/rattler_conda_types/src/match_spec/snapshots/rattler_conda_types__match_spec__parse__tests__test_from_string_Strict.snap
```diff
@@ -28,9 +29,9 @@ blas *.* mkl:
...
 foo==1.0=py27_0:
-  error: "'foo==1.0=py27_0' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or ."
+  error: "The build string '=py27_0' is not valid it can only contain alphanumeric characters and underscores"
```
Hofer-Julian, line 34:
```suggestion
  error: "The build string '=py27_0' is not valid, it can only contain alphanumeric characters and underscores"
```

Hofer-Julian (commented):
Maybe a small test ensuring that `python=3.8` contains to be valid in the future would be nice :)

Hofer-Julian (approved):

PR #772, opened 2024-07-08
chore: bump zip to 2.1.3

jpcorreia99:
Sometimes fetching packages using rattler fails with ` ZipError: "The file length is not available in the local header"`.

This error comes from the `zip` library here:

https://github.com/zip-rs/zip2/blame/b6e0a0693ba2d07f541cd9017b60df7e65817e48/src/types.rs#L688-L696 

As can be seen in a comment 

```
/* FIXME: these were previously incorrect: add testing! */
        /* flags & (1 << 3) != 0 */
```

Bumping the version of zip has solved our issues and we're able to install the desired packages.

### DIFF
--- a/crates/rattler_package_streaming/src/write.rs
+++ b/crates/rattler_package_streaming/src/write.rs
@@ -308,7 +308,7 @@ pub fn write_conda_package<W: Write + Seek>(
             .expect("1-1-2023 00:00:00 should convert into datetime")
     };
 
-    let options: zip::write::FileOptions<'_, ()> = zip::write::FileOptions::default()
+    let options = zip::write::SimpleFileOptions::default()
         .compression_method(zip::CompressionMethod::Stored)
         .last_modified_time(last_modified_time);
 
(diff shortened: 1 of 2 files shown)


crates/rattler_package_streaming/src/write.rs
```diff
@@ -308,7 +308,7 @@ pub fn write_conda_package<W: Write + Seek>(
             .expect("1-1-2023 00:00:00 should convert into datetime")
     };
 
-    let options = zip::write::FileOptions::default()
+    let options: zip::write::FileOptions<'_, ()> = zip::write::FileOptions::default()
```
jpcorreia99, line 311: @baszalmstra I'm quite bad with rust ahaha, compilation failed asking for a type here. This worked but please let me know if there's a better approach

jpcorreia99:
@baszalmstra / @wolfv hey there, I'd appreciate a review here whenever you have the time for it!

wolfv:
nice find. Thanks so much!

wolfv:
@nichmor does this include your fix as well?

crates/rattler_package_streaming/src/write.rs
baszalmstra: I think instead of specifying the type you can use:

```suggestion
    let options = zip::write::SimpleFileOptions::default()
```

baszalmstra (approved):
Nice thanks!

nichmor:
> @nichmor does this include your fix as well?

no

PR #776, opened 2024-07-11
fix: unclobber issue when packages are named differently

wolfv:
cc @baszalmstra - i think that was the test case I wrote for a issues similar to teh one wtih pyarrow now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ tag-prefix = ""
 lto = true
 
 [workspace.dependencies]
-anyhow = "1.0.79"
+anyhow = "1.0.80"
 assert_matches = "1.5.0"
 async-compression = { version = "0.4.6", features = [
     "gzip",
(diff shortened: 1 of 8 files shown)


baszalmstra:
@wolfv I cant tag you as a reviewer, but a review would be appreciated. :)

Hofer-Julian:
I can confirm that it fixes https://github.com/prefix-dev/pixi/issues/1594

crates/rattler/src/install/driver.rs
```diff
@@ -1,25 +1,32 @@
...
-/// process doesn't try to acquire more resources than the system has available.
+use super::{
+    clobber_registry::{ClobberError, ClobberRegistry, ClobberedPath},
+    link_script::{PrePostLinkError, PrePostLinkResult},
+    unlink::{recursively_remove_empty_directories, UnlinkError},
+    Transaction,
+};
+use crate::install::link_script::LinkScriptError;
+
+/// Packages can mostly be installed in isolation and therefor in parallel.
```
Hofer-Julian, line 23:
```suggestion
/// Packages can mostly be installed in isolation and therefore in parallel.
```

Hofer-Julian (approved):
Only found one typo that `typos` missed :)

Apart from that it looks good to me 👍

PR #777, opened 2024-07-11
perf: faster prefix record deser

nichmor:
based on this blog https://blog.datalust.co/deserializing-json-really-fast/ I've tried to choose simd_json for our deserialization to PrefixRecord as it seems that it has a lot of samples in the flamegraph
image

In the benchmarks, having 42 JSON files in the conda_meta folder and using simd_json will give us the following results:

```
process_json_files      time:   [9.2815 ms 9.3741 ms 9.4924 ms]
Found 7 outliers among 100 measurements (7.00%)
5 (5.00%) high mild
2 (2.00%) high severe
```

using old way:
```
process_json_files      time:   [9.7586 ms 9.8881 ms 10.041 ms]
change: [+3.5487% +5.4835% +7.3793%] (p = 0.00 < 0.05)
Performance has regressed.
```

We can see an increase of `+5.4835%`


Cargo.toml
```diff
@@ -127,7 +127,7 @@ smallvec = { version = "1.13.2", features = [
     "const_generics",
     "union",
 ] }
-strum = { version = "0.26.2", features = ["derive"] }
+strum = { version = "0.26.3", features = ["derive"] }
```
nichmor, line 130: contains the fix for this:
https://github.com/Peternator7/strum/issues/358

crates/rattler_conda_types/Cargo.toml
```diff
@@ -23,6 +23,7 @@ purl = { workspace = true, features = ["serde"] }
 rattler_digest = { path = "../rattler_digest", version = "0.19.4", default-features = false, features = ["serde"] }
 rattler_macros = { path = "../rattler_macros", version = "0.19.4", default-features = false }
 regex = { workspace = true }
+simd-json = { version = "*", features = ["serde_impl"]}
```
baszalmstra, line 26: Can you add a minimal version here?

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -12,6 +12,8 @@ use std::io::{BufWriter, Read};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 
+use simd_json;
```
baszalmstra, line 15: I think you can just remove this right?

Hofer-Julian:
I pushed a commit that removes the unsafe usage: https://github.com/mamba-org/rattler/pull/777/commits/11c86003dc3834fe5570734d115ff2d70e4394a1

crates/rattler_conda_types/Cargo.toml
nichmor: made this `workspace=true` and added `simd-json` in root `Cargo.toml`

PR #781, opened 2024-07-17
feat: Add to_path() method to ExplicitEnvironmentSpec

synapticarbors:
Based on feedback from @baszalmstra on discord, I added a `to_path` method to the `ExplicitEnvironmentSpec` struct so that it could be dumped to file. This would be a first step in a `pixi export` PR to be submitted later. 

This is my first rust-based PR to a project (although I've fooled around with it a bit on my own), so feedback is very welcome. For testing, `ExplicitEnvironmentSpec` and `ExplicitEnvironmentEntry` didn't implement `PartialEq`, so I just unrolled the content. This isn't ideal, but I didn't want to start poking around on the main struct definitions, but I can if that's preferred.

### DIFF
--- a/crates/rattler_conda_types/src/explicit_environment_spec.rs
+++ b/crates/rattler_conda_types/src/explicit_environment_spec.rs
@@ -153,7 +153,7 @@ impl ExplicitEnvironmentSpec {
 
     /// Converts an [`ExplicitEnvironmentSpec`] to a string representing a valid explicit
     /// environment file
-    pub fn to_string(&self) -> String {
+    pub fn to_spec_string(&self) -> String {
         let mut s = String::new();
 
         if let Some(plat) = &self.platform {
@@ -171,7 +171,7 @@ impl ExplicitEnvironmentSpec {
 
     /// Writes an explicit environment spec to file
     pub fn to_path(&self, path: impl AsRef<Path>) -> Result<(), std::io::Error> {
-        let s = self.to_string();
+        let s = self.to_spec_string();
 
         fs::write(path, s)?;
 
@@ -276,9 +276,9 @@ mod test {
     #[case::ros_noetic_linux_64("explicit-envs/ros-noetic_linux-64.txt")]
     #[case::vs2015_runtime_win_64("explicit-envs/vs2015_runtime_win-64.txt")]
     #[case::xtensor_linux_64("explicit-envs/xtensor_linux-64.txt")]
-    fn test_to_string(#[case] path: &str) {
+    fn test_to_spec_string(#[case] path: &str) {
         let env = ExplicitEnvironmentSpec::from_path(&get_test_data_dir().join(path)).unwrap();
-        let env_cmp = ExplicitEnvironmentSpec::from_str(&env.to_string()).unwrap();
+        let env_cmp = ExplicitEnvironmentSpec::from_str(&env.to_spec_string()).unwrap();
 
         assert_eq!(env.platform, env_cmp.platform);
         assert_eq!(
--- a/crates/rattler_conda_types/src/explicit_environment_spec.rs
+++ b/crates/rattler_conda_types/src/explicit_environment_spec.rs
@@ -151,23 +151,29 @@ impl ExplicitEnvironmentSpec {
         Self::from_reader(File::open(path)?)
     }
 
-    /// Writes an explicit environment spec to file
-    pub fn to_path(&self, path: impl AsRef<Path>) -> Result<(), std::io::Error> {
-        let plat = match self.platform {
-            Some(p) => p.as_str(),
-            None => "",
-        };
-        let mut out = String::new();
-        out.push_str("# This file may be used to create an environment using:\n");
-        out.push_str("# $ conda create --name <env> --file <this file>\n");
-        out.push_str(&format!("# platform: {plat}\n"));
-        out.push_str("@EXPLICIT\n");
+    /// Converts an [`ExplicitEnvironmentSpec`] to a string representing a valid explicit
+    /// environment file
+    pub fn to_string(&self) -> String {
+        let mut s = String::new();
 
-        for p in self.packages.iter() {
-            out.push_str(&format!("{}\n", p.url.as_str()));
+        if let Some(plat) = &self.platform {
+            s.push_str(&format!("# platform: {plat}\n"));
         }
 
-        fs::write(path, out)?;
+        s.push_str("@EXPLICIT\n");
+
+        for p in &self.packages {
+            s.push_str(&format!("{}\n", p.url.as_str()));
+        }
+
+        s
+    }
+
+    /// Writes an explicit environment spec to file
+    pub fn to_path(&self, path: impl AsRef<Path>) -> Result<(), std::io::Error> {
+        let s = self.to_string();
+
+        fs::write(path, s)?;
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


crates/rattler_conda_types/src/explicit_environment_spec.rs
```diff
@@ -245,6 +266,37 @@ mod test {
...
+                .map(|entry| entry.url.clone())
+                .collect::<Vec<_>>(),
+            round_trip_env
+                .packages
+                .iter()
+                .map(|entry| entry.url.clone())
+                .collect::<Vec<_>>()
+        );
+
+        tmp_dir.close().unwrap()
```
baszalmstra, line 297: No need, this happens on drop.

```suggestion
```

crates/rattler_conda_types/src/explicit_environment_spec.rs
```diff
@@ -245,6 +266,37 @@ mod test {
...
+    #[case::ros_noetic_linux_64("explicit-envs/ros-noetic_linux-64.txt")]
+    #[case::vs2015_runtime_win_64("explicit-envs/vs2015_runtime_win-64.txt")]
+    #[case::xtensor_linux_64("explicit-envs/xtensor_linux-64.txt")]
+    fn test_to_path(#[case] path: &str) {
+        let env = ExplicitEnvironmentSpec::from_path(&get_test_data_dir().join(path)).unwrap();
+
+        let tmp_dir = tempfile::tempdir().unwrap();
+        let file_path = tmp_dir.path().join("temp_explicit_env.txt");
+
+        assert_matches!(env.to_path(file_path.clone()), Ok(()));
```
baszalmstra, line 279:
```suggestion
        env.to_path(file_path.clone()).expect("failed to write to path!")
```

crates/rattler_conda_types/src/explicit_environment_spec.rs
```diff
@@ -150,6 +150,27 @@ impl ExplicitEnvironmentSpec {
     pub fn from_path(path: &Path) -> Result<Self, ParseExplicitEnvironmentSpecError> {
         Self::from_reader(File::open(path)?)
     }
+
+    /// Writes an explicit environment spec to file
+    pub fn to_path(&self, path: impl AsRef<Path>) -> Result<(), std::io::Error> {
```
baszalmstra, line 155: It would be nice to have a method that converts this instance to a string. E.g. `to_string()` which is then used by this function. That way people can also write their own serialization logic if they want to.

baszalmstra (changes requested):

baszalmstra:
Might also be interesting to implement this (still draft) spec completely: https://github.com/conda/ceps/pull/79

synapticarbors:
@baszalmstra -- Thank you for the quick review. I'll make the code changes later today. I did have a few questions:

> Might also be interesting to implement this (still draft) spec completely: [conda/ceps#79](https://github.com/conda/ceps/pull/79)

I read through the CEP and was trying to figure out what was needed to implement this completely. The only ideas that I had were arguments to the proposed `to_string` method like `to_string(&self, include_platform: bool, comments: Option<String>)`. Those some arguments would be passed to the `to_path` method and then would be passed down to `to_string`. Let me know if that sounds reasonable or if you prefer a different way or signature. 

The other requirements in the CEP were:
- Topologically sorted url list: This seems outside of the scope of this struct as it has no information or knowledge about package inter-dependencies
- Expanding paths and no environment variables: I wasn't sure about this as the `Url` type in rust seems to be "absolute" and will error on relative file paths when using `from_path()` for example. 

Please let me know what in the CEP is missing in your mind and I'll try to implement it.

crates/rattler_conda_types/src/explicit_environment_spec.rs
synapticarbors: Just to clarify here, do you mean an explicit `to_string` method, or:

```rust
impl fmt::Display for ExplicitEnvironmentSpec {
...
}
```

so that rust automatically implements `ToString`?

baszalmstra: No I meant a `to_string` method. I dont think it makes sense to implement `Display` because that is expected to be more human-readable.

baszalmstra:
I hadnt really checked what needed to happen to implement the CEP. 😅 I think most of it is related to how we parse the file not so much how we write it. In that sense I think your implementation already contains everything required.

I dont think it makes sense to add a `comments` argument to `to_string`. Users can just call the `to_string` method and prepend whatever if they really want to. The platform is also not really needed because you could just as well set it to `None` in the struct itself.

synapticarbors:
@baszalmstra -- So it sounds like what you think is the best approach is:
1. Add `to_string()`. This should take the `ExplicitEnvironmentSpec` and return `String` containing:
  ```
"# platform: linux-64\n@EXPLICIT\nURL_1\nURL_2\n...."
  ```
if `Some(plat)`, else:
  ```
"@EXPLICIT\nURL_1\nURL_2\n...."
  ```

2. Modify my current `to_path` to use `to_string()`. I guess here maybe we'd still want `comments: Option<String>` to allow custom comments to be written to the file? Or just basically write the output of `to_string()` to file and leave it to the user to write their own custom function if they want any prepended comments.

baszalmstra:
Indeed. And I would indeed forgo the `comments` for now. If people want that they can write their custom function.

synapticarbors:
@baszalmstra -- I made the requested changes. I just wanted to note that `cargo clippy` complains about the `to_string` method:

```
❯ cargo clippy
    Checking rattler_conda_types v0.26.1 (.../rattler/crates/rattler_conda_types)
warning: implementation of inherent method `to_string(&self) -> String` for type `explicit_environment_spec::ExplicitEnvironmentSpec`
   --> crates/rattler_conda_types/src/explicit_environment_spec.rs:156:5
    |
156 | /     pub fn to_string(&self) -> String {
157 | |         let mut s = String::new();
158 | |
159 | |         if let Some(plat) = &self.platform {
...   |
169 | |         s
170 | |     }
    | |_____^
    |
    = help: implement trait `Display` for type `explicit_environment_spec::ExplicitEnvironmentSpec` instead
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#inherent_to_string
```

So maybe an alternative name? Ignore the warning? Or implement `Display`?. Also, I'm wondering if the `to_path()` is worth adding since it does so little.

baszalmstra:
If you can think of a better name be my guest. I would also be fine with suppressing this particular lint.

Even though to_path does little its a nice little utlity function. 👍

synapticarbors:
The best alternatives I could think of at this point are `to_spec_string()` and `render_spec_string()`. You're right that implementing `Display` doesn't seem correct since what this is outputting is a rendering of the struct to the spec, rather than a general representation of the struct. If you like this alternative, I could also change `to_path()` -> `to_spec_path()`/`render_spec_path()`.

baszalmstra:
I think to_path is still fine and to_spec_string also sounds great!

synapticarbors:
@baszalmstra I've renamed the method so everything should be in order now.

baszalmstra (approved):

crates/rattler_conda_types/src/explicit_environment_spec.rs
synapticarbors: I just realized I missed this one suggestion in the tests. Do you want me to open another PR?

baszalmstra: Its fine, dont worry about it! 👍

Thanks again for contributing!

PR #784, opened 2024-07-19
feat: expose `HasPrefixEntry` for public use

kelvinou01:
I was using rattler's `HasPrefix` to read info/has_prefix, and I couldn't unpack the `HasPrefixEntry` structs contained within, because it isn't `pub`:

```rust
let has_prefix = HasPrefix.from_str(...)?;
has_prefix.files.into_iter()
    .map(|HasPrefixEntry { relative_path, file_mode, prefix }| {
        // do stuff with relative_path, file_mode, and prefix
    }
    .collect()
```

I went thru src/package/mod.rs, and I don't think making `HasPrefixEntry` public is bad semantics. `HasPrefix` is `pub` for downstream consumption, so I think `HasPrefixEntry` should be as well? Correct me if I'm wrong, this is my first contrib to rattler :)

### DIFF
--- a/crates/rattler_conda_types/src/package/has_prefix.rs
+++ b/crates/rattler_conda_types/src/package/has_prefix.rs
@@ -15,10 +15,14 @@ use std::{
     sync::OnceLock,
 };
 
+/// Representation of an entry in `info/has_prefix`.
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):
LGTM! Thanks for contributing!

PR #792, opened 2024-07-25
fix: parse channel in matchspec string

ruben-arts:
This now allows for:
- `"https://software.repos.intel.com/python/conda::python[version=3.9]"`: Having a URL as channel in this format.
- `"conda-forge::foo[subdir=linux-64]`: Having a subdir in the brackets
- `"conda-forge/linux-64::foo`: Having a subdir in the channel string, **only if it is a platform**.

### DIFF
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -198,6 +198,10 @@ impl Channel {
                 }
             }
         } else {
+            // Validate that the channel is a valid name
+            if channel.contains(|c| c == ':' || c == '\\') {
+                return Err(ParseChannelError::InvalidName(channel.to_owned()));
+            }
             Channel {
                 platforms,
                 ..Channel::from_name(channel, config)
@@ -366,6 +370,10 @@ pub enum ParseChannelError {
     /// Error when the path is invalid.
     #[error("invalid path '{0}'")]
     InvalidPath(String),
+
+    /// Error when the channel name is invalid.
+    #[error("invalid channel name: '{0}'")]
+    InvalidName(String),
 
     /// The root directory is not an absolute path
     #[error("root directory from channel config is not an absolute path")]
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -199,7 +199,7 @@ impl Channel {
             }
         } else {
             // Validate that the channel is a valid name
-            if channel.contains(|c| c == ':' || c == '\\') {
+            if channel.contains(&[':', '\\']) {
                 return Err(ParseChannelError::InvalidName(channel.to_owned()));
             }
             Channel {
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -549,6 +549,12 @@ mod tests {
             channel.base_url().to_string(),
             "https://conda.anaconda.org/conda-forge/"
         );
+
+        let channel = Channel::from_str(
+            "https://conda.anaconda.org/conda-forge/label/rust_dev",
+            &config,
+        );
+        assert_eq!(channel.unwrap().name(), "conda-forge/label/rust_dev",);
     }
 
     #[test]
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -199,7 +199,7 @@ impl Channel {
             }
         } else {
             // Validate that the channel is a valid name
-            if channel.contains(&[':', '\\']) {
+            if channel.contains([':', '\\']) {
                 return Err(ParseChannelError::InvalidName(channel.to_owned()));
             }
             Channel {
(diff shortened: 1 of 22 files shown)


baszalmstra:
mm tricky, a path is also not technically a url..

ruben-arts:
I'm miss using this PR to fix #793

@baszalmstra I removed the whole URL from path part. I'll make a new PR where I add the `path` field.

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -527,13 +528,9 @@ fn matchspec_parser(
         let channel_config = ChannelConfig::default_with_root_dir(
             std::env::current_dir().expect("Could not get current directory"),
         );
-        if let Some((channel, subdir)) = channel_str.rsplit_once('/') {
```
ruben-arts, line 530: I'm not sure this was correct, as a channel can also be "conda-forge/label/xyz"

This is now why the test fails.

ruben-arts: After discussing with @baszalmstra we're going to reimplement this with a list of "known_subdirs" which is just the `platforms`.

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -198,6 +198,10 @@ impl Channel {
                 }
             }
         } else {
+            // Validate that the channel is a valid name
+            if channel.contains(|c| c == ':' || c == '\\') {
```
baszalmstra, line 202:
```suggestion
            if channel.contains(&[':', '\\']) {
```

baszalmstra: Maybe we can include a list of valid characters? `A-Za-z0-9-_`?

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -89,7 +89,7 @@ use matcher::StringMatcher;
 /// assert_eq!(spec.version, Some(VersionSpec::from_str("1.0.*", Strict).unwrap()));
 /// assert_eq!(spec.channel, Some(Channel::from_str("conda-forge", &channel_config).map(|channel| Arc::new(channel)).unwrap()));
 ///
-/// let spec = MatchSpec::from_str("conda-forge/linux-64::foo >=1.0", Strict).unwrap();
+/// let spec = MatchSpec::from_str(r#"conda-forge::foo >=1.0[subdir="linux-64"]"#, Strict).unwrap();
```
baszalmstra, line 92: Can this be changed back with the new logic? Or is this still invalid?

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -473,49 +495,49 @@ fn matchspec_parser(
...
+    let mut input_split = input.rsplitn(3, ':').fuse();
+    let (input, namespace, channel_str) =
+        match (input_split.next(), input_split.next(), input_split.next()) {
+            (Some(input), None, _) => (input, None, None),
+            (Some(input), Some(namespace), None) => (input, Some(namespace), None),
+            (Some(input), Some(namespace), Some(channel_str)) => {
+                (input, Some(namespace), Some(channel_str))
+            }
+            (None, _, _) => ("", None, None),
+        };
```
baszalmstra, line 540:
```suggestion
    let input = input_split.next().unwrap_or("");
    let namespace = input_split.next();
    let channel_str = input_split.next();
```

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -952,9 +968,10 @@ mod tests {
             "python ==2.7.*.*|>=3.6",
             "python=3.9",
             "python=*",
+            "https://software.repos.intel.com/python/conda::python[version=3.9]",
```
baszalmstra, line 971: Could you maybe add a few more edge cases here?

baszalmstra: E.g.:

```suggestion
            "https://software.repos.intel.com/python/conda::python[version=3.9]",
            "https://software.repos.intel.com/python/conda/linux-64::python[version=3.9]",
            "https://software.repos.intel.com/python/conda::python[version=3.9, subdir=linux-64]"
            "https://software.repos.intel.com/python/conda/linux-64::python[version=3.9, subdir=linux-64]",
```

baszalmstra: Maybe more which you can think of?

crates/rattler_conda_types/src/match_spec/mod.rs
ruben-arts: Did you one better:
https://github.com/conda-incubator/rattler/commit/36a03b3500cde36aa292160a2b8277ea8d5b9d82

crates/rattler_conda_types/src/channel/mod.rs
ruben-arts: Something like this?: 
```rs
if !channel.chars().all(|c| c.is_alphanumeric() || c == '-' || c == '_') {
    return Err(ParseChannelError::InvalidName(channel.to_owned()));
}
```
Without the Lenient option I would be afraid of breaking some working systems. The previous change was needed to avoid breaking further parsing. 
What do you think?

ruben-arts: Added your suggestion in https://github.com/conda-incubator/rattler/pull/792/commits/7f653af74bbabcfb40da0fa1977dc323e264dab

baszalmstra (approved):

PR #795, opened 2024-07-25
fix: constraints on virtual packages were ignored

baszalmstra:
If there was a package that had a constraint on a virtual package, e.g.:

```json
"constrains": ["__cuda >=12.1"],
```

and the virtual dependency was never explicitly required it was silently ignored. So even though your system would only support `cuda 11.8` the solver could select a version of a package that required a higher version. 

This PR solves this by adding explicit requirements on each specified virtual package. This behavior is also added for libsolv.


nichmor:
What would be the error message if I don't have any cuda in my system_requirements, but the package still have a constrain on it?

baszalmstra:
There would be no error. Because at that point there is no virtual package. It basically acts as a constrains on a non-existing package. We already have a test for that. :)

tdejager (approved):

PR #797, opened 2024-07-26
fix: Fallback to fully reading the package stream when downloading before attempting decompression

jpcorreia99:
Fixes https://github.com/conda-incubator/rattler/issues/794.

Now, we detect the error message thrown when attempting to decompress the HTTP stream. When this is the case, we do a second request and read all the data to an in-memory buffer, before attempting decompression.

The chosen approach was the third one discussed in this comment https://github.com/conda-incubator/rattler/issues/794#issuecomment-2250648079 as it is the one that works out of the box + does not have a perf implication on the normal streaming downloads.

Haven't implemented tests yet, as I'd first like to discuss the approach in itself. But I've confirmed this fixes the issue I was facing!


crates/rattler_package_streaming/src/read.rs
```diff
@@ -76,13 +81,73 @@ pub fn extract_conda(reader: impl Read, destination: &Path) -> Result<ExtractRes
...
+    let mut buffer = Vec::new();
+    let sha256_reader = rattler_digest::HashingReader::<_, rattler_digest::Sha256>::new(reader);
+    let mut md5_reader =
+        rattler_digest::HashingReader::<_, rattler_digest::Md5>::new(sha256_reader);
+    md5_reader.read_to_end(&mut buffer)?;
 
+    // The stream must be seekable when decompressing zip archives using data descriptors
+    let cursor = Cursor::new(buffer);
+
+    let mut archive = ZipArchive::new(cursor)?;
```
jpcorreia99, line 108: Not 100% sure of the tradeoff between having the package all in-memory vs first dumping it to a tempfile. Do you guys have an opinion on this?

crates/rattler_package_streaming/src/reqwest/tokio.rs
```diff
@@ -131,12 +140,44 @@ pub async fn extract_conda(
...
+            }
+            Ok(result)
+        }
+        // https://github.com/conda-incubator/rattler/issues/794
+        Err(ExtractError::ZipError(zip_error))
+            if (zip_error
+                .to_string()
+                .contains(DATA_DESCRIPTOR_ERROR_MESSAGE)) =>
+        {
+            tracing::debug!("Failed to stream decompress conda package from '{}' due to the presence of zip data descriptors. Falling back to non streaming decompression;", url);
```
jpcorreia99, line 169: A bit split between using `debug` or `warn` here. I'm tempted to use WARN since this is a non-advisable state to be in.

jpcorreia99:
@wolfv / @nichmor here's my attempted fix for the issue discussed yesterday! Confirmed it fixed my case. Will try to create a zipfile representing this state so that you may reproduce it as well

crates/rattler_package_streaming/src/reqwest/tokio.rs
```diff
@@ -131,12 +139,44 @@ pub async fn extract_conda(
...
+    .await
+    {
+        Ok(result) => {
+            if let Some(reporter) = &reporter {
+                reporter.on_download_complete();
+            }
+            Ok(result)
+        }
+        // https://github.com/conda-incubator/rattler/issues/794
+        Err(ExtractError::ZipError(zip_error))
```
baszalmstra, line 163: Maybe get to the inner error faster:

```
        Err(ExtractError::ZipError(ZipError::UnsupportedArchive(zip_error)))
```

crates/rattler_package_streaming/src/tokio/async_read.rs
```diff
@@ -33,14 +34,18 @@ pub async fn extract_tar_bz2(
 pub async fn extract_conda(
     reader: impl AsyncRead + Send + 'static,
     destination: &Path,
+    extract_fn: fn(Box<dyn Read>, &Path) -> Result<ExtractResult, ExtractError>,
```
baszalmstra, line 37: I would prefer to keep the API the same as the blocking one, so splitting this up into seperate functions instead of passing in a closure.

baszalmstra (commented):
I think the approach is valid using a fallback method.

Im wondering though if we can detect this without having to resort to catching and interpreting string errors? Can we read the header ourself beforehand?

Reading to memory might be risking OOM issues, perhaps use a [SpooledTempFile](https://docs.rs/tempfile/latest/tempfile/struct.SpooledTempFile.html). 

We definitely need an example of such a ZIP file.

jpcorreia99:
@baszalmstra 

>Im wondering though if we can detect this without having to resort to catching and interpreting string errors? Can we read the header ourself beforehand?

The zipfile contains multiple headers, one per file inside, so I think to be able to read the headers we'd need to basically reimplement the zip crate ourselves, which I don't think is viable. 

Will follow the file suggestion and work on creating a repro!

jpcorreia99:
@baszalmstra managed to repro the issue. You can just download a conda package, decompress it and then recompress it with `zip -r <output_file> <input_folder> -fd`.

I've updated the code with a test resource generated exactly this way!

crates/rattler_package_streaming/src/reqwest/tokio.rs
jpcorreia99: thank you! Still have to match the string because the zip crate uses `UnsupportedArchive` for multiple different error messsages

crates/rattler_package_streaming/src/tokio/async_read.rs
jpcorreia99: got it, kept the existing api and pushed this logic a level deeper.

crates/rattler_package_streaming/src/fs.rs
```diff
@@ -32,7 +32,7 @@ pub fn extract_tar_bz2(archive: &Path, destination: &Path) -> Result<ExtractResu
 /// ```
 pub fn extract_conda(archive: &Path, destination: &Path) -> Result<ExtractResult, ExtractError> {
     let file = File::open(archive)?;
-    crate::read::extract_conda(file, destination)
+    crate::read::extract_conda_via_streaming(file, destination)
```
jpcorreia99, line 35: just renamed the function to distinguish from the new buffered solution

jpcorreia99:
hey @baszalmstra would appreciate a new review when possible for you :)

crates/rattler_package_streaming/src/tokio/async_read.rs
```diff
@@ -30,17 +31,49 @@ pub async fn extract_tar_bz2(
 }
 
 /// Extracts the contents of a `.conda` package archive.
+/// This will perform on-the-fly decompression by streamign the reader.
```
nichmor, line 34:
```suggestion
/// This will perform on-the-fly decompression by streaming the reader.
```

crates/rattler_package_streaming/src/tokio/async_read.rs
```diff
@@ -30,17 +31,49 @@ pub async fn extract_tar_bz2(
...
+    extract_conda_internal(
+        reader,
+        destination,
+        crate::read::extract_conda_via_buffering,
+    )
+    .await
+}
+
+/// Extracts the contents of a `.conda` package archive using the provided extraction function
+pub async fn extract_conda_internal(
```
nichmor, line 62: Maybe we can make this function non-public because we already provide `extract_conda` and `extract_conda_via_buffering` as API

crates/rattler_package_streaming/src/tokio/async_read.rs
```diff
@@ -30,17 +31,49 @@ pub async fn extract_tar_bz2(
...
+) -> Result<ExtractResult, ExtractError> {
+    extract_conda_internal(
+        reader,
+        destination,
+        crate::read::extract_conda_via_streaming,
+    )
+    .await
+}
+
+/// Extracts the contents of a `.conda` package archive by first fully reading the streaming, instead
```
nichmor, line 47: maybe something like this:
`/// Extracts the contents of a `.conda` package archive by fully reading the streaming and then decompressing`

so basically we do not mention `on the fly-decompression`

crates/rattler_package_streaming/tests/extract.rs
```diff
@@ -279,6 +280,36 @@ fn test_extract_flaky_conda(#[values(0, 1, 13, 50, 74, 150, 8096, 16384, 20000)]
     assert_matches::assert_matches!(result, ExtractError::IoError(_));
 }
 
+#[rstest]
+fn test_extract_data_descriptor_package_fails_streaming_and_uses_buffering() {
+    let package_path = "tests/resources/ca-certificates-2024.7.4-hbcca054_0.conda";
+
+    let temp_dir = Path::new(env!("CARGO_TARGET_TMPDIR"));
+    println!("Target dir: {}", temp_dir.display());
```
nichmor, line 288: I think we can remove println

crates/rattler_package_streaming/tests/extract.rs
```diff
@@ -111,7 +111,7 @@ fn test_extract_conda(#[case] input: Url, #[case] sha256: &str, #[case] md5: &st
     println!("Target dir: {}", temp_dir.display());
     let file_path = tools::download_and_cache_file(input, sha256).unwrap();
     let target_dir = temp_dir.join(file_path.file_stem().unwrap());
-    let result = extract_conda(
+    let result = extract_conda_via_streaming(
```
nichmor, line 114: can we keep here `extract_conda ? I think it will still call underneath `extract_conda_via_streaming` so we will be sure that we didn't change old implementation somehow

crates/rattler_package_streaming/tests/extract.rs
```diff
@@ -266,7 +267,7 @@ fn test_extract_flaky_conda(#[values(0, 1, 13, 50, 74, 150, 8096, 16384, 20000)]
     let temp_dir = Path::new(env!("CARGO_TARGET_TMPDIR"));
     println!("Target dir: {}", temp_dir.display());
     let target_dir = temp_dir.join(package_path.file_stem().unwrap());
-    let result = extract_conda(
+    let result = extract_conda_via_streaming(
```
nichmor, line 270: I think the same here ^

crates/rattler_package_streaming/tests/extract.rs
```diff
@@ -279,6 +280,36 @@ fn test_extract_flaky_conda(#[values(0, 1, 13, 50, 74, 150, 8096, 16384, 20000)]
...
+            "The file length is not available in the local header"
+        ))
+    );
+
+    let new_result =
+        extract_conda_via_buffering(File::open(package_path).unwrap(), &target_dir).unwrap();
+
+    assert_eq!(
+        &format!("{:x}", new_result.sha256),
+        "6a5d6d8a1a7552dbf8c617312ef951a77d2dac09f2aeaba661deebce603a7a97"
```
nichmor, line 305: maybe we can use `insta::assert_snapshot` for asserting this?

crates/rattler_package_streaming/tests/write.rs
```diff
@@ -209,7 +209,7 @@ fn test_rewrite_conda() {
 
         let name = file_path.file_stem().unwrap().to_string_lossy();
         let target_dir = temp_dir.join(file_path.file_stem().unwrap());
-        extract_conda(File::open(&file_path).unwrap(), &target_dir).unwrap();
+        extract_conda_via_streaming(File::open(&file_path).unwrap(), &target_dir).unwrap();
```
nichmor, line 212: I think we should still test our current `extract_conda`

nichmor (commented):
Looks very good ! Thanks for contributing this
I've left some small comments

nichmor:
I checked the code a little, 
so [here](https://github.com/zip-rs/zip-old/blob/e485cbf5768a3e44685c8e5da8126c541e04d01f/src/read.rs#L587) 
they are reading some flags to determine if an error should be thrown:
`https://github.com/zip-rs/zip-old/blob/e485cbf5768a3e44685c8e5da8126c541e04d01f/src/read.rs#L638`

could we potentially rely on this and do this verification ourselves and correctly choose the the way we should decompress this? ( via streaming or buffering ) cc @baszalmstra

crates/rattler_package_streaming/tests/extract.rs
jpcorreia99: same response as in https://github.com/conda-incubator/rattler/pull/797#discussion_r1698415712. This test was testing the innermost method  that I just renamed to `extract_conda_via_streaming`. I'm happy to rename it back to `extract_conda` to get rid of this confusing state if it helps!

crates/rattler_package_streaming/tests/extract.rs
jpcorreia99: same response as above ^

jpcorreia99:
hey @nichmor thank you for the review! I've addressed them or commented back, would appreciate a new review whenever you have the time :)

crates/rattler_package_streaming/tests/extract.rs
```diff
@@ -279,6 +280,29 @@ fn test_extract_flaky_conda(#[values(0, 1, 13, 50, 74, 150, 8096, 16384, 20000)]
...
+        ExtractError::ZipError(zip::result::ZipError::UnsupportedArchive(
+            "The file length is not available in the local header"
+        ))
+    );
+
+    let new_result =
+        extract_conda_via_buffering(File::open(package_path).unwrap(), &target_dir).unwrap();
+
+    insta::assert_snapshot!("new_result_sha256", &format!("{:x}", new_result.sha256));
+    insta::assert_snapshot!("new_result_md5", &format!("{:x}", new_result.md5));
```
baszalmstra, line 303: Please combine the two snapshot values in a single snapshot (otherwise you have to run the test twice to update the snapshots) and since the output is not large prefer an [inline snapshot](https://insta.rs/docs/snapshot-types/#inline-snapshots).

baszalmstra (commented):
Some small requests but otherwise LGTM

crates/rattler_package_streaming/tests/extract.rs
jpcorreia99: Addressed in https://github.com/conda-incubator/rattler/pull/797/commits/16f584098cd141d193c3876e5e74f0ac0536fc77, please let me know if my approach makes sense!

baszalmstra (approved):

crates/rattler_package_streaming/tests/extract.rs
nichmor: ah, thanks for clarifying this!

nichmor (approved):
thanks! ❤️

PR #800, opened 2024-07-31
chore: add / for forward compatibility of the lockfile

ruben-arts:
This should re-add the backslash to the channel.

This is to be forward compatible with the previous situation.

### DIFF
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -90,16 +90,17 @@ impl NamedChannelOrUrl {
     /// Converts the channel to a base url using the given configuration.
     /// This method ensures that the base url always ends with a `/`.
     pub fn into_base_url(self, config: &ChannelConfig) -> Url {
-        match self {
+        let url = match self {
             NamedChannelOrUrl::Name(name) => {
                 let mut base_url = config.channel_alias.clone();
                 if let Ok(mut segments) = base_url.path_segments_mut() {
                     segments.push(&name);
                 }
-                add_trailing_slash(&base_url).into_owned()
+                base_url
             }
-            NamedChannelOrUrl::Url(url) => add_trailing_slash(&url).into_owned(),
-        }
+            NamedChannelOrUrl::Url(url) => url,
+        };
+        add_trailing_slash(&url).into_owned()
     }
 
     /// Converts this instance into a channel.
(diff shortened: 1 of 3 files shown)


crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -94,7 +97,7 @@ impl NamedChannelOrUrl {
                 }
                 base_url
```
baszalmstra, line 98: This is not guaranteed to have a trailing slash now.

baszalmstra: Maybe you can also add a comment to the function that the returned url will always have a trailing slash?

baszalmstra: And also add some tests to verify that base_url does always have a trailing slash but as_str doesnt.

ruben-arts: I think i've added what you requested in this comment but I'm not sure what you mean with the `as_str`. The canonical name is tested in:
https://github.com/mamba-org/rattler/blob/68c4254de11293639a5800b6f05295405b2d5eff/crates/rattler_conda_types/src/channel/mod.rs#L654-L673

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -85,16 +88,17 @@ impl NamedChannelOrUrl {
...
+    /// This method ensures that the base url always ends with a `/`.
     pub fn into_base_url(self, config: &ChannelConfig) -> Url {
         match self {
             NamedChannelOrUrl::Name(name) => {
                 let mut base_url = config.channel_alias.clone();
                 if let Ok(mut segments) = base_url.path_segments_mut() {
                     segments.push(&name);
                 }
-                base_url
+                add_trailing_slash(&base_url).into_owned()
```
baszalmstra, line 99: This line is a duplicate of the line below. Please return something from the match and then apply removing the trailing slash on the return value.

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -667,4 +671,41 @@ mod tests {
...
+        let channel = Channel::from_str("conda-forge", &channel_config).unwrap();
+        let channel_with_backslash =
+            Channel::from_str("https://conda.anaconda.org/conda-forge/", &channel_config).unwrap();
+        assert_eq!(channel.base_url(), channel_with_backslash.base_url());
+
+        // Channel with backslash should have backslash
+        let channel_with_backslash = Channel::from_str("conda-forge/", &channel_config).unwrap();
+        let channel =
+            Channel::from_str("https://conda.anaconda.org/conda-forge", &channel_config).unwrap();
+        assert_eq!(channel.base_url(), channel_with_backslash.base_url());
```
baszalmstra, line 692: Here you are basically checking if for some input when you parse it to a channel it contains a backslash. But there is a lot of duplication. Maybe write a for loop over some input values and in the for loop check if the base_url contains a backslash. You can then also add more input values. Try with a url that already contains a backslash etc.

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -667,4 +671,41 @@ mod tests {
...
+
+        // The named channel to channel should have backslash
+        let channel = named_channel.into_channel(&channel_config);
+        let channel_with_backslash =
+            Channel::from_str("https://conda.anaconda.org/conda-forge/", &channel_config).unwrap();
+        assert_eq!(channel.base_url(), channel_with_backslash.base_url());
+        assert_eq!(
+            channel.base_url().as_str(),
+            "https://conda.anaconda.org/conda-forge/"
+        );
```
baszalmstra: Some here!

crates/rattler_conda_types/src/channel/mod.rs
ruben-arts: :+1:

crates/rattler_conda_types/src/channel/mod.rs
ruben-arts: :+1:

crates/rattler_conda_types/src/channel/mod.rs
ruben-arts: Cleaned it up

baszalmstra (approved):

PR #801, opened 2024-07-31
fix: redact secrets in the `canonical_name` functions

wolfv:
This is a bit annoying, because I didn't want to depend on either `rattler_networking` or `rattler_conda_types` in either of the crates, so I duplicated the code for now. 

We could factor it out and move it into a `rattler_redaction` or `rattler_utils` crate.
Ideally we would implement it for `Url` somewhere, and the redact implementation for `reqwest::Error` would stay in the networking crate.

I was also not sure if we want the `canonical_name` function to return a `NameOrUrl` or just a `String` (whcih the one on `Channel` does, I believe).

### DIFF
--- a/crates/rattler_conda_types/Cargo.toml
+++ b/crates/rattler_conda_types/Cargo.toml
@@ -37,6 +37,7 @@ tracing = { workspace = true }
 typed-path = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 indexmap = { workspace = true }
+rattler_redaction = { path = "../rattler_redaction" }
 
 [dev-dependencies]
 rand = { workspace = true }
--- a/crates/rattler_conda_types/Cargo.toml
+++ b/crates/rattler_conda_types/Cargo.toml
@@ -37,7 +37,7 @@ tracing = { workspace = true }
 typed-path = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 indexmap = { workspace = true }
-rattler_redaction = { path = "../rattler_redaction" }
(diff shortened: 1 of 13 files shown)


baszalmstra:
Yeah let's move it into a new crate (`rattler_redaction` sounds great!), the crate can also have an optional dependency on `reqwest_middleware`

wolfv:
What about `rattler_utils`? I think some people hate "utils" though.

wolfv:
OK, moved things into their own crateycrate. 🚀

crates/rattler_conda_types/Cargo.toml
```diff
@@ -37,6 +37,7 @@ tracing = { workspace = true }
 typed-path = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 indexmap = { workspace = true }
+rattler_redaction = { path = "../rattler_redaction" }
```
baszalmstra, line 40: You need to specify the version otherwise we cant publish

crates/rattler_repodata_gateway/Cargo.toml
```diff
@@ -53,6 +53,7 @@ tracing = { workspace = true }
 url = { workspace = true, features = ["serde"] }
 zstd = { workspace = true }
 rattler_cache = { version = "0.1.4", path = "../rattler_cache" }
+rattler_redaction = { path = "../rattler_redaction", features = ["reqwest", "reqwest-middleware"] }
```
baszalmstra, line 56: Same here, you have to add a version.

baszalmstra (approved):

baszalmstra (approved):

PR #804, opened 2024-08-06
fix: parse `~=` as version not as path

ruben-arts:
By adding the path parsing to the nameless matchspec I broke the parsing of `~= 1.2.3`. 

This PR changes a few things to better deal with that. 
- It now checks for `~/` to verify if it is a path
- We can now parse `~/channel_name::package` as the tilde is now actually parsed instead of seen as part of the string.
- Added a bunch more test cases to catch more issues in the future.

### DIFF
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -449,14 +449,14 @@ fn absolute_path(path: &str, root_dir: &Path) -> Result<Utf8TypedPathBuf, ParseC
     }
 
     // Parse the `~` as the home folder
-    if path.starts_with("~/") {
+    if let Some(user_path) = path.strip_prefix("~/").ok().or_else(|| path.strip_prefix("~\\").ok()) {
         return Ok(Utf8TypedPathBuf::from(
             dirs::home_dir()
                 .ok_or(ParseChannelError::InvalidPath(path.to_string()))?
                 .to_string_lossy()
                 .as_ref(),
         )
-        .join(path.strip_prefix("~/").unwrap()));
+        .join(user_path);
     }
 
     let root_dir_str = root_dir
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -534,7 +534,7 @@ mod tests {
             .unwrap()
             .into_os_string()
             .into_encoded_bytes();
-        let home_dir = Utf8NativePath::from_bytes_path(&NativePath::new(&home_dir))
+        let home_dir = Utf8NativePath::from_bytes_path(NativePath::new(&home_dir))
             .unwrap()
             .to_typed_path();
         assert_eq!(
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -6,12 +6,11 @@ use std::{
 };
 
 use file_url::directory_path_to_url;
+use rattler_redaction::Redact;
 use serde::{Deserialize, Serialize, Serializer};
 use thiserror::Error;
 use typed_path::{Utf8NativePathBuf, Utf8TypedPath, Utf8TypedPathBuf};
 use url::Url;
-
-use rattler_redaction::Redact;
 
 use super::{ParsePlatformError, Platform};
 use crate::utils::{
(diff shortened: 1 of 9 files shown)


crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -449,6 +448,17 @@ fn absolute_path(path: &str, root_dir: &Path) -> Result<Utf8TypedPathBuf, ParseC
...
+    // Parse the `~` as the home folder
+    if path.starts_with("~/") {
+        return Ok(Utf8TypedPathBuf::from(
+            dirs::home_dir()
+                .ok_or(ParseChannelError::InvalidPath(path.to_string()))?
+                .to_string_lossy()
+                .as_ref(),
+        )
+        .join(path.strip_prefix("~/").unwrap()));
+    }
```
baszalmstra, line 460: You can avoid an `unwrap` and also account for windows paths:

```suggestion
    if let Some(user_path) = path.strip_prefix("~/").ok().or_else(|| path.strip_prefix("~\\").ok()) {
        return Ok(Utf8TypedPathBuf::from(
            dirs::home_dir()
                .ok_or(ParseChannelError::InvalidPath(path.to_string()))?
                .to_string_lossy()
                .as_ref(),
        )
        .join(user_path);
    }
```

crates/rattler_conda_types/src/utils/path.rs
```diff
@@ -9,7 +25,7 @@ pub(crate) fn is_path(path: &str) -> bool {
     // Check if the path starts with a common path prefix
     if path.starts_with("./")
         || path.starts_with("..")
-        || path.starts_with('~')
+        || path.starts_with("~/")
```
baszalmstra, line 28:
```suggestion
        || path.starts_with("~/")
        || path.starts_with("~\\")
```

crates/rattler_conda_types/src/utils/path.rs
```diff
@@ -21,3 +37,34 @@ pub(crate) fn is_path(path: &str) -> bool {
...
+    #[test]
+    fn test_is_path() {
+        assert!(is_path("/foo"));
+        assert!(is_path("/C:/foo"));
+        assert!(is_path("C:/foo"));
+        assert!(is_path("\\\\foo"));
+        assert!(is_path("\\\\server\\foo"));
+
+        assert!(is_path("./conda-forge/label/rust_dev"));
+        assert!(is_path("~/foo"));
```
baszalmstra, line 66:
```suggestion
        assert!(is_path("~/foo"));
        assert!(is_path("~\\foo"));
```

crates/rattler_conda_types/src/utils/path.rs
```diff
@@ -21,3 +37,34 @@ pub(crate) fn is_path(path: &str) -> bool {
...
+    #[test]
+    fn test_is_absolute_path() {
+        assert!(is_absolute_path("/foo"));
+        assert!(is_absolute_path("/C:/foo"));
+        assert!(is_absolute_path("C:/foo"));
+        assert!(is_absolute_path("\\\\foo"));
+        assert!(is_absolute_path("\\\\server\\foo"));
+
+        assert!(!is_absolute_path("conda-forge/label/rust_dev"));
+        assert!(!is_absolute_path("~/foo"));
```
baszalmstra, line 51:
```suggestion
        assert!(!is_absolute_path("~/foo"));
        assert!(!is_absolute_path("~\\foo"));
```

baszalmstra (changes requested):

ruben-arts:
@baszalmstra  The `~\` wasn't something that worked with the `TypedPath::starts_with()`. Removed that code again but added more tests.

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -444,11 +443,27 @@ pub(crate) const fn default_platforms() -> &'static [Platform] {
...
     if path.is_absolute() {
         return Ok(path.normalize());
     }
 
+    // Parse the `~/` as the home folder
+    if let Ok(user_path) = path.strip_prefix("~/") {
+        return Ok(Utf8TypedPathBuf::from(
+            dirs::home_dir()
+                .ok_or(ParseChannelError::InvalidPath(path.to_string()))?
+                .to_string_lossy()
```
baszalmstra, line 461: You should use `to_str` and if that fails return `ParseChannelError::NotUtf8RootDir(..)`

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -444,11 +443,27 @@ pub(crate) const fn default_platforms() -> &'static [Platform] {
...
+                .as_ref(),
+        )
+        .join(user_path));
```
baszalmstra, line 464:
```suggestion
        .join(user_path))
	.normalize();
```

ruben-arts:
The `insta::settings!({filters}...` are sooooooo awesome!

crates/rattler_conda_types/src/channel/mod.rs
ruben-arts: I changed it to return the given path. But the error is not really correct then, as it is going to error on the `dirs::home_dir()` not being available. Which is just weird and not well documented when that would go wrong.

crates/rattler_conda_types/src/channel/mod.rs
ruben-arts: :+1: Done and added a test

baszalmstra (approved):

PR #810, opened 2024-08-11
feat: parse `channel` key and consolidate `NamelessMatchSpec`

baszalmstra:
Fixes https://github.com/conda-incubator/rattler/issues/793

Continuation of https://github.com/conda-incubator/rattler/pull/792 that adds support for `1.2.3[channel=...]` and consolidates logic of parsing matchspec and namelessmatchspec some more.


ruben-arts (approved):
LGTM, added a test for a case I was curious about and it behaves like I expect. So I'm all good

PR #811, opened 2024-08-13
feat: add extra field

nichmor:
Needed for  https://github.com/conda-forge/conda-smithy/issues/2034


crates/rattler_conda_types/src/package/about.rs
```diff
@@ -41,6 +41,10 @@ pub struct AboutJson {
     )]
     pub doc_url: Vec<Url>,
 
+    /// Extra metadata that was passed during the build
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub extra: Option<HashMap<String, String>>,
```
baszalmstra, line 46:
```suggestion
    #[serde(skip_serializing_if = "HashMap::is_empty", default)]
    pub extra: HashMap<String, String>,
```

baszalmstra (changes requested):

nichmor:
it looks like we need to use serde_json::Value as extra can contain different objects types

baszalmstra:
I think we should use an `IndexMap` instead to preserve the insertion order.

wolfv:
Or btree?

nichmor:
I reused FxHashMap as we do in repodata , it is ok?

baszalmstra:
I would prefer to just keep the insertion order using an `IndexMap`. But if you wanna sort them by name I would indeed use a `BTreeMap`.

nichmor:
> I would prefer to just keep the insertion order using an `IndexMap`. But if you wanna sort them by name I would indeed use a `BTreeMap`.

changed to BTreeMap to not have reproducibility issues in future

baszalmstra (approved):

PR #813, opened 2024-08-13
refactor: change links from conda-incubator to conda

baszalmstra:
Change all links from `conda-incubator` to `conda`


baszalmstra:
I need to fix the python build but that's independant of this PR.

Issue #809, opened 2024-08-08
Many `__init__` are missing from documentation


calebzulawski:
I'm working on a `conda-lock`-ish custom tool for my company's build system.  The py-rattler API is great!  But I'm noticing many `__init__` functions are missing from the documentation, so I'm consulting the source.  Is this intentional and I shouldn't be using these constructors?  Maybe they can be added to the docs even without a docstring?

An example: creating a `LockFile` from `Environment`s, and creating `Environment` from `RepoDataRecord`s.

wolfv:
That sounds great. If you have time to send a PR that would be heavily appreciated.

wolfv:
As we merged the PR, let's close this issue! :)

PR #817, opened 2024-08-14
Add missing __init__ docstrings

calebzulawski:
Add some missing `__init__` documentation for #809


baszalmstra (approved):

calebzulawski:
This should be good now

baszalmstra:
I tried to push to your PR to fix the test issue but unfortunately it seems Im not allowed to. 

The fix is very simple, just add an additional `>>>` at the end of the snippets:

```python
>>> NamelessMatchSpec("24")
NamelessMatchSpec("==24")
>>>
```

Would you be able to fix this?

calebzulawski:
Ah, I was confused what was wrong. Thanks!  Looks like you will need to enable auto-merge again

Issue #815, opened 2024-08-14
Glibc detection on solus os not working


wolfv:
Details: https://discord.com/channels/1082332781146800168/1271852940474191924

PR #818, opened 2024-08-14
feat: Add support for `CONDA_OVERRIDE_CUDA`

SobhanMP:
### Description

Hi, I know this is terrible, not elegant, etc, but I do think that having support for something in the spirit of CONDA_OVERRIDE_CUDA is necessary on academic HPC environments. Reasons include 
- only having internet access on the login node with no gpu
- launching jobs with submitit on login nodes with no gpu.

I'm not really used to writing rust, so would be greatful for some guidance.

Thanks

### DIFF
--- a/crates/rattler_virtual_packages/src/cuda.rs
+++ b/crates/rattler_virtual_packages/src/cuda.rs
@@ -13,7 +13,6 @@ use once_cell::sync::OnceCell;
 use rattler_conda_types::Version;
 use std::process::Command;
 use std::{
-    env,
     mem::MaybeUninit,
     os::raw::{c_int, c_uint, c_ulong},
     str::FromStr,
--- a/crates/rattler_virtual_packages/src/cuda.rs
+++ b/crates/rattler_virtual_packages/src/cuda.rs
@@ -27,14 +27,10 @@ pub fn cuda_version() -> Option<Version> {
         .clone()
 }
 
-const CUDA_OVERRIDE_ENV: &str = "CONDA_OVERRIDE_CUDA";
 /// Attempts to detect the version of CUDA present in the current operating system by employing the
 /// best technique available for the current environment.
 pub fn detect_cuda_version() -> Option<Version> {
-    let cuda_override = env::var(CUDA_OVERRIDE_ENV);
-    if cuda_override.is_ok() {
-        Version::from_str(&cuda_override.unwrap()).ok()
-    } else if cfg!(target_env = "musl") {
+    if cfg!(target_env = "musl") {
         // Dynamically loading a library is not supported on musl so we have to fall-back to using
         // the nvidia-smi command.
         detect_cuda_version_via_nvidia_smi()
@@ -273,16 +269,5 @@ mod test {
     pub fn doesnt_crash_nvidia_smi() {
         let version = detect_cuda_version_via_nvidia_smi();
         println!("Cuda {version:?}");
-    }
-
-    #[test]
-    pub fn cuda_override(){
-        let backup = env::var(CUDA_OVERRIDE_ENV);
-        env::set_var(CUDA_OVERRIDE_ENV, "69.42");
-        let version = detect_cuda_version();
-        assert!(version.unwrap() == Version::from_str("69.42").unwrap());
-        if backup.is_ok() {
-            env::set_var(CUDA_OVERRIDE_ENV, backup.unwrap())
-        }
     }
 }
--- a/crates/rattler_virtual_packages/src/cuda.rs
+++ b/crates/rattler_virtual_packages/src/cuda.rs
@@ -13,6 +13,7 @@ use once_cell::sync::OnceCell;
 use rattler_conda_types::Version;
 use std::process::Command;
 use std::{
+    env,
     mem::MaybeUninit,
     os::raw::{c_int, c_uint, c_ulong},
     str::FromStr,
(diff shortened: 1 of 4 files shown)


crates/rattler_virtual_packages/src/cuda.rs
```diff
@@ -26,10 +27,14 @@ pub fn cuda_version() -> Option<Version> {
         .clone()
 }
 
+const CUDA_OVERRIDE_ENV: &str = "CONDA_OVERRIDE_CUDA";
 /// Attempts to detect the version of CUDA present in the current operating system by employing the
 /// best technique available for the current environment.
 pub fn detect_cuda_version() -> Option<Version> {
```
baszalmstra, line 33: This function should be pure functional without environmental side-effects.

Instead I propose we add some sort of `from_env` function to the appropriate virtual packages that allow users to choose whether they want this behavior. Or a `or_override_with_env_var` or something along those lines.

We should also debate the environment variable name. I believe conda has a different name? Im also not sure whether we should hardcode that in rattler itself.

wolfv:
I think this would also be good for all the other virtual packages as a way to hot-fix certain scenarios that we haven't covered yet (#815 for example).

baszalmstra:
My idea would be to add functions to the virtual package implementations. Lets take CUDA as an example, we would add these functions to the `Cuda` impl:

```rust
impl Cuda {
    pub fn from_env_var_name(env_var_name: &str) -> Option<Self> {
		let version_str = env::var(env_var_name)?;
		// Extract versions
		...
    }
    
    pub fn from_default_env_var() -> Option<Self> {
    	Self::from_env_var_name("CONDA_OVERRIDE_CUDA")
    }
    
    ...
}
```

You can then write:

```
let cuda = Cuda::from_default_env_var().or_else(Cuda::current);
```

To get the CUDA version of the system based on env var, or if not present by detecting it.

Similarly a function `VirtualPackage::from_default_env_vars` could be added. 

I think this allows:

* Users of the library to decide whether they want to allow environment overrides.
* Users can specify custom environment variables if they want to.
* Users can use the default environment variable names.

Which I think hits the sweet spot.

And indeed as @wolfv said, it would be nice to add these for the other virtual packages as well.

SobhanMP:
oh, yeah that looks great!

SobhanMP:
Does this make sense as a rough draft?

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -34,16 +34,42 @@ pub mod linux;
...
+
+impl<T: From<Version>> EnvOverride for T {
+    fn from_env_var_name(env_var_name: &str) -> Option<Self> {
+        let var = env::var(env_var_name).ok()?;
+        Some(Self::from(Version::from_str(&var).ok()?))
+    }
+
+    fn default_env_name() -> String {
+        format!("CONDA_{}_OVERIDE", type_name::<T>().to_uppercase())
+    }
```
baszalmstra, line 70: I dont think relying on type names is good here. I prefer just reimplementing this per type instead of using a blanket implementation

baszalmstra: The name should also be [`CONDA_OVERRIDE_CUDA`](https://github.com/conda/conda/blob/77aa3d337963fdbe5c817304a2052ae0125fd326/docs/source/user-guide/tasks/manage-virtual.rst#overriding-detected-packages)

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -229,6 +261,29 @@ impl From<LibC> for VirtualPackage {
     }
 }
 
+impl EnvOverride for LibC {
+    fn default_env_name() -> String {
+        "CONDA_LIBC_OVERIDE".to_string()
```
baszalmstra, line 266: Conda only allows overriding the glibc family version it seems:

https://github.com/conda/conda/blob/77aa3d337963fdbe5c817304a2052ae0125fd326/docs/source/user-guide/tasks/manage-virtual.rst#overriding-detected-packages

baszalmstra (commented):
Yeah that looks a lot better! I left a few comments.

crates/rattler_virtual_packages/src/lib.rs
SobhanMP: Does rattler need to adhere to that?

crates/rattler_virtual_packages/src/lib.rs
SobhanMP: nice catch. The blanket implementation is opt in, is it really a problem?

SobhanMP:
I think there is one issue, finding the environment variable and it being empty is different than the environment variable not being present, should the type be `Option<Option<T>>`? first one option showing that the env was found the second one that the env was none?

SobhanMP:
I think everything should have been handled gracefully.

SobhanMP:
Do we want to override the glibc runtime? like `CONDA_OVERRIDE_GLIBC="2@musl"`, same for `Linux` and `Archspec`. We could also do unix and win but all of these are not overridable by conda.

baszalmstra (approved):
This looks good! Ill take a closer look at the CI failure tomorrow!

SobhanMP:
This should solve some of the failure, pre-commit is not ran by default

SobhanMP:
I don't think pyrattler has any implication for pixi, so I'd rather this be merged w/o py-rattler support. My understanding is that  in `py-rattler/src/virtual_package.rs` the PyVirtualPackage needs to expose some more code, the problem is that this either requires either
1- A special classs for virtual packages that can be overriden (there is already one for arch), 
2- that all virtual pacakges have override, or 
3- using `Cuda::from_default_env_var.unwrap_or(Cuda::current().into()).unwrap()` in the `VirtualPackage::current` or `try_detect_virtual_packages` instead of `Cuda::current()`

SobhanMP:
@baszalmstra It's very inconvenient that the checks require maintainer auth.

baszalmstra:
Yes, but once your first PR lands that restriction is lifted 👍

baszalmstra:
I think it would be good to also expose this functionality to the python bindings, but we can do that in another PR.

Thanks for your first contribution @SobhanMP !

SobhanMP:
@baszalmstra I think this first requires figuring out https://github.com/conda/rattler/pull/818#issuecomment-2297850077 . I think option 2 makes the most sense.

SobhanMP:
The next thing is where this should be used. Should  `try_detect_virtual_packages` be patched? Should it take some structure to rename overrides?

PR #819, opened 2024-08-14
feat: add package names api for gateway

nichmor:
### Description

   
   will be used by new gateway for pixi search

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
@@ -64,4 +64,9 @@ impl SubdirClient for LocalSubdirClient {
         })
         .await
     }
+
+    fn package_names(&self) -> Arc<[String]> {
+        let sparse_repodata: Arc<SparseRepoData> = self.sparse.clone();
+        sparse_repodata.package_names().map(|v| v.into()).collect()
+    }
 }
--- a/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
@@ -65,8 +65,11 @@ impl SubdirClient for LocalSubdirClient {
         .await
     }
 
-    fn package_names(&self) -> Arc<[String]> {
+    fn package_names(&self) -> Vec<String> {
         let sparse_repodata: Arc<SparseRepoData> = self.sparse.clone();
-        sparse_repodata.package_names().map(std::convert::Into::into).collect()
+        sparse_repodata
+            .package_names()
+            .map(std::convert::Into::into)
+            .collect()
     }
 }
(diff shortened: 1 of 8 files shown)


crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
```diff
@@ -64,4 +64,9 @@ impl SubdirClient for LocalSubdirClient {
         })
         .await
     }
+
+    fn package_names(&self) -> Arc<[String]> {
```
baszalmstra, line 68: You are currently not caching this method so there is not need to return an `Arc`. Just use `Vec<String>`.

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -271,11 +271,163 @@ impl GatewayQuery {
...
     fn into_future(self) -> Self::IntoFuture {
         self.execute().boxed()
     }
 }
+
+/// Represents a query for package names to execute with a [`Gateway`].
+///
+/// When executed the query will asynchronously load the package names from all
+/// subdirectories (combination of channels and platforms).
+///
```
baszalmstra, line 287:
```suggestion
```

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -271,11 +271,163 @@ impl GatewayQuery {
...
+///
+/// Repodata is cached by the [`Gateway`] so executing the same query twice
+/// with the same channels will not result in the repodata being fetched
+/// twice.
```
baszalmstra, line 291: This doesnt apply to this function.

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -271,11 +271,163 @@ impl GatewayQuery {
...
+    /// query. This allows reporting progress back to a user.
+    pub fn with_reporter(self, reporter: impl Reporter + 'static) -> Self {
+        Self {
+            reporter: Some(Arc::new(reporter)),
+            ..self
+        }
+    }
+
+    /// Execute the query and return the package names.
+    pub async fn execute(self) -> Result<Vec<String>, GatewayError> {
```
baszalmstra, line 336: This function is way more complex than it needs to be. There is no recursion in the names so you can do all of this concurrently. 

In pseudo code:

```
let package_names = Set::new();
for (channel, platform) in channels_and_platforms {
   subdir = get_or_create_subdir(..).await
   package_names.extend(subdir.package_names().await)
}

return package_names
```

You just need to do this concurrently for each `subdir`. Ill let you figure out how ;)

baszalmstra (changes requested):

crates/rattler_repodata_gateway/src/gateway/query.rs
nichmor: I've removed all unnecessary parts and just used FuturesUnordered to poll results.

Let me know if it's ok

baszalmstra (commented):
lets return `PackageName` as the type. Otherwise looks great!

nichmor:
> lets return `PackageName` as the type. Otherwise looks great!

I've changed NameQuery to return PackageName.

I wasn't sure if the SubDirClient trait should return PackageName instead of strings so I've decided to keep it more simple building block.

baszalmstra (approved):

PR #834, opened 2024-08-22
feat: start adding interface to override

SobhanMP:
Following 721a6c101ecc4a726ab13b4e051adfc266691474, this commit will start adding the interface for python.

### DIFF
--- a/crates/rattler_virtual_packages/src/lib.rs
+++ b/crates/rattler_virtual_packages/src/lib.rs
@@ -70,7 +70,13 @@ pub trait EnvOverride: Sized {
         F: FnOnce() -> Result<Option<Self>, DetectVirtualPackageError>,
     {
         match env::var(env_var_name) {
-            Ok(var) => Ok(Some(Self::from_env_var_name_with_var(env_var_name, &var)?)),
+            Ok(var) => {
+                if var.is_empty() {
+                    Ok(None)
+                } else {
+                    Ok(Some(Self::from_env_var_name_with_var(env_var_name, &var)?))
+                }
+            }
             Err(env::VarError::NotPresent) => f(),
             Err(e) => Err(DetectVirtualPackageError::VarError(e)),
         }
--- a/crates/rattler_virtual_packages/src/lib.rs
+++ b/crates/rattler_virtual_packages/src/lib.rs
@@ -164,6 +164,24 @@ impl VirtualPackage {
             .get_or_try_init(try_detect_virtual_packages)
             .map(Vec::as_slice)
     }
+
+    /// disable overrides
+    pub fn current_no_overrides() -> Result<&'static [Self], DetectVirtualPackageError> {
+        static DETECTED_VIRTUAL_PACKAGES: OnceCell<Vec<VirtualPackage>> = OnceCell::new();
+        DETECTED_VIRTUAL_PACKAGES
+            .get_or_try_init(try_detect_virtual_packages_no_overrides)
+            .map(Vec::as_slice)
+    }
+
+    /// use custom overrides
+    pub fn current_with_overrides(
+        overrides: &VirtualPackageOverride<'_>,
+    ) -> Result<&'static [Self], DetectVirtualPackageError> {
+        static DETECTED_VIRTUAL_PACKAGES: OnceCell<Vec<VirtualPackage>> = OnceCell::new();
+        DETECTED_VIRTUAL_PACKAGES
+            .get_or_try_init(|| try_detect_virtual_packages_with_overrides(overrides))
+            .map(Vec::as_slice)
+    }
 }
 
 /// An error that might be returned by [`VirtualPackage::current`].
@@ -186,14 +204,17 @@ pub enum DetectVirtualPackageError {
     VersionParseError(#[from] ParseVersionError),
 }
 
-struct VirtualPackageOverride<'a> {
+/// Configure the overrides used in in this crate.
+
+pub struct VirtualPackageOverride<'a> {
     osx: Option<&'a str>,
     libc: Option<&'a str>,
     cuda: Option<&'a str>,
 }
 
 impl VirtualPackageOverride<'static> {
-    fn none() -> Self {
+    /// Disable all overrides
+    pub fn none() -> Self {
         Self {
             osx: None,
             libc: None,
@@ -201,6 +222,7 @@ impl VirtualPackageOverride<'static> {
         }
     }
 
+    /// Use the default overrides of Conda.
     fn default() -> Self {
         Self {
             osx: Some(Osx::DEFAULT_ENV_NAME),
@@ -261,8 +283,14 @@ fn try_detect_virtual_packages_with_overrides(
 }
 
 fn try_detect_virtual_packages() -> Result<Vec<VirtualPackage>, DetectVirtualPackageError> {
+    try_detect_virtual_packages_with_overrides(&VirtualPackageOverride::default())
+}
+
+fn try_detect_virtual_packages_no_overrides(
+) -> Result<Vec<VirtualPackage>, DetectVirtualPackageError> {
     try_detect_virtual_packages_with_overrides(&VirtualPackageOverride::none())
 }
+
 /// Linux virtual package description
 #[derive(Clone, Eq, PartialEq, Hash, Debug, Deserialize)]
 pub struct Linux {
--- a/crates/rattler_virtual_packages/src/lib.rs
+++ b/crates/rattler_virtual_packages/src/lib.rs
@@ -105,7 +105,10 @@ pub trait EnvOverride: Sized {
     fn detect_from_host() -> Result<Option<Self>, DetectVirtualPackageError>;
 
     /// Apply the override to the current virtual package. If the override is `None` then use the fallback
-    fn detect_with_fallback<F>(ov: &Override, fallback: F) -> Result<Option<Self>, DetectVirtualPackageError>
+    fn detect_with_fallback<F>(
+        ov: &Override,
+        fallback: F,
+    ) -> Result<Option<Self>, DetectVirtualPackageError>
     where
         F: FnOnce() -> Result<Option<Self>, DetectVirtualPackageError>,
     {
@@ -173,10 +176,7 @@ impl From<VirtualPackage> for GenericVirtualPackage {
 impl VirtualPackage {
     /// Returns virtual packages detected for the current system or an error if the versions could
     /// not be properly detected.
-    #[deprecated(
-        since = "1.0.4",
-        note = "Use `Self::detect` instead"
-    )]
+    #[deprecated(since = "1.0.4", note = "Use `Self::detect` instead")]
     pub fn current() -> Result<Vec<Self>, DetectVirtualPackageError> {
         try_detect_virtual_packages_with_overrides(&VirtualPackageOverrides::none())
     }
--- a/crates/rattler_virtual_packages/src/lib.rs
+++ b/crates/rattler_virtual_packages/src/lib.rs
@@ -62,22 +62,50 @@ pub trait EnvOverride: Sized {
     /// - `None`, then the environment variable did not exist,
     /// - `Some(Err(None))`, then the environment variable exist but was set to zero, so the package should be disabled
     /// - `Some(Ok(pkg))`, then the override was for the package.
-    fn from_env_var_name(env_var_name: &str) -> Option<Result<Self, Option<ParseVersionError>>> {
-        let var = env::var(env_var_name).ok()?;
-        if var.is_empty() {
-            Some(Err(None))
-        } else {
-            Some(Self::from_env_var_name_with_var(env_var_name, &var).map_err(Some))
+    fn from_env_var_name_or<F>(
+        env_var_name: &str,
+        f: F,
+    ) -> Result<Option<Self>, DetectVirtualPackageError>
+    where
+        F: FnOnce() -> Result<Option<Self>, DetectVirtualPackageError>,
+    {
+        match env::var(env_var_name) {
+            Ok(var) => Ok(Some(Self::from_env_var_name_with_var(env_var_name, &var)?)),
+            Err(env::VarError::NotPresent) => f(),
+            Err(e) => Err(DetectVirtualPackageError::VarError(e)),
         }
+    }
+
+    /// Helper method for [`EnvOverride::from_env_var_name_or`] that uses [`EnvOverride::current`] as the default.
+    fn from_env_var_name_or_current(
+        env_var_name: &str,
+    ) -> Result<Option<Self>, DetectVirtualPackageError> {
+        Self::from_env_var_name_or(env_var_name, Self::current)
+    }
+
+    /// Helper method for [`EnvOverride::from_env_var_name_or`] that does not have any default.
+    fn from_env_var_name(env_var_name: &str) -> Result<Option<Self>, DetectVirtualPackageError> {
+        Self::from_env_var_name_or(env_var_name, || Ok(None))
     }
 
     /// Default name of the environment variable that overrides the virtual package.
     const DEFAULT_ENV_NAME: &'static str;
 
     /// Shortcut for `EnvOverride::from_env_var_name(EnvOverride::DEFAULT_ENV_NAME)`.
-    fn from_default_env_var() -> Option<Result<Self, Option<ParseVersionError>>> {
-        Self::from_env_var_name(Self::DEFAULT_ENV_NAME)
+    fn from_default_env_var_or<F>(f: F) -> Result<Option<Self>, DetectVirtualPackageError>
+    where
+        F: FnOnce() -> Result<Option<Self>, DetectVirtualPackageError>,
+    {
+        Self::from_env_var_name_or(Self::DEFAULT_ENV_NAME, f)
     }
+
+    /// Shortcut for `EnvOverride::from_env_var_name(EnvOverride::DEFAULT_ENV_NAME)`.
+    fn from_default_env_var() -> Result<Option<Self>, DetectVirtualPackageError> {
+        Self::from_default_env_var_or(|| Ok(None))
+    }
+
+    /// This method is here so that `<Self as EnvOverride>::current` always returns the same error type.
+    fn current() -> Result<Option<Self>, DetectVirtualPackageError>;
 }
 
 /// An enum that represents all virtual package types provided by this library.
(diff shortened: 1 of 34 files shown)


SobhanMP:
@baszalmstra Should `VirtualPackageOverride` be accessible from python?

SobhanMP:
@iamthebot (we'd have to ask @baszalmstra), but as things are, the overrides are enabled by default, so they work. Do you need to rename the overrides?

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -150,10 +202,46 @@ pub enum DetectVirtualPackageError {
...
+    VarError(#[from] env::VarError),
+
+    #[error(transparent)]
+    VersionParseError(#[from] ParseVersionError),
+}
+
+/// Configure the overrides used in in this crate.
+
+pub struct VirtualPackageOverride<'a> {
+    osx: Option<&'a str>,
```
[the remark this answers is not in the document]
baszalmstra: I wonder if maybe we should change this to a trie state: 

1. dont use any override
2. use the default env var if it exists,
3. use the environment variable with a specific name.

that way we can get rid of all the slightly different functions.

WDYT?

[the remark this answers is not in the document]
SobhanMP: I agree, but how about a 4th option that forces a specific version? (I think @iamthebot would like that)

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -136,6 +170,24 @@ impl VirtualPackage {
             .get_or_try_init(try_detect_virtual_packages)
             .map(Vec::as_slice)
     }
+
+    /// disable overrides
+    pub fn current_no_overrides() -> Result<&'static [Self], DetectVirtualPackageError> {
```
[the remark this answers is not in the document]
SobhanMP: I changed the default to match conda (i.e. use the default overrides)

SobhanMP:
@baszalmstra where would be the right place to put the tests?

```
❯ CONDA_OVERRIDE_CUDA=1984 pixi run rattler create jupyterlab
Pixi task (rattler in default): cargo run --bin rattler --release -- create jupyterlab
    Finished release [optimized] target(s) in 0.13s
     Running `.pixi/target/release/rattler create jupyterlab`
Target prefix: /home/psi/rattler/.prefix
Installing for platform: Linux64
Loaded 20237 records in 520.548978ms
Virtual packages:
  - __unix=0=0
  - __linux=6.6.45=0
  - __glibc=2.40=0
  - __cuda=1984=0
  - __archspec=1=x86_64_v4

 Already up to date
 ```

**Comment by baszalmstra** [comment]
I dont think we need an end-to-end test. I would place some unit tests in the lib.rs that test a few different overrides. Watch out with environment variables though because rust runs all tests in parallel.

**Comment by SobhanMP** [comment]
@baszalmstra I think this is done (minus minor ci failure), unless overrides should also override checks like `platform.is_linux`

File: crates/rattler_virtual_packages/src/lib.rs [hunk 3]
```diff
@@ -48,35 +47,81 @@ use libc::DetectLibCError;
...
+    String(String),
+    /// Disable overrides
+    None,
+}
+
+impl Default for Override {
+    fn default() -> Self {
+        Self::DefaultEnvVar
+    }
+}
```
baszalmstra, line 67:
```suggestion
#[derive(Clone, Debug, PartialEq, Default)]
pub enum Override {
    /// Use the default override env var name
    #[default]
    DefaultEnvVar,
    /// Use custom env var name
    EnvVar(String),
    /// Use a custom override directly
    String(String),
    /// Disable overrides
    None,
}
```

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -48,35 +47,81 @@ use libc::DetectLibCError;
...
+    fn parse_version(value: &str) -> Result<Self, ParseVersionError>;
+
+    /// Helper to convert the output of `parse_version` and handling empty strings.
+    fn parse_version_(value: &str) -> Result<Option<Self>, DetectVirtualPackageError> {
+        if value.is_empty() {
+            Ok(None)
+        } else {
+            Ok(Some(Self::parse_version(value)?))
+        }
+    }
```
baszalmstra, line 82: Maybe rename to `parse_version_opt` (`opt` for optional)? The underscore tells me very little.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -48,35 +47,81 @@ use libc::DetectLibCError;
...
     /// - `Some(Ok(pkg))`, then the override was for the package.
-    fn from_env_var_name(env_var_name: &str) -> Option<Result<Self, Option<ParseVersionError>>> {
-        let var = env::var(env_var_name).ok()?;
-        if var.is_empty() {
-            Some(Err(None))
-        } else {
-            Some(Self::from_env_var_name_with_var(env_var_name, &var).map_err(Some))
+    fn from_env_var_name_or<F>(
+        env_var_name: &str,
+        f: F,
```
baszalmstra, line 91: Maybe we can rename this argument to `default` or something along those lines?

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -48,35 +47,81 @@ use libc::DetectLibCError;
...
         }
     }
 
     /// Default name of the environment variable that overrides the virtual package.
     const DEFAULT_ENV_NAME: &'static str;
 
-    /// Shortcut for `EnvOverride::from_env_var_name(EnvOverride::DEFAULT_ENV_NAME)`.
-    fn from_default_env_var() -> Option<Result<Self, Option<ParseVersionError>>> {
-        Self::from_env_var_name(Self::DEFAULT_ENV_NAME)
+    /// This method is here so that `<Self as EnvOverride>::current` always returns the same error type.
```
baszalmstra, line 106: Maybe also explain in the docs what this should do.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -48,35 +47,81 @@ use libc::DetectLibCError;
...
+    fn current() -> Result<Option<Self>, DetectVirtualPackageError>;
+
+    /// Apply the override to the current virtual package. If the override is `None` then use the fallback
+    fn apply_override_or<F>(ov: &Override, f: F) -> Result<Option<Self>, DetectVirtualPackageError>
```
baszalmstra, line 110: Maybe we can rename `f` to `fallback`?

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -48,35 +47,81 @@ use libc::DetectLibCError;
...
     }
 
     /// Default name of the environment variable that overrides the virtual package.
     const DEFAULT_ENV_NAME: &'static str;
 
-    /// Shortcut for `EnvOverride::from_env_var_name(EnvOverride::DEFAULT_ENV_NAME)`.
-    fn from_default_env_var() -> Option<Result<Self, Option<ParseVersionError>>> {
-        Self::from_env_var_name(Self::DEFAULT_ENV_NAME)
+    /// This method is here so that `<Self as EnvOverride>::current` always returns the same error type.
+    fn current() -> Result<Option<Self>, DetectVirtualPackageError>;
```
baszalmstra, line 107: Maybe we can rename this to `detect_from_host` or something along those lines.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -48,35 +47,81 @@ use libc::DetectLibCError;
...
+        match ov {
+            Override::None => f(),
+            Override::String(str) => Self::parse_version_(str),
+            Override::DefaultEnvVar => Self::from_env_var_name_or(Self::DEFAULT_ENV_NAME, f),
+            Override::EnvVar(name) => Self::from_env_var_name_or(name, f),
+        }
+    }
+
+    /// Shortcut for `apply_override_or` with `Self::current`
+    fn apply_override(ov: &Override) -> Result<Option<Self>, DetectVirtualPackageError> {
```
baszalmstra, line 123: And rename this simply to `detect`?

py-rattler/rattler/virtual_package/virtual_package.py
```diff
@@ -1,11 +1,154 @@
 from __future__ import annotations
 
-from rattler.rattler import PyVirtualPackage
+from rattler.rattler import PyVirtualPackage, PyOverride, PyVirtualPackageOverrides
 from typing import List
 
 from rattler.virtual_package.generic import GenericVirtualPackage
 
 
+class Override:
```
baszalmstra, line 9: This needs some documentation.

py-rattler/rattler/virtual_package/virtual_package.py
```diff
@@ -1,11 +1,154 @@
...
+    def _from_py_override(cls, py_override: PyOverride) -> Override:
+        """Construct Rattler Override from FFI PyOverride object."""
+        override = cls.__new__(cls)
+        override._override = py_override
+        return override
+
+    @classmethod
+    def default_env_var(cls) -> Override:
+        """
+        Returns the default environment variable override.
```
baszalmstra, line 22: This is a bit confusing. Maybe we can write it as:

```suggestion
        Returns a new instance to indicate that the default environment variable should overwrite the detected information from the host if specified.
```

And something similar for the other functions.

py-rattler/rattler/virtual_package/virtual_package.py
```diff
@@ -1,11 +1,154 @@
...
+    def _from_py_virtual_package_overrides(
+        cls, py_virtual_package_overrides: PyVirtualPackageOverrides
+    ) -> VirtualPackageOverrides:
+        """Construct Rattler VirtualPackageOverrides from FFI PyVirtualPackageOverrides object."""
+        virtual_package_overrides = cls.__new__(cls)
+        virtual_package_overrides._overrides = py_virtual_package_overrides
+        return virtual_package_overrides
+
+    @classmethod
+    def default(cls) -> VirtualPackageOverrides:
```
baszalmstra, line 81: I think this should just be the `__init__` function.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -130,11 +175,15 @@ impl From<VirtualPackage> for GenericVirtualPackage {
 impl VirtualPackage {
     /// Returns virtual packages detected for the current system or an error if the versions could
     /// not be properly detected.
-    pub fn current() -> Result<&'static [Self], DetectVirtualPackageError> {
-        static DETECTED_VIRTUAL_PACKAGES: OnceCell<Vec<VirtualPackage>> = OnceCell::new();
-        DETECTED_VIRTUAL_PACKAGES
-            .get_or_try_init(try_detect_virtual_packages)
-            .map(Vec::as_slice)
+    pub fn current() -> Result<Vec<Self>, DetectVirtualPackageError> {
+        try_detect_virtual_packages_with_overrides(&VirtualPackageOverrides::default())
```
baszalmstra, line 179: I think this is a breaking change because now all of a sudden environment variables can override the previous behavior. I think we should restrict the behavior to `current_with_overrides` or perhaps deprecate this function.

We can also rename `current_with_overrides` to `detect`?

crates/rattler_virtual_packages/src/lib.rs
baszalmstra: Or `current_with_overrides` to stick to convention?

baszalmstra (commented):
This is starting to shape up nicely!

crates/rattler_virtual_packages/src/lib.rs
SobhanMP: I see, makes sense.

crates/rattler_virtual_packages/src/lib.rs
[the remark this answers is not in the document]
SobhanMP: I think it doesn't hurt to expose this, especially useful when `current` is bugged.

SobhanMP:
@baszalmstra, I did the renaming and the deprecation. I think that's everything. Apologies if I missed something.

right now:
- `current` is the old `current`
- `detect` is detect with defaults
- `VirtualPackageOverrides.__init__` replaces `VirtualPackageOverrides.default`
- Some documentation on the python side
- The functions are `detect_with_overrides` and `detect_with_fallback`.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -130,11 +176,21 @@ impl From<VirtualPackage> for GenericVirtualPackage {
...
-        DETECTED_VIRTUAL_PACKAGES
-            .get_or_try_init(try_detect_virtual_packages)
-            .map(Vec::as_slice)
+    #[deprecated(since = "1.0.4", note = "Use `Self::detect` instead")]
+    pub fn current() -> Result<Vec<Self>, DetectVirtualPackageError> {
+        try_detect_virtual_packages_with_overrides(&VirtualPackageOverrides::none())
+    }
+
+    /// Detect the virtual packages of the current system with the default overrides.
+    pub fn detect() -> Result<Vec<Self>, DetectVirtualPackageError> {
```
baszalmstra, line 185: Lets just remove this function in favor of `detect_with_overrides`. I dont think we need an overload without an argument.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -150,10 +206,39 @@ pub enum DetectVirtualPackageError {
...
     #[error(transparent)]
     DetectLibC(#[from] DetectLibCError),
+
+    #[error(transparent)]
+    VarError(#[from] env::VarError),
+
+    #[error(transparent)]
+    VersionParseError(#[from] ParseVersionError),
+}
+/// Configure the overrides used in in this crate.
```
baszalmstra, line 216:
```suggestion
/// Configure the overrides used in this crate.
```

py-rattler/tests/unit/test_override.py
```diff
@@ -0,0 +1,28 @@
+from rattler import VirtualPackage, VirtualPackageOverrides, Override, Version, PackageName
+
+def test_stuff() -> None:
```
baszalmstra, line 3: Maybe a little more descriptive name 😄

py-rattler/rattler/virtual_package/virtual_package.py
```diff
@@ -1,11 +1,163 @@
...
+    @classmethod
+    def _from_py_virtual_package_overrides(
+        cls, py_virtual_package_overrides: PyVirtualPackageOverrides
+    ) -> VirtualPackageOverrides:
+        """Construct Rattler VirtualPackageOverrides from FFI PyVirtualPackageOverrides object."""
+        virtual_package_overrides = cls.__new__(cls)
+        virtual_package_overrides._overrides = py_virtual_package_overrides
+        return virtual_package_overrides
+
+    def __init__(self) -> None:
```
baszalmstra, line 90: Can you add named arguments with defaults for osx, cuda, etc?

py-rattler/rattler/virtual_package/virtual_package.py
```diff
@@ -22,8 +174,27 @@ def current() -> List[VirtualPackage]:
...
+    @staticmethod
+    def detect() -> List[VirtualPackage]:
+        """
+        Returns virtual packages detected for the current system or an error
+        if the versions could not be properly detected.
+        """
+        return [VirtualPackage._from_py_virtual_package(vp) for vp in PyVirtualPackage.detect()]
+
+    @staticmethod
+    def detect_with_overrides(overrides: VirtualPackageOverrides) -> List[VirtualPackage]:
```
baszalmstra, line 189: Same here, we can just have one function with a default argument (or None)

crates/rattler_virtual_packages/src/lib.rs
SobhanMP: I think
`let virtual_packages = rattler_virtual_packages::VirtualPackage::detect_with_overrides(&rattler_virtual_packages::VirtualPackageOverrides::default())?;`
is not very concise :sweat_smile:

baszalmstra: `VirtualPackage::detect(&Default::default())` seems fine to me.

SobhanMP: hmm, yeah, that works. Somehow, `Default` is more robust than what I imagined.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -130,11 +176,19 @@ impl From<VirtualPackage> for GenericVirtualPackage {
...
     /// Returns virtual packages detected for the current system or an error if the versions could
     /// not be properly detected.
-    pub fn current() -> Result<&'static [Self], DetectVirtualPackageError> {
-        static DETECTED_VIRTUAL_PACKAGES: OnceCell<Vec<VirtualPackage>> = OnceCell::new();
-        DETECTED_VIRTUAL_PACKAGES
-            .get_or_try_init(try_detect_virtual_packages)
-            .map(Vec::as_slice)
+    #[deprecated(
+        since = "1.0.4",
+        note = "Use `Self::detect(&Default::default)` instead"
```
baszalmstra, line 181: mm I guess this should be none right?

py-rattler/rattler/virtual_package/virtual_package.py
```diff
@@ -22,7 +181,19 @@ def current() -> List[VirtualPackage]:
         Returns virtual packages detected for the current system or an error
         if the versions could not be properly detected.
         """
-        return [VirtualPackage._from_py_virtual_package(vp) for vp in PyVirtualPackage.current()]
+        warnings.warn("Use `detect` instead")
+        return VirtualPackage.detect_with_overrides(VirtualPackageOverrides.none())
+
+    @staticmethod
+    def detect_with_overrides(overrides: VirtualPackageOverrides | None = None) -> List[VirtualPackage]:
```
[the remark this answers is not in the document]
SobhanMP: I think this is outdated right?

crates/rattler_virtual_packages/src/lib.rs
SobhanMP: I'd argue no, because the expectation is to be conda compatible :shrug:

baszalmstra: I disagree, rattler is also used in applications where this outside environment override behavior is unwanted. For an application this behavior makes sense but for a library I think reading from environment variables should be opt in.

baszalmstra: Furthermore, this suggestion changes the default current behavior of the function. I think the depreciation should tell people how to migrate to something which exhibits the same behavior.

py-rattler/rattler/virtual_package/virtual_package.py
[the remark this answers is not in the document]
baszalmstra: No I think we should just have one function called detect which you can pass an override or not. I think its superfluous to also add with_override to the name.

baszalmstra (approved):

baszalmstra:
Thanks! Sorry it took so long!

SobhanMP:
No worries, thanks for all the helpful comments!

Issue #205, opened 2023-06-06
Test: Check invalidation of matchspecs that are formatted incorrect


ruben-arts:
We're looking to bolster the robustness of our Rattler's matchspec implementation, particularly in the area of input parsing. While we currently have some testing mechanisms in place, we'd like to ensure we are fully covered, especially with respect to certain error scenarios.

Two primary errors that need to be checked for are:
[StringMatcherParseError](https://github.com/mamba-org/rattler/blob/main/crates/rattler_conda_types/src/match_spec/matcher.rs#L45)
[ParseMatchSpecError](https://github.com/mamba-org/rattler/blob/main/crates/rattler_conda_types/src/match_spec/parse.rs#L23)

We recommend taking a look at Conda's tests for some inspiration:
https://github.com/conda/conda/blob/9e8425844a28ffad0c4a3adcf28a2e769f965947/tests/models/test_match_spec.py

This is a fantastic opportunity for anyone looking to make their first contribution, as this issue is primarily about adding more tests. We welcome incremental progress, so don't feel pressured to craft a massive Pull Request - a single test at a time is perfectly fine!

If you have any questions or need any assistance, feel free to ask here or join our conversation on  [discord](https://discord.gg/kKV8ZxyzY4).  We're excited to see your contributions!

Ehtesham-Zahid:
Hi maintainers,
I see that some work has already been done here, but it looks like the issue is still open.
I'd love to contribute by adding more tests for StringMatcherParseError and ParseMatchSpecError.
Could you please assign this to me if it's still open?
Thanks a lot!

baszalmstra:
Yeah, I apologize, I think this has already been fixed.  Your welcome to tackle other issues! Or even just improve rubustness of existing tests.

Ehtesham-Zahid:
> Yeah, I apologize, I think this has already been fixed. Your welcome to tackle other issues! Or even just improve rubustness of existing tests.

Thanks for clarifying!
I'll explore ways to improve the tests.

PR #839, opened 2024-08-30
feat: add edge case tests for `StringMatcher`

kelvinou01:
### Description

- Adds some edge case testing for `StringMatcher` (not exactly https://github.com/conda/rattler/issues/205, but related)


baszalmstra (approved):
Thanks!

PR #840, opened 2024-08-31
feat: add `rattler_menuinst` crate

wolfv:
TODO:

- [x] run `precreate` code if specified in the menuinst
- [x] fix the XML editing for Linux!
- [x] Windows activation running
- [x] add method that takes a prefix record, and installs or removes menus based on that (and writes new prefix record)

Serialization

- [x] use internally tagged enum for prefix record serialization
- [ ] decide on writing to LinuxXML menu categories (and subsequent removal)

Testing

- [ ] create a terminal_profile test package for windows
- [ ] create a activation test package for windows, mac and linux (eg. using bash)
- [ ] create a test for URL handlers on Windows
- [ ] create a test for file association on Windows
- [ ] create a test for pre-create code

Some notes:

- on linux, there are no icons that are installed for mime types. I think it could be done relatively easily with something like https://stackoverflow.com/questions/30931/register-file-extensions-mime-types-in-linux
- we call xdg-mime to install the xml file. This automatically also updates the mime database. ~~But it makes it harder to test, because that installs to user directory~~ we should test by overriding XDG_DATA_DIRECTORY and XDG_CONFIG_DIRECTORY 💡


wolfv:
Apparently the name can be a dict: 

```
            "name": {
                "target_environment_is_base": "superspyder 1.2 ({{ DISTRIBUTION_NAME }})",
                "target_environment_is_not_base": "superspyder 1.2 ({{ ENV_NAME }})"
            },
```

Hofer-Julian:
Linux is now *nearly* working.

Running `cargo run -- create napari-menu` gives the following desktop file in `.local/share/applications/napari (0.5.4).desktop`

```
[Desktop Entry]
Type=Application
Encoding=UTF-8
Name=napari (0.5.4)
Exec=unset PYTHONHOME && unset PYTHONPATH && /var/home/julian/Projekte/github.com/conda/rattler/.prefix/bin/python -m napari
Terminal=false
Icon=/var/home/julian/Projekte/github.com/conda/rattler/.prefix/Menu/napari.png
Comment=a fast n-dimensional image viewer in Python
Categories=Graphics;Science;
StartupWMClass=napari
```

`Exec` doesn't support complicated expressions like this one. We should extract that to a script or trampoline.

Hofer-Julian:
We went with directly calling the quoted command with `bash -c` instead.

One thing that still needs to be investigated if we need to run `update-desktop-database` after installing, my system gives unclear results. We could run it just to be sure. Mimetypes would be also great to have.

dhirschfeld:
Cool! Super excited about this for `pixi global install`-ing applications

dhirschfeld:
:tada:

crates/rattler_menuinst/src/lib.rs
```diff
@@ -0,0 +1,150 @@
...
+                macos::remove_menu_item(tracker).unwrap();
+            }
+            Tracker::Linux(tracker) => {
+                #[cfg(target_os = "linux")]
+                linux::remove_menu_item(tracker).unwrap();
+            }
+            Tracker::Windows(tracker) => {
+                #[cfg(target_os = "windows")]
+                windows::remove_menu_item(tracker)?;
+            }
```
ruben-arts, line 135: I think you want to just return one error here instead of an unwrap

wolfv: Good point! I think there is one more unwrap in the windows specific code that I marked with a TODO.

ruben-arts: Let's align the return type with all the functions, and then also make this function return the removed paths. This is useful in pixi for logging.

wolfv: It's not just paths, though, it's also registry entries on Windows, for example, or entries in the terminal profile configuration file.

Might be more useful to just render the `tracker` information in pixi.

crates/rattler_menuinst/data/menuinst.default.json
```diff
@@ -0,0 +1,68 @@
+{
+  "id_": "https://schemas.conda.io/menuinst-1.schema.json",
```
Hofer-Julian, line 2: This link seems to be broken

Hofer-Julian: Also, why is this file needed? :)

crates/rattler_menuinst/data/menuinst.schema.json
```diff
@@ -0,0 +1,731 @@
+{
```
Hofer-Julian, line 1: What will this schema be used for?

crates/rattler_menuinst/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+    XmlError(#[from] quick_xml::Error),
+
+    #[cfg(target_os = "windows")]
+    #[error("Failed to install menu item: {0}")]
+    WindowsError(#[from] ::windows::core::Error),
+
+    #[cfg(target_os = "linux")]
+    #[error("Menu config location is not a file: {0:?}")]
+    MenuConfigNotAFile(PathBuf),
+}
```
Hofer-Julian: > The error message should be lowercase

From https://rust-lang.github.io/api-guidelines/interoperability.html?highlight=lower#error-types-are-meaningful-and-well-behaved-c-good-err

Hofer-Julian (commented):
A small set of first comments :)

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -821,4 +821,9 @@ mod tests {
             MatchSpec::from_nameless(NamelessMatchSpec::from_str(">=12", Strict).unwrap(), None);
         assert!(!spec.is_virtual());
     }
+
+    #[test]
+    fn test_size_stays_the_same() {
+        assert_eq!(std::mem::size_of::<MatchSpec>(), 488);
+    }
```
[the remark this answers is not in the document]
wolfv: reverted. Purpose is to make sure to not make the type too big, but I wasn't the one to remove the test in the first place.

crates/rattler_menuinst/data/menuinst.schema.json
wolfv: I removed it.

crates/rattler_menuinst/data/menuinst.default.json
wolfv: We test that we can parse the defaults.json from upstream menuinst: https://github.com/conda/menuinst/blob/main/menuinst/data/menuinst-1-1-0.default.json

crates/rattler_menuinst/src/lib.rs
wolfv: fixed

crates/rattler_menuinst/data/menuinst.default.json
Hofer-Julian: I can't seem to find that test. Can you point me to where it is? :)

wolfv:
```
    #[test]
    fn test_deserialize_defaults() {
        let test_data = test_data();
        let schema_path = test_data.join("defaults/defaults.json");
        let schema_str = std::fs::read_to_string(schema_path).unwrap();
        let schema: super::MenuInstSchema = serde_json::from_str(&schema_str).unwrap();
        insta::assert_debug_snapshot!(schema);
    }

```

Hofer-Julian: That uses `test-data/defaults/defaults.json`, not `data/menuinst.default.json`

PR #841, opened 2024-08-31
chore: apply clippy's suggetions

SobhanMP:
I think it's sometimes helpful to use Clippy to avoid common mistakes. Unfortunately, this repo generates way too many of these.

The only one I could not figure out was 

```
#[allow(clippy::ptr_arg)]
fn get_bytes_offset(lines: &Vec<&str>)
```

This commit is not supposed to change any behavior whatsoever.

### DIFF
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -51,7 +51,7 @@ jobs:
       - name: Run rustfmt
         uses: actions-rust-lang/rustfmt@v1
       - name: Run clippy
-        run: cargo clippy --all-targets -- -D warnings
+        run: cargo clippy --all-targets
   build:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -51,8 +51,7 @@ jobs:
       - name: Run rustfmt
         uses: actions-rust-lang/rustfmt@v1
       - name: Run clippy
-        run: cargo clippy --all-targets
-
+        run: cargo clippy --all-targets -- -D warnings
   build:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -52,6 +52,7 @@ jobs:
         uses: actions-rust-lang/rustfmt@v1
       - name: Run clippy
         run: cargo clippy --all-targets
+
   build:
     name: ${{ matrix.name }}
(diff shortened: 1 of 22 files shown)


baszalmstra:
Thats strange since we run clippy in CI! Would you be able to figure out why these werent caught and adapt the workflow so we catch these things earlier in the future?

SobhanMP:
I think that precommit only runs clippy on changed files

SobhanMP:
but then precommit doesn't run clippy https://github.com/conda/rattler/blob/8a9fed3044a55b3d639f12d3f0db2f3d89bb1294/.github/workflows/pre-commit.yaml#L21

baszalmstra:
We run clippy here:

https://github.com/conda/rattler/blob/8a9fed3044a55b3d639f12d3f0db2f3d89bb1294/.github/workflows/rust-compile.yml#L53

SobhanMP:
yeah, that's what i realized

SobhanMP:
i think we need to run it with `cargo clippy -- -D warnings` so that the job fails on warnings

baszalmstra:
We are because of:

https://github.com/conda/rattler/blob/8a9fed3044a55b3d639f12d3f0db2f3d89bb1294/.github/workflows/rust-compile.yml#L23

baszalmstra:
Which version of rust are you using?

We are still using 1.75.0 which is pinned through the rust-toolchain file. Are you perhaps using a higher version?

SobhanMP:
```
rustc --version
rustc 1.80.1 (3f5fd8dd4 2024-08-06)
```

Ah

baszalmstra:
I assume you are not using pixi or rustup? How did you install the rust compiler?

SobhanMP:
rustup?

SobhanMP:
ps. https://releases.rs/ show that 1.80 is stable

SobhanMP:
maybe it's a good idea to bump https://github.com/conda/rattler/blob/c02c9926318867fc61f25cb247588ae24263f9cb/rust-toolchain?

baszalmstra:
> maybe it's a good idea to bump [c02c992/rust-toolchain](https://github.com/conda/rattler/blob/c02c9926318867fc61f25cb247588ae24263f9cb/rust-toolchain)? or not setting it so it always stays at stable

We want to pin the rust compiler version because otherwise every time a new version comes out we have a lot of clippy warnings (exactly what you ran into). Bumping it is a good idea, this is already happening in #837

baszalmstra:
#837 has been merged which bumps the rust toolchain to 1.80.0. This introduces a number of conflicts. Can you check if there are still missing clippy warnings?

SobhanMP:
Yep, every thing looks good!

Issue #43, opened 2023-02-07
Implement layered package caches


baszalmstra:
Conda and Mamba both support having multiple cache directories. Rattler currently only supports a single directory. The "layered" cache can have multiple readable directories and only one writable directory. Conda/Mamba tries to [write a magical file to all these directories at startup to determine which one is writable](https://github.com/mamba-org/mamba/blob/31e5f861b91ca9819e92ee899e894a31f318f836/libmamba/src/core/package_cache.cpp#LL64C36-L64C36). 

Rattler should also facilitate "layered" package caches. We could introduce a trait for `PackageCache` that is implemented for both the current implementation as well as a layered version.

gzm55:
any updates?

baszalmstra:
This is not being worked on atm. Contributions are welcome!

kelvinou01:
@baszalmstra hey, I'm trying to work out a PR for this. Do we need to enforce the user to have strictly one writable cache directory? 

1.  Looking at [conda's config class docs](https://github.com/conda/conda/blob/62d23648319e743f6248093d19647c936ed84e83/conda/base/context.py#L1657): 

```python
pkgs_dirs=dals(
    """
    The list of directories where locally-available packages are linked from at
    install time. Packages not locally available are downloaded and extracted
    into the first writable directory.
    """
)
```

Seems like multiple writable package caches are allowed. [Packages are both downloaded and extracted to the first writable cache](https://github.com/conda/conda/blob/62d23648319e743f6248093d19647c936ed84e83/conda/core/package_cache_data.py#L625)

2. I think it makes sense for Rattler to allow multiple writable caches. 

Consider the scenario where you have a multi-user linux machine with a system-wide rattler package cache.  This system-wide cache is read-only to users, and only writable by the system admin. e.g. normal user `alice` has RW access to /home/alice/.cache/rattler/cache, but only R access to /.cache/rattler/cache (I think where this system-wide cache dir is located is arbitrary?). 

If a system admin `bob` has their own local cache dir (e.g. /home/bob/.cache/rattler/cache), `bob` would now have two writable caches. Enforcing strictly one writable cache would break this.

wolfv:
I think a user could have more than one writable cache, but we should probably write to the "first" one. If we would have a nice API to list all available package caches, and those that are writable, it would be nice to be able to show that to users (e.g. with `pixi info`).

@baszalmstra is also currently working on much improving the package cache concurrency.

PR #843, opened 2024-09-03
refactor: put PackageCache into an enum in preparation for a multi-layered package cache

kelvinou01:
### Description

As per https://github.com/conda/rattler/issues/43, we want to implement layered package caches for Rattler. Putting PacakgeCache into an enum allows switching between implementations (`SingletonPackageCache` and `LayeredPackageCache`)

I'm considering both traits and enums. Leaning towards enum because:
- My impression is that we don't allow downstream users to add their own PackageCache types, thus a closed set of types (enum) is more suitable. 
- Enums are more performant

Stuff to note
- This will be a breaking change
- To let PackageCache be public, I had to make PackageCacheInner public as well (even though `inner` is a private struct field), not sure if there's a way not to.

### DIFF
--- a/crates/rattler-bin/Cargo.toml
+++ b/crates/rattler-bin/Cargo.toml
@@ -27,13 +27,13 @@ clap = { workspace = true, features = ["derive"] }
 console = { workspace = true, features = ["windows-console-colors"] }
 indicatif = { workspace = true }
 once_cell = { workspace = true }
-rattler = { path="../rattler", version = "0.27.6", default-features = false, features = ["indicatif"] }
-rattler_conda_types = { path="../rattler_conda_types", version = "0.27.2", default-features = false }
+rattler = { path="../rattler", version = "0.27.7", default-features = false, features = ["indicatif"] }
+rattler_conda_types = { path="../rattler_conda_types", version = "0.27.3", default-features = false }
 rattler_networking = { path="../rattler_networking", version = "0.21.2", default-features = false }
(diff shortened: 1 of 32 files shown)


baszalmstra:
Hey! Thanks for making this PR! Im wondering what your design considerations are. Why have two separate implementations? Isnt a non-layered cache just cache with one layer?

kelvinou01:
> Isnt a non-layered cache just cache with one layer?

Hey @baszalmstra  i was actually referencing this

> We could introduce a trait for PackageCache that is implemented for both the current implementation as well as a layered version.

Maybe we can allow users to pick between the singleton and layered implementations, while we polish the latter? Or we can just create a layered implementation and have it be shipped to users (with the default being a cache with one layer)

baszalmstra:
> Maybe we can allow users to pick between the singleton and layered implementations, while we polish the latter? Or should we just create a layered implementation and have it be shipped to users (with the default being a cache with one layer)

Yeah, I think (in hindsight) that makes the most sense.

FYI I'm also about to merge a relatively large refactor of the PackageCache in `https://github.com/conda/rattler/pull/837`.

kelvinou01:
Thanks for the heads up on the PR! 

Also just to be sure, which idea did you think made the most sense, using the enum or not using it?

baszalmstra:
> Also just to be sure, which idea did you think made the most sense, using the enum or not using it?

I think a single implementation makes more sense. But if you feel uncertain about the implementation using an enum is fine for me as well.

baszalmstra:
#837 was just merged, you might have a few conflicts.

PR #845, opened 2024-09-03
fix: allow `gcs://` and `oci://` in gateway

baszalmstra:
Fixes https://github.com/prefix-dev/pixi/issues/1922


baszalmstra:
@clement-chaneching Would you be able to verify if this works for you using something like?

```
rattler create --dry-run -c gcs://... <some package>
```

PR #850, opened 2024-09-05
fix: typo in py-rattler

Hofer-Julian:


baszalmstra:
I dont think this change is breaking, I think that type is only used internally.

Hofer-Julian:
> I dont think this change is breaking, I think that type is only used internally.

Even better :)

Will change the PR description and title accordingly

wolfv (approved):

PR #855, opened 2024-09-07
fix: publish MatchSpecOrSubSection for env yaml

abkfenris:
### Description

With the addition of `EnvironmentYaml` to rattler, I'm taking another swing at adding exporting from Pixi (replacing https://github.com/prefix-dev/pixi/pull/1427 ), and using `EnvironmentYaml` to reuse code.

Currently the `MatchSpecOrSubSection` enum isn't published at the library level, so dependencies can't be added from `MatchSpecs` or Pip subsections. This publishes the enum so that it can be used for constructing `EnvironmentYaml`s from other representations.


wolfv:
Awesome, thanks for the PR! :)

baszalmstra (approved):

baszalmstra:
Thanks! Looks good to me!

wolfv (approved):

PR #863, opened 2024-09-13
refactor: migrate from typed-path to camino

Hofer-Julian:


Hofer-Julian:
I don't confident anymore this is worth pursuing. The biggest annoyance is that `typed-path` has no serde support. For pixi, we might want to go for camino anyway.

baszalmstra:
Understandable.

Perhaps we could contribute serde support back?

Hofer-Julian:
> Understandable.
> 
> Perhaps we could contribute serde support back?

That would be good, yes

PR #860, opened 2024-09-12
refactor: smaller lockfiles, source dependencies and bug fixes

baszalmstra:
Significantly refactors the lock file format to reduce the size of the lock-file and support source dependencies. 

#### File size reduction

The size of the lock file is decreased by removing some fields that can be derived from the location of the package:

```yaml
- kind: conda
  name: bzip2
  version: 1.0.8
  build: h93a5062_5
  build_number: 5
  subdir: osx-arm64
  url: https://conda.anaconda.org/conda-forge/osx-arm64/bzip2-1.0.8-h93a5062_5.conda
  sha256: bfa84296a638bea78a8bb29abc493ee95f2a0218775642474a840411b950fe5f
  md5: 1bbc659ca658bfd49a481b5ef7a0f40f
  license: bzip2-1.0.6
  license_family: BSD
  size: 122325
  timestamp: 1699280294368
```

becomes:

```yaml
- conda: "https://conda.anaconda.org/conda-forge/osx-arm64/bzip2-1.0.8-h93a5062_5.conda"
  sha256: bfa84296a638bea78a8bb29abc493ee95f2a0218775642474a840411b950fe5f
  md5: 1bbc659ca658bfd49a481b5ef7a0f40f
  license: bzip2-1.0.6
  license_family: BSD
  size: 122325
  timestamp: 1699280294368
```

Since this is a breaking change the lock-file format has increased to `6`.

#### Source packages

Source dependencies can now also be specified. This changes the location of the package and multiple environments can use different variants from the same source package. To differentiate between those environments also get the ability to add selection parameters to the packages they reference:

```yaml
version: 6
environments:
  default:
    channels:
      - url: "https://conda.anaconda.org/conda-forge/"
    packages:
      linux-64:
        - conda: "../minimal-project"
          subdir: linux-64
      win-64:
        - conda: "../minimal-project"
          subdir: win-64
packages:
  - conda: "../minimal-project"
    name: minimal-project
    version: "0.1"
    build: first
    subdir: linux-64
    depends:
      - __unix
  - conda: "../minimal-project"
    name: minimal-project
    version: "0.1"
    build: first
    subdir: win-64
    depends:
      - __win
```

The serialization code automatically selects the fewest selects to disambiguate different packages. 

#### Roundtrip tests

While debugging I also added a test to roundtrip serialize lock-files. I discovered a lot of bugs in the old code which I patched up to the best of my abilities.

This PR also includes the changes from #858 so Im closing that one in favor of this one.

PR #866, opened 2024-09-20
feat: lock file update for source dependencies

tdejager:
### Description
  This includes changes to the lock format, as well as making it a lot smaller.                                       
  That we need to include source dependencies in pixi


tdejager (commented):
LGTM!

wolfv:
Is this not the same as #860 ?

tdejager:
Indeed it is @baszalmstra told me to create it in this way, so we need to either close this or #860. cc @baszalmstra

Issue #869, opened 2024-09-20
Regression in fix for #793? `MatchSpec` parsing appears to fail when channel is passed as url 


cisaacstern:
### Checklist

- [X] I added a descriptive title
- [X] I searched open reports and couldn't find a duplicate

### What happened?

Using the latest `py-rattler==0.6.3` I am unable to parse `MatchSpec`s when the channel is passed as a url. In  https://github.com/conda/rattler/issues/793#issuecomment-2262269859 the following example is given (copying for visibility here):

<details>

```python
>>> from rattler import MatchSpec
>>> m = MatchSpec("https://conda.anaconda.org/conda-forge::python[version=3.9]")
>>> m
MatchSpec("conda-forge::python ==3.9")
>>> m.channel
Channel(name="conda-forge", base_url="https://conda.anaconda.org/conda-forge/")
```

</details>

However on my machine I cannot reproduce this. I have just done the following in an empty directory (on mac M1):

1. ✨ Create a new pixi project and add `py-rattler`
    ```console
    $ pixi init
    $ pixi add --manifest-path pixi.toml py-rattler
    ✔ Added py-rattler >=0.6.3,<0.7
    ```

2. 🔧 Add `test_matchspec.py` to the project directory (adapted from the previously-mentioned example in https://github.com/conda/rattler/issues/793#issuecomment-2262269859):

    ```python
    # test_matchspec.py
    
    from rattler import MatchSpec
    
    if __name__ == "__main__":
        with_url_channel = MatchSpec("https://conda.anaconda.org/conda-forge::python[version=3.9]")
        print(f"{with_url_channel = }")
        print(f"{with_url_channel.channel = }")
        with_named_channel = MatchSpec("conda-forge::python[version=3.9]")
        print(f"{with_named_channel = }")
        print(f"{with_named_channel.channel = }")
    ```

3. 🤔 Run the script. Note that the `with_url_channel` does not parse correctly (contrary to the linked example in #793):

    ```console
    $ pixi run --manifest-path pixi.toml -e default python test_matchspec.py
    with_url_channel = MatchSpec("*")
    with_url_channel.channel = None
    with_named_channel = MatchSpec("conda-forge::python ==3.9")
    with_named_channel.channel = Channel(name="conda-forge", base_url="https://conda.anaconda.org/conda-forge/")
    ```

For reference, here is the output of `pixi list` for this environment:

<details>

```console
$ pixi list --manifest-path pixi.toml -e default    
[...]

PR #870, opened 2024-09-20
test: add tests for py-rattler MatchSpec channel parsing

cisaacstern:
### Description

Tests for MatchSpec channel parsing as requested in https://github.com/conda/rattler/issues/869#issuecomment-2364170101.

Contrary to the observations in #869, both of these tests pass for me locally when running 

```console
$ cd py-rattler
$ pixi run --manifest-path pixi.toml -e test test
```

Which perhaps means the url parsing regressing is solved in the current `main`, but broken in the `0.6.3.` release?

### DIFF
--- a/py-rattler/tests/unit/test_matchspec.py
+++ b/py-rattler/tests/unit/test_matchspec.py
@@ -1,0 +1,13 @@
+from rattler import MatchSpec
+
+
+def test_parse_channel_from_canonical_name():
+    m = MatchSpec("conda-forge::python[version=3.9]")
+    assert m.channel.name == "conda-forge"
+    assert m.channel.base_url == "https://conda.anaconda.org/conda-forge/"
+
+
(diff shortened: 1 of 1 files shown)


baszalmstra:
These have indeed been fixed, but we have to cut a new release. Ill try to do that today.

baszalmstra:
I just released 0.7.0 which should include all these fixes.

cisaacstern:
Thank you so much for the prompt fix!

> I just released 0.7.0 which should include all these fixes.

Am I correct in guessing that this release will become available on conda-forge following merge of #878 ?

wolfv:
Not quite - it will be on conda-forge once this is built and uploaded, and propagated through the CDN: https://github.com/conda-forge/py-rattler-feedstock

In ~1 hour or so it should be ready! :)

PR #871, opened 2024-09-21
fix: normalize channel in lock file

baszalmstra:
Normalizes channels in lock-files to never contain a trailing slash.


tdejager (approved):
Nice!

PR #879, opened 2024-09-30
fix: topological sort when cycles appear in leaf nodes

wolfv:
There were two issues with our topological sort implementation:

- cycles between two noarch packages or two non-noarch packages were never "broken". This is fixed by this PR (going to add a test next)
- When the root node (or leaf node) contained a cycle, it was not part of the result. The code is changed to first detect _any and all_ cycles in the graph, and then find the root nodes with these cycles in mind.

The changes came up because a user was installing `holoviews` which has a circular dependency on `panel`. It failed later in the un-clobbering (because some packages were just missing after the topological sort because of this faulty "root" detection).


crates/rattler_conda_types/src/repo_data/topological_sort.rs
```diff
@@ -158,10 +133,12 @@ fn break_cycles<T: AsRef<PackageRecord>>(
             // prefer arch packages over noarch packages
             let p1_noarch = p1.as_ref().noarch.is_none();
             let p2_noarch = p2.as_ref().noarch.is_none();
+
             if p1_noarch && !p2_noarch {
                 cycle_breaks.insert((pi1.clone(), pi2.clone()));
                 break;
-            } else if !p1_noarch && p2_noarch {
+            } else if !p1_noarch && p2_noarch || i == cycle.len() - 1 {
```
wolfv, line 140: Note the added condition here to also return a cycle break if none of the packages is noarch and we're in the last iteration.

baszalmstra (approved):

Issue #884, opened 2024-10-01
`MatchSpec.channel.base_url` always returns conda forge URL (even if `._match_spec.channel` has another url)


cisaacstern:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

As discussed in #869, in `py-rattler >=0.7.0`, the following parsing is possible:

https://github.com/conda/rattler/blob/8e86a8aa09b95ec2a08b6921b6d41b4ac01b02ef/py-rattler/tests/unit/test_matchspec.py#L11-L15

I subsequently found, however, that neither of the following cases work as expected:

```python
def test_parse_channel_from_url_filesystem() -> None:
    m = MatchSpec("file:///Users/rattler/channel0::python[version=3.9]")
    assert m.channel is not None
    assert m.channel.name == "channel0"
    assert m.channel.base_url == "file:///Users/rattler/channel0/"


def test_parse_channel_from_url_localhost() -> None:
    m = MatchSpec("http://localhost:8000/channel0::python[version=3.9]")
    assert m.channel is not None
    assert m.channel.name == "channel0"
    assert m.channel.base_url == "http://localhost:8000/channel0/"
```

In both cases `m.channel.base_url` is incorrectly returned as `"https://conda.anaconda.org/conda-forge/"`, when IIUC it should in fact be, `"file:///Users/rattler/channel0/"` and `"http://localhost:8000/channel0/"`, respectively.

Notably, I found that the "private" container `._match_spec.channel` _does_ know the correct URL in both cases (or at least, a version of it), so the issue would seem not to be parsing per se, but rather how the parsed result is surfaced at the Python layer. Specifically, there is no code path for setting a `ChannelConfig` here:

https://github.com/conda/rattler/blob/8e86a8aa09b95ec2a08b6921b6d41b4ac01b02ef/py-rattler/rattler/match_spec/match_spec.py#L145

so the [default of conda-forge](https://github.com/conda/rattler/blob/8e86a8aa09b95ec2a08b6921b6d41b4ac01b02ef/py-rattler/rattler/channel/channel_config.py#L7) is used in all cases.

I have a PR that adds these tests and a (reference) fix. I will open that now. Thanks all!

_**Edit**_: linked PR opened here [test: return correct channel urls from MatchSpec.channel.base_url #885](https://github.com/conda/rattler/pull/885)

PR #885, opened 2024-10-01
fix: return correct channel urls from `MatchSpec.channel.base_url`

cisaacstern:
### Description

This PR adds the tests described in #884.

It also provides a (reference) fix in `MatchSpec.channel`. I say "reference" because I imagine the maintainers may prefer to handle this in some other way (perhaps by pushing the url parsing I've done in Python here to the Rust layer?), in which case I will certainly have no objection to my "fix" being discarded. But I provide it nonetheless to indicate the general shape of a possible solution.

Thanks all! Happy to contribute to this further if I am able.

### DIFF
--- a/py-rattler/Cargo.lock
+++ b/py-rattler/Cargo.lock
@@ -2634,7 +2634,7 @@ dependencies = [
 
 [[package]]
 name = "rattler"
-version = "0.27.12"
+version = "0.27.13"
 dependencies = [
  "anyhow",
  "console",
@@ -2747,7 +2747,7 @@ dependencies = [
 
 [[package]]
 name = "rattler_index"
-version = "0.19.29"
+version = "0.19.30"
 dependencies = [
  "fs-err",
  "rattler_conda_types",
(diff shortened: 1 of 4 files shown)


wolfv:
Thanks for the fix! It looks pretty good to me, but I am wondering if there might be an easier way. @baszalmstra recently refactored some parts of this so on the Rust side we have distinct types for channels from (relative) Path, URL, or file URL. Maybe we can expose that in a clean way and have a function on the Rust side that returns the right thing. 

But you have clearly identified a nice bug!

I'll wait with merging until we hear from Bas.

cisaacstern:
> do you mind if I amend your PR?

Yes of course please do, however you see fit!

baszalmstra:
I modified you code slighty, the rust side already returned everything you needed we just had to convert it properly to python. Your tests seem to succeed with these changes. Thanks for adding those!

cisaacstern:
Amazing, tysm!

PR #887, opened 2024-10-02
fix: fix-up shebangs with spaces

wolfv:
This changes a little bit how we handle shebangs in text files.

When a shebang (after replacement) contains a `" "` space in the result we need to replace it with `/usr/bin/env ...` to make things work. This breaks execution in "non-activated" environments though. 

For Python scripts we can use another small trick: we can execute `/usr/bin/sh` and then run `exec "/path/to python with spaces/python" {args} ...` (which is also wrapped in a python comment). 

We are already doing this when we create Python entrypoints from scratch for noarch packages. I added this functionality here as well (duplicates code from `python.rs` though).

### DIFF
--- a/crates/rattler-bin/src/commands/create.rs
+++ b/crates/rattler-bin/src/commands/create.rs
@@ -91,6 +91,9 @@ pub async fn create(opt: Opt) -> anyhow::Result<()> {
     let target_prefix = opt
         .target_prefix
         .unwrap_or_else(|| current_dir.join(".prefix"));
+
+    // Make the target prefix absolute
+    let target_prefix = std::path::absolute(target_prefix)?;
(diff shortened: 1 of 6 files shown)


crates/rattler-bin/src/commands/create.rs
```diff
@@ -91,6 +91,9 @@ pub async fn create(opt: Opt) -> anyhow::Result<()> {
     let target_prefix = opt
         .target_prefix
         .unwrap_or_else(|| current_dir.join(".prefix"));
+
+    // Make the target prefix absolute
+    let target_prefix = std::path::absolute(target_prefix)?;
```
wolfv, line 96: This is somewhat unrelated but fixes an issue in the `rattler` binary.

baszalmstra (approved):

PR #892, opened 2024-10-04
feat: total ordering for the dependency provider

tdejager:


crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -90,6 +90,16 @@ pub enum NamedChannelOrUrl {
...
 
+impl std::hash::Hash for NamedChannelOrUrl {
+    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
+        match self {
+            NamedChannelOrUrl::Name(name) => name.hash(state),
+            NamedChannelOrUrl::Url(url) => url.as_str().hash(state),
+            NamedChannelOrUrl::Path(path) => path.as_str().hash(state),
+        }
+    }
+}
```
baszalmstra, line 101: Why was this needed?

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -1127,6 +1130,10 @@ mod test {
             }
             previous = Some(version);
         }
+
+        // Try to see if the sort works
+        let mut cloned_versions = versions.clone();
+        cloned_versions.sort();
```
baszalmstra, line 1136: Is this still needed?

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -1397,4 +1404,87 @@ mod test {
...
+        //                     ord_ac != Ordering::Greater,
+        //                     "Transitivity violated between {:?}, {:?}, and {:?}",
+        //                     a,
+        //                     b,
+        //                     c
+        //                 );
+        //             }
+        //         }
+        //     }
+        // }
```
baszalmstra, line 1488: Can this be removed?

crates/rattler_solve/src/resolvo/conda_util.rs
```diff
@@ -1,17 +1,344 @@
...
+use super::SolverPackageRecord;
+
 #[derive(Copy, Clone, Debug, Eq, PartialEq)]
 pub(super) enum CompareStrategy {
     Default,
     LowestVersion,
 }
 
+/// Sorts the candidates based on the strategy.
+/// and some different rules
```
baszalmstra, line 22: Can you elaborate?

crates/rattler_solve/src/resolvo/conda_util.rs
```diff
@@ -1,17 +1,344 @@
...
+
+        // Otherwise, select the variant with the highest build number
+        match a_record.build_number().cmp(&b_record.build_number()) {
+            Ordering::Less => return Ordering::Greater,
+            Ordering::Greater => return Ordering::Less,
+            Ordering::Equal => return Ordering::Equal,
+        };
+    }
+
+    pub fn sort_by_dependencies(
```
baszalmstra, line 83: The API of this function is maybe slight weird because it assumes that the input is already sorted by `sort_by_name_version_build`. Would it make sense to have one `pub` function in this `impl` that calls both function in succession and make the other functions not-pub? Or have `sort_by_name_version_build` return a wrapper type that guarentees the sorted properties that is consumed by this function?

crates/rattler_solve/src/resolvo/conda_util.rs
```diff
@@ -1,17 +1,344 @@
...
+    /// Sorts the solvables by the highest version of the dependencies shared by the solvables.
+    /// what this function does is:
+    /// 1. Find the first unsorted solvable in the list
+    /// 2. Get the dependencies for each solvable
+    /// 3. Get the known dependencies for each solvable, filter out the unknown dependencies
+    /// 4. Retain the dependencies that are shared by all the solvables
+    /// 5. Create a max vector which is the maximum version of each of the shared dependencies, per dependency name
+    /// 6. Calculate a total score  by counting the position of the solvable in the list with sorted dependencies
+    /// 7. Sort by the total score and use timestamp of the record as a tie breaker
+    fn sort_by_highest_version(
```
baszalmstra, line 120: Maybe we should rename this function `sort_by_highest_dependency_versions`?

crates/rattler_solve/src/resolvo/conda_util.rs
```diff
@@ -1,17 +1,344 @@
...
+            max_solvables.sort_by(|a, b| a.version.cmp(&b.version));
+            // dbg!(max_solvables);
+        }
+
+        Self { max_map }
+    }
+
+    /// Per dependency, score the solvables based on the highest version of the dependency
+    fn score_solvables(&self) -> HashMap<SolvableId, u32> {
+        let mut scores = HashMap::new();
```
baszalmstra, line 302: Maybe we can preallocate all the entries with `with_capacity`?

crates/rattler_solve/src/resolvo/conda_util.rs
```diff
@@ -1,17 +1,344 @@
...
+fn unique_name_ids<'a>(vectors: impl IntoIterator<Item = HashSet<NameId>>) -> HashSet<NameId> {
+    let mut iter = vectors.into_iter();
+    if let Some(first_set) = iter.next() {
+        iter.fold(first_set.clone(), |mut acc: HashSet<NameId>, set| {
+            acc.retain(|item| set.contains(item));
+            acc
+        })
+    } else {
+        HashSet::new() // Return empty set if input is empty
+    }
```
baszalmstra, line 340:
```suggestion
   iter
   	.into_iter()
 	.reduce(|mut acc, hs| {
            acc.retain(|name| hs.contains(name));
            acc
        })
        .unwrap_or_default()
```

baszalmstra (commented):
Looks really good!

I feel like this would really benefit from a specific test but Im unsure how to do that without too much effort. @wolfv Maybe you have some thoughts?

wolfv:
Are there any two packages that would have resulted in a non-total order before? That might make a good test case then?

crates/rattler_conda_types/src/channel/mod.rs
tdejager: This was missing and got a compile error for the new rust version.

crates/rattler_conda_types/src/version/mod.rs
tdejager: I think we can remove it indeed.

crates/rattler_solve/src/resolvo/conda_util.rs
tdejager: Yes I'll refactor.

crates/rattler_solve/src/resolvo/conda_util.rs
tdejager: Yes will change this later.

crates/rattler_solve/src/resolvo/conda_util.rs
tdejager: Good one!

tdejager:
I could try making traits to get the required fields and make a minimal test like the one wolf suggested. But that would delay the merge a bit. Let me know what you think!

wolfv:
It's not possible to make a vector and call sort on it in a test?

tdejager:
> It's not possible to make a vector and call sort on it in a test?

I think you might think it was one of the trait implementations, which was not the problem. It needs to touch the code in the sorter functions.

Note that the code that was panicking begore is still being run (and not panicking now). This is essentially what you are proposing, albeit not in minimal form.

crates/rattler_solve/src/resolvo/conda_util.rs
tdejager: Done!

crates/rattler_conda_types/src/channel/mod.rs
baszalmstra: But wasnt the derive good enough?

crates/rattler_conda_types/src/utils/serde.rs
```diff
@@ -144,60 +140,60 @@ impl SerializeAs<chrono::DateTime<chrono::Utc>> for Timestamp {
...
-    fn serialize_as<S>(source: &Vec<T>, serializer: S) -> Result<S::Ok, S::Error>
-    where
-        S: Serializer,
-    {
-        let mut elements = source.iter().collect_vec();
-        elements.sort();
-        SerializeAsWrap::<Vec<&T>, Vec<&TAs>>::new(&elements).serialize(serializer)
-    }
-}
+// #[derive(Default)]
```
baszalmstra, line 143: Can this code be removed?

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -1079,18 +1079,21 @@ mod test {
             " < 2!0.4.1", // epoch increased again
         ];
 
-        let ops = versions.iter().map(|&v| {
-            let (op, version) = if let Some((op, version)) = v.trim().split_once(' ') {
+        let mut versions = Vec::new();
```
baszalmstra, line 1082: This seems to be unused now.

tdejager:
I've made most of the changes, I'm not sure if the added complexity of the traits would be worth it. I could work on that if that is needed in a seperate PR that would not block this change, and the homebrew error. If we can think of an easy integration test to add, I could add that for sure

crates/rattler_solve/src/resolvo/conda_sorting.rs
```diff
@@ -0,0 +1,426 @@
...
+
+    /// Sorts the solvables by the highest version of the dependencies shared by the solvables.
+    /// what this function does is:
+    /// 1. Find the first unsorted solvable in the list
+    /// 2. Get the dependencies for each solvable
+    /// 3. Get the known dependencies for each solvable, filter out the unknown dependencies
+    /// 4. Retain the dependencies that are shared by all the solvables
+    /// 6. Calculate a total score by counting the position of the solvable in the list with sorted dependencies
+    /// 7. Sort by the score per solvable and use timestamp of the record as a tie breaker
+    fn _sort_by_highest_dependency_versions(
```
baszalmstra, line 145: The underscore would suggest its unused. Maybe rename to sort_subset_..

crates/rattler_solve/src/resolvo/conda_sorting.rs
```diff
@@ -0,0 +1,426 @@
...
+                        v.push(max_solvable.clone());
+                    })
+                    .or_insert_with(|| vec![max_solvable]);
+            }
+        }
+
+        // Sort all vectors of dependencies by version
+        for max_solvables in max_map.values_mut() {
+            max_solvables.sort_by(|a, b| a.version.cmp(&b.version));
+            // dbg!(max_solvables);
```
baszalmstra, line 336: Remove maybe?

baszalmstra (commented):
Some small comments but the docs look great!

baszalmstra:
Perhaps we can make a simpler integration test by loading our sample repodata.json, sorting sprcific packages like libuuid and pytorch and creating some snapshots out of that?

crates/rattler_conda_types/src/channel/mod.rs
tdejager: I'm actually unsure how that derive could have ever worked because afaic `Utf8TypedPathBuf` does not derive Hash?

baszalmstra: It does implement Hash: https://docs.rs/typed-path/latest/typed_path/enum.Utf8TypedPathBuf.html#impl-Hash-for-Utf8TypedPathBuf

tdejager:
```
error[E0277]: the trait bound `Utf8TypedPathBuf: Hash` is not satisfied
  --> crates/rattler_conda_types/src/channel/mod.rs:90:10
   |
81 | #[derive(Debug, Clone, Eq, PartialEq, Hash)]
   |                                       ---- in this derive macro expansion
...
90 |     Path(Utf8TypedPathBuf),
   |          ^^^^^^^^^^^^^^^^ the trait `Hash` is not implemented for `Utf8TypedPathBuf`
```

Looking at the docs, https://docs.rs/camino/latest/camino/struct.Utf8PathBuf.html#impl-Hash-for-Utf8PathBuf it seems it might do so now, we could also bump?

crates/rattler_conda_types/src/utils/serde.rs
tdejager: It seemed to be unused, that was the error in CI. Or what do you mean :)?

tdejager:
> Perhaps we can make a simpler integration test by loading our sample repodata.json, sorting sprcific packages like libuuid and pytorch and creating some snapshots out of that?

how is that better than the comparison with libsolv that we are currently doing?

crates/rattler_solve/src/resolvo/conda_sorting.rs
tdejager: Oops got some python confusion there.

crates/rattler_conda_types/src/version/mod.rs
tdejager: Removed.

crates/rattler_solve/src/resolvo/conda_sorting.rs
tdejager: Removed.

crates/rattler_conda_types/src/channel/mod.rs
tdejager: Bumped to 0.9.2, 0.9.0 did not implement Hash.

crates/rattler_conda_types/src/utils/serde.rs
tdejager: Oh yeah I commented to see if it cause any failures, removed it because it seemed alright to do so now.

baszalmstra:
That is testing whether we solve similar which indeed requires sorting but doesnt necessarily hit these edge cases. Besides that I think it would be nice to have a test that highlights how we sort similar packages. Having a snapshot that shows how all pytorch variants are sorted for instance would be quite useful and help guide any future changes.

baszalmstra:
One more final thing: Could you also run the rattler_solve benchmarks to see if we didn't accidentally introduce a performance regression?

tdejager:
Thanks makes sense will do both. Let me make those test in main first in a separate PR, we can make sure the snapshots stay the same.

PR #894, opened 2024-10-07
add failing test for self-clobbering

wolfv:
@baszalmstra this is a test that currently fails and should be fixed by #893


wolfv:
Fixed by #893

PR #900, opened 2024-10-12
feat: use sha in cache lock file, needed for source builds.

tdejager:
### Description

This PR adds the ability in the cache to use the sha256 of a key as the cache lock file as well.
This is needed so that source build archives with the same paths but different contents are correctly replaced in the cache.

### DIFF
--- a/crates/rattler_cache/src/package_cache/cache_key.rs
+++ b/crates/rattler_cache/src/package_cache/cache_key.rs
@@ -8,10 +8,10 @@ use std::fmt::{Display, Formatter};
 /// TODO: Wouldn't it be better to cache based on hashes?
 #[derive(Debug, Hash, Clone, Eq, PartialEq)]
 pub struct CacheKey {
-    name: String,
-    version: String,
-    build_string: String,
-    sha256: Option<Sha256Hash>,
+    pub(crate) name: String,
+    pub(crate) version: String,
+    pub(crate) build_string: String,
+    pub(crate) sha256: Option<Sha256Hash>,
 }
 
 impl CacheKey {
--- a/crates/rattler_cache/src/package_cache/cache_key.rs
+++ b/crates/rattler_cache/src/package_cache/cache_key.rs
@@ -5,7 +5,6 @@ use std::fmt::{Display, Formatter};
 
 /// Provides a unique identifier for packages in the cache.
 /// TODO: This could not be unique over multiple subdir. How to handle?
-/// TODO: Wouldn't it be better to cache based on hashes?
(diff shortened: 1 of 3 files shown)


crates/rattler_cache/src/package_cache/cache_lock.rs
```diff
@@ -196,7 +221,14 @@ impl CacheRwLock {
 }
 
 impl CacheRwLock {
+    /// Reads the revision from the cache lock file.
     pub fn read_revision(&mut self) -> Result<u64, PackageCacheError> {
+        self.file.lock().rewind().map_err(|e| {
```
baszalmstra, line 226: Would be good to lock once so that another thread doesnt move the position in the file.

crates/rattler_cache/src/package_cache/cache_lock.rs
```diff
@@ -212,6 +244,34 @@ impl CacheRwLock {
...
+    pub fn read_sha256(&mut self) -> Result<Option<Sha256Hash>, PackageCacheError> {
+        let mut file = self.file.lock();
+        file.rewind().map_err(|e| {
+            PackageCacheError::LockError(
+                "failed to rewind cache lock for reading sha256".to_string(),
+                e,
+            )
+        })?;
+        let mut buf = [0; 32];
+        let _ = file.seek(SeekFrom::Start(8)).map_err(|e| {
```
baszalmstra, line 258: Maybe we can make this 8 not a magic number?

baszalmstra (commented):
Nice with all the tests! Left two small comments

crates/rattler_cache/src/package_cache/cache_lock.rs
tdejager: Good point!

tdejager:
Ok just tested it out with `pixi-build` and seems to work like a charm.

crates/rattler_cache/src/package_cache/cache_key.rs
```diff
@@ -8,10 +8,10 @@ use std::fmt::{Display, Formatter};
 /// TODO: Wouldn't it be better to cache based on hashes?
```
wolfv, line 8: Can we remove this TODO then? :)

baszalmstra (approved):

wolfv (approved):

crates/rattler_cache/src/package_cache/cache_key.rs
tdejager: Yes

baszalmstra (approved):

PR #910, opened 2024-10-23
perf: enable using sharded repodata for custom channels

baszalmstra:
Adds the ability to use sharded repodata for all or specific channels. 

This PR also introduces the `ChannelUrl` type because I was sick and tired of dealing with trailing slashes. The `ChannelUrl` type represents the base url of a channel and always includes a trailing slash.


crates/rattler_repodata_gateway/src/gateway/sharded_subdir/mod.rs
```diff
@@ -125,39 +119,50 @@ impl SubdirClient for ShardedSubdir {
...
+        }
 
         // Download the shard
         let shard_url = self
             .shards_base_url
             .join(&format!("{shard:x}.msgpack.zst"))
             .expect("invalid shard url");
 
-        let mut shard_request = self
+        tracing::warn!("{:?}", &shard_url);
```
wolfv, line 158: Is this a debug statement? Otherwise, maybe add some context to the warning?

PR #915, opened 2024-10-30
feat: Add middlewares to Client in py-rattler

### LINKED ISSUE #859: py-rattler: add support for mirror middleware
### Checklist

- [X] I added a descriptive title
- [X] I searched open requests and couldn't find a duplicate

### What is the idea?

Both pixi and rattler-build allow you to specify mirrors: https://pixi.sh/latest/reference/pixi_configuration/#mirror-configuration

It would be nice if we could have this feature exposed in py-rattler as well.

### Why is this needed?

We currently have a python script that installs a `pixi.lock` file into a custom prefix (as pixi only installs into `.pixi/envs/default`).
It would be nice to have this working with mirror configuration as well.

The following example currently doesn't do this but maybe adding it to `AuthenticatedClient` could be a way forward.

```py
#!/usr/bin/env -S pixi exec --spec py-rattler --spec typer -- python

import asyncio
from pathlib import Path

from rattler import install as rattler_install
from rattler import LockFile, Platform
from rattler.networking.authenticated_client import AuthenticatedClient
import typer


app = typer.Typer()


async def _install(
    lock_file_path: Path,
    environment_name: str,
    platform: Platform,
    target_prefix: Path,
) -> None:
    lock_file = LockFile.from_path(lock_file_path)
    environment = lock_file.environment(environment_name)
    records = environment.conda_repodata_records_for_platform(platform)
    await rattler_install(
        records=records,
        target_prefix=target_prefix,
        client=AuthenticatedClient(),
    )


@app.command()
def install(
    lock_file_path: Path = Path("pixi.lock").absolute(),
    environment_name: str = "default",
    platform: str = str(Platform.current()),
    target_prefix: Path = Path("env").absolute(),
):
    asyncio.run(
        _install(
            lock_file_path=lock_file_path,
            environment_name=environment_name,
            platform=Platform(platform),
            target_prefix=target_prefix,
        )
    )


if __name__ == "__main__":
    app()
```

### What should happen?

_No response_

### Additional Context

_No response_

pavelzw:
### Description

Closes #859


py-rattler/rattler/networking/client.py
```diff
@@ -0,0 +1,58 @@
...
+        ```python
+        >>> Client()
+        Client()
+        >>>
+        ```
+        """
+        return f"{type(self).__name__}()"
+
+
+class AuthenticatedClient(Client):
```
pavelzw, line 40: not sure whether we should still keep `AuthenticatedClient` now that we have `Client`. Any opinions? Since it's not really much code i would be inclined to keeping it.

py-rattler/src/networking/client.rs
```diff
@@ -0,0 +1,42 @@
+use crate::networking::middleware::PyMiddleware;
+use pyo3::{pyclass, pymethods};
+use rattler_networking::{AuthenticationMiddleware, MirrorMiddleware};
+use reqwest_middleware::ClientWithMiddleware;
+
+#[pyclass]
+#[repr(transparent)]
+#[derive(Clone)]
+pub struct PyClientWithMiddleware {
```
pavelzw, line 9: not sure what is a better name, `PyClientWithMiddleware` or `PyClient`?

pavelzw:
@baszalmstra this is ready for review

pavelzw:
@delsner feel free to take a look as well

py-rattler/rattler/networking/client.py
```diff
@@ -0,0 +1,60 @@
+from __future__ import annotations
+from rattler.rattler import PyClientWithMiddleware
+from rattler.networking.middleware import AuthenticationMiddleware, MirrorMiddleware
+
+
+class Client:
+    """
+    A client that can be used to make authenticated requests.
```
delsner, line 8:
```suggestion
    A client that can be used to make requests.
```

py-rattler/rattler/networking/middleware.py
```diff
@@ -0,0 +1,48 @@
+from __future__ import annotations
+from rattler.rattler import PyMirrorMiddleware, PyAuthenticationMiddleware
+
+
+class MirrorMiddleware:
+    def __init__(self, middlewares: dict[str, list[str]]) -> None:
+        self._middleware = PyMirrorMiddleware(middlewares)
```
delsner, line 7:
```suggestion
    def __init__(self, mirrors: dict[str, list[str]]) -> None:
        self._middleware = PyMirrorMiddleware(mirrors)
```
It would probably also make sense to docoument the format of `mirrors` here in a docstring.

py-rattler/src/networking/client.rs
delsner: Given the convention so far, I'd stick to `PyClientWithMiddleware`.

py-rattler/src/networking/client.rs
```diff
@@ -0,0 +1,39 @@
...
+impl PyClientWithMiddleware {
+    #[new]
+    pub fn new(middlewares: Option<Vec<PyMiddleware>>) -> Self {
+        let middlewares = middlewares.unwrap_or_default();
+        let client = reqwest_middleware::ClientBuilder::new(reqwest::Client::new());
+        let client = middlewares
+            .into_iter()
+            .fold(client, |client, middleware| match middleware {
+                PyMiddleware::MirrorMiddleware(middleware) => {
+                    client.with(MirrorMiddleware::from(middleware))
```
delsner, line 23: Can we also use `middleware.into()` here and below?

py-rattler/src/networking/middleware.rs
```diff
@@ -0,0 +1,70 @@
...
+            .map(|(k, v)| {
+                (
+                    Url::parse(&k).unwrap(),
+                    v.into_iter()
+                        .map(|url| Mirror {
+                            url: Url::parse(&url).unwrap(),
+                            no_zstd: false,
+                            no_bz2: false,
+                            no_jlap: false,
+                            max_failures: None,
```
delsner, line 36: Are these options irrelevant for defining mirrors? Would it maybe make sense to create another struct to pass them from Python?

py-rattler/src/networking/middleware.rs
```diff
@@ -0,0 +1,70 @@
...
+    #[new]
+    pub fn __init__(inner: HashMap<String, Vec<String>>) -> Self {
+        let map = inner
+            .into_iter()
+            .map(|(k, v)| {
+                (
+                    Url::parse(&k).unwrap(),
+                    v.into_iter()
+                        .map(|url| Mirror {
+                            url: Url::parse(&url).unwrap(),
```
delsner, line 32: Can we confidently unwrap here?

py-rattler/src/networking/middleware.rs
```diff
@@ -0,0 +1,70 @@
...
+#[repr(transparent)]
+#[derive(Clone)]
+pub struct PyAuthenticationMiddleware {}
+
+#[pymethods]
+impl PyAuthenticationMiddleware {
+    #[new]
+    pub fn __init__() -> Self {
+        Self {}
+    }
```
delsner, line 63: Do we need this or does pyo3 generate a default ctor?

py-rattler/src/networking/mod.rs
```diff
@@ -30,7 +31,7 @@ pub fn py_fetch_repo_data<'a>(
     callback: Option<&'a PyAny>,
 ) -> PyResult<&'a PyAny> {
     let mut meta_futures = Vec::new();
-    let client = PyAuthenticatedClient::new();
+    let client = PyClientWithMiddleware::new(None);
```
delsner, line 34: Is it desired that this change removes the auth middleware here?

delsner (commented):
Nice, thanks for finishing this!

py-rattler/src/networking/mod.rs
pavelzw: Yes I think so, `AuthenticatedClient` is now a Python class that just has a default middleware passed in Python

py-rattler/src/networking/client.rs
pavelzw: nope

```
error[E0283]: type annotations needed
  --> src/networking/client.rs:23:28
   |
23 |                     client.with(middleware.into())
   |                            ^^^^ ----------------- type must be known at this point
   |                            |
   |                            cannot infer type of the type parameter `M` declared on the method `with`
```

also i find `MirrorMiddleware::from` a bit easier to understand in this scenario

pavelzw: although `client.with::<MirrorMiddleware>(middleware.into())` works

pavelzw: but i still find `::from` a bit nicer to read so i'll leave it as is since we are not able to collapse it all into one line using `into` anyway

py-rattler/src/networking/middleware.rs
pavelzw: i would say they are irrelevant until somebody actually needs them in python.
then, the API can be extended IMO but in order to keep py-rattler somewhat simple i would suggest we leave them like this for now

py-rattler/src/networking/middleware.rs
pavelzw: nope

```
051         >>> middleware = AuthenticationMiddleware()
UNEXPECTED EXCEPTION: TypeError('No constructor defined')
```

py-rattler/src/networking/mod.rs
pavelzw: i would say to go in line with `rattler.install` where by default we don't have any authentication and need to specify `client=AuthenticatedClient()` we should change this here

pavelzw: added the ability to specify `client` in `py_fetch_repo_data`

py-rattler/src/networking/middleware.rs
pavelzw: [`e8c631c` (#915)](https://github.com/conda/rattler/pull/915/commits/e8c631c2118c3c8be9df05b13b8ba107d9e552a2)

py-rattler/rattler/networking/client.py
baszalmstra: Id prefer to only keep `Client`. Im not too well versed in python common practices but instead of subclasses we could also add a static method to `Client` to initialize with the authentication middleware?

Im fine either way.

baszalmstra (approved):

py-rattler/rattler/networking/client.py
pavelzw: yeah, let's create a static method instead. this will be a breaking change, can you take care of adding a note in the changelog afterwards?

baszalmstra:
Thank you very much @pavelzw !

PR #917, opened 2024-10-31
fix: matchspec build / version from brackets and string serialization

wolfv:
We were parsing brackets correctly, but we overrode it with `None`.

This fixes matchspecs like `foo * [build="*bla*"]`

For now this is no error when the build string is set twice (e.g. `foo * bar*[build="baz"]`). I am not sure if we _should_ treat that as an error (at least in strict mode?). Same goes for the version.

cc @jaimergp thanks for discovering this today!

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -489,7 +489,7 @@ mod tests {
 
     use crate::{
         match_spec::Matches, MatchSpec, NamelessMatchSpec, PackageName, PackageRecord,
-        ParseStrictness::*, RepoDataRecord, StringMatcher, Version,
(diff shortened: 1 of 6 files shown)


wolfv:
Interestingly in conda, it looks like brackets are always preferred (which is the inverse of what I did here).

For example:

```
>>> a = MatchSpec("foo [version=5.5.5] 1.2.3 foo")
>>> a
MatchSpec("foo==5.5.5=foo")
```

We can easily match conda behavior here. We could also add the option to fail on this in strict mode and allow version / build only _once_ in strict mode (or only allow duplication if the value is exactly the same?).

baszalmstra:
That seems like a good approach!

wolfv:
I implemented that strict parsing fails when encountering a key multiple times (in brackets, or outside brackets & inside brackets). 
IMO this is good to go! :)

baszalmstra (approved):

PR #930, opened 2024-11-05
fix: gateway recursive records

baszalmstra:
Fix an issue where all records where returned by the gateway even when they didnt match the root specs when using recursive mode.

### DIFF
--- a/py-rattler/rattler/package/package_name.py
+++ b/py-rattler/rattler/package/package_name.py
@@ -101,10 +101,15 @@ class PackageName:
         >>> PackageName("test-abc") == PackageName("test-ABC")
         True
         >>> PackageName("test-abc") == "test-abc"
+        True
+        >>> PackageName("test-abc") == "not-test-abc"
         False
         >>>
         ```
         """
+        if isinstance(other, str):
+            return self._name == PyPackageName(other)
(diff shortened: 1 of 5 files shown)


py-rattler/rattler/package/package_name.py
```diff
@@ -124,10 +129,15 @@ def __ne__(self, other: object) -> bool:
...
+        False
+        >>> PackageName("test-abc") != "not-test-abc"
         True
         >>>
         ```
         """
+        if isinstance(other, str):
+            return self._name != PyPackageName(other)
+
         if not isinstance(other, PackageName):
```
wolfv, line 141: Let's make this a TypeError?

py-rattler/rattler/package/package_name.py
```diff
@@ -101,10 +101,15 @@ def __eq__(self, other: object) -> bool:
...
+        True
+        >>> PackageName("test-abc") == "not-test-abc"
         False
         >>>
         ```
         """
+        if isinstance(other, str):
+            return self._name == PyPackageName(other)
+
         if not isinstance(other, PackageName):
```
wolfv, line 113: TypeError

PR #933, opened 2024-11-08
feat: use fs-err in rattler_shell

wolfv:


baszalmstra:
Woops! Can you run cargo fmt?

PR #934, opened 2024-11-10
feat: set cache directory

hadim:
This PR allows setting the cache directory using `RATTLER_CACHE_DIR` or `XDG_CACHE_HOME` and fallback to the current `dirs::cache_dir()` if needed.

I think it makes sense to have this logic into rattler directly but the use case was for rattler-build: https://github.com/prefix-dev/rattler-build/issues/1171

### DIFF
--- a/crates/rattler_cache/src/lib.rs
+++ b/crates/rattler_cache/src/lib.rs
@@ -10,8 +10,8 @@ pub use consts::{PACKAGE_CACHE_DIR, REPODATA_CACHE_DIR};
 
 /// Determines the default cache directory for rattler.
 /// It first checks the environment variable `RATTLER_CACHE_DIR`.
-/// If not set, it falls back to the environment variable `XDG_CACHE_HOME`.
-/// If not set, it falls back to the standard cache directory provided by `dirs::cache_dir()`.
+/// If not set, it falls back to `XDG_CACHE_HOME/rattler/cache`.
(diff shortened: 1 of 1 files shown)


crates/rattler_cache/src/lib.rs
```diff
@@ -7,9 +8,26 @@ pub mod validation;
...
-    Ok(dirs::cache_dir()
-        .ok_or_else(|| anyhow::anyhow!("could not determine cache directory for current platform"))?
-        .join("rattler/cache"))
+    Ok(std::env::var("RATTLER_CACHE_DIR")
+        .map(PathBuf::from)
+        .or_else(|_| {
+            std::env::var("XDG_CACHE_HOME")
+                .map(PathBuf::from)
+                .or_else(|_| {
+                    dirs::cache_dir().ok_or_else(|| {
```
baszalmstra, line 22: Im pretty sure that this function already takes the XDG_ variable into account.

Its also linux specific so wouldnt work for windows or macos.

crates/rattler_cache/src/lib.rs
```diff
@@ -1,3 +1,4 @@
+use anyhow::Ok;
```
baszalmstra, line 1: Ok is always available so this seems redundant.

baszalmstra (commented):
I left a few comment s but otherwise looks good!

baszalmstra (approved):
Thanks!

PR #937, opened 2024-11-13
feat: Add `link_package_from_archive`

pavelzw:
### Description

It would be nice to be able to install directly from a package file instead of needing to have the package available in the cache.


crates/rattler/src/install/mod.rs
```diff
@@ -876,4 +905,40 @@ mod test {
...
+            let paths = link_package_from_archive(
+                &package_path,
+                environment_dir.path(),
+                &install_driver,
+                InstallOptions::default(),
+            )
+            .await
+            .unwrap();
+
+            insta::assert_yaml_snapshot!(paths);
```
baszalmstra, line 941: Can you give the snapshot a name? This guarentees that the snapshot is properly linked to the loop iteration. Otherwise we might get flaky tests later.

crates/rattler/src/install/mod.rs
```diff
@@ -876,4 +905,40 @@ mod test {
...
+
+    #[tracing_test::traced_test]
+    #[tokio::test]
+    async fn test_link_package_from_archive() {
+        let packages = vec![(
+            "https://conda.anaconda.org/conda-forge/win-64/ruff-0.0.171-py310h298983d_0.conda",
+            "25c755b97189ee066576b4ae3999d5e7ff4406d236b984742194e63941838dcd"
+        ),
+        (
+            "https://conda.anaconda.org/conda-forge/linux-aarch64/bzip2-1.0.8-hf897c2e_4.tar.bz2",
```
baszalmstra, line 917: I think this package contains symlinks which won't work on windows. Can you use another package?

baszalmstra: Or if possible a smaller package stored in the repo? Premature opimization perhaps, but I want the tests to run very fast.

pavelzw: [`48d9c20` (#937)](https://github.com/conda/rattler/pull/937/commits/48d9c204e093e896919120c5a229fd3385688318)

pavelzw:
closing this for now as this implementation is so trivial that it doesn't make much sense to include it here. installing it directly without any package cache would be nice but i won't have the time in the near future to do so

PR #938, opened 2024-11-13
fix:  Release downloader lock before re-attempting fallback download 

jpcorreia99:
### Description
In https://github.com/conda/rattler/pull/837 locking logic was added to the download reporter to prevent multiple downloads.

This was clashing with the logic added in https://github.com/conda/rattler/pull/797 where, if we detect we're trying to download a zip that has data descriptors, we fallback to doing a full download and interrupt the streaming decompression download.

The issue comes from us not releasing the lock before attempting the second download, hitting this line https://github.com/conda/rattler/blob/8bb883536df52de46fd2c040228bb77d39c92607/crates/rattler_cache/src/package_cache/mod.rs#L439

As a fix, we report the previous download as completed before attempting the second one! Have verified in my custom setup that this works.

### DIFF
--- a/crates/rattler_package_streaming/src/reqwest/tokio.rs
+++ b/crates/rattler_package_streaming/src/reqwest/tokio.rs
@@ -159,6 +159,9 @@ pub async fn extract_conda(
             if (zip_error.contains(DATA_DESCRIPTOR_ERROR_MESSAGE)) =>
         {
             tracing::warn!("Failed to stream decompress conda package from '{}' due to the presence of zip data descriptors. Falling back to non streaming decompression", url);
+            if let Some(reporter) = &reporter {
+                reporter.on_download_complete();
+            }
             let new_reader =
                 get_reader(url.clone(), client, expected_sha256, reporter.clone()).await?;
             crate::tokio::async_read::extract_conda_via_buffering(new_reader, destination).await
--- a/crates/rattler_package_streaming/src/reqwest/tokio.rs
+++ b/crates/rattler_package_streaming/src/reqwest/tokio.rs
@@ -164,7 +164,18 @@ pub async fn extract_conda(
             }
             let new_reader =
                 get_reader(url.clone(), client, expected_sha256, reporter.clone()).await?;
-            crate::tokio::async_read::extract_conda_via_buffering(new_reader, destination).await
+
+            match crate::tokio::async_read::extract_conda_via_buffering(new_reader, destination)
+                .await
+            {
+                Ok(result) => {
+                    if let Some(reporter) = &reporter {
+                        reporter.on_download_complete();
+                    }
+                    Ok(result)
+                }
+                Err(e) => Err(e),
+            }
         }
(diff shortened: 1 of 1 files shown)


crates/rattler_package_streaming/src/reqwest/tokio.rs
```diff
@@ -159,6 +159,9 @@ pub async fn extract_conda(
             if (zip_error.contains(DATA_DESCRIPTOR_ERROR_MESSAGE)) =>
         {
             tracing::warn!("Failed to stream decompress conda package from '{}' due to the presence of zip data descriptors. Falling back to non streaming decompression", url);
+            if let Some(reporter) = &reporter {
+                reporter.on_download_complete();
+            }
             let new_reader =
                 get_reader(url.clone(), client, expected_sha256, reporter.clone()).await?;
             crate::tokio::async_read::extract_conda_via_buffering(new_reader, destination).await
```
jpcorreia99, line 167: Do you think I should also be reporting the conclusion of this download?

baszalmstra: If the called function doesnt so this internally then that makes sense yeah!

jpcorreia99: Updated!

jpcorreia99:
@baszalmstra what would we want to do with this extra method/argument? We already have the  

` tracing::warn!("Failed to stream decompress conda package from '{}' due to the presence of zip data descriptors. Falling back to non streaming decompression", url);`  

message which is very visible.

baszalmstra:
Sure thats for users, but I can imagine that a Reporter doesnt expect a download to be completed twice. If we add an additional parameter to the function we can tell the reporter that a download failed and it could handle the completion of the second properly.

jpcorreia99:
@baszalmstra In this flow, the sequence 
`reporter. on_download_start()` and `on_download_complete()` is called twice, sequentially (so there aren't two  completions for a single start). So, logically, one download starts and completes and another one starts and completes. Does that still match your concern? 

Currently, when calling `on_download_complete`, the reporter releases the lock and calls `on_download_complete` for an inner reporter:

https://github.com/conda/rattler/blob/8bb883536df52de46fd2c040228bb77d39c92607/crates/rattler_cache/src/package_cache/mod.rs#L449-L457

This inner reporter seems to do something for the visual progress bar:
https://github.com/conda/rattler/blob/8bb883536df52de46fd2c040228bb77d39c92607/crates/rattler/src/install/installer/indicatif.rs#L608-L629.

I understand the proposition of trying to change from 

- on_download_start
- on_download_finish
- on_download_start
- on_download_finish

to 
- start
- failure 
- finish

However, it will be pretty hard (and possibly mess up existing logic) to get rid of the second `on_download_start` because we must call it to get a new reader to start the download here

https://github.com/conda/rattler/b


Also, visually, I even with this existing logic it all looked completely fine to me. I didn't notice any strangeness.

PR #940, opened 2024-11-15
perf: validate using threads

baszalmstra:
A small improvement where we validate the contents of packages using multiple threads.


wolfv (approved):

wolfv:
I think this is good but it would also still be good to make it configurable in `pixi` or other tools.

baszalmstra:
Yes I want to makenit configurable whether the content is actually checked or not.

baszalmstra:
yes it was build on this!

PR #943, opened 2024-11-16
WIP: url with trailing slash mirror

wolfv:
@baszalmstra I refactored things to use the URLWithSlash in the mirror.

Unfortunately, that type currently lives in `rattler_conda_types`. I need it in `rattler_networking` though. They both currently don't have a relationship.

We could add another utils crate or make one dependent on the other. Not sure. Happy about suggestions. 

This PR builds on top of #939 and I'd suggest merging that one first.


baszalmstra:
I suppose adding another crate lije url-utils or whatever would help? We could also rename file_url and add it there?

wolfv:
Do you think file-url and urlwithslash could be in a `UrlExt` crate or something? Or something rattler-specific.

PR #944, opened 2024-11-16
chore: allow `ChannelUrl` in `Channel::from_url` interface

wolfv:
Not sure if it's a good idea. LMK @baszalmstra :)


baszalmstra:
If you change it Into<ChannelUrl> then both types remain usable.

wolfv:
Great point!

baszalmstra (approved):

PR #945, opened 2024-11-17
feat: py-rattler - add more types to construct a paths.json type

wolfv:
Adds lots of setters to the paths_json and related types.

Some of the `enum`'s could be exposed in a more Pythonic way but we can also do that later :)


baszalmstra:
Amazing work!

baszalmstra (approved):

Issue #578, opened 2024-03-22
Add retry/backoff to file removing on Windows


wolfv:
When overwriting a file on prefix updates it might try to overwrite a file that is still in use. The workaround is to stop all usages of the prefix before an update. 

### Where is this important
A common use-case is in a pixi project that is activated/used in an IDE. These IDE's might be using the Python Interpreter for the language server. 

The error you'll get on a package update will then look like this:
```
❯ pixi add python>3.11
  ⠁ validate cache       [00:00:00] [━━━━━━━━━━━━━━━━━╾──] python (+1)
  ⠁ download & extract   [00:00:00] [────────────────────]      0 B @ 0 B/s    pydantic-core (+1)
  ⠁ installing packages  [00:00:00] [━╾──────────────────]     2/30 python (+14)
  × failed to unlink python-3.10.15-hfaddaf0_2_cpython.conda
  ╰─▶ failed to delete file: DLLs/_ctypes.pyd
```
Which is hard to read and UX unfriendly. A way to avoid this error entirely would be an amazing improvement for the UX. 

### Solution
This has been done before in `mamba`, by renaming it to a `xxx.conda_trash`. Then another run could look for all the trash files and actually remove them.

Initial issue:
_Sometimes, removing a file that was (just) in use fails on Windows. I think it's good practice to add a retry after a few ms or more (we also hit something like that in pixi and fixed it with a sleep: https://github.com/prefix-dev/pixi/pull/985/commits/0ee497fde01133f96d2030eb83687eaee034d00b_

adament:
If I were to take a stab at this, would you prefer to have a design discussion here first or that I just make an attempt at it and then we can have a discussion on the preferred solution in the pull request?

baszalmstra:
Awesome! The issue already contains a possible solution, but if you have other ideas feel free to discuss it here or implement it right away.

PR #953, opened 2024-11-19
feat: Move files to .trash folder if they are in use

adament:
### Description

Attempt to move files to .trash folder if they are in use, this should solve #578. The issue is that on Windows we cannot delete files that are in use. When using a global package cache and hardlinking files into environments this means that if a given DLL or exe is in use in any environment with the same package the deletion files. But we can rename the hardlink. Conda and mamba have similar fallback behaviour.

The solution in this PR is slightly different than the ones used in conda and mamba as far as I am aware. This PR:

* Creates a folder in `.trash` in the root of the environment rather than having a file like mamba_trash.txt to keep track of all the files which could not be deleted. Then a separate empty_trash procedure should simply attempt to delete all files in the .trash folder and if successful, it should remove the `.trash` folder.
* When moving a file to `.trash` it is renamed to `uuid.rattler_trash`. Mamba and conda kept the original filename and appended `.mamba_trash`, but then they had to worry about filename collisions and having some fallback procedure for coming up with new names on a collision. It thus seemed simpler to me to just give a unique name to each file in the trash folder. Though in this way we lose information about the origin of the file.

Please tell me if you think the approach or the implementation is wrong.

### DIFF
--- a/crates/rattler/src/install/mod.rs
+++ b/crates/rattler/src/install/mod.rs
@@ -60,7 +60,7 @@ use simple_spawn_blocking::Cancelled;
 use tokio::task::JoinError;
 use tracing::instrument;
 pub use transaction::{Transaction, TransactionError, TransactionOperation};
-pub use unlink::unlink_package;
+pub use unlink::{empty_trash, unlink_package};
 
 use crate::install::entry_point::{
     create_unix_python_entry_point, create_windows_python_entry_point,
--- a/crates/rattler/src/install/unlink.rs
+++ b/crates/rattler/src/install/unlink.rs
@@ -5,6 +5,7 @@ use std::{
     io::ErrorKind,
     path::{Path, PathBuf},
 };
+use uuid::Uuid;
 
 use rattler_conda_types::PrefixRecord;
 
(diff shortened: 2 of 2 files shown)


wolfv:
Awesome! I am wondering - did you check that "moving" to a different folder works when the file is in use? If yes, then this is a great solution. 

I also thought that we could try to have `filename.<uuid>.trash` as a filename but it might get longer than the allowed length. 

I think going for UUID is fine with me. We would probably also be on the safe side with, idk, filename + 4 random characters or so.

There are also ways to add metadata to files that are NTFS specific (similar to xattrs on Unix) but not sure if that would be worth it to investigate.

baszalmstra:
Lets add the filename indeed to make it slightly easier to debug. If filename becomes to long we can worry about it then.

adament:
I have changed it to the filename.<uuid>.trash pattern and attempted to make the test more robust.

baszalmstra:
Thank you so much for this PR and your patience reviewing our comments! Highly appreciated! :)

adament:
Thank you for taking the time to review and comment.

PR #956, opened 2024-11-22
feat: add gcs middleware

tdejager:
## Adds GCS Middleware to python bindings
This PR adds the ability to use the GCS middleware from python, it uses the underlying gcs middlware type from rattler directly.
This feature also adds the gcs feature as a default to the networking crate in py-rattler.

## Test
A simple instantiation test has been written. Any other test that could be added?

## Misc
I've also added a `repl` task to start an ipython session which you can use to mess around with the API

### DIFF
--- a/py-rattler/Cargo.lock
+++ b/py-rattler/Cargo.lock
@@ -2693,7 +2693,7 @@ dependencies = [
 
 [[package]]
 name = "rattler"
-version = "0.28.1"
+version = "0.28.3"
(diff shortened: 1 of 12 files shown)


baszalmstra (approved):

tdejager:
Oh final thing, did you try `cargo insta --unreferenced=delete` that should clear up any unused snapshots because the code has moved around.

PR #958, opened 2024-11-25
feat: use `fs-err` also for tokio

Hofer-Julian:
This should improve error messages for file operations driven by tokio


wolfv:
I think we can even get rid of tokio for any file system activity cc @baszalmstra

wolfv:
plain `fs` could also do the trick since async `fs` doesn't seem to increase speeds a lot. But we can do one change at a time. I think this is good!

baszalmstra (approved):

PR #959, opened 2024-11-25
feat: upgrade reqwest middleware and retry

tdejager:
## Description

Upgrade reqwest middleware and retry crates. This upgrade is needed so that we can start using these versions in pixi, in conjuction with the uv crates.


wolfv:
Looks good to me!

PR #964, opened 2024-12-02
refactor: parse matchspec url into NamedChannelOrUrl [skip ci]

wolfv:
I think it would be good to stop the weird channel parsing in matchspecs because it's subtly wrong. We always create a "default" channel config. 

However, now we have two problems:

- we need a channel config in the resolvo context so that we can continue comparing matchspecs
- the `matches(...)` function needs a second `matches_with_channel_config(...)` function or a second parameter with a optional channel config to properly evaluate the channels from config.

Wanted to ask for early feedback before going down this path further.


baszalmstra:
Perhaps we should have a "ResolvedMatchSpec" and a non-resolved one. Also a big refactor..

PR #965, opened 2024-12-04
fix: merge purls from matching records in lock-file

baszalmstra:
When two packages are added or read from a v5 lock-file they are deduplicated. However, even though the packages may refer to the same package, they might include optional fields like `purls`. This PR merges these optional fields from deduplicated records before building or deserializing lock-files.

I added two tests:

1. A v5 lock-file with two similar records where one has purls and the other doesnt.
2. Building a lock-file through the API will deduplicate the package data but merge the purls.

I believe this does not need a bump of the lock-file format because this only adds data that was already read correctly but was not properly serialized/constructed.


crates/rattler_lock/src/parse/deserialize.rs
```diff
@@ -219,37 +220,89 @@ fn parse_from_lock<P>(
...
-                                                    let conda_package = &conda_packages[*idx];
-                                                    name.as_ref().map_or(true, |name| {
-                                                        name == &conda_package.record().name
-                                                    }) && version.as_ref().map_or(true, |version| {
-                                                        version == &conda_package.record().version
-                                                    }) && build.as_ref().map_or(true, |build| {
-                                                        build == &conda_package.record().build
-                                                    }) && subdir.as_ref().map_or(true, |subdir| {
-                                                        subdir == &conda_package.record().subdir
+                                                all_packages
```
nichmor, line 275: small request: could we please add a comment what this part of branching is doing?

crates/rattler_lock/src/builder.rs
```diff
@@ -362,3 +396,77 @@ impl From<PypiPackageEnvironmentData> for HashablePypiPackageEnvironmentData {
...
+    use std::{collections::BTreeSet, str::FromStr};
+
+    use rattler_conda_types::{PackageName, PackageRecord, Platform, Version};
+    use url::Url;
+
+    use crate::{CondaBinaryData, LockFile};
+
+    #[test]
+    fn test_merge_records_and_purls() {
+        let record = PackageRecord {
```
nichmor, line 411: One small suggestion for the test:

maybe we could test here that indeed a package with purl will win in case of merging with one that doesn't have purl.
This is my idea:

nichmor, line 411:
```suggestion
        let mut record = PackageRecord::new(
            PackageName::new_unchecked("foobar"),
            Version::from_str("1.0.0").unwrap(),
            "build".into(),
        
        );

        let mut record_without_purl = PackageRecord::new(
            PackageName::new_unchecked("foobar"),
            Version::from_str("1.0.0").unwrap(),
            "build".into(),
        
        );

        record.purls = Some(
            vec![
                "pkg:pypi/foobar@1.0.0".parse().unwrap(),
            ]
            .into_iter()
            .collect(),
        );
        
        let record = PackageRecord {
            subdir: "linux-64".into(),
            ..record
        };

        let record_without_purl = PackageRecord {
            subdir: "linux-64".into(),
            ..record_without_purl
        };
```

crates/rattler_lock/src/builder.rs
```diff
@@ -362,3 +396,77 @@ impl From<PypiPackageEnvironmentData> for HashablePypiPackageEnvironmentData {
...
+                    file_name: "foobar-1.0.0-build.tar.bz2".to_string(),
+                    channel: None,
+                }
+                .into(),
+            )
+            .with_conda_package(
+                "foobar",
+                Platform::Linux64,
+                CondaBinaryData {
+                    package_record: PackageRecord {
```
nichmor, line 455:
```suggestion
                    package_record: record_without_purl,
```

crates/rattler_lock/src/builder.rs
```diff
@@ -362,3 +396,77 @@ impl From<PypiPackageEnvironmentData> for HashablePypiPackageEnvironmentData {
...
+                    )
+                    .unwrap()
+                    .into(),
+                    file_name: "foobar-1.0.0-build.tar.bz2".to_string(),
+                    channel: None,
+                }
+                .into(),
+            )
+            .finish();
+        insta::assert_snapshot!(lock_file.render_to_string().unwrap());
```
nichmor, line 470: so the snapshot will be like this:
```
---
source: crates/rattler_lock/src/builder.rs
expression: lock_file.render_to_string().unwrap()
---
version: 6
environments:
  default:
    channels: []
    packages:
      linux-64:
      - conda: https://prefix.dev/example/linux-64/foobar-1.0.0-build.tar.bz2
  foobar:
    channels: []
    packages:
      linux-64:
      - conda: https://prefix.dev/example/linux-64/foobar-1.0.0-build.tar.bz2
packages:
- conda: https://prefix.dev/example/linux-64/foobar-1.0.0-build.tar.bz2
  arch: x86_64
  platform: linux
  channel: null
  purls:
  - pkg:pypi/foobar@1.0.0

```

which will give us confidence that we indeed merge purls , even when the package is in different environments

baszalmstra:
@nichmor I added all your feedback, let me know what you think!

nichmor (commented):
I like it!

PR #966, opened 2024-12-05
fix: gcs access and add test

wolfv:
Unfortunately a bug slipped in where we were sending `Bearer Bearer <token>` (ie. one Bearer too much). 

This fixes it and adds a test. I also added credential for a test account in the `GOOGLE_CLOUD_TEST_KEY_JSON` secret. This can access a bucket in the prefix-dev GCS (`conda-channel-test/test-channel`).

## How to use GCS channels with rattler / pixi:

To use GCS channels, you need to first login using the `gcloud` CLI tool. For automatic discovery, rattler expects the `~/.config/gcloud/application_default_credentials.json` file to be available (or the `GOOGLE_APPLICATION_CREDENTIALS` env var to contain the JSON contents). 

To login, you can use `gcloud auth application-default login` which will automatically create the JSON file in the right place.

## Logs on how the service account was created:

```
gcloud iam service-accounts create gh-action-test-bucket-reader \
    --display-name "Github Actions Test Bucket Reader"

gcloud iam roles create bucketReader \
    --project=conda-channel-test \
    --title="Bucket Reader" \
    --permissions=storage.objects.get,storage.objects.list


gsutil iam ch \
    serviceAccount:gh-action-test-bucket-reader@conda-channel-test.iam.gserviceaccount.com:roles/storage.objectViewer \
    gs://test-channel

gcloud iam service-accounts keys create key.json \
    --iam-account=gh-action-test-bucket-reader@conda-channel-test.iam.gserviceaccount.com
```

### DIFF
--- a/crates/rattler_networking/src/gcs_middleware.rs
+++ b/crates/rattler_networking/src/gcs_middleware.rs
@@ -68,27 +68,32 @@ async fn authenticate_with_google_cloud(mut req: Request) -> MiddlewareResult<Re
 mod tests {
     use super::*;
     use reqwest::Client;
+    use tempfile;
 
     #[tokio::test]
     async fn test_gcs_middleware() {
-        let Ok(credentials) = std::env::var("GOOGLE_CLOUD_TEST_KEY_JSON") else {
-            eprintln!("GOOGLE_CLOUD_TEST_KEY_JSON is not set - skipping test");
-            return;
-        };
+        let credentials = std::env::var("GOOGLE_CLOUD_TEST_KEY_JSON").unwrap();
 
-        // We have to set GOOGLE_APPLICATION_CREDENTIALS to the content of the JSON key file
-        std::env::set_var("GOOGLE_APPLICATION_CREDENTIALS", credentials);
+        // We have to set GOOGLE_APPLICATION_CREDENTIALS to the path of the JSON key file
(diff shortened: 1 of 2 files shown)


crates/rattler_networking/src/gcs_middleware.rs
```diff
@@ -63,3 +63,37 @@ async fn authenticate_with_google_cloud(mut req: Request) -> MiddlewareResult<Re
...
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use reqwest::Client;
+    use tempfile;
+
+    #[tokio::test]
+    async fn test_gcs_middleware() {
```
baszalmstra, line 74: Should we only run this test if this variable is available?

baszalmstra (commented):
Looks good and awesome that you added a test!

crates/rattler_networking/src/gcs_middleware.rs
wolfv: We only run this test when the `gcs` feature is enabled. So that already makes the test run pretty seldomly. I havent' found a way to print a big fat warning, so I prefer to fail the test.

baszalmstra:
Can you reopen this PR as a branch so we are sure that the test works?

PR #974, opened 2024-12-10
feat: allow using an authenticated `Client` in `Gateway`

### LINKED ISSUE #973: How to use authentication with py-rattler's solver?
I see it's possible to pass an authenticated Client to `rattler.install`, but I don't see any equivalent way to do it with `rattler.solve`

calebzulawski:
### Description

I am using a private conda channel and I am able to `rattler.install`, but not `rattler.solve`!  This change allows authenticating the solver as well. Fixes #973 

I can't think of a reasonable way to unit test this, but I did test it via `pixi run repl`.


wolfv:
Awesome!

PR #975, opened 2024-12-10
feat: add editable key to lock file

Hofer-Julian:


baszalmstra:
Im wondering if it makes sense to add this information to the lockfile. 

"Editable" only means something in the context of pypi packages. I think we should generalize this.

But on a more general note cant we derive this information from the project itself too?

Hofer-Julian:
We decided against this approach

PR #977, opened 2024-12-12
fix: package cache lock file_name was incorrect

baszalmstra:
This PR fixes an issue where the package cache would be repopulated even though this was not required. This was because the individual cache entries shared the same lock file. For example these packages would all use the same lock file name (`cffi-1.17.lock`):

```
cffi-1.17.0-py38h4cb3324_0
cffi-1.17.1-py311he736701_0
cffi-1.17.1-py312h4389bb4_0
cffi-1.17.1-py313ha7868ed_0
```

This PR fixes this by correctly computing the lock file name for each entry (e.g. `cffi-1.17.1-py313ha7868ed_0.lock`).


ruben-arts (approved):
Crazy that it was so simple, nice find!

PR #980, opened 2024-12-12
perf: reduce ideal concurrency limit

baszalmstra:
@wolfv found out that the default io concurrency limit of 100 is WAY to high and actually hurts performance quite a bit. This PR uses the number of physical cores to compute an better value.


wolfv:
Let me make this configurable in `rattler` create and run a few tests on macOS.

PR #985, opened 2024-12-16
feat: improve performance when linking files using `rayon`

baszalmstra:


crates/rattler/src/install/installer/indicatif.rs
```diff
@@ -375,7 +375,7 @@ impl<F: ProgressFormatter> IndicatifReporterInner<F> {
             .iter()
             .map(|&idx| (self.package_sizes[idx], &self.package_names[idx]));
 
-        let largest_package = package_iter.max_by_key(|(size, _)| *size);
+        let largest_package = package_iter.min_by_key(|(size, _)| *size);
```
wolfv, line 378: ?

baszalmstra: Yeaaah I added this to make it look for the users as if more stuff is happening. The biggest package will take the longest. This changes that the user instead sees way more packages in the loading bar.

wolfv (approved):

PR #989, opened 2024-12-17
feat: make strict parsing even stricter

wolfv:
This PR makes parsing even stricter. With these changes, `foo 1.2.3` throws an error and suggests either `==1.2.3`, or `1.2.3.*`.

Note: there seems to be a bug (not sure if new?) when parsing `1!1.2.3.*` - it says "ExpectedEof".


baszalmstra (approved):

baszalmstra:
You still have some test failures, but otherwise looks good to me!

PR #991, opened 2024-12-19
perf: read json to memory before parsing

wolfv:
I saw a lot of file system activity on macOS related to reading package cache files. 

The way we were reading them was not optimal. `std::fs::read_to_string` does a better job: it looks at the file, checks the size, and uses the size as a hint to first allocate the string and then allocate the buffer for the reading. So I hope this will improve this a bit.

Screenshot 2024-12-19 at 09 41 39


baszalmstra:
Yes from_reader is notoriously slow! 👍

PR #997, opened 2024-12-21
perf: use `asm` feature for sha hashing in rattler-bin

wolfv:


baszalmstra:
As I understand it, this doesnt work on windows because https://github.com/RustCrypto/asm-hashes doesnt support windows msvc? That means we cannot just enable the feature in rattler because it would break compilation downstream.

Should we close this and enable it in pixi specifically?

PR #998, opened 2024-12-22
[wip] tiered cache

wolfv:
This is a starting point for a tiered cache where one could have multiple package caches, where some of the lower ones might be read-only.

A challenge is that we probably cannot do "read-locking" on read-only caches since we wouldn't be allowed to create lockfiles at all.

Another goal is that we can say: can-reflink or can-symlink per-cache (I want to store a HashMap in the cache with prefix mapping to capabilities).

### DIFF
--- a/crates/rattler-bin/Cargo.toml
+++ b/crates/rattler-bin/Cargo.toml
@@ -36,6 +36,8 @@ rattler_virtual_packages = { path="../rattler_virtual_packages", version = "1.1.
 rattler_cache = { path="../rattler_cache", version = "0.3.1", default-features = false }
 reqwest = { workspace = true }
 reqwest-middleware = { workspace = true }
+# we enable the asm feature for sha2 to speed up hashing in the binary
+sha2 = { workspace = true, features = ["asm"] }
 tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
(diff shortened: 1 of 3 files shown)


kelvinou01:
@wolfv Hello from https://github.com/conda/rattler/pull/1003! I may be missing some context, so I’d appreciate your clarification:

> A challenge is that we probably cannot do "read-locking" on read-only caches since we wouldn't be allowed to create lockfiles at all.

Isn't it guaranteed that each package in the cache has a lockfile? If so, how does that affect the read-locking concern?

> Another goal is that we can say: can-reflink or can-symlink per-cache (I want to store a HashMap in the cache with prefix mapping to capabilities).

Could you provide more details on this? I’m not entirely clear on how the can-reflink and can-symlink capabilities would be used per cache.

Thanks!

wolfv:
@kelvinou01 ah I see. Maybe you are right! I wasn't sure if we could aquire a lock in a read-only filesystem but your argument makes sense. I think if we cannot aquire a lock, we can ignore that, too.

We are linking each package individually, and we are testing wether we can reflink from package cache to target_prefix for each package. Instead of performing this check for every package, we could perform the check once, per package cache, and store this information in the cache. 

We have functions that try to create a hardlink or symlink from package cache to prefix already in the `link.rs` file (I think).

PR #1003, opened 2024-12-23
Add support for layered package cache

kelvinou01:
### Description

As per https://github.com/conda/rattler/issues/43, we want to implement layered package caches for Rattler.

Stuff to note
- I tried to make this not a break, but alas. Specifically, i made `PackageCacheError` a `#[non_exhaustive]` enum, and modified some variants. 

[Old PR for ref](https://github.com/conda/rattler/pull/843)

### DIFF
--- a/crates/rattler_cache/src/package_cache/mod.rs
+++ b/crates/rattler_cache/src/package_cache/mod.rs
@@ -6,6 +6,7 @@ use std::{
     fmt::Debug,
     future::Future,
     path::{Path, PathBuf},
+    pin::Pin,
     sync::Arc,
     time::{Duration, SystemTime},
 };
@@ -40,7 +41,7 @@ mod reporter;
 /// Instead, this is left up to the user when the package is requested. If the
 /// package is found in the cache it is returned immediately. However, if the
 /// cache is stale a user defined function is called to populate the cache. This
-/// separates the corners between caching and fetching of the content.
+/// separates the concerns between caching and fetching of the content.
 #[derive(Clone)]
 pub struct PackageCache {
     inner: Arc<PackageCacheInner>,
@@ -48,6 +49,10 @@ pub struct PackageCache {
 
 #[derive(Default)]
 struct PackageCacheInner {
+    layers: Vec<PackageCacheLayer>,
+}
+
+struct PackageCacheLayer {
     path: PathBuf,
     packages: DashMap<BucketKey, Arc<tokio::sync::Mutex<Entry>>>,
 }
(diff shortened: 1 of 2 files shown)


wolfv:
Hi @kelvinou01 - I also started hacking on this a bit but it looks like you are already much further progressed :)

Here was my attempt: https://github.com/conda/rattler/pull/998

gzm55:
@kelvinou01 is this pr ready?

gzm55:
it seems the pr does not contains the changes of the config key.

kelvinou01:
> @kelvinou01 is this pr ready?

@gzm55 not yet, is there an urgent need for it? I plan to work on it soon.

gzm55:
no urgent need. but expect this feature for multi-user environment

kelvinou01:
@wolfv hey, could you help me review this? 

Also, I did try implementing storing the capabilities cache like #998. However, I thought this PR was large enough already, and the capabilities cache can be introduced in a subsequent PR?

gzm55:
when select a read or write cache, can we prefer the one from which we can make a hard link?

kelvinou01:
@gzm55 I see how that can be useful! Do you think this feature is necessary to introduce with this PR, or can we do it in a subsequent one as i mentioned above?

gzm55:
> @gzm55 I see how that can be useful! Do you think this feature is necessary to introduce with this PR, or can we do it in a subsequent one as i mentioned above?

I think a subsequent pr for the feature mentioned above is ok.

crates/rattler_cache/src/package_cache/mod.rs
```diff
@@ -98,17 +131,151 @@ impl From<Cancelled> for PackageCacheError {
...
+            .get(&cache_key.clone().into())
+            .ok_or(PackageCacheLayerError::PackageNotFound)?
+            .clone();
+        let mut cache_entry = cache_entry.lock().await;
+        let cache_path = self.path.join(cache_key.to_string());
+
+        match validate_package_common::<
+            fn(PathBuf) -> _,
+            Pin<Box<dyn Future<Output = Result<(), _>> + Send>>,
+            std::io::Error,
```
baszalmstra, line 172: Is this needed?

crates/rattler_cache/src/package_cache/mod.rs
```diff
@@ -98,17 +131,151 @@ impl From<Cancelled> for PackageCacheError {
...
+    /// Determine if the layer is read-only in the filesystem
+    pub fn is_readonly(&self) -> bool {
+        self.path
+            .metadata()
+            .map(|m| m.permissions().readonly())
+            .unwrap_or(false)
+    }
+
+    /// Validate the package, and throw if invalid.
+    pub async fn validate_or_throw(
```
baszalmstra, line 157: This doesnt really "throw" anything. Maybe call `try_validate`?

crates/rattler_cache/src/package_cache/mod.rs
```diff
@@ -98,17 +131,151 @@ impl From<Cancelled> for PackageCacheError {
...
+    /// Constructs a new [`PackageCache`] with only one layer.
     pub fn new(path: impl Into<PathBuf>) -> Self {
-        Self {
-            inner: Arc::new(PackageCacheInner {
+        Self::new_layered(std::iter::once(path.into()))
+    }
+
+    /// Constructs a new [`PackageCache`] located at the specified paths.
+    /// Read-only layers are queried first.
+    /// Within read-only layers, the ordering is defined in this constructor. Ditto for writable layers.
```
baszalmstra, line 239: I dont fully understand this comment and I also dont see how it relates to the code? To me it makes most sense to pass in the paths in the order the user wants the layers to be ordered.

I would assume you look through the layers in the order they are passed in, and only write to the first writable layer.

crates/rattler_cache/src/package_cache/mod.rs
```diff
@@ -98,17 +131,151 @@ impl From<Cancelled> for PackageCacheError {
...
         }
     }
 
+    /// Returns a tuple containing two sets of layers:
+    /// - A collection of read-only layers.
+    /// - A collection of writable layers.
+    ///
+    /// The permissions are checked at the time of the function call.
+    pub fn split_layers(&self) -> (Vec<&PackageCacheLayer>, Vec<&PackageCacheLayer>) {
+        let readonly_layers = self
```
baszalmstra, line 264: You can use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.partition

baszalmstra (changes requested):

crates/rattler_cache/src/package_cache/mod.rs
kelvinou01: You're right, thanks!

crates/rattler_cache/src/package_cache/mod.rs
kelvinou01: Yes unfortunately, as the compiler has to infer the types somehow if we're passing in `None` for `fetch` and `reporter`. 

One way around this would be to assign default types to `F`, `Fut`, `E` during the declaration of `validate_package_common<F, Fut, E>`, but that feature is being phased out of Rust.

crates/rattler_cache/src/package_cache/mod.rs
kelvinou01: Seems like I didn't make the comment clear. What I meant was, the querying order is defined by 
- Primary key: read-only before write-only
- Secondary key: order it was passed into the constructor

But now that I think about it, querying all layers in the order passed into the constructor makes more sense.

crates/rattler_cache/src/package_cache/mod.rs
kelvinou01: Thank you for the tip ^^ integrated this fix

gzm55:
@baszalmstra is there other suggestions about this pr?

baszalmstra (approved):
This looks good to me. Would you be able to fix the merge conflict? We can merge afterwards unless @wolfv has specific comments.

kelvinou01:
> Would you be able to fix the merge conflict

@baszalmstra done!

baszalmstra:
@wolfv Would you be able to take a look at this PR?

chrisburr:
Is there anything left blocking this PR?

wolfv:
I actually could use this feature well in `rattler-build` for temporary package caches during testing - so I'll fix the conflicts and merge it!

wolfv:
Thank you @kelvinou01 and sorry for the extremely long wait!

kelvinou01:
No worries at all. Happy to contribute!

Issue #960, opened 2024-11-25
Add `s3://` as possible package URLs


pavelzw:
@delsner and I are planning to add S3 support to pixi/rattler.
Following ideas for modifying the config files

pixi.toml
```toml
[project]
name = "my-project"
channels = [
  "s3://my-bucket/channel",
  "conda-forge",
]

[s3]
region = "us-west-2"
endpoint = "http://localhost:9000"
url-style = "subdomain" # or "path" (used by Cloudflare R2 or minio)

[s3."my-bucket"]
region = "eu-west-1"
```

config.toml
```toml
[s3]
region = "eu-west-1" # globally (used in pixi init, pixi seach, pixi exec)

[s3.conda-forge] # globally (used in pixi init, pixi search, pixi exec)
# not used when not specified in pixi.toml but used in s3://
endpoint = "http://localhost:9000"
region = "us-east-1"
url-style = "subdomain" # or "path"
```

```bash
pixi auth login --access-key-id <access-key-id> --secret-access-key <secret-access-key> --session-token <session-token> s3://my-bucket
```
credentials.json
```json
{
    "s3://my-bucket": {
        "S3": {
            "AccessKeyId": "<access-key-id>",
            "SecretAccessKey": "<secret-access-key>",
            "SessionToken": "<session-token>" // optional
        }
    }
}
```

pixi.lock
```yaml
version: 6
environments:
  default:
    channels:
    - url: https://conda.anaconda.org/conda-forge/
    - url: s3://my-bucket/channel/
      s3:
        region: eu-west-1
        endpoint: http://localhost:9000
        url-style: subdomain # or path
```

Any thoughts on this proposal?

pavelzw:
One open question would be:

do we want to use the rust-s3 crate (which might introduce additional dependencies) that also provides some sort of native credential reading (through env vars, ...) or should we only rely on our own implementation where we store access keys, ... in pixi's credential stores?

I personally would like to be able to handle pixi's credential stores for S3 credentials as well.
We could theoretically default to aws's defaults if we didn't specify anything else.

For reproducibily's sake we should IMO definitely add the aws region and endpoint (if not default) into pixi.lock (and pixi.toml)

pavelzw:
AFAICT the designed above should be able to handle other providers like minio, cloudflare, gcs (for which we already have a middleware, maybe deprecate it afterwards?) as well.

olivier-lacroix:
@pavelzw , any reason for not using the same approach as for gcs?

[...]

PR #1008, opened 2025-01-02
feat: Add S3 support

pavelzw:
### Description

xref #960

https://github.com/conda/rattler/issues/960#issuecomment-2568213511

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,6 @@
+[test-groups]
+serial-integration = { max-threads = 1 }
+
+[[profile.default.overrides]]
+filter = 'package(rattler_networking) and (test(#test_minio*) | test(s3_middleware::tests))'
+test-group = 'serial-integration'
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,6 +1,0 @@
-[test-groups]
-serial-integration = { max-threads = 1 }
-
-[[profile.default.overrides]]
-filter = 'package(rattler_networking) and (test(#test_minio*) | test(s3_middleware::tests))'
-test-group = 'serial-integration'
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -146,12 +146,16 @@ jobs:
           # on windows, the minio server doesn't get shutdown properly so there is still a lock on this file
           post-cleanup: false
 
-      - name: Start minio for integration tests
+      - name: Install minio for integration tests
         if: ${{ !matrix.skip-tests }}
         run: |
           pixi global install minio-client
           pixi global install minio-server
-          pixi run minio server --address 0.0.0.0:9000 ${{ runner.temp }}/minio-data &
+
+      - name: Start minio server
+        if: ${{ !matrix.skip-tests }}
+        run: |
+          minio${{ startsWith(matrix.os, 'windows') && '.bat' || '' }} server --address 0.0.0.0:9000 ${{ runner.temp }}/minio-data &
 
       - name: Run tests
         if: ${{ !matrix.skip-tests }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -158,7 +158,8 @@ jobs:
           minio server --help
           mkdir -p ${{ runner.temp }}/minio-data
           minio server --address 127.0.0.1:9000 ${{ runner.temp }}/minio-data &
-          curl -i localhost:9000
+          sleep 5
+          curl -I http://localhost:9000/minio/health/live
 
       - name: Run tests
         if: ${{ !matrix.skip-tests }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -138,33 +138,19 @@ jobs:
         with:
           tool: cargo-nextest
 
-      - name: Set up pixi
-        if: ${{ !matrix.skip-tests }}
-        uses: prefix-dev/setup-pixi@v0.8.1
-        with:
-          run-install: false
-          # on windows, the minio server doesn't get shutdown properly so there is still a lock on this file
-          post-cleanup: false
-
-      - name: Install minio for integration tests
-        if: ${{ !matrix.skip-tests }}
-        run: |
-          pixi global install minio-client
-          pixi global install minio-server
-
       - name: Run tests
         if: ${{ !matrix.skip-tests }}
         env:
           GOOGLE_CLOUD_TEST_KEY_JSON: ${{ secrets.GOOGLE_CLOUD_TEST_KEY_JSON }}
           RATTLER_TEST_R2_ACCESS_KEY_ID: ${{ secrets.RATTLER_TEST_R2_ACCESS_KEY_ID }}
           RATTLER_TEST_R2_SECRET_ACCESS_KEY: ${{ secrets.RATTLER_TEST_R2_SECRET_ACCESS_KEY }}
-        run: |
-          minio server --help
-          mkdir -p ${{ runner.temp }}/minio-data
-          minio server --address 127.0.0.1:9000 ${{ runner.temp }}/minio-data &
-          sleep 5
-          curl -I http://localhost:9000/minio/health/live
-          cargo nextest run --workspace --features ${{ env.DEFAULT_FEATURES }},experimental_extras --target ${{ matrix.target }} ${{ steps.build-options.outputs.CARGO_BUILD_OPTIONS}} ${{ steps.test-options.outputs.CARGO_TEST_OPTIONS}} 
+        run: >
+          cargo nextest run
+          --workspace
+          --features ${{ env.DEFAULT_FEATURES }},experimental_extras
+          --target ${{ matrix.target }}
+          ${{ steps.build-options.outputs.CARGO_BUILD_OPTIONS}}
+          ${{ steps.test-options.outputs.CARGO_TEST_OPTIONS}} 
 
       - name: Run doctests
         if: ${{ !matrix.skip-tests }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -152,27 +152,6 @@ jobs:
           pixi global install minio-client
           pixi global install minio-server
 
-      # - name: Start minio server
-      #   if: ${{ !matrix.skip-tests && !startsWith(matrix.os, 'windows-') }}
-      #   run: |
-      #     minio server --help
-      #     mkdir -p ${{ runner.temp }}/minio-data
-      #     minio server --address 127.0.0.1:9000 ${{ runner.temp }}/minio-data &
-      #     sleep 5
-
-      # - name: Start minio server
-      #   if: ${{ !matrix.skip-tests && startsWith(matrix.os, 'windows-') }}
-      #   shell: powershell
-      #   run: |
-      #     minio server --help
-      #     mkdir ${{ runner.temp }}\minio-data
-      #     Start-Process "minio" -ArgumentList "server --address 127.0.0.1:9000 ${{ runner.temp }}\minio-data" -NoNewWindow
-      #     Start-Sleep -Seconds 5
-
-      # - name: Test minio server connection
-      #   if: ${{ !matrix.skip-tests }}
-      #   run: curl -I http://localhost:9000/minio/health/live
-
       - name: Run tests
         if: ${{ !matrix.skip-tests }}
         env:
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -185,12 +185,7 @@ jobs:
           minio server --address 127.0.0.1:9000 ${{ runner.temp }}/minio-data &
           sleep 5
           curl -I http://localhost:9000/minio/health/live
-          cargo nextest run \
-          --workspace \
-          --features ${{ env.DEFAULT_FEATURES }} \
-          --target ${{ matrix.target }} \
-          ${{ steps.build-options.outputs.CARGO_BUILD_OPTIONS }} \
-          ${{ steps.test-options.outputs.CARGO_TEST_OPTIONS }}
+          cargo nextest run --workspace --features ${{ env.DEFAULT_FEATURES }} --target ${{ matrix.target }} ${{ steps.build-options.outputs.CARGO_BUILD_OPTIONS }} ${{ steps.test-options.outputs.CARGO_TEST_OPTIONS }}
 
       - name: Run doctests
         if: ${{ !matrix.skip-tests }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -166,10 +166,11 @@ jobs:
         run: |
           minio server --help
           mkdir ${{ runner.temp }}\minio-data
-          Start-Process "minio" -ArgumentList "server --address 127.0.0.1:9000 ${{ runner.temp }}\minio-data" -Wait -NoNewWindow
+          Start-Process "minio" -ArgumentList "server --address 127.0.0.1:9000 ${{ runner.temp }}\minio-data" -NoNewWindow
           Start-Sleep -Seconds 5
 
       - name: Test minio server connection
+        if: ${{ !matrix.skip-tests }}
         run: curl -I http://localhost:9000/minio/health/live
 
       - name: Run tests
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -153,13 +153,24 @@ jobs:
           pixi global install minio-server
 
       - name: Start minio server
-        if: ${{ !matrix.skip-tests }}
+        if: ${{ !matrix.skip-tests && !startsWith(matrix.os, 'windows-') }}
         run: |
           minio server --help
           mkdir -p ${{ runner.temp }}/minio-data
           minio server --address 127.0.0.1:9000 ${{ runner.temp }}/minio-data &
           sleep 5
-          curl -I http://localhost:9000/minio/health/live
+
+      - name: Start minio server
+        if: ${{ !matrix.skip-tests && startsWith(matrix.os, 'windows-') }}
+        shell: powershell
+        run: |
+          minio server --help
+          mkdir ${{ runner.temp }}\minio-data
+          Start-Process "minio" -ArgumentList "server --address 127.0.0.1:9000 ${{ runner.temp }}\minio-data" -Wait -NoNewWindow
+          Start-Sleep -Seconds 5
+
+      - name: Test minio server connection
+        run: curl -I http://localhost:9000/minio/health/live
 
       - name: Run tests
         if: ${{ !matrix.skip-tests }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -152,8 +152,6 @@ jobs:
           pixi global install minio-client
           pixi global install minio-server
           pixi run minio server --address 0.0.0.0:9000 ${{ runner.temp }}/minio-data &
-          MINIO_PID=$(echo $!)
-          echo "MINIO_PID=${MINIO_PID}" >> $GITHUB_ENV
 
       - name: Run tests
         if: ${{ !matrix.skip-tests }}
@@ -166,10 +164,6 @@ jobs:
           --target ${{ matrix.target }}
           ${{ steps.build-options.outputs.CARGO_BUILD_OPTIONS }}
           ${{ steps.test-options.outputs.CARGO_TEST_OPTIONS }}
-
-      - name: Stop minio
-        if: ${{ always() && !matrix.skip-tests }}
-        run: kill -9 -$MINIO_PID
 
       - name: Run doctests
         if: ${{ !matrix.skip-tests }}
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -165,6 +165,8 @@ jobs:
         if: ${{ !matrix.skip-tests }}
         env:
           GOOGLE_CLOUD_TEST_KEY_JSON: ${{ secrets.GOOGLE_CLOUD_TEST_KEY_JSON }}
+          RATTLER_TEST_R2_ACCESS_KEY_ID: ${{ secrets.RATTLER_TEST_R2_ACCESS_KEY_ID }}
+          RATTLER_TEST_R2_SECRET_ACCESS_KEY: ${{ secrets.RATTLER_TEST_R2_SECRET_ACCESS_KEY }}
         run: >
           cargo nextest run
           --workspace
--- a/.github/workflows/rust-compile.yml
+++ b/.github/workflows/rust-compile.yml
@@ -149,8 +149,8 @@ jobs:
           --workspace
           --features ${{ env.DEFAULT_FEATURES }},experimental_extras
           --target ${{ matrix.target }}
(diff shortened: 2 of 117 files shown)


crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,320 @@
...
+use aws_config::BehaviorVersion;
+use aws_sdk_s3::presigning::{PresignedRequest, PresigningConfig};
+use reqwest::{Request, Response};
+use reqwest_middleware::{Middleware, Next, Result as MiddlewareResult};
+use url::Url;
+
+/// S3 middleware to authenticate requests.
+pub struct S3Middleware {
+    config_file: Option<PathBuf>,
+    profile: Option<String>,
```
pavelzw, line 14: while theoretically you might be able to configure S3 with only env variables like `AWS_CONFIG_PATH` and `AWS_PROFILE`, there might be a scenario where you want to let pixi use a specific profile while still having a different `AWS_PROFILE` (which then would get used by things like terraform in `pixi run`)

that's why we opted to add these two variables (which can be set in `pixi.toml` under `s3.config-file` and `s3.profile`) to override `AWS_CONFIG_PATH` and `AWS_PROFILE`

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,320 @@
...
+    force_path_style: Option<bool>,
```
pavelzw, line 15: services like minio don't use `https://<bucket-name>.hostname/item.txt` but `https://hostname/bucket-name/item.txt` to provide objects. by default, aws doesn't use the path style.

interestingly, awscli manages to infer this itself; will take a look how it does this soon and whether we should implement this too.

AFAIK you can't specify `use-path-style` in the aws config file :/

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,320 @@
...
+        }
+    }
+
+    /// Generate a presigned S3 `GetObject` request.
+    async fn generate_presigned_s3_request(
+        &self,
+        bucket_name: &str,
+        key: &str,
+    ) -> MiddlewareResult<PresignedRequest> {
+        let client = create_client(
```
pavelzw, line 71: if this wasn't async, we could also implement it in `S3iddleware::new` :/
now we need to instantiate this on every call which isn't ideal

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,320 @@
...
+            std::env::set_var(key, value);
+        }
+        f().await;
+        for (key, _) in env {
+            std::env::remove_var(key);
+        }
+    }
+
+    #[tokio::test]
+    #[serial]
```
pavelzw, line 133: these tests need to be serial in order for the env variables not overwriting each other

crates/rattler_networking/src/s3_middleware.rs
pavelzw, line 1: a current limitation is that we need to have one shared credential for each bucket we access with this setup.
we could also make the `config-file` and `profile` channel-dependent but i think as a first version, this is okay, wdyt?

.github/workflows/rust-compile.yml
```diff
@@ -139,17 +138,27 @@ jobs:
...
 
+      - name: Set up pixi
+        uses: prefix-dev/setup-pixi@v0.8.1
+        with:
+          run-install: false
+
+      - name: Install minio
+        run: |
+          pixi global install minio-server
+          pixi global install minio-client
```
pavelzw, line 149: depends on https://github.com/conda-forge/staged-recipes/pull/28736

crates/rattler_networking/Cargo.toml
```diff
@@ -11,10 +11,11 @@ license.workspace = true
 readme.workspace = true
 
 [features]
-default = ["native-tls"]
+default = ["native-tls", "s3"]
```
pavelzw, line 14:
```suggestion
default = ["native-tls"]
```

probably this makes more sense?

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,341 @@
...
+pub struct S3 {
+    auth_storage: AuthenticationStorage,
+    config: S3Config,
+    expiration: std::time::Duration,
+}
+
+/// S3 middleware to authenticate requests.
+#[derive(Clone, Debug)]
+pub struct S3Middleware {
+    s3: S3,
```
pavelzw, line 41: we created this extra abstraction instead of putting auth_storage, config, expiration in `S3Middleware` directly since we plan to re-use it in `rattler-build upload s3`

pavelzw:
this should be ready for a first round of reviews @baszalmstra. not much missing anymore here

crates/rattler/src/cli/auth.rs
```diff
@@ -24,6 +24,18 @@ pub struct LoginArgs {
     /// The token to use on anaconda.org / quetz authentication
     #[clap(long)]
     conda_token: Option<String>,
+
+    /// The AWS access key ID
+    #[clap(long)]
+    aws_access_key_id: Option<String>,
```
pavelzw, line 30: maybe `access_key_id` would be more fitting...

crates/rattler_networking/src/authentication_storage/storage.rs
```diff
@@ -159,6 +159,32 @@ impl AuthenticationStorage {
...
+
+                        match possible_rest {
+                            Some(rest) => {
+                                if let Ok(new_url) = Url::parse(rest) {
+                                    current_url = new_url;
+                                } else {
+                                    return Ok((url, None));
+                                }
+                            }
+                            _ => return Ok((url, None)), // No more subdomains to check
```
delsner, line 179:
```suggestion
                            _ => return Ok((url, None)),  // No more subpaths to check
```

crates/rattler_networking/src/s3_middleware.rs
delsner: This is usually specified in the AWS config file via `s3.addressing_style`, but is currently not supported by the rust AWS SDK: https://github.com/awslabs/aws-sdk-rust/issues/1230

crates/rattler/src/cli/auth.rs
delsner: Changed to `s3_*`.

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -80,41 +83,62 @@ impl S3 {
...
+                            None,
+                            "pixi",
+                        ))
+                }
                 (_, Some(_)) => {
-                    panic!("Unsupported authentication method"); // todo proper error message
+                    return Err(anyhow::anyhow!("Unsupported authentication method"));
+                }
+                (_, None) => {
+                    eprintln!("No authentication found");
```
pavelzw, line 104: this should probably be a logging message instead of stderr

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -80,41 +83,62 @@ impl S3 {
...
-            let config = aws_sdk_s3::config::Builder::from(&sdk_config).build();
-            aws_sdk_s3::Client::from_conf(config)
+            let sdk_config = aws_config::defaults(BehaviorVersion::latest()).load().await;
+            let mut s3_config_builder = aws_sdk_s3::config::Builder::from(&sdk_config);
+            // Infer if we expect path-style addressing from the endpoint URL.
+            if let Some(endpoint_url) = sdk_config.endpoint_url() {
+                // If the endpoint URL is localhost, we most likely have to use path-style addressing.
+                if endpoint_url.starts_with("http://localhost") {
+                    s3_config_builder = s3_config_builder.force_path_style(true);
+                }
```
pavelzw, line 125: this seems overly specific; there are more cases where this happens; i would just add a comment to https://github.com/awslabs/aws-sdk-rust/issues/1230 here

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -325,12 +340,78 @@ region = eu-central-1
...
+        );
+        assert_eq!(presigned.scheme(), "http");
+        assert_eq!(presigned.host_str().unwrap(), "localhost");
+        assert!(presigned.query().unwrap().contains("X-Amz-Credential"));
+    }
+
+    #[rstest]
+    #[tokio::test]
+    #[serial]
+    async fn test_presigned_s3_request_public_bucket() {
```
pavelzw, line 395: we should also have a test that ensures public buckets work with `S3Config::FromAWS`

crates/rattler_networking/tests/integration_test.rs
```diff
@@ -0,0 +1,220 @@
...
+aws_secret_access_key = minioadmin
+endpoint_url = http://localhost:9000
+region = eu-central-1
+"#;
+    let aws_config_path = temp_dir.path().join("aws.config");
+    std::fs::write(&aws_config_path, aws_config).unwrap();
+    (temp_dir, aws_config_path)
+}
+
+/* -------------------------------------- TESTS ------------------------------------- */
```
pavelzw, line 148: we should add a test for public buckets here as wll (both using `S3Config::FromAWS` and `S3Config::Custom`

crates/rattler_networking/src/s3_middleware.rs
pavelzw: ah, i'm assuming you added this because otherwise the minio integration tests are failing, right?

delsner: Yes, that's why I need it

delsner: I still added an xref here.

crates/rattler_networking/tests/integration_test.rs
delsner: Done

crates/rattler/src/cli/auth.rs
```diff
@@ -110,6 +130,18 @@ fn login(args: LoginArgs, storage: AuthenticationStorage) -> Result<(), Authenti
         }
     } else if let Some(token) = args.token {
         Authentication::BearerToken(token)
+    } else if let Some(access_key_id) = args.s3_access_key_id {
```
delsner, line 133:
```suggestion
    } else if let (Some(access_key_id), Some(secret_access_key)) = (args.s3_access_key_id, args.s3_secret_access_key) {
```

maybe like this @pavelzw? It's a bit asymmetric right now that we raise a dedicated error for `access_key_id != None && secret_access_key == None` and not for `secret_access_key != None && access_key_id == None`. This way, we wouldn't raise a custom error, but just fall back to `AuthenticationCLIError::NoAuthenticationMethod`.

crates/rattler_networking/src/s3_middleware.rs
delsner: Done

crates/rattler_networking/src/s3_middleware.rs
delsner: Done

crates/rattler_networking/tests/integration_test.rs
```diff
@@ -0,0 +1,311 @@
...
+            args
+        );
+        MinioServer { handle, directory }
+    }
+}
+
+impl Drop for MinioServer {
+    fn drop(&mut self) {
+        eprintln!("Shutting down Minio server (PID={})", self.handle.id());
+        match self.handle.kill() {
```
pavelzw, line 45: it looks like this isn't working properly on windows :/

Cargo.toml
```diff
@@ -72,6 +72,9 @@ getrandom = { version = "0.2.15", default-features = false }
 glob = "0.3.2"
 google-cloud-auth = { version = "0.17.2", default-features = false }
 google-cloud-token = "0.1.2"
+aws-config = "1.5.12"
+aws-sdk-s3 = "1.67.0"
+aws-runtime = "1.5.2"
```
baszalmstra, line 77: Should we add `default-features = false` to ensure we dont pull in many subdependencies?

crates/rattler/src/cli/auth.rs
```diff
@@ -24,6 +24,18 @@ pub struct LoginArgs {
     /// The token to use on anaconda.org / quetz authentication
     #[clap(long)]
     conda_token: Option<String>,
+
+    /// The S3 access key ID
+    #[clap(long)]
+    s3_access_key_id: Option<String>,
```
baszalmstra, line 30: We can mark these as mutually exclusive with `conda_token` etc. See https://docs.rs/clap/latest/clap/struct.Arg.html#method.conflicts_with. Similarly we can also use [`required_if`](https://docs.rs/clap/latest/clap/struct.Arg.html#method.requires_if) to link the key id and access together.

crates/rattler_networking/src/s3_middleware.rs
baszalmstra: If you would add specific config per channel what would that look like? Does it change the API much?

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,468 @@
...
+    }
+
+    /// Generate a presigned S3 `GetObject` request.
+    async fn generate_presigned_s3_url(&self, url: Url) -> MiddlewareResult<Url> {
+        let client = self.create_s3_client(Some(url.clone())).await?;
+
+        let bucket_name = url.host_str().ok_or_else(|| {
+            reqwest_middleware::Error::middleware(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "Host should be present in S3 URL",
```
baszalmstra, line 154: please start library errors with lowercase. That makes it easier to concatenate errors like: "Something when wrong because, {}".

```suggestion
                "host is missing from s3:// URL",
```

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,468 @@
...
+            ))
+        })?;
+        let key = url.path().strip_prefix("/").ok_or_else(|| {
+            reqwest_middleware::Error::middleware(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                "Missing prefix",
+            ))
+        })?;
```
baszalmstra, line 162: When would this happen? If there is no path? The error is a bit unclear. Would it make sense to unwrap instead?

crates/rattler_networking/src/s3_middleware.rs
baszalmstra: Lets use this crate instead: https://crates.io/crates/temp-env

crates/rattler_networking/tests/integration_test.rs
delsner: Hm, then I'm going to rewrite the integration tests so that we no longer start minio as a subprocess from rust, but start it as a background service before running the integration tests.

baszalmstra (commented):
Thanks everyone for this PR, I left some small comments but overall this looks great!

.config/nextest.toml
```diff
@@ -0,0 +1,6 @@
+[test-groups]
+serial-integration = { max-threads = 1 }
+
+[[profile.default.overrides]]
+filter = 'package(rattler_networking) and (test(#test_minio*) | test(s3_middleware::tests))'
+test-group = 'serial-integration'
```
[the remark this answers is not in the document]
delsner: removed in favor of a global minio server instance for testing. Having a global instance has the advantage that we can run tests in parallel and is more robust especially on Windows, where killing of a subprocess sometimes fails.

crates/rattler/src/cli/auth.rs
```diff
@@ -26,15 +26,15 @@ pub struct LoginArgs {
     conda_token: Option<String>,
 
     /// The S3 access key ID
-    #[clap(long)]
+    #[clap(long, requires_all = ["s3_secret_access_key"], conflicts_with_all = ["host", "token","username", "password", "conda_token"])]
```
pavelzw, line 29:
```suggestion
    #[clap(long, requires_all = ["s3_secret_access_key"], conflicts_with_all = ["host", "token", "username", "password", "conda_token"])]
```

crates/rattler_networking/src/s3_middleware.rs
delsner: Yes, I think unwrapping makes more sense here.

crates/rattler_networking/src/s3_middleware.rs
delsner: Done!

crates/rattler_networking/src/s3_middleware.rs
delsner: > we could also make the `config-file` and `profile` channel-dependent

I think this comment was based on an outdated version of the code. Still, we do not support channel-specific configuration yet. Did you already think about an API @pavelzw?

pavelzw: from the API side, I could think of something like

```rs
#[derive(Clone, Debug)]
pub struct S3 {
    auth_storage: AuthenticationStorage,
    default_config: S3Config,
    // channel -> config map, overrides default_config
    channel_config: HashMap<String, S3Config>,
    expiration: std::time::Duration,
}
```

In `create_s3_client` we would gather the config for the respective bucket or falling back to the default config.

It can be configured in `pixi.toml` (or global config respectively)

```toml
[s3-config]
endpoint-url = "https://s3.aws..."
region = "eu-central-1"
force-path-style = false

[channel."s3://my-bucket".s3-config]
endpoint-url = "https://localhost:9000"
region = "eu-west-1"
force-path-style = true
```

From a pixi.toml perspective, this should be backwards compatible so I think it's fine to not implement this yet but only if the need arises.

crates/rattler_networking/Cargo.toml
delsner, line 14: We should probably remove this before merging.

crates/rattler_networking/src/s3_middleware.rs
delsner: Follow-up discussion here: https://github.com/prefix-dev/pixi/pull/2825#discussion_r1932413189

crates/rattler_networking/tests/integration_test.rs
pavelzw, line 1: it probably is a bit nicer if we refactor the integration tests to use cloudflare r2 instead of minio similar to how we do it in the pixi PR
this then wouldn't require additional setup

crates/rattler_networking/src/s3_middleware.rs
pavelzw: as discussed in the other PR, we will now do the config on a bucket level: [`876fbbe` (#1008)](https://github.com/conda/rattler/pull/1008/commits/876fbbeed7a1dec9ea6db762329d3e013e42cf00)

crates/rattler_networking/tests/integration_test.rs
```diff
@@ -0,0 +1,241 @@
...
+    assert_eq!(result.status(), 200);
+    let body = result.text().await.unwrap();
+    assert!(
+        body.contains("my-webserver-0.1.0-pyh4616a5c_0.conda"),
+        "body does not contain package: {}",
+        body
+    );
+}
+
+#[ignore]
```
pavelzw, line 127: this still needs the creation of a public R2 bucket with the permissions `s3:GetObject` and `s3:ListBucket`

.github/workflows/rust-compile.yml
```diff
@@ -139,17 +138,33 @@ jobs:
...
-          ${{ steps.build-options.outputs.CARGO_BUILD_OPTIONS}} 
-          ${{ steps.test-options.outputs.CARGO_TEST_OPTIONS}} 
+          RATTLER_TEST_R2_ACCESS_KEY_ID: ${{ secrets.RATTLER_TEST_R2_ACCESS_KEY_ID }}
+          RATTLER_TEST_R2_SECRET_ACCESS_KEY: ${{ secrets.RATTLER_TEST_R2_SECRET_ACCESS_KEY }}
+        run: |
+          minio server --help
+          mkdir -p ${{ runner.temp }}/minio-data
+          minio server --address 127.0.0.1:9000 ${{ runner.temp }}/minio-data &
+          sleep 5
+          curl -I http://localhost:9000/minio/health/live
```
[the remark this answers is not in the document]
pavelzw: we decided to use Cloudflare R2 for the integration tests as this requires less setup. https://github.com/conda/rattler/pull/1008#discussion_r1935658214

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,432 @@
...
+            endpoint_url,
+            region,
+            force_path_style,
+        } = self
+            .config
+            .get(bucket_name)
+            .unwrap_or(&S3Config::FromAWS)
+            .clone()
+        {
+            let auth = self.auth_storage.get_by_url(url).unwrap(); // todo
```
wolfv, line 85: woops! :)

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,432 @@
...
+    }
+
+    /// Generate a presigned S3 `GetObject` request.
+    async fn generate_presigned_s3_url(&self, url: Url) -> MiddlewareResult<Url> {
+        let client = self.create_s3_client(url.clone()).await?;
+
+        let bucket_name = url
+            .host_str()
+            .ok_or_else(|| anyhow::anyhow!("host should be present in S3 URL"))?;
+        let key = url.path().strip_prefix("/").unwrap();
```
wolfv, line 164: Maybe do `let Some(...)` instead of `unwrap`?

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,432 @@
...
+    async fn handle(
+        &self,
+        mut req: Request,
+        extensions: &mut http::Extensions,
+        next: Next<'_>,
+    ) -> MiddlewareResult<Response> {
+        if req.url().scheme() == "s3" {
+            let url = req.url().clone();
+            let presigned_url = self.s3.generate_presigned_s3_url(url).await?;
+            // panic!("{}", presigned_url);
```
wolfv, line 194: remove?

crates/rattler_networking/src/s3_middleware.rs
pavelzw: i think there we should do something `.ok_or_else(|| anyhow::anyhow!("invalid s3 url: {}", url))?` here instead

crates/rattler_networking/tests/integration_test.rs
pavelzw: done

wolfv (commented):
one more unwrap

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -0,0 +1,434 @@
...
+                            session_token,
+                            None,
+                            "pixi",
+                        ))
+                }
+                (_, Some(_)) => {
+                    return Err(anyhow::anyhow!("unsupported authentication method"));
+                }
+                (_, None) => {
+                    tracing::info!("No authentication found, assuming bucket is public");
```
pavelzw, line 112: tbh i don't think the concept of public buckets makes much sense 🤔
for public buckets you could just use the http url of the bucket in your `channels` instead. for example, you could use `https://my-public-bucket.s3.eu-central-1.amazonaws.com/channel`.

i think we should just fail on no credentials found and advise people to use https for public buckets

pavelzw: [`52add4e` (#1008)](https://github.com/conda/rattler/pull/1008/commits/52add4e02b370bb24a4604f68adc8534fda06b67)

PR #1010, opened 2025-01-03
feat: wire up network config for sandbox

wolfv:
Looks like this was forgotten which makes network always disabled.


wolfv:
I've tested this in rattler-build and it's working :)

PR #1012, opened 2025-01-05
feat: Add `rattler-index-s3`

delsner:


baszalmstra:
Im not sure if this is ready for review but Im wondering if it makes sense to use the `https://github.com/apache/opendal` instead and adapting `rattler-index`. That would allow us to plug loads of storage providers in there. (We also use this on `prefix.dev`).

delsner:
@baszalmstra this is not ready to review yet. Thanks for the hint on opndal, I'll check it out later.

crates/rattler-index-s3/src/main.rs
```diff
@@ -0,0 +1,79 @@
...
+    // Parse the command line arguments
+    let opt = Args::parse();
+
+    // Determine the logging level based on the the verbose flag and the RUST_LOG environment
+    // variable.
+    let default_filter = if opt.verbose {
+        LevelFilter::DEBUG
+    } else {
+        LevelFilter::INFO
+    };
```
pavelzw, line 58: we should IMO use clap_verbosity_flag here

delsner: I copied this from rattler-bin, but I agree with you.

delsner:
Closing this as I'll start working on an implementation using opendal on a separate branch.

PR #1019, opened 2025-01-08
feat: Add support for optional dependencies

prsabahrami:
### Description

I have updated the `matchspec` parser to parse the list of features. (e.g. `sympy[version=1.2, optional_features=[feat1, feat2]]` into [feat1, feat2]).
In addition, I added `SolverPackageRecord::RecordWithFeature` which represents a package with a single feature. Hence, the goal is to turn `sympy[feat1, feat2]` into `sympy[feat1], sympy[feat2]`. We also need to assure that both of these adhere to the correct version of sympy.


crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -146,6 +146,11 @@ pub struct PackageRecord {
     #[serde(skip_serializing_if = "NoArchType::is_none")]
     pub noarch: NoArchType,
 
+    /// Specifications of optional or dependencies. These are dependencies that are
+    /// only required if certain features are enabled or if certain conditions are met.
+    #[serde(default)]
+    pub optional_depends: BTreeMap<String, Vec<String>>,
```
baszalmstra, line 152: Using a `BTreeMap` is an interesting choice here. The `BTreeMap` guarentees a certain order for the keys (alphabetical), does this matter do you think? If not you could just use a `HashMap`. But maybe the input order matters, in which case you might want to use an `IndexMap`.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -543,7 +577,53 @@ impl<'a> DependencyProvider for CondaDependencyProvider<'a> {
...
+                    &mut parse_match_spec_cache,
+                ) {
+                    Ok(version_set_id) => version_set_id,
+                    Err(e) => {
+                        let reason = self.pool.intern_string(format!(
+                            "failed to create base package dependency: {e}",
+                        ));
+                        return Dependencies::Unknown(reason);
+                    }
+                };
```
baszalmstra, line 621: Instead of parsing you could also instantiate a MatchSpec where you set the name, version, build_string and number, and subdir. Or perhaps just the sha256 hash, or all?

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -651,16 +747,33 @@ impl super::SolverImpl for Solver {
                 .intern_version_set(name_id, NamelessMatchSpec::default().into())
         });
 
-        let root_requirements = task.specs.iter().map(|spec| {
+        let root_requirements = task.specs.iter().flat_map(|spec| {
             let (name, nameless_spec) = spec.clone().into_nameless();
+            let features = spec.optional_features.clone();
```
baszalmstra, line 752: Is this clone required?

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -651,16 +747,33 @@ impl super::SolverImpl for Solver {
...
+                for feature in features {
+                    // Create a version set that matches the feature-enabled package
+                    let package_name_with_feature =
+                        format!("{}[{}]", name.as_normalized(), feature);
+                    let feature_name_id =
+                        provider.pool.intern_package_name(package_name_with_feature);
+                    let feature_version_set = provider
+                        .pool
+                        .intern_version_set(feature_name_id, nameless_spec.clone().into());
+                    reqs.push(feature_version_set);
```
baszalmstra, line 770: You could also just clone the `spec` and add the optional dependency perhaps? Instead of parsing.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -543,7 +577,53 @@ impl<'a> DependencyProvider for CondaDependencyProvider<'a> {
             dependencies.requirements.push(version_set_id.into());
```
baszalmstra, line 577: I think you need to split each matchspec into multiple solver specs if you have optional dependencies. E.g. the spec (in python terms) `boltons[foo, bar]` should result in 3 requirements: boltons, boltons with `foo`, and boltons with `bar`. Because one requirement can only ever match a single solvable, so to get the 3 solvables mentioned above you also need to add 3 requirements.

tdejager:
@baszalmstra @prsabahrami  Do you think we need to version the `MatchSpec` implementation in rust with this feature addition? Or maybe have an explicit cargo feature that marks this as an extension? That way consumers of the libary cannot seralize into matchspecs that cannot be parse by e.g conda?

wolfv:
I think we might even want to gate the whole feature behind a compile time option for now? Until we have a CEP that passed?

crates/rattler_conda_types/src/repo_data/mod.rs
prsabahrami: Oh fair point! That's true. It would be important for the lockfile. I believe `IndexMap` takes O(1) on avg and keeps the elements in the order they were inserted whereas `BTreeMap` sorts them. I'll switch to `IndexMap` for this.

wolfv: Indeed: https://github.com/indexmap-rs/indexmap/issues/288
So maybe just stick to BTreeMap.

wolfv: Although we might also change this in the future to have a different format altogether :)

prsabahrami:
@baszalmstra I have updated the pool to take in `NameType` which an enum of either single `String` or a pair of name and feature. Please let me know what you think!

crates/rattler_conda_types/src/snapshots/rattler_conda_types__prefix_record__test__pysocks-1_7_1-pyh0701188_6_json.snap
```diff
@@ -14,6 +13,7 @@ license_family: BSD
 md5: 56cd9fe388baac0e90c7149cfac95b60
 name: pysocks
 noarch: python
+optional_depends: {}
```
baszalmstra, line 16: mm strange, since you made it so that empty optional dependencies are skipped I would not expect this here.

crates/rattler_lock/src/parse/models/v6/conda_package_data.rs
```diff
@@ -157,6 +157,7 @@ impl<'a> TryFrom<CondaPackageDataModel<'a>> for CondaPackageData {
             build_number,
             constrains: value.constrains.into_owned(),
             depends: value.depends.into_owned(),
+            optional_depends: std::collections::BTreeMap::new(),
```
baszalmstra, line 160: I think you should also add this to the serialization logic so that we also have optional dependencies in the lock-files!

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -555,7 +714,9 @@ impl<'a> DependencyProvider for CondaDependencyProvider<'a> {
                         return Dependencies::Unknown(reason);
                     }
                 };
-            dependencies.constrains.push(version_set_id);
+            for version_set_id in version_set_id {
+                dependencies.constrains.push(version_set_id);
+            }
```
baszalmstra, line 719: Might be better to use `extend` to ensure we dont push individual items and allocate multiple times.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -522,28 +621,88 @@ impl<'a> DependencyProvider for CondaDependencyProvider<'a> {
...
+                                "the dependency '{depends}' failed to parse: {e}",
+                            ));
+
+                            return Dependencies::Unknown(reason);
+                        }
+                    };
+
+                for version_set_id in version_set_id {
+                    dependencies.requirements.push(version_set_id.into());
+                }
```
baszalmstra, line 701: Might be better to use `extend` to ensure we dont push individual items and allocate multiple times.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -297,94 +382,108 @@ impl<'a> CondaDependencyProvider<'a> {
...
+                // Add feature-enabled solvables
+                for feature in record.package_record.optional_depends.keys() {
+                    let feature_solvable = pool.intern_solvable(
+                        package_name,
+                        SolverPackageRecord::RecordWithFeature(record, feature.clone()),
+                    );
+                    let package_name_with_feature = pool.intern_package_name((
+                        record.package_record.name.as_normalized(),
+                        feature.clone(),
+                    ));
```
baszalmstra, line 398: I think you need to use the WithFeature variant here.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -522,28 +610,89 @@ impl<'a> DependencyProvider for CondaDependencyProvider<'a> {
...
+                        }
+                    };
+
+                dependencies
+                    .requirements
+                    .extend(version_set_id.into_iter().map(Requirement::from));
+            }
         }
 
-        for constrains in rec.package_record.constrains.iter() {
```
baszalmstra, line 546: This only needs to happen for base packages.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -49,17 +50,22 @@ impl<'a> FromIterator<&'a RepoDataRecord> for RepoData<'a> {
...
+    feature: Option<String>,
     _marker: PhantomData<&'a PackageRecord>,
 }
 
 impl<'a> From<NamelessMatchSpec> for SolverMatchSpec<'a> {
     fn from(value: NamelessMatchSpec) -> Self {
+        let feature = value
+            .optional_features
+            .as_ref()
+            .and_then(|features| features.first().cloned());
```
baszalmstra, line 65: I guess this should be `None`?

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -575,7 +726,23 @@ impl<'a> DependencyProvider for CondaDependencyProvider<'a> {
...
+                    SolverPackageRecord::Record(rec) => {
+                        // Base package matches if spec matches and no features are required
+
+                        spec.matches(*rec) != inverse
+                    }
+                    SolverPackageRecord::RecordWithFeature(rec, feature) => {
+                        // Feature-enabled package matches if spec matches and feature is required
+
+                        if spec.matches(*rec) {
+                            spec.optional_features
```
baszalmstra, line 738: This should be `SolverMatchSpec::feature`!

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -708,23 +893,47 @@ impl super::SolverImpl for Solver {
...
+) -> Result<Vec<VersionSetId>, ParseMatchSpecError> {
+    let match_spec = MatchSpec::from_str(spec_str, ParseStrictness::Lenient)?;
+    let (name, spec) = match_spec.into_nameless();
+
+    let mut version_set_ids = vec![];
+
+    if let Some(ref features) = spec.optional_features {
+        for feature in features {
+            let mut spec_with_feature = spec.clone();
+            spec_with_feature.optional_features = Some(vec![feature.to_string()]);
```
baszalmstra, line 909: This should be `SolverMatchSpec::feature`

crates/rattler_solve/src/resolvo/mod.rs
prsabahrami: We need to have an assertion here to make sure that only 1 feature is in the `optional_features` of `value` and then select the only one as the `SolverMatchSpec::feature`. I'm not sure why it would be None.

baszalmstra: It should be None initially. In parse_match_spec we then explicitly add a feature.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -708,23 +897,47 @@ impl super::SolverImpl for Solver {
...
+            return Ok(vec![*spec_id]);
+        }
+
         let dependency_name = pool.intern_package_name(
             name.as_ref()
-                .expect("match specs without names are not supported")
+                .expect("Packages with no name are not supported")
                 .as_normalized(),
         );
         let version_set_id = pool.intern_version_set(dependency_name, spec.into());
```
baszalmstra, line 936: Because of the `From` impl this will also insert a set a feature for the SolverMatchSpec!

I would instead change this function such that you create a `SolverMatchSpec` for the NamelessMatchSpec. Without a feature. Then clone that `SolverMatchSpec` for every feature and explicitly set the `feature` field.

prsabahrami:
@baszalmstra I wrote a function `with_feature` for `SolveMatchSpec` which would clone the spec and set the feature directly.

wolfv:
I think my idea wrt to making this optional was also to introduce a third `ParseStrictness` mode: `Experimental` - and only parse the brackets in experimental mode. Then later we could enable this mode when ingesting MatchSpecs from environment files or in `rattler::create`.

On the other hand that might make things ugly in the meantime so guarding it behind a feature flag might be better indeed.

crates/rattler_conda_types/src/match_spec/matcher.rs
```diff
@@ -45,6 +45,12 @@ impl PartialEq for StringMatcher {
     }
 }
 
+impl From<String> for StringMatcher {
```
wolfv, line 48: Isn't this confusing with the `from_str` implementation?

You can call `let foo: StringMatcher = "foo".parse().unwrap();` (same for a String instance I believe).

crates/rattler_conda_types/src/repo_data_record.rs
```diff
@@ -24,6 +24,17 @@ pub struct RepoDataRecord {
...
     /// TODO: Refactor this into `Source` which can be a "name", "channelurl", or "direct url".
     pub channel: Option<String>,
+
+    /// The selected feature set for this package.
+    pub selected_feature: Option<String>,
+}
+
+impl RepoDataRecord {
+    /// Set the selected feature set for this package.
+    pub fn set_selected_feature(&mut self, selected_feature: String) -> &mut Self {
```
wolfv, line 34: why is this singular and only a String? Do we even need to store this information?

prsabahrami: This is used for printing the transactions after resolving along with the name.

crates/rattler_conda_types/src/match_spec/matcher.rs
prsabahrami: Ah yes, I forgot to remove it. We don't need it actually.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -697,8 +905,13 @@ impl super::SolverImpl for Solver {
         let required_records = solvables
```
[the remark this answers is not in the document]
prsabahrami: I initially tried creating an enum for `RepoDataRecord` which would be either a `RepoDataRecord` or a `(RepoDataRecord, String)` but it turns out to be quite considerably messy as `print_transaction` would then require cloning the objects. I can give it another try and see if it becomes easier with a separate list of `Option<String>`.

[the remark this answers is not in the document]
wolfv: You could first gather all the repodata records, and in a second run you could associate all the features so that it's one repodata record with a list of features.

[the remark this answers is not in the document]
prsabahrami: @wolfv @baszalmstra Just changed the return type of `solve` to return a HashMap of `RepoDataRecord` and list of features. Let me know what you think!

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -694,37 +902,72 @@ impl super::SolverImpl for Solver {
...
-    pool: &Pool<SolverMatchSpec<'a>>,
-    spec_str: &'a str,
-    parse_match_spec_cache: &mut HashMap<&'a str, VersionSetId>,
-) -> Result<VersionSetId, ParseMatchSpecError> {
-    if let Some(spec_id) = parse_match_spec_cache.get(spec_str) {
-        Ok(*spec_id)
+fn parse_match_spec(
+    pool: &Pool<SolverMatchSpec<'_>, NameType>,
+    spec_str: &str,
+    parse_match_spec_cache: &mut HashMap<String, VersionSetId>,
```
baszalmstra, line 931:
```suggestion
    parse_match_spec_cache: &mut HashMap<String, Vec<VersionSetId>>,
```

we can cache all the version sets that are returned by this function. Not just the versionset for the base requirement.

baszalmstra (commented):
I think we are close! This review is mostly nitpicks!

crates/rattler_conda_types/src/repo_data_record.rs
```diff
@@ -31,3 +34,6 @@ impl AsRef<PackageRecord> for RepoDataRecord {
         &self.package_record
     }
 }
+
+/// Type alias for the solver result containing records and their features
+pub type SolverResult = HashMap<RepoDataRecord, Option<Vec<String>>>;
```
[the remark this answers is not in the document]
prsabahrami: Just to confirm, you mean repodata records with a HashMap that maps packages names to selected features?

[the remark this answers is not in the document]
baszalmstra: I mean something like:

```rust
struct Solution {
    record: Vec<RepodataRecord>,
    features: HashMap<PackageName, Vec<String>>,
}
```

Since package names must be unique I think this would work just fine.

prsabahrami:
@baszalmstra Should I update the CI to test the tests under the feature flag as well?

baszalmstra:
Yeah makes sense! You could add them to a seperate file/module

crates/rattler_solve/src/resolvo/mod.rs
prsabahrami: Wouldn't be this a bit inefficient if lets say we have `pack[feat1, feat2]` and then `pack[feat2, feat3]` and we are essentially adding the version set for `pack[feat2]` twice. 
I have made the change but let me know if I am missing something.

baszalmstra: The most important thing to cache is the parsing which is relatively slow. Since the optional features are part of the input string it makes sense to me to cache all the versionsets the result from it. 

The current implementation also doesnt cache the parsing, I think that part should be fixed.

prsabahrami: Makes sense.
Yes – I just pushed a fix. Let me know what you think!

test-data/channels/dummy-optional-dependencies/noarch/repodata.json
```diff
@@ -46,6 +46,15 @@
           ],
           "with-latest-bors": [
             "bors >=2.0"
+          ],
+          "legacy-only": [
+            "bar >=1.0"
+          ],
+          "with-recursive": [
+            "foo[optional_features=[with-bar]]"
```
wolfv, line 54: interesting! Just to be sure, this package itself doesn't seem to have the feature. Should this then fail to resolve?

prsabahrami: The package foo has the feature `with-bar` but because they are different versions and there is no version that overlaps between the feature `with-recursive` and `with-bar`, it should fail to resolve.

.github/workflows/rust-compile.yml
```diff
@@ -155,3 +155,88 @@ jobs:
         if: ${{ !matrix.skip-tests }}
         run: >
           cargo test --doc
+
+  test-experimental:
```
baszalmstra, line 159: Ah I think I misunderstood what you meant initially. I think we should just enable the experimental feature for the regular tests. We dont need a whole new job.

baszalmstra (commented):
Im still thinking about the name optional_features. I wonder if its descriptive enough.

Im a little on the fence about the name features. It implies that a certain "feature" is enabled but in reality the only thing that changes is that an extra set of dependencies is required. This might ofcourse cause new functionality to be unlocked but not necessarily.

Im starting to think that the name extras might be better.. Wdyt?

Also optional is a bit strange in this context. Optional to me would imply that a solution without the optional features is also valid, where it is in fact not. Perhaps in the matchspec we could just drop this part? E.g. `foobar[features=[baz]]`

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -173,6 +176,15 @@ impl Display for MatchSpec {
...
         }
 
+        match &self.extras {
+            Some(dependencies) => {
+                write!(f, "[")?;
+                write!(f, "{}", dependencies.iter().format(", "))?;
+                write!(f, "]")?;
+            }
+            None => {}
+        }
```
baszalmstra, line 186: THis will conflict if you also have for instance an `md5`. You should add `dependencies.iter...` as an elements into the `keys` vec below.

crates/rattler_conda_types/Cargo.toml
```diff
@@ -12,6 +12,7 @@ readme.workspace = true
 
 [features]
 default = ["rayon"]
+extras = []
```
baszalmstra, line 15: Lets rename this to `experimental_extras` to indicate that this is still experimental. We will remove the feature if we deem it stable enough.

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -265,6 +278,8 @@ pub struct NamelessMatchSpec {
     pub build_number: Option<BuildNumberSpec>,
     /// Match the specific filename of the package
     pub file_name: Option<String>,
+    /// The selected optional features of the package
```
baszalmstra, line 281:
```suggestion
    /// Optional extra dependencies to select for the package
```

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -1375,4 +1425,39 @@ mod tests {
...
     }
+
+    #[cfg(feature = "extras")]
+    #[test]
+    fn test_simple_extras() {
+        let spec = MatchSpec::from_str("foo[extras=[bar]]", Strict).unwrap();
+
+        assert_eq!(spec.extras, Some(vec!["bar".to_string()]));
+        assert!(MatchSpec::from_str("foo[extras=[bar,baz]", Strict).is_err());
+    }
```
baszalmstra, line 1436: Lets also add some test cases for failing cases. E.g. `foo=[extras=]` or `foo=[extras=bar,baz]` or `foo=[extras=[bar,baz]`, `foo=[extras=[bar,]]`

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -146,6 +146,11 @@ pub struct PackageRecord {
     #[serde(skip_serializing_if = "NoArchType::is_none")]
     pub noarch: NoArchType,
 
+    /// Specifications of optional or dependencies. These are dependencies that are
+    /// only required if certain features are enabled or if certain conditions are met.
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub optional_depends: BTreeMap<String, Vec<String>>,
```
baszalmstra, line 152: We can also call this `extras` I think?

crates/rattler_solve/Cargo.toml
```diff
@@ -43,6 +44,7 @@ url = { workspace = true }
 default = ["resolvo"]
 libsolv_c = ["rattler_libsolv_c", "libc"]
 resolvo = ["dep:resolvo", "dep:futures"]
+extras = []
```
baszalmstra, line 47: Same rename here.

crates/rattler_solve/src/libsolv_c/mod.rs
```diff
@@ -94,7 +94,7 @@ impl super::SolverImpl for Solver {
     >(
         &mut self,
         task: SolverTask<TAvailablePackagesIterator>,
-    ) -> Result<Vec<RepoDataRecord>, SolveError> {
+    ) -> Result<SolverResult, SolveError> {
```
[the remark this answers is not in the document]
baszalmstra: Can you create an issue for this so we dont forget?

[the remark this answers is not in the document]
baszalmstra: Furthermore, maybe it would be good to error out if a matchspec with extras is used. That way we provide a clear error message when people try it.

crates/rattler_solve/tests/snapshots/backends__resolvo__solve_dummy_repo_optional_depends_feature_conflict.snap
```diff
@@ -0,0 +1,11 @@
+---
+source: crates/rattler_solve/tests/backends.rs
+expression: result.unwrap_err()
+---
+Cannot solve the request because of: The following packages are incompatible
+├─ foo >=4.0 can be installed with any of the following options:
+│  └─ foo 4.0.2
+└─ foo[with-bar] * cannot be installed because there are no viable options:
+   └─ foo[with-bar] 3.0.2 would require
+      └─ foo ==3.0.2 py36h1af98f8_1[md5=fb731d9290f0bcbf3a054665f33ec94f, sha256=67a63bec3fd3205170eaad532d487595b8aaceb9814d13c6858d7bac3ef24cd4], which cannot be installed because there are no viable options:
```
baszalmstra, line 10: Mm this error is pretty uninformative. Especially if there will be multiple solvables that provide that extra. It will probably list every single solvable and extra here. 

Perhaps it would be possible to merge all the SolvableWithFeature solvables of a single version into a single solvable? I assume that will result in a much clearer error message and reduce the total number of solvables..

crates/rattler_solve/tests/snapshots/backends__resolvo__solve_dummy_repo_optional_depends_recursive_feature.snap
```diff
@@ -0,0 +1,13 @@
+---
+source: crates/rattler_solve/tests/backends.rs
+expression: result.unwrap_err()
+---
+Cannot solve the request because of: The following packages are incompatible
+└─ foo[with-recursive] * cannot be installed because there are no viable options:
+   └─ foo[with-recursive] 2.0.2 would require
+      ├─ foo ==2.0.2 py36h1af98f8_1[md5=d65ab674acf3b7294ebacaec05fc5b54, sha256=1154fceeb5c4ee9bb97d245713ac21eb1910237c724d2b7103747215663273c2], which can be installed with any of the following options:
+      │  └─ foo 2.0.2
+      └─ foo[with-bar] *, which cannot be installed because there are no viable options:
```
baszalmstra, line 10: This error message is also kinda vague.. If I read it, it states that `foo[with-recursive] *` requires, `foo[with-bar]`??

PR #1030, opened 2025-01-22
feat: lock when indexing

wolfv:
I shamefully copied the code from `rattler_repodata_gateway` to add a `LockedFile` to the indexing. 

We would like to prevent multiple instances of e.g. `rattler-build` to write to the `repodata.json` at the same time.

A [companion PR is in rattler-build](https://github.com/prefix-dev/rattler-build/pull/1372) about compressing the packages into a temporary file, and then atomically renaming them once they are finished.

Both are made in order to improve the experience when running multiple builds in parallel.

### DIFF
--- a/crates/rattler_index/Cargo.toml
+++ b/crates/rattler_index/Cargo.toml
@@ -11,7 +11,9 @@ license.workspace = true
 readme.workspace = true
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


wolfv:
Superseded by #1031

PR #1034, opened 2025-01-29
feat: Expose FetchRepoDataOptions to py-rattler

tl-hbk:
### Description

I sometimes run into zstd decoding errors when trying to fetch repodata and was wanting to be able to disable zstd from py-rattler.  This PR exposes the FetchRepoDataOptions struct in rust to python.

I originally had the python class in `gateway.py` but that was causing a circular import when trying to import it in `fetch_repo_data.py` so I moved it.


baszalmstra (approved):

baszalmstra:
Thanks!

PR #1038, opened 2025-02-01
fix: add track-features conda type

wolfv:
This adds a new `TrackFeatures` wrapper type.

Track features need to be serialized as a comma delimited string.

This still needs tests!

PR #1039, opened 2025-02-01
fix: track feature serialization and deserialization

baszalmstra:
This fixes serialization and deserialization of track features. We serialized/deserialized this as either a string or a list of strings but in fact, [the "standard"](https://github.com/conda/schemas/blob/main/repodata-record-1.schema.json#L115) is to use a _**space-separated string**_. Deserialization is backward compatible with a list. But serialization will now always produce a space-separated string.

This _only_ affects repodata and index.json format. Lock-files will still function the same way since they have their own parsing logic, and I think a list makes much more sense.

https://github.com/conda/conda-build/issues/5608


baszalmstra:
Supersedes #1038

wolfv:
Indeed, I just tried with conda-build:

```json
        "ttfeat-0.1.0-0.tar.bz2": {
            "build": "0",
            "build_number": 0,
            "depends": [],
            "md5": "f2dc829d0f36e4c125015591d07db7fc",
            "name": "ttfeat",
            "sha256": "f98a973b21a4943fc2872c7cb907d07e6219386509551d953b2e8062b1c37a01",
            "size": 2779,
            "subdir": "osx-arm64",
            "timestamp": 1738476674481,
            "track_features": "python tensorflow foobar",
            "version": "0.1.0"
        },
```

crates/rattler_conda_types/src/utils/serde.rs
```diff
@@ -152,3 +155,45 @@ pub(crate) fn sort_map_alphabetically<T: Serialize, S: serde::Serializer>(
         .collect::<BTreeMap<_, _>>()
         .serialize(serializer)
 }
+
+/// A helper to serialize and deserialize `track_features` in repodata. Track
+/// features are expected to be a comma seperated list. However, in the past we
```
wolfv, line 160: This is then wrong?

wolfv (approved):
LGTM! Should we call it `serde::TrackFeatures` instead of `Features`?

crates/rattler_conda_types/src/utils/serde.rs
baszalmstra, line 160:
```suggestion
/// features are expected to be a space seperated list. However, in the past we
```

baszalmstra: Yes thats before I learned that it was crazier then I thought.. I changed it.

baszalmstra:
> LGTM! Should we call it `serde::TrackFeatures` instead of `Features`?

Its actually rhe same behavior for `features`. We dont parse that as a list yet because it would break the api but I thought Id already name it properly.

wolfv:
OK! Features are really deprecated and broken so I don't think we'll ever need them but good to know :)

PR #1047, opened 2025-02-05
fix: make linter happy and add license-files

wolfv:
Not sure if this helps the version sdist situation.


baszalmstra:
I dont like this behavior at all, apparently if there is a LICENSE file the sdist will not include it but will reference it ..

Although I think this method of yours works, it makes the licensing of the project slightly more complicated because now someone has to parse the file instead of being able to refering to an SPDX. Let me try something today.

baszalmstra:
If I read [PEP 639](https://peps.python.org/pep-0639/) correctly we should just be able to use just

```toml
license = "BSD-3-Clause"
```

But [this part of the maturin codebase](https://github.com/PyO3/maturin/blob/e84790857c1ca902576533e832b264059b8a3978/src/metadata.rs#L217-L239) ALWAYS adds the LICENSE.

I think the easiest step for now, is just to remove the `LICENSE` file on build to not trigger this behavior??

wolfv:
I copied the code from Ruff. But feel free to change things around :)

baszalmstra:
So, my solution is to simply remove the LICENSE file. It serves no purpose since there is one at the repository's root.

wolfv:
Should we try?

Issue #1044, opened 2025-02-05
Add a `rattler_pypi_interop` crate


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

We would like to add better support to the rattler library for working with packages from PyPI. Much of this functionality exists the [rip project](https://github.com/prefix-dev/rip), but because this is no longer being actively developed, we would like to bring it over to the rattler project.

### Why is this needed?

Another project that could provide this functionality in Rust is the [uv project](https://github.com/astral-sh/uv), but unfortunately the crates that provide it are not available on crates.io and we're not sure if they ever will be made available there.

To continue publishing the rattler crates on crates.io, we need to ensure that all of its dependencies are available there.

We also need this crate to fill in the gaps for pypi interoperability. For example, the `rattler_lock` crate is currently unable to create lock files with pypi packages using components exclusively available from rattler. This is a bad developer experience, and we would like to fix it.

### What should happen?

We should first focus on bringing components from the [rattler_installs_packages](https://github.com/prefix-dev/rip/tree/main/crates/rattler_installs_packages) crate to a new create in rattler called `rattler_pypi_interop`.

Initially, it probably makes sense to just focus on the following folders in this crate:

- [ ] [types](https://github.com/prefix-dev/rip/tree/main/crates/rattler_installs_packages/src/types)
- [ ] [artifacts](https://github.com/prefix-dev/rip/tree/main/crates/rattler_installs_packages/src/artifacts)
- [ ] [index](https://github.com/prefix-dev/rip/tree/main/crates/rattler_installs_packages/src/index)

### Additional Context

See original discussion surrounding this issue here:

- https://github.com/conda/rattler/issues/833

jacopoabramo:
@iamthebot 

There is a [blog post](https://prefix.dev/blog/uv_in_pixi) about this decision.

travishathaway:
Would be great to help the `uv` folks get some of their crates published. Not sure how much work is involved their though.

In the meantime, I think it makes sense to keep pursing this crate and just keep it as light-weight as possible. When I wrote this issue, I was envisioning a library with "read-only" capabilities for pypi packages. This would primarily be used to fetch metadata about packages installed via `pip` or `uv`.

As I've said elsewhere, my immediate use case for it is being able to generate `rattle_lock` lock files with out the need of `uv` as a dependency.

[...]

PR #1054, opened 2025-02-06
Add a `rattler_pypi_interop` crate

travishathaway:
### Description

Fixes: #1044 

This is still a draft and so far, I have just copied over the [rattler_installs_package](https://github.com/prefix-dev/rip/tree/main/crates/rattler_installs_packages) crate from rip.  I'm initially planning to prune this crate and remove some unnecessary pieces. So far, this consists of the pieces of the code which perform solving of pypi dependencies.


baszalmstra:
I really hope you can trim it down to the bare essentials. This is a lot of code to maintain.

tdejager:
Hey @travishathaway, excellent initial work on the first steps so far. I still feel as @iamthebot said, that it might be better to split up the parts into different `pypi-*` crates that we could review better and integrate better. Browsing through the PR now, its quite hard to have a high quality review, and I even helped write some of that code.

I think a structure could potentially be:
* `pypi-types`: containing artifact types and such.
* `pypi-gateway`: for working with the PyPI indexes.
* `python-env` (not sure about this name) for inspecting the python environment.

We might need to also write down the goals for the crates. AFAIK from your current work you need to mostly translate an existing environment into a locked down version, in which case you would I think only need to hit the PyPI index for a simple request or not at all. That said having functionality to hit the index, cache the results, and build wheels might be interesting for other use cases. So I'd guess its just a matter of what we want in `rattler` not really sure if thats been written down yet? Like what actual functionality do we actually require in rattler.

We might also want a test that verifies if the metadata can be correctly retrieved for the 500 most popular python projects ore something.

A last note, not sure if I'm reading the code correctly, but it seems that with the current code we won't be able to retrieve metadata for sdists that do not have these statically available right?

Feel free to chime in @baszalmstra @wolfv

travishathaway:
Hi @tdejager,

To answer you first question, the code that retrieved `sdist` metadata from the index relied on building to the `sdist` in order to get it. I think this is because it needed to do that in order to transform the `sdist` into a `wheel`.  I could go back and look at this and see how I could re-write the code in order to make that not necessary.

I do like your idea about coming up with a design document for this as something that would give us a high level overview of these crates and what they are meant to be used for. I would be happy to write something like this. Is there a particular format you prefer for this? I usually use hackmd.io

Regarding the adding additional crates, I'm not opposed to it, but I'm also not sure if it's necessary. I could create a series of separate pull requests though. I'll just have to do it in the right order because the modules within the crates depend on each other.

PR #1057, opened 2025-02-10
chore: Add presigned URL to debug logs

pavelzw:
### Description

In some cases you might want to debug why your pixi call to S3 didn't work.
For this, I added a debug log with the presigned URL.
Since the URL expires in a fixed amount of time (iirc 5min by default) anyway, I wouldn't consider this leaking secrets, wdyt?


wolfv:
Hmm not the biggest fan. Could we just make sure that we print meaningful error when the download does not succeed?

pavelzw:
i for now don't have the need for this anymore. eventually, it might be nice to also print this stuff but i don't plan on working on this in the near future

PR #1060, opened 2025-02-11
feat: add archive cache

nichmor:
### Description
This PR aims to add `ArchiveCache` entity. It will be responsible for storing `unarchived` conda packages on disk.

This, later, will be used by `rattler-build` to extract only `run_exports.json` from archive instead of the entire pkg, which fixes some situation when unpacking pkg on windows.

Where it will be used: https://github.com/prefix-dev/rattler-build/blob/main/src/run_exports.rs#L99

`ArchiveCache` exposes only getting package from URL with retrying policy ( like `PackageCache` does ) but without exposing getting conda package from path ( unlike `PackageCache` ).

Filesystem syncrhonisation with locks is absent on purpose because `NamedTempFile::persist` will atomically replace it.

### DIFF
--- a/crates/rattler_cache/Cargo.toml
+++ b/crates/rattler_cache/Cargo.toml
@@ -23,7 +23,9 @@ rattler_digest = { version = "1.0.5", path = "../rattler_digest", default-featur
 rattler_networking = { version = "0.22.1", path = "../rattler_networking", default-features = false }
 rattler_package_streaming = { version = "0.22.25", path = "../rattler_package_streaming", default-features = false, features = ["reqwest"] }
 reqwest.workspace = true
+tempfile.workspace = true
 tokio = { workspace = true, features = ["macros"] }
+tokio-util = { workspace = true, features = ["io-util"] }
 tracing.workspace = true
 url.workspace = true
 thiserror.workspace = true
@@ -39,7 +41,7 @@ axum.workspace = true
 bytes.workspace = true
 futures.workspace = true
 rstest.workspace = true
-tempfile.workspace = true
+# tempfile.workspace = true
 tokio-stream.workspace = true
 tower-http = { workspace = true, features = ["fs"] }
 tools = { path = "../tools" }
--- a/crates/rattler_cache/src/archive_cache/cache_key.rs
+++ b/crates/rattler_cache/src/archive_cache/cache_key.rs
@@ -1,0 +1,64 @@
+use rattler_conda_types::PackageRecord;
+use rattler_digest::Sha256Hash;
+use std::fmt::{Display, Formatter};
+
+/// Provides a unique identifier for packages in the cache.
+/// TODO: This could not be unique over multiple subdir. How to handle?
+#[derive(Debug, Hash, Clone, Eq, PartialEq)]
+pub struct CacheKey {
+    pub(crate) name: String,
+    pub(crate) version: String,
+    pub(crate) build_string: String,
+    pub(crate) sha256: Option<Sha256Hash>,
+}
+
+impl CacheKey {
+    /// Adds a sha256 hash of the archive.
+    pub fn with_sha256(mut self, sha256: Sha256Hash) -> Self {
+        self.sha256 = Some(sha256);
+        self
+    }
+
+    /// Potentially adds a sha256 hash of the archive.
+    pub fn with_opt_sha256(mut self, sha256: Option<Sha256Hash>) -> Self {
+        self.sha256 = sha256;
+        self
+    }
+}
+
+impl CacheKey {
+    /// Return the sha256 hash of the package if it is known.
+    pub fn sha256(&self) -> Option<Sha256Hash> {
+        self.sha256
+    }
+
+    pub fn sha256_str(&self) -> String {
+        self.sha256()
+            .map(|hash| format!("{hash:x}"))
+            .unwrap_or_default()
+    }
+}
+
+impl From<&PackageRecord> for CacheKey {
+    fn from(record: &PackageRecord) -> Self {
+        Self {
+            name: record.name.as_normalized().to_string(),
+            version: record.version.to_string(),
+            build_string: record.build.clone(),
+            sha256: record.sha256,
+        }
+    }
+}
+
+impl Display for CacheKey {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}-{}-{}-{}",
+            &self.name,
+            &self.version,
+            &self.build_string,
+            self.sha256_str()
+        )
+    }
+}
--- a/crates/rattler_cache/src/archive_cache/download.rs
+++ b/crates/rattler_cache/src/archive_cache/download.rs
@@ -1,0 +1,64 @@
+use std::sync::Arc;
+
+use ::tokio::io::AsyncWriteExt;
+use fs_err::tokio;
+use futures::StreamExt;
+use rattler_package_streaming::DownloadReporter;
+use tempfile::NamedTempFile;
+// use tokio_stream::StreamExt;
+use url::Url;
+
+/// Download the contents of the archive from the specified remote location
+/// and store it in a temporary file.
+pub(crate) async fn download(
+    client: reqwest_middleware::ClientWithMiddleware,
+    url: Url,
+    reporter: Option<Arc<dyn DownloadReporter>>,
+) -> Result<NamedTempFile, DownloadError> {
+    let temp_file = NamedTempFile::new()?;
+
+    // Send the request for the file
+    let response = client.get(url.clone()).send().await?.error_for_status()?;
+
+    if let Some(reporter) = &reporter {
+        reporter.on_download_start();
+    }
+
+    let total_bytes = response.content_length();
+    // Convert the named temp file into a tokio file
+    let mut file = tokio::File::from_std(fs_err::File::from_parts(
+        temp_file.as_file().try_clone().unwrap(),
+        temp_file.path(),
+    ));
+
+    let mut stream = response.bytes_stream();
+
+    let mut bytes_received = 0;
+    while let Some(chunk_result) = stream.next().await {
+        let chunk = chunk_result?;
+
+        if let Some(reporter) = &reporter {
+            bytes_received += chunk.len() as u64;
+            reporter.on_download_progress(bytes_received, total_bytes);
+        }
+        file.write_all(&chunk).await?;
+    }
+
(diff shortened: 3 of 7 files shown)


wolfv:
One quick remark: I think we shoudl fall back to the `md5` hash if `sha256` is not available.

crates/rattler_cache/src/run_exports_cache/cache_key.rs
```diff
@@ -0,0 +1,76 @@
...
+pub enum CacheKeyError {
+    #[error("Could not identify the archive type from the name: {0}")]
+    InvalidArchiveIdentifier(String),
+}
+
+impl Display for CacheKey {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "{}-{}-{}-{}{}",
```
wolfv, line 68: I would prefer to do this in a way that it looks really nice. Now, if there is no sha256 hash, it will end with a `-`.

We can structure the code a little different and remove the `sha256_str` function. Just do all that in this function here and have perfect formatting.

crates/rattler_cache/src/run_exports_cache/download.rs
```diff
@@ -0,0 +1,67 @@
+use std::sync::Arc;
+
+use ::tokio::io::{AsyncSeekExt, AsyncWriteExt};
+use fs_err::tokio;
+use futures::StreamExt;
+use rattler_package_streaming::DownloadReporter;
+use tempfile::NamedTempFile;
+// use tokio_stream::StreamExt;
```
wolfv, line 8: we can remove this?

crates/rattler_package_streaming/src/lib.rs
```diff
@@ -62,6 +63,12 @@ impl From<ZipError> for ExtractError {
     }
 }
 
+impl From<Cancelled> for ExtractError {
```
wolfv, line 66: Is this related?

crates/rattler_package_streaming/src/seek.rs
```diff
@@ -100,19 +121,26 @@ fn get_file_from_archive(
...
-                .map_err(|e| ExtractError::ArchiveMemberParseError(P::package_path().to_owned(), e))
-        }
-    }
+    P::from_str(&String::from_utf8_lossy(&content))
+        .map_err(|e| ExtractError::ArchiveMemberParseError(P::package_path().to_owned(), e))
+}
+
+/// Get a [`PackageFile`] from temporary archive and return it as a temporary file.
+pub fn get_package_file<P: PackageFile>(
+    archive_temp_file: NamedTempFile,
```
wolfv, line 132: You can just take the `path` for the NamedTempFile? I am not sure why we need this in the signature?

crates/rattler_repodata_gateway/src/fetch/mod.rs
```diff
@@ -725,7 +725,10 @@ async fn stream_and_decode_to_file(
     // Clone the file handle and create a hashing writer so we can compute a hash
     // while the content is being written to disk.
     let file = tokio_fs::File::from_std(fs_err::File::from_parts(
-        temp_file.as_file().try_clone().unwrap(),
+        temp_file
+            .as_file()
+            .try_clone()
+            .map_err(FetchRepoDataError::IoError)?,
```
wolfv, line 731: Can this happen? Just wondering :)

crates/rattler_package_streaming/src/seek.rs
wolfv: Or let me rephrase this. I think the interface should be slightly more generic. It should be something like:

```
extract_package_file<P: PackageFile, W: Write>(path: &Path, writer: &mut W) -> io::Result<...> {
   ... write the extracted file to the writer ...
}
```

crates/rattler_package_streaming/src/lib.rs
nichmor:
```
rattler_package_streaming::seek::get_package_file
``` is a blocking operation ( function ) 

so I'm running it like this:
```
simple_spawn_blocking::tokio::run_blocking_task(move || {
                                rattler_package_streaming::seek::get_package_file::<RunExportsJson>(result)
                            }).await;
                            ```
and the error should be created from `Cancelled`, so I need to implement this conversion

crates/rattler_repodata_gateway/src/fetch/mod.rs
nichmor: I guess the clone operation can fail if https://man7.org/linux/man-pages/man2/fcntl.2.html will return something unexpected which will be converted by https://docs.rs/cvt/latest/cvt/. 
don't know what are the `triggers` for this

crates/rattler_cache/src/run_exports_cache/cache_key.rs
```diff
@@ -0,0 +1,87 @@
+use rattler_conda_types::{package::ArchiveIdentifier, PackageRecord};
+use rattler_digest::{Md5Hash, Sha256Hash};
+use std::fmt::{Display, Formatter};
+
+/// Provides a unique identifier for packages in the cache.
+/// TODO: This could not be unique over multiple subdir. How to handle?
```
tdejager, line 6: Is this smething that should still be handled?

crates/rattler_cache/src/run_exports_cache/cache_key.rs
```diff
@@ -0,0 +1,87 @@
...
+            sha256: record.sha256,
+            md5: record.md5,
+            extension: archive_identifier.archive_type.extension().to_string(),
+        })
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum CacheKeyError {
+    #[error("Could not identify the archive type from the name: {0}")]
```
tdejager, line 67:
```suggestion
    #[error("could not identify the archive type from the name: {0}")]
```

crates/rattler_cache/src/run_exports_cache/cache_key.rs
```diff
@@ -0,0 +1,87 @@
...
+
+    /// Return the sha256 hash string of the package if it is known.
+    pub fn sha256_str(&self) -> String {
+        self.sha256()
+            .map(|hash| format!("{hash:x}"))
+            .unwrap_or_default()
+    }
+
+    /// Try to create a new cache key from a package record and a filename.
+    pub fn new(record: &PackageRecord, filename: &str) -> Result<Self, CacheKeyError> {
```
tdejager, line 50: Maybe `try_new` would be better as it returns a result? or `create`?

crates/rattler_cache/src/run_exports_cache/download.rs
```diff
@@ -0,0 +1,66 @@
...
+        if let Some(reporter) = &reporter {
+            bytes_received += chunk.len() as u64;
+            reporter.on_download_progress(bytes_received, total_bytes);
+        }
+        file.write_all(&chunk).await?;
+    }
+
+    file.flush().await?;
+
+    file.rewind().await?;
```
tdejager, line 49: Do we need the `rewind` as we are not returning the `file` and do no operations on it afterwards?

crates/rattler_package_streaming/src/seek.rs
```diff
@@ -84,6 +85,25 @@ fn get_file_from_archive(
     Err(ExtractError::MissingComponent)
 }
 
+/// Read a package file content from archive
+fn read_package_file_content<'a, P: PackageFile>(
```
tdejager, line 89: Maybe we can just pass the `package_path` in, as afaics it does not use the generic for anything else?

tdejager (changes requested):
Looking, good just a few minor nits.

crates/rattler_cache/src/run_exports_cache/cache_key.rs
nichmor: good catch! I think it has already been handled because we are using sha256/md5 to identify them uniquely.

crates/rattler_cache/src/run_exports_cache/download.rs
nichmor: We read later from this file when unarchiving

crates/rattler_package_streaming/src/seek.rs
nichmor: `file` is used by `stream_tar_bz2` to read form file, and `path` is used to determine `ArchiveType` so we need both

crates/rattler_cache/src/run_exports_cache/download.rs
tdejager: I think we dont need the async file, as we could run a blocking path for it I suppose, most of these async operations are not faster.

tdejager: And If you do want to keep the async API, I'm not sure if `temp_file.as_file().try_clone()?` returns a new file handle or the same one? Otherwise the rewind would still not be needed.

crates/rattler_package_streaming/src/seek.rs
tdejager: Sorry I dont get what you mean? AFAICS the only place that generic is sued in this function is for the `P::package_path`? And if you just pass that in as a str it makes the method more generic by default.

crates/rattler_cache/src/run_exports_cache/download.rs
```diff
@@ -0,0 +1,66 @@
...
+    let response = client.get(url.clone()).send().await?.error_for_status()?;
+
+    if let Some(reporter) = &reporter {
+        reporter.on_download_start();
+    }
+
+    let total_bytes = response.content_length();
+    // Convert the named temp file into a tokio file
+    let mut file = tokio::File::from_std(fs_err::File::from_parts(
+        temp_file.as_file().try_clone()?,
```
baszalmstra, line 30: You can use NamedTempFile into_parts to get the file handle without cloning. Cloning may fail.

At the end of the function you can call from_parts to reconstruct the namedtempfile

crates/rattler_repodata_gateway/src/fetch/mod.rs
baszalmstra: Same here! Use into_parts to forgo to the try_clone

nichmor: I think I cannot do it here.

```HashingWriter::<_, Blake2b256>::new(file);``` take ownership of the file. 
If I will do `temp_file.into_parts()` I will lose this file handle when creating it `from_parts` so later, I will not be able to create NamedTempFile again.

PR #1063, opened 2025-02-13
feat: write prefix record atomically

wolfv:
This should help when we hit issues with empty files.

### DIFF
--- a/crates/rattler_conda_types/src/prefix_record.rs
+++ b/crates/rattler_conda_types/src/prefix_record.rs
@@ -234,8 +234,6 @@ impl PrefixRecord {
         path: impl AsRef<Path>,
         pretty: bool,
     ) -> Result<(), std::io::Error> {
-        const BUFFER_SIZE: usize = 64 * 1024;
-
         let path = path.as_ref();
         let parent = path.parent().ok_or_else(|| {
             std::io::Error::new(
@@ -255,7 +253,7 @@ impl PrefixRecord {
         })?;
 
         // Write to temp file with buffered writer
-        let writer = BufWriter::with_capacity(BUFFER_SIZE, &temp_file);
+        let writer = BufWriter::with_capacity(64 * 1024, &temp_file);
         self.write_to(writer, pretty)?;
 
         // Ensure all data is written before persisting
--- a/crates/rattler_conda_types/src/prefix_record.rs
+++ b/crates/rattler_conda_types/src/prefix_record.rs
@@ -238,26 +238,28 @@ impl PrefixRecord {
         let parent = path.parent().ok_or_else(|| {
             std::io::Error::new(
                 std::io::ErrorKind::Other,
-                "Failed to get parent directory of path",
+                format!(
+                    "Failed to get parent directory of path '{}'",
+                    path.display()
+                ),
             )
         })?;
-
-        fs_err::create_dir_all(parent)?;
 
         // Use a temporary file in the same directory for atomic writes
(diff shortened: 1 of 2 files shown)


crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -234,8 +235,41 @@ impl PrefixRecord {
         path: impl AsRef<Path>,
         pretty: bool,
     ) -> Result<(), std::io::Error> {
-        let file = File::create(path.as_ref())?;
-        self.write_to(BufWriter::with_capacity(50 * 1024, file), pretty)
+        const BUFFER_SIZE: usize = 64 * 1024;
```
ruben-arts, line 238: What is this based on?

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -234,8 +235,41 @@ impl PrefixRecord {
...
+
+        let path = path.as_ref();
+        if let Some(parent) = path.parent() {
+            std::fs::create_dir_all(parent)?;
```
ruben-arts, line 242: I believe we try to use `fs_err` to get better errors.

crates/rattler_conda_types/src/prefix_record.rs
wolfv: Used to be 50 * 1024, now it's 64 * 1024 - doesn't really matter that much but it saves us some calls to `write` by having a slightly larger buffer. This buffer size is 64Kb.

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -234,8 +235,41 @@ impl PrefixRecord {
...
+            temp_file.as_file().sync_all()?;
+
+            // Atomically rename the temp file to the target path
+            temp_file.persist(path).map_err(|e| {
+                std::io::Error::new(
+                    std::io::ErrorKind::Other,
+                    format!("Failed to persist file: {}", e),
+                )
+            })?;
+        } else {
```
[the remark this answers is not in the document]
wolfv: Yeah we could also error if there is no parent, which indeed should never happen.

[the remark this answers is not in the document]
nichmor: I think it's ok to not have parent, we could just directly write the file ( because we are in the root and it exists ) and use prefix_record_root as prefix for namedtempfile

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -234,8 +234,42 @@ impl PrefixRecord {
         path: impl AsRef<Path>,
         pretty: bool,
     ) -> Result<(), std::io::Error> {
-        let file = File::create(path.as_ref())?;
-        self.write_to(BufWriter::with_capacity(50 * 1024, file), pretty)
+        const BUFFER_SIZE: usize = 64 * 1024;
+
+        let path = path.as_ref();
+        let parent = path.parent().ok_or_else(|| {
```
ruben-arts, line 240: `parent()` will return `Some("")` if the file is not a absolute path. Is that okay?

ruben-arts (commented):
I'm not sure how I can test if this solves the issue, but it the happy flow keeps working.

crates/rattler_conda_types/src/prefix_record.rs
wolfv: Yeah that should be OK since then it will jsut be a relative path (relative to CWD) and the rest of the types should also work with relative paths.

PR #1066, opened 2025-02-13
fix: another clobbering issue

### LINKED ISSUE #1061: Had a clobber
Once more I hit a clobber ... 

I need to take another look (this came when setting up the pixi environment):

```
Error: 
  × failed to unclobber clobbered files
  ├─▶ failed to rename lib/gcc/arm64-apple-darwin20.0.0/13.2.0/include/ISO_Fortran_binding.h__clobber-from-gfortran_impl_osx-arm64 to lib/gcc/arm64-apple-
  │   darwin20.0.0/13.2.0/include/ISO_Fortran_binding.h
  ╰─▶ failed to rename file from /Users/wolfv/Programs/pixi/.pixi/envs/default/lib/gcc/arm64-apple-darwin20.0.0/13.2.0/include/ISO_Fortran_binding.h__clobber-
      from-gfortran_impl_osx-arm64 to /Users/wolfv/Programs/pixi/.pixi/envs/default/lib/gcc/arm64-apple-darwin20.0.0/13.2.0/include/ISO_Fortran_binding.h: No
      such file or directory (os error 2)
```

wolfv:
This should fix #1061

### DIFF
--- a/crates/rattler-bin/src/commands/create.rs
+++ b/crates/rattler-bin/src/commands/create.rs
@@ -19,8 +19,8 @@ use rattler::{
     package_cache::PackageCache,
 };
 use rattler_conda_types::{
-    Channel, ChannelConfig, GenericVirtualPackage, MatchSpec, Matches, PackageName,
-    ParseStrictness, Platform, PrefixRecord, RepoDataRecord, Version,
+    Channel, ChannelConfig, GenericVirtualPackage, MatchSpec, PackageName, ParseStrictness,
+    Platform, PrefixRecord, RepoDataRecord, Version,
 };
 use rattler_networking::{AuthenticationMiddleware, AuthenticationStorage};
 use rattler_repodata_gateway::{Gateway, RepoData, SourceConfig};
@@ -254,7 +254,7 @@ pub async fn create(opt: Opt) -> anyhow::Result<()> {
     let solver_task = SolverTask {
         locked_packages,
         virtual_packages,
-        specs: specs.clone(),
+        specs,
         timeout: opt.timeout.map(Duration::from_millis),
         strategy: opt.strategy.map_or_else(Default::default, Into::into),
         ..SolverTask::from_iter(&repo_data)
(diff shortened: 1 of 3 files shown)


crates/rattler-bin/src/commands/create.rs
```diff
@@ -271,6 +271,16 @@ pub async fn create(opt: Opt) -> anyhow::Result<()> {
 
     let required_packages: Vec<RepoDataRecord> = solver_result.records;
 
+    let no_deps = false;
```
baszalmstra, line 274: Should this be removed?

crates/rattler/src/install/mod.rs
```diff
@@ -328,16 +333,25 @@ pub async fn link_package(
...
+            let clobber_path = clobber_paths.get(&entry.relative_path).cloned();
+            LinkPath {
+                entry,
+                computed_path,
+                clobber_path,
+            }
+        })
+        .collect();
+
+    // drop(clobber_paths);
```
baszalmstra, line 349: Remove?

crates/rattler/src/install/mod.rs
```diff
@@ -639,25 +655,32 @@ pub fn link_package_sync(
...
+        let clobber_path = clobber_paths.get(&entry.relative_path).cloned();
+        LinkPath {
+            entry,
+            computed_path,
+            clobber_path,
         }
-    }
+    });
+
+    // We can already drop this
```
baszalmstra, line 676: Remove?

crates/rattler/src/install/mod.rs
```diff
@@ -737,19 +760,20 @@ pub fn link_package_sync(
         // files that are either in the clobber map or contain a placeholder,
         // we defer to the regular linking that comes after this block
         // and re-add them to the paths_by_directory map
-        for file in files {
-            if clobber_paths.contains_key(&file.1) || file.0.prefix_placeholder.is_some() {
```
wolfv, line 741: The problem was that we were using the computed path here, which already contained the "clobber"-modifications. And thus, we never re-added the path here...

crates/rattler-bin/src/commands/create.rs
wolfv: Removed all changes to this file.

PR #1068, opened 2025-02-14
feat: implement --no-deps and --only-deps

wolfv:
Somewhat useful CLI flags, potentially for testing etc.


crates/rattler-bin/src/commands/create.rs
```diff
@@ -60,6 +60,12 @@ pub struct Opt {
 
     #[clap(long)]
     strategy: Option<SolveStrategy>,
+
+    #[clap(long)]
```
baszalmstra, line 64: can you add a conflicts_with ?

wolfv: done by using a group

PR #1075, opened 2025-02-19
feat: implement shell completion helpers

wolfv:
This is a bit of a starting point to add support for sourcing shell completions in rattler shell.


crates/rattler_shell/src/shell/mod.rs
```diff
@@ -527,6 +546,10 @@ impl Shell for Fish {
         cmd.arg(path);
         cmd
     }
+
+    fn completion_script_location(&self) -> Option<PathBuf> {
+        Some(PathBuf::from("share/fish/vendor_completions"))
```
wolfv, line 551: this should be `vendor_completions.d`

wolfv, line 551:
```suggestion
        Some(PathBuf::from("share/fish/vendor_completions.d"))
```

ruben-arts:
Let's merge this while we're working on the `pixi` implementation, this is untested code that can easily be extended with more features. So let's do it when there is something/one that uses it.

crates/rattler_shell/src/completion.rs
```diff
@@ -0,0 +1,70 @@
+//! Module for handling shell completion scripts.
```
wolfv, line 1: I think we can remove this.

wolfv:
We need to decide on the API here. We now have a high-level command that is able to source completion files from a given directory. 

I think the question is whether the input to the API should be a the final directory (ie. `prefix.join(completion_dir)` or whether it should be just the prefix, and `rattler` will stick things together. 

The benefit of giving the _final_ dir directly would be that for `pixi global` we could arrange folders a bit different (e.g. `.pixi/completions/bash`, `.pixi/completions/fish`, ...).

Otherwise we would be forced to do: `.pixi/completions/share/bash/bash-completions/...`. If we give the directory directly, the calling code for `pixi shell` should look something like:

```
if let Some(completions_dir) = shell.completions_dir() {
    shell_script.source_completions(prefix.join(completions_dir));
}
```
WDYT @ruben-arts / @Hofer-Julian ?

wolfv:
I decided to do the following API:

shell has an `completions_dir() -> Option<PathBuf>` that returns the default completion folder location.

The sourcing of completions requires to give the full folder to the folder with completions then though. So we can put them into a different location for `pixi global`.

baszalmstra (approved):

PR #1076, opened 2025-02-19
feat: Use `opendal` in `rattler-index` and add executable

delsner:
### Description

Final piece of work for implementing https://github.com/conda/rattler/issues/960.


baszalmstra:
Exciting!!

pavelzw:
showcase:


https://github.com/user-attachments/assets/a1a9d8e5-fb5f-444a-acb7-e2f0e1c9acb9

Cargo.toml
```diff
@@ -67,6 +68,7 @@ futures = "0.3.31"
 futures-util = "0.3.31"
 fxhash = "0.2.1"
 # lots of other crates are still stuck on older version which breaks `deserialize`
+
```
pavelzw, line 71:
```suggestion
```

crates/rattler_index/src/lib.rs
```diff
@@ -82,13 +103,22 @@ pub fn package_record_from_tar_bz2(file: &Path) -> std::io::Result<PackageRecord
 /// and extract the package record from it.
 pub fn package_record_from_conda(file: &Path) -> std::io::Result<PackageRecord> {
     let reader = fs::File::open(file)?;
+    package_record_from_conda_reader(reader)
+}
+
+/// Extract the package record from a `.conda` package file content.
+/// /// This function will look for the `info/index.json` file in the conda package
```
pavelzw, line 110:
```suggestion
/// This function will look for the `info/index.json` file in the conda package
```

crates/rattler_index/src/lib.rs
```diff
@@ -97,117 +127,317 @@ pub fn package_record_from_conda(file: &Path) -> std::io::Result<PackageRecord>
...
+            tokio::spawn(index_subdir(
+                *subdir,
+                op.clone(),
+                force,
+                multi_progress.clone(),
+                semaphore.clone(),
+            ))
+        })
+        .collect::<Vec<_>>();
+    try_join_all(tasks).await?;
```
pavelzw, line 440: we should add a `println!("Finished indexing ...")` here as otherwise we won't have any output in non-tty environments

py-rattler/src/index.rs
```diff
@@ -1,21 +1,65 @@
...
+            session_token,
+            target_platform,
+            force,
+            max_parallel,
+        )
+        .await
+        {
             Ok(_v) => Ok(()),
             Err(e) => Err(PyRattlerError::from(e).into()),
         }
```
pavelzw, line 65: i think we can simplify this using `map_err(PyRattlerError::from)?`

py-rattler/tests/unit/test_index.py
```diff
@@ -30,23 +39,109 @@ def test_index(package_directory):
...
+
+
+@pytest.fixture()
+def s3_config() -> S3Config:
+    access_key_id = os.environ.get("RATTLER_TEST_R2_ACCESS_KEY_ID")
+    if not access_key_id:
+        pytest.skip("RATTLER_TEST_R2_ACCESS_KEY_ID environment variable is not set")
+    secret_access_key = os.environ.get("RATTLER_TEST_R2_SECRET_ACCESS_KEY")
+    if not secret_access_key:
+        pytest.skip("RATTLER_TEST_R2_SECRET_ACCESS_KEY environment variable is not set")
```
pavelzw, line 83: @delsner these secrets need to be present in the github workflow as env vars, @wolfv could you add them as repo secrets?

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+                force_path_style,
+                access_key_id,
+                secret_access_key,
+                session_token,
+                cli.target_platform,
+                cli.force,
+                cli.max_parallel,
+            )
+            .await?;
+            Ok(())
```
pavelzw, line 130:
```suggestion
            .await
```

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+    let cli = Cli::parse();
+
+    tracing_subscriber::FmtSubscriber::builder()
+        .with_max_level(cli.verbose.log_level_filter().as_trace())
+        .init();
+
+    let result = match cli.command {
+        Commands::FileSystem { channel } => {
+            index_fs(channel, cli.target_platform, cli.force, cli.max_parallel).await?;
+            Ok(())
```
pavelzw, line 106:
```suggestion
            index_fs(channel, cli.target_platform, cli.force, cli.max_parallel).await
```

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+                secret_access_key,
+                session_token,
+                cli.target_platform,
+                cli.force,
+                cli.max_parallel,
+            )
+            .await?;
+            Ok(())
+        }
+    };
```
pavelzw, line 132:
```suggestion
    }?;
```

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+    println!("Finished indexing channel...");
```
pavelzw, line 133:
```suggestion
    println!("Finished indexing channel.");
```
`...` usually indicates that something is still coming 😄

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+    result
```
pavelzw, line 134: I would use `?` and return `Ok(())` here

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,135 @@
...
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // Parse the command line arguments
+    let cli = Cli::parse();
+
+    tracing_subscriber::FmtSubscriber::builder()
+        .with_max_level(cli.verbose.log_level_filter().as_trace())
+        .init();
+
+    let result = match cli.command {
```
delsner, line 103:
```suggestion
    match cli.command {
```

crates/rattler_index/src/main.rs
delsner, line 134:
```suggestion
    Ok(())
```

py-rattler/tests/unit/test_index.py
```diff
@@ -30,23 +39,109 @@ def test_index(package_directory):
...
+# ---------------------------------------- S3 ---------------------------------------- #
+
+
+@dataclass
+class S3Config:
+    access_key_id: str
+    secret_access_key: str
+    region: str = "auto"
+    endpoint_url: str = "https://e1a7cde76f1780ec06bac859036dbaf7.r2.cloudflarestorage.com"
+    bucket_name: str = "rattler-build-upload-test"
```
delsner, line 72: @wolfv let me know if you prefer to create a separate bucket for this project and I will adjust the code.

crates/rattler_index/src/lib.rs
```diff
@@ -63,12 +75,21 @@ pub fn package_record_from_index_json<T: Read>(
...
+    package_record_from_tar_bz2_reader(reader)
+}
+
+/// Extract the package record from a `.tar.bz2` package file.
+/// This function will look for the `info/index.json` file in the conda package
+/// and extract the package record from it.
+pub fn package_record_from_tar_bz2_reader(reader: impl Read) -> std::io::Result<PackageRecord> {
+    let bytes = reader.bytes().collect::<Result<Vec<u8>, _>>()?;
+    let reader = Cursor::new(bytes.clone());
     let mut archive = read::stream_tar_bz2(reader);
```
delsner, line 87: note that this duplicates the file content in-memory. We didn't really find an elegant solution since `stream_tar_bz2` consumes the reader. But we're also still on amateur rust level, so happy to implement any suggested improvement 😅

crates/rattler_index/src/lib.rs
```diff
@@ -97,117 +127,317 @@ pub fn package_record_from_conda(file: &Path) -> std::io::Result<PackageRecord>
...
+            Ok(bytes) => serde_json::from_slice(&bytes.to_vec())?,
+            Err(e) => {
+                if e.kind() != opendal::ErrorKind::NotFound {
+                    return Err(e.into());
+                }
+                tracing::info!("Could not find repodata.json. Creating new one.");
+                RepoData {
+                    info: Some(ChannelInfo {
+                        subdir: subdir.to_string(),
+                        base_url: None,
```
pavelzw, line 151: we should be able to infer the base url from the cli args, right?

crates/rattler_index/src/lib.rs
```diff
@@ -97,117 +127,317 @@ pub fn package_record_from_conda(file: &Path) -> std::io::Result<PackageRecord>
...
-            }),
-            packages: HashMap::default(),
-            conda_packages: HashMap::default(),
-            removed: HashSet::default(),
-            version: Some(2),
-        };
+    let repodata = RepoData {
+        info: Some(ChannelInfo {
+            subdir: subdir.to_string(),
+            base_url: None,
```
pavelzw, line 294: we should use the old repodata's base_url here

crates/rattler_index/src/lib.rs
```diff
@@ -97,117 +127,317 @@ pub fn package_record_from_conda(file: &Path) -> std::io::Result<PackageRecord>
...
+                access_key_id,
+                secret_access_key,
+                session_token,
+            }),
+        ) = auth
+        {
+            s3_config.access_key_id = Some(access_key_id);
+            s3_config.secret_access_key = Some(secret_access_key);
+            s3_config.session_token = session_token;
         }
```
pavelzw, line 365: is it intended that we don't fail when we don't find any credentials? to me, a public s3 bucket doesn't make much sense as then it's just an http server

delsner: We don't fail because there might still be the option of passing credentials via environment variables (e.g., `AWS_*`). Opendal claims it handles these as well.

crates/rattler_index/src/lib.rs
```diff
@@ -97,117 +127,317 @@ pub fn package_record_from_conda(file: &Path) -> std::io::Result<PackageRecord>
...
-                ArchiveType::TarBz2 => repodata
-                    .packages
-                    .insert(file_name.to_string_lossy().to_string(), record),
-                ArchiveType::Conda => repodata
-                    .conda_packages
-                    .insert(file_name.to_string_lossy().to_string(), record),
-            };
+    let repodata_path = format!("{subdir}/repodata.json");
+    let repodata_bytes = serde_json::to_vec(&repodata)?;
+    op.write(&repodata_path, repodata_bytes).await?;
```
pavelzw, line 304:
```suggestion
    let repodata_path = format!("{subdir}/repodata.json");
    let repodata_bytes = serde_json::to_vec(&repodata)?;
    op.write(&repodata_path, repodata_bytes).await?;
    // todo: also write repodata.json.bz2, repodata.json.zst, repodata.json.jlap and sharded repodata once available in rattler
    // https://github.com/conda/rattler/issues/1096
```

crates/rattler_index/src/lib.rs
delsner: would that be `s3://my-bucket/my-channel` then? This field is never set in the repodata.json files that I have access to.

delsner: We decided to omit this, similar to what conda-forge and artifactory do.

pavelzw: this field actually does something else, you can point to a remote location for packages. for our use case, we can keep it `None`

pavelzw (approved):

crates/rattler_index/Cargo.toml
pavelzw, line 1: we still need to do the whole native-tls / rustls-tls and native-roots dance

crates/rattler_index/Cargo.toml
```diff
@@ -3,22 +3,49 @@ name = "rattler_index"
...
 categories.workspace = true
 homepage.workspace = true
 repository.workspace = true
 license.workspace = true
 readme.workspace = true
+default-run = "rattler-index"
+
+[[bin]]
+name = "rattler-index"
+path = "src/main.rs"
```
pavelzw, line 16: i would suggest we don't publish binaries for rattler-index in this repo as this will probably complicate the release-plz workflow in this repo quite a bit. instead, we can only build binaries on conda-forge (and other package managers like brew), imo this should be enough

crates/rattler_index/Cargo.toml
pavelzw: done in [`c7785b8` (#1076)](https://github.com/conda/rattler/pull/1076/commits/c7785b83c0897a8e934db406b7044197bb7286aa)

```
❯ cargo tree --package rattler_index --features rustls-tls --no-default-features | grep native
│   │   │   │   ├── rustls-native-certs v0.8.1
│   │   │   ├── rustls-native-certs v0.8.1 (*)
rattler on  rattler-index-opendal [$!?] via 🦀 v1.84.1 via 📦 v0.41.4
❯ cargo tree --package rattler_index --features native-tls --no-default-features -i rustls
warning: nothing to print.

To find dependencies that require specific target platforms, try to use option `--target all` first, and then narrow your search scope accordingly.

```

crates/rattler_conda_types/Cargo.toml
```diff
@@ -11,7 +11,15 @@ license.workspace = true
...
 [features]
-default = ["rayon"]
+default = ["rayon", "rustls-tls"]
+native-tls = [
+  "rattler_redaction/native-tls",
+  "rattler_package_streaming/native-tls",
+]
+rustls-tls = [
+  "rattler_redaction/rustls-tls",
+  "rattler_package_streaming/rustls-tls",
```
wolfv, line 21: this is not used in `rattler_conda_types`, correct?

crates/rattler_conda_types/Cargo.toml
```diff
@@ -11,7 +11,15 @@ license.workspace = true
 readme.workspace = true
 
 [features]
-default = ["rayon"]
+default = ["rayon", "rustls-tls"]
+native-tls = [
+  "rattler_redaction/native-tls",
```
wolfv, line 16: Is this really necessary?

pavelzw:
not sure in what way the python failures are related or not, but https://github.com/conda-forge/rust-activation-feedstock/pull/72 is the conda-forge update to 1.85.0

wolfv:
Things are fixed on main now but looks like I broke this PR. Sorry :(

wolfv (approved):

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,133 @@
+use clap::{arg, Parser, Subcommand};
+use clap_verbosity_flag::Verbosity;
+use rattler_conda_types::Platform;
+use rattler_index::{index_fs, index_s3};
+use tracing_log::AsTrace;
```
wolfv, line 5: Remove thsi

crates/rattler_index/src/main.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+/// Entry point of the `rattler-index` cli.
+#[tokio::main]
+async fn main() -> anyhow::Result<()> {
+    // Parse the command line arguments
+    let cli = Cli::parse();
+
+    tracing_subscriber::FmtSubscriber::builder()
+        .with_max_level(cli.verbose.log_level_filter().as_trace())
```
wolfv, line 100: You can enable the `tracing` feature on `clap_verbose_...` and then use the right function.

wolfv (changes requested):

crates/rattler_index/src/main.rs
wolfv: https://github.com/clap-rs/clap-verbosity-flag/blob/master/examples/tracing.rs

crates/rattler_index/src/lib.rs
```diff
@@ -97,117 +127,319 @@ pub fn package_record_from_conda(file: &Path) -> std::io::Result<PackageRecord>
...
+                            .acquire()
+                            .await
+                            .expect("Semaphore was unexpectedly closed");
+                        pb.set_message(format!(
+                            "Indexing {} {}",
+                            subdir.as_str(),
+                            console::style(filename.clone()).dim()
+                        ));
+                        let file_path = format!("{subdir}/{filename}");
+                        let buffer = op.read(&file_path).await?;
```
Xuanwo, line 246: The returning `Buffer` implements `bytes::Buf` so we can use:

```rust
package_record_from_tar_bz2_reader(buffer.reader())
```

In this way, we can avoid some extra copy of data.

pavelzw: thanks! implemented in #1123

PR #1081, opened 2025-02-20
fix: support file URL for run exports cache

nichmor:
## Overview

This PR adds support of  `file` scheme `URL's` in `run-exports cache`.  This means that we can now also pass some local conda archive to extract and cache `run_exports.json` from it


crates/rattler_cache/src/run_exports_cache/mod.rs
```diff
@@ -681,4 +689,42 @@ mod test {
 
         assert_ne!(first_cache_path.path(), second_package_cache.path());
     }
+
+    #[tokio::test]
+    // Test caching a run exports file by file:// URL
```
wolfv, line 694: This doesn't look like that's what the test does/

wolfv (commented):
Test needs adjustment and the title of the PR / description could be better ("handle filepath" doesn't really tell me much) :)

PR #1083, opened 2025-02-20
fix(js): use api-extractor to fix typings

baszalmstra:
Use https://api-extractor.com/ to also rollup the typing definition file.


js-rattler/package.json
```diff
@@ -38,5 +38,8 @@
         "webpack": "^5.96.1",
         "webpack-cli": "^5.1.4"
     },
-    "type": "module"
+    "type": "module",
```
trim21, line 41: this is wrong, we build umd output so this option should not be set.

baszalmstra: 👍

baszalmstra:
@trim21 Would love another review from you 😄

trim21:
LGTM

PR #1084, opened 2025-02-20
build(js): add esm output

trim21:
### Description

we only build umd output, which doesn't with with esm only environment. this add esm target.

### DIFF
--- a/js-rattler/package.json
+++ b/js-rattler/package.json
@@ -26,7 +26,7 @@
     },
     "scripts": {
         "build:wasm": "wasm-pack build --target bundler --mode normal --release",
-        "build": "webpack --mode=production && api-extractor run --verbose",
+        "build": "rimraf dist && webpack --mode=production && api-extractor run --verbose",
         "test": "cross-env NODE_OPTIONS=--experimental-vm-modules jest",
         "fmt": "prettier --write .",
         "fmt:check": "prettier --check ."
@@ -39,6 +39,7 @@
         "jest": "^29.7.0",
         "prettier": "^3.3.3",
         "prettier-plugin-jsdoc": "^1.3.0",
+        "rimraf": "^6.0.1",
         "ts-jest": "^29.2.5",
         "ts-loader": "^9.5.1",
         "ts-node": "^10.9.2",
--- a/js-rattler/package.json
+++ b/js-rattler/package.json
@@ -12,11 +12,18 @@
         "url": "https://github.com/conda/rattler/issues"
     },
     "homepage": "https://github.com/conda/rattler#readme",
-    "main": "./dist/main.js",
+    "main": "./dist/umd/main.js",
     "types": "./dist/index.d.ts",
     "files": [
         "dist/**/*"
     ],
+    "exports": {
+        ".": {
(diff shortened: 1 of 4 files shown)


baszalmstra:
Should we merge #1083 first?

trim21:
> Should we merge #1083 first?

Yes, I'll resolve the conflicts.

js-rattler/package.json
```diff
@@ -1,5 +1,5 @@
 {
-    "name": "@baszalmstra/rattler",
+    "name": "@conda-org/rattler",
```
baszalmstra, line 2: Unfortunately we dont have consensus within the steering council yet where this package should go, so for now we should keep it under my username.

trim21: ok

js-rattler/webpack.config.js
```diff
@@ -40,3 +35,30 @@ export default {
...
 };
+
+export default [
+    merge(options, {
+        output: {
+            assetModuleFilename: "[name][ext]",
+            clean: true,
+            path: path.resolve(__dirname, "dist/umd/"),
+            filename: "[name].js",
+            library: "js-rattler",
```
baszalmstra, line 46: Shouldnt this be `rattler`?

trim21: I don't know, I didn't touch it.

wolfv:
Merged #1083

trim21:
sadly with multiple output, wasm file will be included multiple times, have no idea how to avoid this:

image

baszalmstra:
@trim21 Maybe just having esm is fine?

js-rattler/webpack.config.mjs
```diff
@@ -40,3 +35,30 @@ export default {
...
 };
+
+export default [
+    merge(options, {
+        output: {
+            assetModuleFilename: "[name][ext]",
+            clean: true,
+            path: path.resolve(__dirname, "dist/umd/"),
+            filename: "[name].js",
+            library: "js-rattler",
```
baszalmstra, line 46:
```suggestion
            library: "rattler",
```

Doesnt matter when it was introduced, lets use the proper name.

baszalmstra:
Or maybe we can output both to the same folder but name the entry point file differently? Would that work?

trim21:
> Or maybe we can output both to the same folder but name the entry point file differently? Would that work?

that should work

trim21:
good idea

trim21:
I remove @wasm-tool/wasm-pack-plugin and add `npm run build:wasm` in npm build script, which is broken with new webpack config.

baszalmstra:
> I remove @wasm-tool/wasm-pack-plugin and add `npm run build:wasm` in npm build script, which is broken with new webpack config.

Sounds good. Can you also bump the version to `0.1.2`, Ill publish after this is merged.

baszalmstra:
We are currently only running jest for node. Does it make sense to also run the tests with a headless browser?

trim21:
I personally think running with nodejs only is fine, let's consider add it when someone report a browser only issue?

Issue #1089, opened 2025-02-20
js: umd build is broken


trim21:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

funny story:

this simple script doesn't work: `const { Version, VersionSpec } = require('@baszalmstra/rattler');`

```text
C:\Users\Trim21\proj\renovate\node_modules\.pnpm\@baszalmstra+rattler@0.1.2\node_modules\@baszalmstra\rattler\dist\umd.js:10
})(self, () => {
   ^

ReferenceError: self is not defined
    at Object.<anonymous> (C:\Users\Trim21\proj\renovate\node_modules\.pnpm\@baszalmstra+rattler@0.1.2\node_modules\@baszalmstra\rattler\dist\umd.js:10:4)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Object..js (node:internal/modules/cjs/loader:1689:10)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
    at Function._load (node:internal/modules/cjs/loader:1128:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:218:24)
    at Module.require (node:internal/modules/cjs/loader:1340:12)
    at require (node:internal/modules/helpers:141:16)
    at Object.<anonymous> (C:\Users\Trim21\proj\renovate\a.js:1:34)

Node.js v22.11.0
```

PR #1090, opened 2025-02-20
fix(js): add nodejs common js build

trim21:
### Description

close https://github.com/conda/rattler/issues/1089


baszalmstra:
Can you clarify a little bit maybe? why donwe need commonjs?

trim21:
sorry, my fault. we don't need to add new commonjs output

Issue #1091, opened 2025-02-20
js: can not load esm build or umd build in nodejs env


trim21:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

I'm trying with both require or import to load our npm package `@baszalmstra/rattler` and non of them work.

I think this is becasue webpack assume it's web env and call `fetch` to load the wasm file, instead of `fs.readFileSync`

![Image](https://github.com/user-attachments/assets/8f424a10-0da5-4e0d-9359-a7e63c2f831a)

baszalmstra:
mm maybe we can do something with our invocation to wasmpack?

trim21:
sorry, I'm not a native english speaker, what do you mean by "our invocation to wasmpack"?

trim21:
I have a working nodejs common js build at https://github.com/conda/rattler/pull/1093

baszalmstra:
thats exactly what I meant! 👍

PR #1093, opened 2025-02-20
build(js): replace webpack with rollup and build universal esm package

trim21:
### Description

our umd build and esm can't be loaded on nodejs env, webpack actually build a umd/esm that use `fetch` to load the wasm file asynchronously.

So we need a new bundler(rollup here) to build a output for nodejs.

It currently build nodejs commonjs and esm files.

close https://github.com/conda/rattler/issues/1091


baszalmstra:
Nice! Ill take a closer look later today!

trim21:
you should be take to test it with this script after run `npm run build`

`js-rattler/main.cjs`

```js
const { Version, VersionSpec } = require('@baszalmstra/rattler');

new Version('1.2.3');
```

trim21:
I updated it to make it build both cjs and esm output for nodejs env.

trim21:
> We still need some fix for downstream bundler users.

It get very ditry to support this case, let's discuss this after this pr is merged

baszalmstra:
Why do we need a second bundler for this? Cant webpack be instructed to use something other than fetch?

trim21:
I think we should replace webpack with rollup and drip umd output， build 2 targets nodejs cjs and universal esm only

baszalmstra:
Would you be up to doing that in this PR? Im not a fan of having two different bundlers.

trim21:
> Would you be up to doing that in this PR? Im not a fan of having two different bundlers.

Yes, it's possible, but in a very anti-pattern way

trim21:
this should also work with downstream bundler users

js-rattler/tsconfig.rollup.json
```diff
@@ -0,0 +1,17 @@
+{
```
baszalmstra, line 1: Maybe we can use [extends](https://www.typescriptlang.org/tsconfig/#extends) to remove some duplication here?

baszalmstra (commented):
This looks much better!

js-rattler/tsconfig.rollup.json
trim21: done

baszalmstra:
When I now run `npm run build` I think the `dist` folder contains more files than we need (`Version.js`, `VersionSpec.js`). Is that correct? I assumed we only need `rattler.bundler.mjs`, `rattler.node.cjs` and the wasm file (and types).

image

trim21:
That looks like output of tsc, which is unexpected and should be removed.

I'll fix this tomorrow.

baszalmstra:
No worries! Ill give it a try!

trim21:
Webpack emit ts types in the past, and now we need to run tsc to generate types, but I forgot it also emit js files.

trim21:
Looks good, maybe we can drop rimraf too

baszalmstra:
Good idea, I removed it in favor of shx. If CI turns green Ill merge and publish.

baszalmstra:
Published as [0.1.3](https://www.npmjs.com/package/@baszalmstra/rattler).

baszalmstra:
Thanks for all the work @trim21 ! Looking forward to your next contribution. 😄

trim21:
I do need some new methods on VersionSpec, I'll add it later

PR #1097, opened 2025-02-24
fix: compiling with pixi on linux

Hofer-Julian:
I mostly had to switch from native-tls to rust-tls.

Annoyingly, I didn't manage to convince the build-process to find libclang needed to compile libsolv so I skip that test for now

Tagging @pavelzw since you also hit this problem recently IIRC

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -16,7 +16,7 @@ build = "cargo build"
 fmt = "cargo fmt"
 lint = "cargo clippy"
 check = "cargo check"
-test = "cargo test"
+test = "cargo nextest run --workspace --no-default-features --features=rustls-tls -E 'not test(libsolv_bindings_up_to_date)'"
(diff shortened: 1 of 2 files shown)


pixi.toml
```diff
@@ -16,7 +16,7 @@ build = "cargo build"
 fmt = "cargo fmt"
 lint = "cargo clippy"
 check = "cargo check"
-test = "cargo test"
+test = "cargo nextest run --workspace --no-default-features --features=rustls-tls -E 'not test(libsolv_bindings_up_to_date)'"
```
Hofer-Julian, line 19:
```suggestion
# libsolv compilation cannot find pixi's clang for some reason
# so we skip that test for now
test = "cargo nextest run --workspace --no-default-features --features=rustls-tls -E 'not test(libsolv_bindings_up_to_date)'"
```

ruben-arts (approved):

PR #1105, opened 2025-02-26
fix: clippy lint

Hofer-Julian:


Hofer-Julian:
It improves *some* of the clippy lints, so I will merge this now

PR #1108, opened 2025-02-27
feat(js): compile `rattler_solve` and `rattler_repodata_gateway`

baszalmstra:
This PR adds the ability to use both `rattler_solve` and `rattler_repodata_gateway`. 

WASM does not support filesystem access without extensions like emscripten or wasi, neither are supported by our build. But rattler aggressively caches data to disk so this PR adds additional code paths to *not* cache anything (except for maybe in memory). This allows compiling rattler_repodata_gateway to WASM. 

Its also not possible to spawn futures or run blocking code on a different thread. Concurrency is supported, and its also used. But spawning is not supported because there is no tokio runtime active. I worked around this by conditionally running blocking code on the current thread.

I did not yet expose these crates to the javascript API yet, that should be done in a separate PR. I did add a proof of concept that fetches repodata and performs a solve from javascript. This also supports sharded repodata! 

Running a solve for `python`, `matplotlib`, `numpy` for `emscripten-wasm32` in the browser takes 0.2 seconds (without any cache) and results in this nice view in the chrome network debugger:

image

(Uses `https://prefix.dev/emscripten-forge-dev` and `https://prefix.dev/conda-forge`)


crates/rattler_repodata_gateway/src/gateway/subdir.rs
```diff
@@ -107,40 +108,28 @@ impl SubdirData {
...
 
-        // At this point we have exclusive write access to this specific entry. All other tasks
-        // will find a pending entry and will wait for the records to become available.
+        // At this point we have exclusive write access to this specific entry. All
+        // other tasks will find a pending entry and will wait for the records
+        // to become available.
         //
-        // Let's start by fetching the records. If an error occurs we immediately return the error.
-        // This will drop the sender and all other waiting tasks will receive an error.
-        let records = match tokio::spawn({
```
baszalmstra, line 115: I removed the spawn here because I dont think its needed. All implementations fan out to spawn blocking when blocking operations are performed.

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -55,50 +55,53 @@ enum SparseRepoDataInner {
...
-    /// legal.
-    #[borrows(bytes)]
-    #[covariant]
-    repo_data: LazyRepoData<'this>,
-}
+// A struct that holds a memory map of a `repodata.json` file and also a
+// self-referential field which indexes the data in the memory map with a
+// sparsely parsed json struct. See [`LazyRepoData`].
+#[cfg(any(unix, windows))]
+self_cell::self_cell!(
```
baszalmstra, line 69: I replaced ouroborous with self_cell because I couldnt get to conditionally compile ouroboros. self_cell should also compile faster anyway.

js-rattler/Cargo.toml
```diff
@@ -31,14 +32,29 @@ wee_alloc = { version = "0.4.5", optional = true }
...
-# Tell `rustc` to optimize for small code size.
+## Tell `rustc` to optimize for small code size.
 opt-level = "s"
 lto = true
 
+[patch.crates-io]
+# We want to use version 0.4.19 but its not released yet.
+# https://github.com/Nullus157/async-compression/pull/319
+async-compression = { git = "https://github.com/Nullus157/async-compression.git" }
+bzip2 = { git = "https://github.com/baszalmstra/bzip2-rs.git", branch = "fix/wasm" }
```
baszalmstra, line 55: I need to upstream this before merging.

baszalmstra: See: https://github.com/trifectatechfoundation/bzip2-rs/pull/131

wolfv: Done, right?!

baszalmstra: Yeah, now all we need is an async compression release! But we can merge without it

baszalmstra: Lets go! 👍

wolfv (approved):

wolfv:
I merged main in preparation of merging this!

PR #1109, opened 2025-02-27
fix: Code review in rattler-index and test fix

pavelzw:
### Description

Follow-up of #1076


crates/rattler_index/src/main.rs
```diff
@@ -2,7 +2,6 @@ use clap::{arg, Parser, Subcommand};
 use clap_verbosity_flag::Verbosity;
 use rattler_conda_types::Platform;
 use rattler_index::{index_fs, index_s3};
-use tracing_log::AsTrace;
```
wolfv, line 5: remove from Cargo.toml, too, pls :)

PR #1117, opened 2025-02-27
chore: Add Cargo.lock

### LINKED ISSUE #1103: Checking in `Cargo.lock`
Like many Rust libraries, we don't check `Cargo.lock` into git. I think we should re-evalutate that decision.

Lock files have many advantages. For downstream `Cargo.lock` is disregarded, so only advantages for rattler developers matter:
- Multiple checkouts of the same commit have the same dependency tree (you and CI, you and your colleague)
- You can go back to older commits and have the same dependency tree as when the commit was made

The main argument against checking in the lock file is, that as mentioned before, the dependencies of downstream users will be locked differently. A new version of your dependencies might break your users' setup when they update, but you don't notice since the dependencies are locked.

With Dependabot that problem is easily avoided. You just let it update your `Cargo.lock` on a regular basis and merge the PR if CI is green and deal with the problem if CI is red. It's even nicer since you see which crates updated and are therefore responsible for the breakage.

That is one of the reasons why the Cargo team moved away from their recommendation to "commit Cargo.lock file for packages with binaries but not libraries": https://blog.rust-lang.org/2023/08/29/committing-lockfiles.html

The new [msrv aware resolver](https://doc.rust-lang.org/nightly/cargo/CHANGELOG.html#cargo-184-2025-01-09) should avoid even further breakage for downstream users that we have to keep in mind.

pavelzw:
### Description

Closes #1103


baszalmstra:
Maybe we can also take the opportunity to update the dependabot config? 

Ideally we also update the lock files for js-rattler and py-rattler and group updates together in a single pr.

pavelzw:
like this @baszalmstra?

.github/dependabot.yml
```diff
@@ -1,34 +1,60 @@
...
+    labels:
+      - dependencies
+    commit-message:
+      prefix: bump
+    groups:
+      rust-py-rattler:
+        patterns:
+          - "*"
+  - package-ecosystem: cargo
+    directory: /
```
baszalmstra, line 28:
```suggestion
    directory: /js-rattler
```

baszalmstra (commented):
Yes! Thanks!

baszalmstra (approved):

Issue #1073, opened 2025-02-18
Arch and platform fields of RepoDataRecord do not round trip through lockfiles


apmorton:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

```python
import rattler
import asyncio

recs = asyncio.run(
    rattler.solve(
        channels=["conda-forge"],
        specs=[
            "_libgcc_mutex ==0.1",
        ],
        locked_packages=None,
    )
)
rattler.LockFile(
    {"env": rattler.Environment("env", {rattler.Platform.current(): recs}, [])}
).to_path("env.lock1")

recs = asyncio.run(
    rattler.solve(
        channels=["conda-forge"],
        specs=[
            "_libgcc_mutex ==0.1",
        ],
        locked_packages=rattler.LockFile.from_path("env.lock1")
        .environment("env")
        .conda_repodata_records_for_platform(rattler.Platform.current()),
    )
)
rattler.LockFile(
    {"env": rattler.Environment("env", {rattler.Platform.current(): recs}, [])}
).to_path("env.lock2")
```

```
$ diff -u env.lock1 env.lock2
--- env.lock1   2025-02-18 13:00:28.502774319 -0600
+++ env.lock2   2025-02-18 13:00:28.503774312 -0600
@@ -9,8 +9,6 @@
 - conda: https://conda.anaconda.org/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
   sha256: fe51de6107f9edc7aa4f786a70f4a883943bc9d39b3bb7307c04c41410990726
   md5: d7c89558ba9fa0495403155b64376d81
-  arch: x86_64
-  platform: linux
   license: None
   size: 2562
   timestamp: 1578324546067
```

### Additional Context

I don't really know where the root of this issue lies, or if I'm using something incorrectly here.

This is a minimal repro of an issue I'm running into with round-tripping lockfiles.

When we solve an environment with no `locked_packages`, write a lockfile, then re-solve every single package drops the above two lines in the resulting lockfile.

wolfv:
I think these issues might be related: https://github.com/prefix-dev/pixi/issues/3233

baszalmstra:
This has been fixed by: https://github.com/conda/rattler/pull/1124

We need a py-rattler release.

apmorton:
This landed in py-rattler [v0.11.0](https://github.com/conda/rattler/releases/tag/py-rattler-v0.11.0)

PR #1124, opened 2025-02-28
fix: roundtrip of arch/platform in lock files

wolfv:
We were filling in the `arch` and `platform` fields in the lockfile if they didn't come from the repodata, but we didn't serialize this information on the first go. 

When loading the data from lockfile and re-serializing, we did write it out leading to roundtrip issues as noted in #1073

In this PR we just remove the writing of `arch` and `platform`.


crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -457,33 +463,17 @@ fn determine_subdir(
...
-            "x86" => Ok(Platform::Win32),
-            "x86_64" => Ok(Platform::Win64),
-            "arm64" => Ok(Platform::WinArm64),
-            _ => Err(ConvertSubdirError::NoKnownCombination { platform, arch }),
-        },
-        _ => Err(ConvertSubdirError::NoKnownCombination { platform, arch }),
-    }?;
-    // Convert back to Platform string which should correspond to known subdirs
-    Ok(plat.to_string())
+    match arch.parse::<Arch>() {
```
wolfv, line 466: the previous function was not covering e.g. `emscripten-wasm32` so I modernized it a bit.

baszalmstra (approved):
Looks great!

PR #1126, opened 2025-02-28
chore: update dependencies

wolfv:
Catching us up before the dependabot does. `nom` is still breaking, as is `generic-array`.


Hofer-Julian (approved):
Looks good!

PR #1128, opened 2025-02-28
feat: add reinstallation method to installer and transaction

tdejager:
This adds a re-installation method to the installer so that we can "force" reinstall packages which is something we want to do in pixi.

I've only added a method for testing the transaction, not sure if this is the right way to go about it :)

### DIFF
--- a/crates/rattler/src/install/unlink.rs
+++ b/crates/rattler/src/install/unlink.rs
@@ -364,6 +364,10 @@ mod tests {
     #[cfg(windows)]
     #[tokio::test]
     async fn test_unlink_package_in_use() {
+        use crate::get_repodata_record;
+        use crate::install::link_package;
+        use crate::install::InstallOptions;
+        use rattler_conda_types::PrefixRecord;
         use std::{
             env::{join_paths, split_paths, var_os},
             io::{BufRead, BufReader},


wolfv (approved):

crates/rattler/src/install/installer/mod.rs
```diff
@@ -216,6 +218,15 @@ impl Installer {
         }
     }
 
+    /// Set the packages that we want explicitly to be reinstalled.
+    #[must_use]
+    pub fn with_reinstall_packages(self, reinstall: Vec<PackageName>) -> Self {
```
baszalmstra, line 223: Can you also add a set_...

crates/rattler/src/install/installer/mod.rs
```diff
@@ -314,6 +325,9 @@ impl Installer {
         let transaction = Transaction::from_current_and_desired(
             installed,
             records.into_iter().collect::<Vec<_>>(),
+            self.reinstall_packages
```
baszalmstra, line 328: option implements into_iter, so you can do .into_iter().flatten()

crates/rattler/src/install/installer/mod.rs
```diff
@@ -50,6 +51,7 @@ pub struct Installer {
     target_platform: Option<Platform>,
     apple_code_sign_behavior: AppleCodeSignBehavior,
     alternative_target_prefix: Option<PathBuf>,
+    reinstall_packages: Option<Vec<PackageName>>,
```
baszalmstra, line 54: Maybe this should be a hashset set we only test for inclusion.

crates/rattler/src/install/installer/mod.rs
tdejager: Oh wait I get it :)

crates/rattler/src/install/installer/mod.rs
tdejager: Done

crates/rattler/src/install/installer/mod.rs
tdejager: Do you want to pass around the `HashSet`? If so we should get rid of the iterator as well in the `Transaction` I think :)

baszalmstra: Yeah makes sense!

tdejager: Ok I've changed it to accept and pass `HashSet`'s around :)

tdejager:
> This would also be useful in `py-rattler`. Would you be open to exposing this in the `py-rattler` bindings as well?

For sure, let me do that end of the week :)

Issue #1096, opened 2025-02-23
convenience functions for writing `RepoData`


pavelzw:
after reworking `rattler_index` in #1076, i think it would be nice to add some convenience functions to generate the different repodata formats, including:

- [x] repodata.json
- [ ] ~repodata.json.bz2~
- [ ] repodata.json.zst (#1312)
- [ ] ~repodata.jlap~
- [ ] sharded repodata (#1312)

rattler-index (and conda-mirror, which i'm currently also working on right now) could then use these functions to generate all repodata variants that are supported by pixi/conda/mamba right now.

- [x] we also should support repodata patching #1129

baszalmstra:
I dont think we need to support jlap. The CEP never passed and it has been superseded by sharded repodata.

pavelzw:
i also don't see much value in `repodata.json.bz2` so let's skip that one as well...

pavelzw:
#1312 should solve this

PR #1129, opened 2025-03-03
feat: Add support for repodata patching in rattler-index, fix silent failures

pavelzw:
### Description

xref #1096


crates/rattler_index/src/lib.rs
```diff
@@ -443,22 +532,57 @@ pub async fn index<T: Configurator>(
...
-            tokio::spawn(index_subdir(
-                *subdir,
-                op.clone(),
-                force,
-                multi_progress.clone(),
-                semaphore.clone(),
-            ))
-        })
-        .collect::<Vec<_>>();
-    try_join_all(tasks).await?;
```
pavelzw, line 461: this was silently failing before because it only checked whether the join handles were okay but not whether the actual return values of the tasks were

crates/rattler_index/src/lib.rs
```diff
@@ -229,53 +269,80 @@ async fn index_subdir(
...
+        match join_result {
+            Ok(Ok(result)) => results.push(result),
+            Ok(Err(e)) => {
+                tasks.clear();
+                tracing::error!("Failed to process package: {}", e);
+                pb.abandon_with_message(format!(
+                    "{} {}",
+                    console::style("Failed to index").red(),
+                    console::style(subdir.as_str()).dim()
+                ));
```
pavelzw, line 318: this is not perfect as the progress bar can get overwritten by other tasks that are still executed after this failure. haven't seen a nice way to avoid this behavior

image

i think we can keep it for now like this since the error messages are printed anyway and i expect this script to be executed in non-interactive environments anyway most of the time

crates/rattler_index/src/main.rs
```diff
@@ -32,13 +32,18 @@ struct Cli {
 
     /// The maximum number of packages to process in-memory simultaneously.
     /// This is necessary to limit memory usage when indexing large channels.
-    #[arg(long, default_value = "128", global = true)]
+    #[arg(long, default_value = "32", global = true)]
```
pavelzw, line 35: when testing it out with 128 on aws s3, i regularly got errors that weren't there when limiting ourselves to 32 maximum concurrent downloads

pavelzw:
the failing JS tests don't look related, right?

pavelzw:
@wolfv @baszalmstra this is ready for review

wolfv:
Just FYI, a PatchInstructions v2 file will be landing soon: https://github.com/conda/ceps/pull/108

pavelzw:
Will keep this in mind when implementing sharding in rattler-index 👍🏻

wolfv (approved):

PR #1130, opened 2025-03-03
feat(js): add Gateway object

baszalmstra:
Adds the `Gateway` object to the javascript bindings. I did not yet expose the most important method `query` in this PR yet, Ill do that in a followup PR. I left this out because it also requires exposing `RepoDataRecord` and there are still a lot of types missing to properly do that.

I also added types for `Platform` and `PackageName`. Both are completely defined in typescript instead of using bindings because I think that provides the best ergonomics. One cool feature that I also implemented is that we can check at compile time that a string literal is actually a valid package name!

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6463,13 +6463,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "wasm-bin"
-version = "0.1.0"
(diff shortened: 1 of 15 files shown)


wolfv:
Awesome PR by the way!

PR #1136, opened 2025-03-04
fix: shortcut filename

Hofer-Julian:
Otherwise, Windows will not know what to do with it


wolfv:
Looks good! we should also double check if we should keep the `(env)` part in here.

ruben-arts:
This works on my windows machine compiled into pixi

wolfv:
This was indeed changed in upstream menuinst: https://github.com/conda/menuinst/blob/fa02b329a15346728664edc7372bd43691ca915f/menuinst/platforms/win.py#L242-L244

Let's remove the `env` stuff.

ruben-arts (approved):

PR #1138, opened 2025-03-04
refactor(js): remove double wrap

baszalmstra:
This PR removes the double wrapping of `Version` and `VersionSpec` and moves all the logic to Rust. This avoids creating an  additional indirection in javascript which is easier to maintain and should be faster.

Initially I double wrapped the types meaning we expose type A from rust, then create type B in typescript that wraps A and only expose B to the user. The advantage of this approach is that we can provide a more ergonomic API to the user. But I found wasm-bindgen flexible enough that we can provide the same level of ergonomics directly from Rust. wasm-bindgen allows overriding types which allows manual control in places where the generated API is a bit ugly.


wolfv:
nice!

wolfv (approved):

PR #1143, opened 2025-03-09
fix: Add location to the cache key

nilskch:
### Description

This PR is trying to solve an issue that was kind of already mentioned in this comment (cc @baszalmstra):
https://github.com/conda/rattler/blob/c7e2197d0bf8cd50a629e352e644ced7c752d9e7/crates/rattler_cache/src/package_cache/cache_key.rs#L7

Over at Palantir, users have the ability to generate custom conda packages and publish them to a private conda channel. The cache key currently consists of a `name`, `version`, and a `build_string` (and an optional `sha256` that we do not use in our custom conda packages).

Unfortunately, we set the `build_string` to a hardcoded value of `py_0`, so realistically the cache key only consists of the `name` and the `version` for those packages.

One of our forward deployed engineers ran into an issue where they tried to install a conda package with the same `name` and the same `version` of a package that they previously installed and they got a false cache hit. They installed a wrong package from a completely different customer and environment.

This is not only a problem because of the missing `sha256` or the hardcoded `build_string`, but this also opens the door for cache poison attacks if you control a conda channel.

### Offered solution

You can either install a conda package from a path ([get_or_fetch_from_path](https://github.com/conda/rattler/blob/c7e2197d0bf8cd50a629e352e644ced7c752d9e7/crates/rattler_cache/src/package_cache/mod.rs#L186)) or from an url ([get_or_fetch_from_url](https://github.com/conda/rattler/blob/c7e2197d0bf8cd50a629e352e644ced7c752d9e7/crates/rattler_cache/src/package_cache/mod.rs#L170)). This PR introduces a new property `location_hash` in the cache key, which is either a hash of the Url without the basic auth, or a hash of the path.

This way we also encode into the key where the package is coming from and therefore avoid running into cache hits with different packages with the same properties.

It might be also good to add the same logic to the

### Important notes
* This solution will invalidate all existing caches and users will need to install all the packages from the source again.
* We might want to add the same logic to the [run_exports_cache](https://github.com/conda/rattler/tree/main/crates/rattler_cache/src/run_exports_cache). As far as I can tell the `run_exports_cache` is currently not being used anywhere in the code and it is also quite new, so not sure what the plans are here - cc @nichmor

### DIFF
--- a/crates/rattler_cache/src/package_cache/cache_key.rs
+++ b/crates/rattler_cache/src/package_cache/cache_key.rs
@@ -39,7 +39,7 @@ impl CacheKey {
 
     /// Adds a hash of the Path to the cache key
     pub fn with_path(mut self, path: &Path) -> Self {
-        let path_hash = compute_bytes_digest::<Sha256>(path.to_os_str().as_encoded_bytes());
+        let path_hash = compute_bytes_digest::<Sha256>(path.as_os_str().as_encoded_bytes());
         self.origin_hash = Some(format!("{path_hash:x}"));
         self
     }
--- a/crates/rattler_cache/src/package_cache/cache_key.rs
+++ b/crates/rattler_cache/src/package_cache/cache_key.rs
@@ -39,7 +39,7 @@ impl CacheKey {
 
     /// Adds a hash of the Path to the cache key
     pub fn with_path(mut self, path: &Path) -> Self {
-        let path_hash = compute_bytes_digest::<Sha256>(path.to_string_lossy().as_bytes());
+        let path_hash = compute_bytes_digest::<Sha256>(path.to_os_str().as_encoded_bytes());
         self.origin_hash = Some(format!("{path_hash:x}"));
         self
     }
--- a/crates/rattler_cache/src/package_cache/cache_key.rs
+++ b/crates/rattler_cache/src/package_cache/cache_key.rs
@@ -1,7 +1,10 @@
 use rattler_conda_types::package::ArchiveIdentifier;
 use rattler_conda_types::PackageRecord;
-use rattler_digest::Sha256Hash;
-use std::fmt::{Display, Formatter};
+use rattler_digest::{compute_bytes_digest, compute_url_digest, Sha256, Sha256Hash};
+use std::{
+    fmt::{Display, Formatter},
+    path::Path,
+};
 
 /// Provides a unique identifier for packages in the cache.
 /// TODO: This could not be unique over multiple subdir. How to handle?
@@ -11,6 +14,7 @@ pub struct CacheKey {
     pub(crate) version: String,
     pub(crate) build_string: String,
     pub(crate) sha256: Option<Sha256Hash>,
+    pub(crate) origin_hash: Option<String>,
 }
 
 impl CacheKey {
@@ -23,6 +27,20 @@ impl CacheKey {
     /// Potentially adds a sha256 hash of the archive.
     pub fn with_opt_sha256(mut self, sha256: Option<Sha256Hash>) -> Self {
         self.sha256 = sha256;
+        self
+    }
+
+    /// Adds a hash of the Url to the cache key
+    pub fn with_url(mut self, url: url::Url) -> Self {
+        let url_hash = compute_url_digest::<Sha256>(url);
+        self.origin_hash = Some(format!("{url_hash:x}"));
+        self
+    }
+
+    /// Adds a hash of the Path to the cache key
+    pub fn with_path(mut self, path: &Path) -> Self {
+        let path_hash = compute_bytes_digest::<Sha256>(path.to_string_lossy().as_bytes());
+        self.origin_hash = Some(format!("{path_hash:x}"));
         self
     }
 }
@@ -41,6 +59,7 @@ impl From<ArchiveIdentifier> for CacheKey {
             version: pkg.version,
             build_string: pkg.build_string,
             sha256: None,
+            origin_hash: None,
         }
     }
 }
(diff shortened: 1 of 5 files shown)


crates/rattler_cache/src/package_cache/cache_key.rs
```diff
@@ -1,7 +1,10 @@
...
-use rattler_digest::Sha256Hash;
-use std::fmt::{Display, Formatter};
+use rattler_digest::{compute_bytes_digest, compute_url_digest, Sha256, Sha256Hash};
+use std::{
+    fmt::{Display, Formatter},
+    path::Path,
+};
 
 /// Provides a unique identifier for packages in the cache.
 /// TODO: This could not be unique over multiple subdir. How to handle?
```
nilskch, line 10:
```suggestion
```

I am not sure what you mean with "over multiple subdir"? Does this fix the `TODO`?

baszalmstra (approved):

baszalmstra:
Unfortunately this will invalidate all caches because it changes the location of the cache entries. We have been trying to avoid that, and we also tried to remain consistent with condas cache. I think we can forgo the latter requirement but I do think we need a migration scheme where old cache entries are reused if possible.

baszalmstra (changes requested):
I was too quick with my approval.

As stated in my comment, we need a migration scheme where old cache entries are reused if possible.

Alternatively, we could also store the location in the cache lock files. That will at least make sure the cache is invalidated if a package is requested from somewhere else. We did the same thing when we added sha hashes.

crates/rattler_cache/src/package_cache/cache_key.rs
baszalmstra: A package with the same name, version, and build string can exist in multiple subdirs. When we added sha hashes this problem was already fixed.

However, since only one entry is added in the cache it can often be overwritten which is bad. Your change fixes that problem too so we can indeed remove the todo.

nilskch:
> I do think we need a migration scheme where old cache entries are reused if possible.

This would be super nice, but I think we have no way of knowing where the package from the cache came from. Do you have suggestions here?

> Alternatively, we could also store the location in the cache lock files.

If we do this, we cannot cache all packages with the same properties but different origins at the same time, so we would invalidate the cache a lot. We also have the problem that we do not know where the package in the current cache is coming from.

baszalmstra:
I think we'll just have to assume that if no location is known the location is equal. That basically acts the same as it does now. New cache entries will have the location added so over time this problem will be solved.

Alternatively we rename the cache entries as soon as we encounter an entry with a location.

nilskch:
I like the rename migration strategy a little bit better, but I am happy to implement either now. The only problem that I see is that this might lead to writing the wrong location to a cache entry. I think most users of rattler don't use private conda channels, so they won't have this issue. I think at Palantir we will add a migration strategy where we clean all caches.

Which migration strategy do you prefer? I can implement it now.

baszalmstra:
Cleaning the cache can easily lead to bloating of disk space because previously linked files will be stored twice, once it the linked environments and once in the new cache entry.

I think renaming makes the most sense. It will essentially act the same as the current behavior. And in pixi (and presumably most other cases) we also have the sha256 hash to work around the issue you are facing.

Care has to be taken with renaming because multiple processes may be involved. You will essentially have to lock both entries in the cache before renaming.

nilskch:
Sounds good!

nilskch:
Renaming the cache folder does not really work cross platform - you would need to move every file over individually and recreate the folder structure. It's quite scary to get this right with symlinks; do you know a nice way to make a copy/rename work here?

nilskch:
> I think we'll just have to assume that if no location is known the location is equal. That basically acts the same as it does now. New cache entries will have the location added so over time this problem will be solved.

I now think this is the better alternative here given that moving the files is not trivial.

baszalmstra:
Can you elaborate? What doesnt really work cross platform?

baszalmstra:
Mmm another issue I just thought of is how we deal with mirrors.. Now the location url is based on where you download the package, but in case of multiple mirrors that might be problematic. Maybe we can change the API to allow the user to pass "original" url?

nilskch:
I ran into issues where I was not able to rename some of the folders and I don't fully understand why. The docs of `std::fs::rename` state that the behavior differs between platforms and also might change in the future.

> Maybe we can change the API to allow the user to pass "original" url?
Ok sounds good to me. This way we wouldn't need to implement a migration strategy. I will push an implementation this evening.

nilskch:
Sorry things got in the way - I will continue working on this next week.

nilskch:
Hey @baszalmstra, sorry for the delay, the PR is ready for a review now. I added the method `with_cached_origin` to the `PackageCache`. If enabled, the path or url is hashed into the cache key.

nilskch:
@baszalmstra do you know why checks fail? They passed and I just rebased on main

baszalmstra:
Its a problem on main. We’ll fix it today or tomorrow.

nilskch:
Thanks! Do you think you have time this week to look into this PR?

nilskch:
Hey, I am not sure which comment you mean :D is there anything left to do to get this PR merged? It won't impact consumers of the library unless they explicitly enable `with_cached_origin`

crates/rattler_cache/src/package_cache/cache_key.rs
```diff
@@ -25,6 +29,20 @@ impl CacheKey {
...
+    /// Adds a hash of the Url to the cache key
+    pub fn with_url(mut self, url: url::Url) -> Self {
+        let url_hash = compute_url_digest::<Sha256>(url);
+        self.origin_hash = Some(format!("{url_hash:x}"));
+        self
+    }
+
+    /// Adds a hash of the Path to the cache key
+    pub fn with_path(mut self, path: &Path) -> Self {
+        let path_hash = compute_bytes_digest::<Sha256>(path.to_string_lossy().as_bytes());
```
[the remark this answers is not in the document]
nilskch: sounds good!

crates/rattler_cache/src/package_cache/cache_key.rs
```diff
@@ -25,6 +29,20 @@ impl CacheKey {
...
+    /// Adds a hash of the Url to the cache key
+    pub fn with_url(mut self, url: url::Url) -> Self {
+        let url_hash = compute_url_digest::<Sha256>(url);
+        self.origin_hash = Some(format!("{url_hash:x}"));
+        self
+    }
+
+    /// Adds a hash of the Path to the cache key
+    pub fn with_path(mut self, path: &Path) -> Self {
+        let path_hash = compute_bytes_digest::<Sha256>(path.to_os_str().as_encoded_bytes());
```
baszalmstra, line 42:
```suggestion
        let path_hash = compute_bytes_digest::<Sha256>(path.as_os_str().as_encoded_bytes());
```

baszalmstra (approved):

PR #1148, opened 2025-03-10
feat: package record

baszalmstra:
Adds the `PackageRecord` type.


js-rattler/crate/package_name.d.ts
```diff
@@ -0,0 +1,188 @@
...
+ * - Lowercase letters (`a-z`)
+ * - Uppercase letters (`A-Z`)
+ * - Digits (`0-9`)
+ * - Underscore (`_`)
+ * - Dash (`-`)
+ * - Dot (`.`)
+ *
+ * @public
+ */
+export declare type PackageNameChar =
```
wolfv, line 36: Would it be better to use a regex?

wolfv (approved):

js-rattler/crate/package_name.d.ts
baszalmstra: Yeah, all the runtime functions use a regex, but the type definitions cannot use regexes so we have to write them out unfortunately.

PR #1155, opened 2025-03-12
refactor: add the remove_from_backup function and update the prefix

prsabahrami:
### Description

As part of a fix for https://github.com/prefix-dev/pixi/issues/3324.
This uses the function from https://github.com/rust-lang/cargo/pull/7192 to remove the environment prefix from backup on macos. 

We get the below results now: 
```sh
cargo r --bin rattler -- create python numpy -c conda-forge
```

```sh
tmutil isexcluded .prefix
>>> [Excluded]  /Users/parsabahraminejad/Desktop/..../rattler/.prefix
```


crates/rattler/src/install/mod.rs
```diff
@@ -587,6 +614,9 @@ pub fn link_package_sync(
     // Ensure target directory exists
     fs_err::create_dir_all(target_dir).map_err(InstallError::FailedToCreateTargetDirectory)?;
 
+    #[cfg(target_os = "macos")]
+    exclude_from_backups(Path::new(&target_prefix));
```
wolfv, line 618: hmm, this function will be called for each package that is going to be linked (which might be 100+ times) while we only need to call it once.

I think the ideal solution would be to port over the `Prefix` type from `pixi` that would take care of all of these initializations upon calling `Prefix::create(opts)`.

All functions that curerntly take a `target_prefix: &Path` shoudl then take a `target_prefix: &Prefix` and `Prefix` should implemnet something like `AsRef<&Path>` or have a `prefix.path()` function.

I would also be OK with merging this without the `Prefix` type refactor, but then I think we should call this function from another place where it is only called once when linking packages.

wolfv (changes requested):

prsabahrami:
@wolfv I have created `Prefix` type which upon creation would remove the folder from backups as you mentioned. Let me know what you think!

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -145,6 +145,76 @@ impl RecordFromPath for PrefixRecord {
...
+}
+
+impl Prefix {
+    /// Create a new prefix, initializing it properly (creating directories, excluding from backups, etc.)
+    pub fn create(path: impl Into<PathBuf>) -> Result<Self, std::io::Error> {
+        let path = path.into();
+
+        // Create the target directory if it doesn't exist
+        fs_err::create_dir_all(&path).map_err(|e| {
+            std::io::Error::other(format!("failed to create target directory: {e}"))
```
wolfv, line 169: Do we even need to map the error? I think `fs_err` should already produce a good looking error.

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -357,6 +357,7 @@ impl PackageRecord {
         topological_sort::sort_topologically(records)
     }
 
+    #[allow(clippy::result_large_err)]
```
wolfv, line 360: I think you can Box the error type that is large to make this go away (cc @baszalmstra to hear his opinion).

wolfv:
Generally speaking looks really good.
We should definitely hear @baszalmstra input as well.

There are a few unrelated (clippy?) files that should go to a different PR :)

crates/rattler/src/install/installer/mod.rs
```diff
@@ -303,14 +305,17 @@ impl Installer {
             )
         });
 
+        let prefix =
+            Prefix::create(prefix.as_ref().to_path_buf()).expect("failed to create prefix");
```
baszalmstra, line 309: Can you turn this into a proper error? Never unwrap unless you are certain this can never fail.

crates/rattler/src/install/mod.rs
```diff
@@ -1233,7 +1224,7 @@ mod test {
         let target_dir = tempdir().unwrap();
         stream::iter(urls)
             .for_each_concurrent(Some(50), |package_url| {
-                let prefix_path = target_dir.path();
+                let prefix_path = Prefix::create(target_dir.path()).unwrap();
```
baszalmstra, line 1227: I guess you want to create the prefix outside of the iterator?

crates/rattler/src/install/unlink.rs
```diff
@@ -154,8 +154,8 @@ pub async fn empty_trash(target_prefix: &Path) -> Result<(), UnlinkError> {
     Ok(())
 }
 
-async fn move_to_trash(target_prefix: &Path, path: &Path) -> Result<(), UnlinkError> {
-    let mut trash_dest = target_prefix.join(".trash");
+async fn move_to_trash(target_prefix: &Prefix, path: &Path) -> Result<(), UnlinkError> {
+    let mut trash_dest = target_prefix.path().join(".trash");
```
baszalmstra, line 158: Maybe we can make a function on `Prefix` that returns the expected trash dir?

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -145,6 +145,76 @@ impl RecordFromPath for PrefixRecord {
     }
 }
 
+/// Represents a conda environment prefix with initialization handling
+#[derive(Debug, Clone)]
```
baszalmstra, line 149: This type deserves its own file.

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -145,6 +145,76 @@ impl RecordFromPath for PrefixRecord {
     }
 }
 
+/// Represents a conda environment prefix with initialization handling
```
baszalmstra, line 148: `initialization handling` is a bit ambiguous here, is it installing packages? Can you explain this more?

crates/rattler_conda_types/src/repo_data/mod.rs
baszalmstra: Yeah boxing is most often the right approach. I just dont understand how this changed since previous versions?

baszalmstra (commented):
Overall looks good! I left a few small comments.

I was also thinking, does it also make sense to handle the CACHEDIR.TAG? https://github.com/prefix-dev/pixi/pull/3353

PR #1160, opened 2025-03-13
feat: remove menu item and trackers in one function

ruben-arts:
### Description
The function is required to comfortably remove the menu items in one go from pixi.


crates/rattler_menuinst/src/lib.rs
```diff
@@ -173,6 +173,29 @@ pub fn install_menuitems(
     Ok(trackers)
 }
 
+/// Remove all menu items from a given prefix record.
+/// This function will remove the menu items from the system,
+/// and update the prefix record with removing the tracker entries
+pub fn remove_menuitems_for_record(
+    target_prefix: &Path,
+    prefix_record: &PrefixRecord,
```
wolfv, line 181: Maybe we should take it by value as it will be out of date when the function returns?

Or we could return the updated PrefixRecord.

crates/rattler_menuinst/src/lib.rs
```diff
@@ -173,6 +173,29 @@ pub fn install_menuitems(
...
+    let mut record = prefix_record.clone();
+    record.installed_system_menus = Vec::new();
+
+    // Save the updated prefix record
+    record
+        .write_to_path(
+            target_prefix.join("conda-meta").join(record.file_name()),
+            true,
+        )
+        .expect("Failed to write prefix record");
```
wolfv, line 195: `?` doesn't work here?

crates/rattler_menuinst/src/lib.rs
ruben-arts: Used by value

crates/rattler_menuinst/src/lib.rs
ruben-arts: It does 👍

wolfv:
JS test failure looks unrelated.

PR #1169, opened 2025-03-18
feat: allow to pass a semaphore for concurrency control

baszalmstra:
To control the maximum number of concurrent requests you can pass a number to `with_max_concurrent_requests`. This will limit the maximum number of requests that are executed concurrently. 

However, sometimes another part of the application might also be performing requests which could cause more requests to be made. This PR modifies the code to allow passing in a semaphore to control the concurrency which can be shared with other parts of the code.


nichmor (approved):
I couldn't find any other place where we need to respect `Semaphore`  apart from  `RemoteSubdirClient` and `RepodataQuery`. If that's ok we can merge it

PR #1175, opened 2025-03-19
fix: Don't clone package bytes in rattler-index

pavelzw:
### Description

cloning package bytes could get out of hand quickly if you have large packages


delsner (approved):
Nice, thanks!

wolfv (approved):

PR #1177, opened 2025-03-20
fix(JS): Remove non-needed top-level awaits

martinRenou:
### Description

Rattler was using a top-level await (actually not even needed as it was awaiting a synchronous function), which is not allowed in some contexts like service workers.

### DIFF
--- a/js-rattler/src/esm.ts
+++ b/js-rattler/src/esm.ts
@@ -6,4 +6,4 @@ import mod from "../pkg/js_rattler_bg.wasm";
 import { initSync } from "../pkg/js_rattler";
 
 //@ts-ignore
-await initSync({ module: await mod() });
+initSync({ module: mod() });


baszalmstra:
Would it be possible to lazily load the wasm only when its being used perhaps? I do not like to have an explicit initialization call. 

Lets just keep the name `initSync` instead of `initRattler`.

js-rattler/src/esm.ts
```diff
@@ -5,5 +5,6 @@ import mod from "../pkg/js_rattler_bg.wasm";
 //@ts-ignore
 import { initSync } from "../pkg/js_rattler";
 
-//@ts-ignore
-await initSync({ module: await mod() });
+export function initRattler() {
+    initSync({ module: mod() });
```
baszalmstra, line 9: I think the `mod` function here returns a Promise so it must be awaited.

trim21: I think it's synchronous after we rollup it.

martinRenou:
Thanks for commenting!

> initSync take about 10ms, is that to slow to you?

Oh right, I assumed that would be slow but haven't checked really 😄 We can live with this.

I will update the PR and just remove the two `await`s that's not needed!

Issue #1179, opened 2025-03-23
Allow empty "info" key for repodata.json


dahakawang:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

Hello, Rattler team! I've started to use `pixi` at work after trying it for some time at home, but within the context of our on-prem Artifactory, it always fails with the below error:

```
  × failed to parse repodata.json
  ╰─▶ missing field `subdir` at line 2 column 14
```

After digging into it for a bit I start to believe that this is because Artifactory emits a `repodata.json` with a slight schema difference that `rattler` is not comfortable with:

```json
{
  "info" : { },    <--- for whatever reason, this is empty but still presented
  "packages" : {
  }
}
```

while it looks like in `rattler`, if "info" is presented, it would require "subdir" to be also there: https://github.com/conda/rattler/blob/main/crates/rattler_conda_types/src/repo_data/mod.rs#L74

I'm not sure what the spec says about this (or if there's a spec?), but to me it's unlikely to change Artifactory's behavior, plus both conda/mamba work in the same context. I thus want to turn to you guys for some help, would it be possible to model this on the `rattler` side in a way that such `repodata.json` is allowed?

baszalmstra:
Sure, Id be happy to accept a PR!

PR #1181, opened 2025-03-24
fix: allow empty info key in repodata.json (#1179)

dahakawang:
### Description

A follow-up of this issue https://github.com/conda/rattler/issues/1179 to unblock myself from using `pixi` at work. I'm not really a Rust developer but it looks like this change is easy enough for me to handle, feedback is welcome.

Did some preliminary tests integrating with `pixi` seems to suggest this fixes my issue, no test was added because this is a data model change, and should be protected by any existing ones.


baszalmstra (approved):

baszalmstra:
Thanks! Much appreciated!

PR #1203, opened 2025-03-27
chore: fix js bindings

wolfv:
I think they broke because of updating `getrandom`. Let's see. cc @prsabahrami


prsabahrami:
@wolfv I don't think I have the permission to merge this. Perhaps @baszalmstra Can confirm this?

baszalmstra (approved):

PR #1206, opened 2025-03-28
feat: add support for md5 hash in direct url requests and add it to cache key

wolfv:
This would help a lot when installing explicit environments, because they usually come with `md5` hashes (from the old days...).

The environments look something like this:

```
# This file may be used to create an environment using:
# $ conda create --name <env> --file <this file>
# platform: osx-arm64
@EXPLICIT
https://conda.anaconda.org/conda-forge/osx-arm64/libcxx-20.1.1-ha82da77_0.conda#833c4899914bf96caf64b52ef415e319
https://conda.anaconda.org/conda-forge/osx-arm64/micromamba-1.5.12-0.tar.bz2#ef854aa57525cb8d9a87cc1f1d59bd92
```

From which we can construct matchspecs like:

`libcxx[url="https://conda.anaconda.org/conda-forge/osx-arm64/libcxx-20.1.1-ha82da77_0.conda", md5="833c4899914bf96caf64b52ef415e319"]`

Which rattler / py-rattler would then install perfectly fine.

cc @johanneskoester

### DIFF
--- a/crates/rattler_cache/src/package_cache/cache_key.rs
+++ b/crates/rattler_cache/src/package_cache/cache_key.rs
@@ -1,6 +1,6 @@
 use rattler_conda_types::package::ArchiveIdentifier;
 use rattler_conda_types::PackageRecord;
-use rattler_digest::Sha256Hash;
+use rattler_digest::{Md5Hash, Sha256Hash};
 use std::fmt::{Display, Formatter};
 
 /// Provides a unique identifier for packages in the cache.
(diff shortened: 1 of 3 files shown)


johanneskoester (approved):

baszalmstra:
Did you also change writing the md5 to the cache lock file?

wolfv:
That should happen since I add it to the `CacheKey` and we use the information from `RepodataRecord` if present (i think we fill it in if not present, when downloading).

wolfv:
superseded by #1293

PR #1215, opened 2025-04-01
Refactor RepoDataRecord::Channel from a simple string to `Channel`

### LINKED ISSUE #146: Channel in `RepoDataRecord` is a simple String
Should we either create a `Channel` type that we can use in `RepoDataRecord` and store it or a smart pointer to a shared / cached channel? Or should we type it as an URL?

It would be nice to be able to call a `.url()` and `.name()` method on the channel to get the short or expanded versions.

kelvinou01:
### Description

Resolves #146


baszalmstra:
Maybe it makes sense to use `ChannelUrl` instead of `Channel` since thats the more canonical form.

baszalmstra:
This PR is very much out of date. Do you want to bring it up to date again?

PR #1219, opened 2025-04-02
feat: py-rattler example tool

### LINKED ISSUE #355: Create a py-rattler example tool
The tool should take dependencies(matchspec) and create an environment from scratch.  

Why: 
- As a potential user of py-rattler, I would like to have an example usage of the library for reference.
- As a contributor of py-rattler, I would like to have an actual tool where I could test the API and immediatly see how it is integrating with the whole library in general.

kelvinou01:
### Description

Resolves #355

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -48,6 +48,10 @@ updates:
       - dependencies
     commit-message:
       prefix: chore(ci)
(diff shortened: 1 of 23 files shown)


wolfv:
Looks great!

PR #1246, opened 2025-04-09
add browser config

wolfv:
This adds a "iife" config so that we can load the JS file and access rattler through `window.Rattler ...`.

It works for my usecase in Retool, although I am not totally sure if there is another way with the modules we have already.


baszalmstra (commented):
Cant the normal esm build just be in that format? 

@trim21 Thoughts?

trim21:
> Cant the normal esm build just be in that format?
> 
> @trim21 Thoughts?

Although I would prefer downstream projects use a bundler to pack them, but it's not a bad idea to add this, I don't have strong opinion on this.

wolfv:
Any idea why the tests are failing here? I triggered them on main and that doesn't show the same failure (https://github.com/conda/rattler/pull/1249).

Is `jest` trying to use the browser files for some reason? Should we put them in a different folder? Like `dist-browser`? So that `jest` doesn't discover them?

trim21:
run build-cjs after run build-wasm should fix it

we should consider fix this in other pr

PR #1255, opened 2025-04-15
feat: add resolver options to lock-file

baszalmstra:
Adds resolver options to the lock file. These options strongly modify the outcome of the solver. They are stored along the locked package so programs can choose to invalidate the content if these options changed.

The options stored are:

* channel-priority: disabled or strict.
* solve strategy: whether the solver selected the highest versions or the lowest.
* exclude-newer: whether packages after a certain date were excluded.

These correspond to options used by `rattler_solve`.

These changes should both be backward- and forward-compatible. E.g. older clients will ignore the new fields, newer clients with older lock-files will use sane defaults.

### DIFF
--- a/crates/rattler_lock/src/options.rs
+++ b/crates/rattler_lock/src/options.rs
@@ -1,0 +1,20 @@
+use rattler_solve::{ChannelPriority, SolveStrategy};
+
+/// Options that were used during the resolution of the packages stored in the
+/// lock-file. These options strongly influence the outcome of the solve and are
+/// therefore stored along with the locked packages.
+#[derive(Debug, Clone, PartialEq, Eq, Default, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct ResolverOptions {
+    /// The strategy that was used to solve the dependencies.
+    #[serde(default, skip_serializing_if = "crate::utils::serde::is_default")]
+    pub strategy: SolveStrategy,
+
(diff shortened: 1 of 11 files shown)


crates/rattler_lock/src/options.rs
```diff
@@ -0,0 +1,20 @@
...
+#[derive(Debug, Clone, PartialEq, Eq, Default, serde::Serialize, serde::Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct ResolverOptions {
+    /// The strategy that was used to solve the dependencies.
+    #[serde(default, skip_serializing_if = "crate::utils::serde::is_default")]
+    pub strategy: SolveStrategy,
+
+    /// The channel priority that was used to solve the dependencies.
+    #[serde(default, skip_serializing_if = "crate::utils::serde::is_default")]
+    pub channel_priority: ChannelPriority,
```
Hofer-Julian, line 15: What is the advantage of this approach over wrapping them into an `Option` like with `exclude_newer`?

Hofer-Julian (approved):
Looks good in general, just added one question

crates/rattler_lock/src/options.rs
baszalmstra: I think we should have some default values here and not have the user have to figure out what those are.

Exclude-newer is wrapped in an option because otherwise we cannot represent "not" exclude-newer.

Hofer-Julian: Okay!

PR #1262, opened 2025-04-17
fix: rename struct and reexport

baszalmstra:
Renames the `options` struct to be more in line with our other crates and reexports some fields so that an import of `rattler_solve` is not perse needed in downstream crates.


Hofer-Julian (approved):
Looks good!

PR #1264, opened 2025-04-18
feat(js): bind PackageMetadata

prathmesh703:
### Description
WASM binding for `PackageMetadata` file in `crates/rattler_conda_types/src/package/package_metadat.rs`


baszalmstra:
Thanks for contributing! We actually already have bindings in the js-rattler folder. This also include a build system to create a npm package. We use this additional crate to not litter the main crates with wasm bindgen types.

prathmesh703:
@baszalmstra  My mistake thanks for pointing me to js-rattler. I will update the PR shortly
.

js-rattler/crate/package_metadata.rs
```diff
@@ -0,0 +1,44 @@
...
+        self.inner.conda_pkg_format_version as f64
+    }
+
+    #[wasm_bindgen(setter)]
+    pub fn set_conda_pkg_format_version(&mut self, version: f64) {
+        self.inner.conda_pkg_format_version = version as u64;
+    }
+
+    #[wasm_bindgen(js_name = toJSON)]
+    pub fn to_json(&self) -> Result<String, JsValue> {
```
baszalmstra, line 34: Javascript already contains a lot of functionality to parse to and from json. The only thing is that our objects might not be fully json compatible. Instead of going directly from string and back I instead opted to be able to convert our bound objects to a plain javascript object which van immediately be converted to json. See PackageRecord for an example.

prathmesh703: @baszalmstra Thanks for the clarification , I will look into this after a week my university exams are currently going on. I hope this won't be a problem.

baszalmstra:
@prathmesh703 Do you think you'll still continue with this?

prathmesh703:
Sorry i wont be able to continue

PR #1265, opened 2025-04-18
feat: Add MemoryStorage as authentication backend

pavelzw:


wolfv:
It would be nice if we had a way to run this backend as a daemon :)

baszalmstra (approved):

PR #1273, opened 2025-04-23
fix: menuinst windows shorcut path

Glatzel:
In recent pixi 0.46.0, windows desktop, start menu, quick launch shortcut are all installed successfully. But start menu and quick launch shortcut ink file do not point to the right command, while desktop shortcut do the right thing.

tasks:
- [x] The command path will set to `command` in creating start menu and quick launch shortcut, just like what it does in creating desktop shortcut. 
- [x] Start menu shortcut will be installed into the `menu_name` subdir. This is what conda menuinst does.
- [x] When removing a shortcut, the `menu_name` subdir in start menu will be removed if it exists and is empty.

## test with pixi in my pc
- [x] Install shortcuts and all types of shortcuts point to the right command/path.
- [x] Install shortcuts and remove env from `pixi-global.toml`, then run `pixi global  update`. The shortcuts ink files should be removed, and the `menu_name` subdir in start menu will also be removed. Also test removing env with `pixi global uninstall`.
- [x] Install shortcuts and remove env from `pixi-global.toml`, **add arbitrary file to `menu_name` subdir in start menu**, then run `pixi update`. The shortcuts ink files should be removed, and the `menu_name` subdir in start menu will still **exist**. Also test removing env with `pixi global uninstall`.

### DIFF
--- a/crates/rattler_menuinst/src/windows.rs
+++ b/crates/rattler_menuinst/src/windows.rs
@@ -340,6 +340,7 @@ impl WindowsMenu {
         let start_menu_subdir_path = self.directories.start_menu.join(&self.menu_name);
         if !start_menu_subdir_path.exists() {
             std::fs::create_dir(&start_menu_subdir_path)?;
+            tracker.start_menu_subdir_path = Some(start_menu_subdir_path.clone());
         }
         let start_menu_link_path = start_menu_subdir_path.join(&link_name);
         let shortcut = Shortcut {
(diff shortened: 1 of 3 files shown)


crates/rattler_menuinst/src/windows.rs
```diff
@@ -371,6 +376,7 @@ impl WindowsMenu {
         if let Some(quick_launch_dir) = self.directories.quick_launch.as_ref() {
             if self.item.quicklaunch.unwrap_or(false) {
                 let quicklaunch_link_path = quick_launch_dir.join(&self.name).join(link_name);
+                std::fs::create_dir(quicklaunch_link_path.parent().unwrap())?;
```
wolfv: Can you use `expect` here? Or perhaps even better, don't use `parent()`, and instead create the `quick_launch_dir_name = quick_launch_dir.join(&self.name)` and then `quick_launch_link_path = quick_launch_dir_name.join(link_name)` so we don't need expect / unwrap.

Glatzel: Thank for your advice. I have changed the subdir creation code.

PS: I should mention that the subdir only be created in the start menu not the quick launch. I do not really know the reason, just follow what conda menuinst does.

baszalmstra:
@wolfv Can you rereview this PR?

wolfv:
I made it a little more backward compatible (by adding `serde::default` to the new Option field).

With my changes, we also only track the directory if we actually created it. 

Looks good to me now, thanks @Glatzel!

wolfv (approved):

PR #1274, opened 2025-04-23
chore: move to edition 2024

Hofer-Julian:
There are a few things left to discuss:

- A couple of tests set env vars which is unsafe in a multi-threaded context. On Pixi, I just enforced current_thread with `tokio`, In rattler, it's also "normal" tests, so I am unsure what's the best way forward here.
- A few lifetime problems that are non-obvious to me
- Clippy dislikes unnecessary `continue`. I think I'd rather disable the warning. What do you say?


Hofer-Julian:
This is too old now to be useful, closing

PR #1279, opened 2025-04-24
refactor: lock workspace member dependencies

baszalmstra:
Pin all the workspace dependencies. 

release-plz automatically bumps any workspace member whose dependencies have changed. This means that crates are always bumped in tandem. This is useful because it means we also always test the crates together that will be used together.

So given that crates are always updated in tandem we choose to pin the member crates together as well.


Hofer-Julian (approved):
@baszalmstra and I discussed this in detail and I agree that this will improve the current situation

PR #1298, opened 2025-05-10
refactor(py): improve `Gateway` typing

trim21:
### Description

if we use `List[T | K]` as argument type then it doesn't accept `list[T]` or `list[K]`, which is actually allowed.


baszalmstra (approved):

baszalmstra:
Thank you!

PR #1303, opened 2025-05-15
feat: Add purls to IndexJson

pavelzw:

### DIFF
--- a/crates/rattler_conda_types/src/package/index.rs
+++ b/crates/rattler_conda_types/src/package/index.rs
@@ -1,11 +1,11 @@
-use std::path::Path;
+use std::{collections::BTreeSet, path::Path};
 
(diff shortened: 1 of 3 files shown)


wolfv (commented):
Looks good!

pavelzw:
like this @baszalmstra: [`be00977` (#1303)](https://github.com/conda/rattler/pull/1303/commits/be00977580397eb77980371d4186f46f85ae2ed2)

xhochy:
Shouldn't the CEP be first accepted before we implement this?

wolfv:
The way it is implemented now it will not change any existing metadata (since it will not render into the JSON when the Option is None). We'll be careful to only allow this in `rattler-build` with tge `experimental` flag enabled.

PR #1305, opened 2025-05-15
fix: clean up URL before serializing or printing

wolfv:
This fixes the root cause of the issue in rattler-build where URLs with secrets leaked into the package: https://github.com/prefix-dev/rattler-build/issues/1636

I think this should be "OK" to do like this, although now users would have a hard time serializing this value _with_ credentials. But I think that should only happen if someone "really" wants it.


baszalmstra (changes requested):
Id prefer to add another serde_with type to serialize without credentials. We use this in pixi build protocol where we do want to keep credentials.

wolfv:
Decided to do the cleaning in rattler-build instead.

PR #1312, opened 2025-05-15
feat: Sharded repodata, zst, add purls and run_exports support

### LINKED ISSUE #1096: convenience functions for writing `RepoData`

pavelzw:
### Description

closes #1096


crates/rattler_conda_types/src/repo_data/sharded.rs
```diff
@@ -19,7 +19,7 @@ pub struct ShardedRepodata {
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct ShardedSubdirInfo {
     /// The name of the subdirectory
-    pub subdir: String,
+    pub subdir: Platform,
```
baszalmstra, line 22: Is this important? The reason we use a string instead of a platform is that by using a string we can also support records and shards for platforms that we dont support directly.

crates/rattler_index/src/lib.rs
```diff
@@ -379,22 +406,141 @@ async fn index_subdir(
...
-            .await?;
+        patched_repodata
+    } else {
+        repodata
+    };
+
+    let repodata_bytes = serde_json::to_vec(&repodata)?;
+    if write_zst {
+        tracing::info!("Compressing repodata bytes");
+        let repodata_zst_bytes = zstd::stream::encode_all(&repodata_bytes[..], 19)?;
```
baszalmstra, line 458: Should we add a const for 19?

crates/rattler_index/src/lib.rs
```diff
@@ -433,6 +583,8 @@ pub async fn index_s3(
     session_token: Option<String>,
     target_platform: Option<Platform>,
     repodata_patch: Option<String>,
+    write_zst: bool,
+    write_shards: bool,
     force: bool,
```
baszalmstra, line 588: I kinda agree with the clippy error here. This function is growing out of proportions. Perhaps we can introduce a struct here?

crates/rattler_index/tests/test_index.rs
```diff
@@ -63,6 +63,8 @@ async fn test_index() {
         Some(Platform::Win64),
         None,
         true,
+        true,
+        true,
```
baszalmstra, line 67: Referring to the clippy error, this is indeed very unclear.

crates/rattler_conda_types/src/repo_data/sharded.rs
pavelzw: not really, will revert

crates/rattler_index/src/lib.rs
```diff
@@ -544,18 +547,43 @@ pub async fn write_repodata(
     Ok(())
 }
 
+/// Configuration for `index_fs`
+pub struct IndexFsConfig<T>
+{
+    /// The channel to index.
+    pub channel: T,
```
baszalmstra, line 554: I would just use a `PathBuf` here. We can get rid of the generic argument then.

baszalmstra (approved):

PR #1315, opened 2025-05-16
chore: bump aws dependencies

wolfv:


wolfv:
Thanks for pushing on this. Looks good to me! I think my main concern is that we might (in the future) copy over more of the rattler network configuration to configure the S3 HTTP client the same way (using the same `proxy` or `ssl` certificate configuration as our main client). However, I expect that this was already different and the AWS SDK just decided to make the HTTP client construction explicit so I guess this is fine to merge until someone complains! :)

baszalmstra:
The construction of the client is not perse explicit. You can use the "default client" but that will always construct a client that is configured to use rustls + aws-lc-rs instead of rust + ring, which we use in all other places. The changes to the client were only needed to make sure that we dont have to pull in the aws-ls-rs dependency, which is painful to compile on more exotic linux architectures. 

aws-ls-rc does seem to be more secure than ring, so I can see a future where we might switch to that.

wolfv:
understood!

PR #1324, opened 2025-05-16
chore: add rustls-tls / nativetls feature to rattler_cache

wolfv:
I believe this was missing


baszalmstra (changes requested):
You should also activate this feature in the crates that depend on this. 

But I wonder if this is really needed? Cant we just use default-features false?

baszalmstra (approved):

PR #1325, opened 2025-05-16
feat: Adding locked packages to the js-rattler solver

AnastasiaSliusar:
### Description

Now, the js-rattler solver does not accept previously installed packages but it should take care about it. 

This PR has next changes:
- adding locked_packages to the solver
- if a locked package does not have depends then this data will be taken from repodata

### The result
For example, using this code
```
const test = await simpleSolve(
  ["xeus-python ", "numpy"],
  ["https://repo.prefix.dev/emscripten-forge-dev","conda-forge"
  ],
  ["emscripten-wasm32", "noarch"],
  [{
      build: "py313h027658c_2",
      buildNumber,
      filename: "xeus-python-0.17.4-py313h027658c_2.tar.bz2",
      packageName: "xeus-python",
      repoName: "https://repo.prefix.dev/emscripten-forge-dev/",
      url: "https://repo.prefix.dev/emscripten-forge-dev/emscripten-wasm32/xeus-python-0.17.4-py313h027658c_2.tar.bz2",
      version: "0.17.4",
  }]
);
console.log('test',test);
```
we will have this result
image

### DIFF
--- a/js-rattler/crate/solve.rs
+++ b/js-rattler/crate/solve.rs
@@ -60,56 +60,54 @@ pub async fn simple_solve(
         .into_iter()
         .map(|p| serde_wasm_bindgen::from_value(p.into()))
         .collect::<Result<Vec<_>, _>>()?;
-
-    let mut installed_packages: Vec<RepoDataRecord> =
+    let js_locked_packages: Vec<SolvedPackage> =
         if locked_packages.is_null() || locked_packages.is_undefined() {
             vec![]
         } else {
-            let js_locked_packages: Vec<SolvedPackage> =
-                serde_wasm_bindgen::from_value(locked_packages).map_err(JsError::from)?;
-
-            js_locked_packages
-                .into_iter()
-                .map(|pkg| {
-                    let url = Url::parse(&pkg.url)
-                        .map_err(|e| JsError::from(ParseChannelError::from(e)))?;
-
-                    let rec = PackageRecord {
-                        name: PackageName::try_from(pkg.package_name.clone())?,
-                        version: Version::from_str(&pkg.version)?.into(),
-                        build: pkg.build.clone(),
-                        build_number: pkg.build_number.unwrap_or_default(),
-                        md5: None,
-                        sha256: None,
-                        size: None,
-                        arch: None,
-                        platform: None,
-                        depends: pkg.depends.unwrap_or_default(),
-                        subdir: pkg.subdir.unwrap_or_else(|| "unknown".to_string()),
-                        extra_depends: BTreeMap::new(),
-                        constrains: vec![],
-                        track_features: vec![],
-                        features: None,
-                        noarch: NoArchType::none(),
-                        license: None,
-                        license_family: None,
-                        timestamp: None,
-                        python_site_packages_path: None,
-                        legacy_bz2_md5: None,
-                        legacy_bz2_size: None,
-                        purls: None,
-                        run_exports: None,
-                    };
-
-                    Ok(RepoDataRecord {
-                        url,
-                        file_name: pkg.filename,
-                        channel: pkg.repo_name,
-                        package_record: rec.clone(),
-                    })
-                })
-                .collect::<Result<Vec<_>, JsError>>()?
+            serde_wasm_bindgen::from_value(locked_packages).map_err(JsError::from)?
         };
+    let mut installed_packages: Vec<RepoDataRecord> = js_locked_packages
+        .clone()
+        .into_iter()
+        .map(|pkg| {
+            let url =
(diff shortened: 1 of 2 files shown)


js-rattler/crate/solve.rs
```diff
@@ -61,20 +81,107 @@ pub async fn simple_solve(
...
+            }
+
+            rec
+        })
+        .collect();
+
     let task = SolverTask {
         specs,
+        locked_packages: enriched_installed_packages,
+        pinned_packages: vec![],
```
baszalmstra, line 179:
```suggestion
```

js-rattler/crate/solve.rs
```diff
@@ -61,20 +81,107 @@ pub async fn simple_solve(
...
+            let name = rec.package_record.name.as_normalized().to_string();
+            let version = rec.package_record.version.to_string();
+            let build = rec.package_record.build.clone();
+            ((name, version, build), rec.package_record.depends.clone())
+        })
+        .collect();
+
+    let installed_packages: Option<Vec<RepoDataRecord>> =
+        if locked_packages.is_null() || locked_packages.is_undefined() {
+            None
```
baszalmstra, line 105: I think you can just use `vec[]!` here.

js-rattler/crate/solve.rs
```diff
@@ -61,20 +81,107 @@ pub async fn simple_solve(
...
+
+            if rec.package_record.depends.is_empty() {
+                if let Some(deps) = repodata_keys.get(&key) {
+                    rec.package_record.depends = deps.clone();
+                }
+            }
+
+            rec
+        })
+        .collect();
```
baszalmstra, line 174: You are recollecting the data instead of just modifying the existing record. You could instead use `iter_mut`

js-rattler/crate/solve.rs
```diff
@@ -61,20 +81,107 @@ pub async fn simple_solve(
...
+                .map(|pkg| {
+                    let url = Url::parse(&pkg.url)
+                        .map_err(|e| JsError::from(ParseChannelError::from(e)))?;
+
+                    let rec = PackageRecord {
+                        name: PackageName::try_from(pkg.package_name.clone())?,
+                        version: Version::from_str(&pkg.version)?.into(),
+                        build: pkg.build.clone(),
+                        build_number: pkg.build_number,
+                        subdir: "unknown".to_string(),
```
baszalmstra, line 121: Packages can also come from different subdirectories (e.g. `noarch` vs `emscripten-wasm`). The unique key of a package is `(name, version, build, subdir)`. You also do not have the `subdir` available to you from the input. This might cause problems down the road but Id be fine merging as is.

baszalmstra (commented):
Can you also add a test that verifies the behavior that you expect? So we dont break it in the future?

js-rattler/crate/solve.rs
```diff
@@ -51,41 +74,138 @@ pub async fn simple_solve(
...
-            version: r.package_record.version.to_string(),
+        .filter(|spec| {
+            let matched = installed_packages
+                .iter()
+                .any(|rec| spec.matches(&rec.package_record));
+
+            !matched
         })
-        .collect())
+        .collect();
```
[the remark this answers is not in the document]
AnastasiaSliusar: > Use https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.any to check if there is `any` spec that isnt matched.

@baszalmstra I pass filtered_spec to the solver after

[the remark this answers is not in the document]
baszalmstra: But then those matchspecs will not be part of the solution? Installed packages might actually be removed from the solution then. I think you should pass `specs`.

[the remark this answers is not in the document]
baszalmstra: The following could happen. You initially have the following spec `numpy <2`. You get `numpy 1.26`.  Then you add another dependency which has a dependency on `numpy >1`. 

Since `numpy 1.26` matches our explicit spec `numpy <2`. The spec is removed from the solve. `numpy` is now unconstrained so the solution might contain a numpy that doesnt respect `numpy <2`. 

You always need to give the solver the complete problem.

js-rattler/crate/solve.rs
```diff
@@ -51,6 +61,66 @@ pub async fn simple_solve(
...
+
+    //if we do not need to solve the same packages, then filter them
+    let all_matched = specs.clone().iter().all(|spec| {
+        installed_packages
+            .iter()
+            .any(|rec| spec.matches(&rec.package_record))
+    });
+
+    if all_matched {
+        return Ok(vec![]);
```
baszalmstra, line 122: This is inconcistent with the result when a resolve is performed which might be very confusing. I think you should just return the installed packages again.

PR #1342, opened 2025-05-20
fix: properly dedup package names

### LINKED ISSUE #1330: SparseRepoData: duplicate package names
### Checklist

### What happened?

Looks like there's no deduplication in `rattler.SparseRepoData.package_names()`

```python
>>> repodata = rattler.SparseRepoData(
  rattler.Channel("https://repo.anaconda.com/pkgs/main"),
  "noarch",
  ".../pkgs/cache/3e39a7aa1.json",
)
>>> names = repodata.package_names()
>>> len(names)
2829
>>> len(set(names))
1642
```

### Additional Context

_No response_

baszalmstra:
Fixes #1330 

Packages names werent properly deduplicated in the sparse repodata when both .conda and .tar.bz2 packages were present. This PR fixes that. Also added a test to verify this behavior.

### DIFF
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -12,7 +12,6 @@ use std::{
 
 use bytes::Bytes;
 use fs_err as fs;
-use itertools::Itertools;
 use rattler_conda_types::{
     compute_package_url, Channel, ChannelInfo, PackageName, PackageRecord, RepoDataRecord,
 };
@@ -165,14 +164,15 @@ impl SparseRepoData {
     /// This works by iterating over all elements in the `packages` and
     /// `conda_packages` fields of the repodata and returning the unique
     /// package names.
-    pub fn package_names(&self) -> impl Iterator<Item = &'_ str> + '_ {
+    pub fn package_names(&self) -> impl Iterator<Item = &'_ str> {
         let repo_data = self.inner.borrow_repo_data();
         repo_data
             .packages
             .iter()
             .chain(repo_data.conda_packages.iter())
             .map(|(name, _)| name.package)
-            .dedup()
+            .collect::<HashSet<_>>()
+            .into_iter()
     }
 
     /// Returns all the records for the specified package name.
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -12,6 +12,7 @@ use std::{
 
 use bytes::Bytes;
 use fs_err as fs;
+use itertools::Itertools;
 use rattler_conda_types::{
     compute_package_url, Channel, ChannelInfo, PackageName, PackageRecord, RepoDataRecord,
 };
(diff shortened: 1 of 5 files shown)


test-data/channels/dummy/linux-64/repodata.json
```diff
@@ -368,6 +368,55 @@
...
+      "subdir": "linux-64",
+      "timestamp": 1715610974000,
+      "version": "2.1"
+    },
+    "issue_717-2.1-bla_1.conda": {
+      "build": "issue_717",
+      "build_number": 0,
+      "depends": [],
+      "constrains": [
+        "ray[default,data] >=2.9.0,<3.0.0"
```
wolfv, line 408: uff

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -165,14 +165,14 @@ impl SparseRepoData {
...
-            .packages
+        let tar_baz2_packages = repo_data.packages.iter().map(|(name, _)| name.package);
+        let conda_packages = repo_data
+            .conda_packages
             .iter()
-            .chain(repo_data.conda_packages.iter())
-            .map(|(name, _)| name.package)
-            .dedup()
+            .map(|(name, _)| name.package);
+        tar_baz2_packages.merge(conda_packages).dedup()
```
wolfv, line 175: Do you know if this prefers the `conda` package or the `tar.bz2`?

baszalmstra: So how it works is that we have two lists, one for tar.bz2 and one for .conda. Both lists are already sorted by package name. `merge` creates a new iterator that is also already sorted. Then `dedup` removes duplicate _repeated_ items.

baszalmstra: Oh and the names are just `foobar` in `foobar-0.1.0-blabla.conda`. So it doesnt matter which we take first.

test-data/channels/dummy/linux-64/repodata.json
baszalmstra: Yeah its just a copy of another record but with another extension.

crates/rattler_repodata_gateway/src/sparse/mod.rs
baszalmstra: Previously the two sorted lists would be concattenated (or chained) and then deduped. So, you could have `[a,b,c,a,b]`, the result would not be a sorted list. We then called `dedup` on that list which for this example would not do anything because there are no _repeated_ duplicates. Now, instead of `chain` we use `merge` and you get this list `[a,a,b,b,c]`.

wolfv: I think ideally we'd have a preference for `.conda` packages and make this deterministic if possible?

wolfv: Or am I missing something and it doesn't matter for this particular function?

wolfv: Ok, looks like this function is only concerned with `package_names` so it doesn't matter where it comes from. Sorry for the noise.

PR #1347, opened 2025-05-22
feat(py): query based on matchspec and package count

baszalmstra:
Adds the ability to load all repodata records from a SparseRepoData, query packages based on matchspec, and return the total number of packages.

### DIFF
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -243,7 +243,7 @@ impl SparseRepoData {
     }
 
     /// Returns all the records that matches any of the specified match spec.
-    pub fn load_matching_records<'m>(
+    pub fn load_matching_records(
         &self,
         spec: impl IntoIterator<Item = impl Borrow<MatchSpec>>,
         variant_consolidation: PackageFormatSelection,
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -221,8 +221,8 @@ impl SparseRepoData {
         }
     }
 
-    /// Returns the number of packages in this instance.
-    pub fn package_count(&self, package_format_selection: PackageFormatSelection) -> usize {
+    /// Returns the number of records in this instance.
+    pub fn record_count(&self, package_format_selection: PackageFormatSelection) -> usize {
         match package_format_selection {
             PackageFormatSelection::PreferConda => {
                 let repo_data = self.inner.borrow_repo_data();
--- a/crates/rattler_repodata_gateway/src/sparse/mod.rs
+++ b/crates/rattler_repodata_gateway/src/sparse/mod.rs
@@ -295,7 +295,7 @@ impl SparseRepoData {
         )
     }
 
-    /// Returns all the records for the specified package name.
+    /// Returns all the records for the specified package format(s).
     pub fn load_all_records(
         &self,
         variant_consolidation: PackageFormatSelection,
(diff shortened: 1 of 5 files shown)


crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -197,14 +197,81 @@ impl SparseRepoData {
     /// This works by iterating over all elements in the `packages` and
     /// `conda_packages` fields of the repodata and returning the unique
     /// package names.
-    pub fn package_names(&self) -> impl Iterator<Item = &'_ str> {
+    pub fn package_names(
+        &self,
+        package_format_selection: PackageFormatSelection,
```
baszalmstra, line 202: This is new! This functionality was missing in previous implementations.

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -216,14 +283,35 @@ impl SparseRepoData {
...
+            variant_consolidation,
+            base_url,
+            &self.channel,
+            &self.subdir,
+            self.patch_record_fn,
+            |_| true, // Dont filter anything out
+        )
+    }
+
+    /// Returns all the records for the specified package name.
```
jaimergp, line 298:
```suggestion
    /// Returns all the records for the specified package format(s).
```

I think?

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -197,14 +197,81 @@ impl SparseRepoData {
...
+            }
+            PackageFormatSelection::OnlyTarBz2 => {
+                itertools::Either::Right(tar_baz2_packages.dedup())
+            }
+            PackageFormatSelection::OnlyConda => itertools::Either::Right(conda_packages.dedup()),
+        }
+    }
+
+    /// Returns the number of packages in this instance.
+    pub fn package_count(&self, package_format_selection: PackageFormatSelection) -> usize {
```
wolfv, line 225: I wonder if this is confusing. I think I initially thought it would be a count of package names.

Maybe package_record_count would be more explicit?

baszalmstra: Or maybe just `record_count`?

wolfv (approved):

jaimergp:
Looks like `.record_count()` is buggy in some cases. See this reproducer:

```python
from tempfile import NamedTemporaryFile
from urllib.request import urlretrieve

from rattler import Channel, SparseRepoData, PackageFormatSelection

channel_url = "https://conda.anaconda.org/conda-test"
subdir = "noarch"
repodata_url = "https://conda.anaconda.org/conda-test/noarch/repodata.json"
# This channel only contains .tar.bz2 files, a total of 12

with NamedTemporaryFile(suffix=".json", delete=False) as tmp:
    urlretrieve(repodata_url, tmp.name)

repo = SparseRepoData(Channel(channel_url), subdir, tmp.name)
for pkg_format in PackageFormatSelection:
    print(pkg_format)
    print(repo.record_count(pkg_format))
    print(len(repo.load_all_records(pkg_format)))
    print("---")
```

It should always report 12 (this channel only lists .tar.bz2 artifacts), but for some reason `.record_count()` reports 10 with PackageFormatSelection.PREFER_CONDA. `.load_all_records()` does report the correct thing all the time.

cc @baszalmstra

baszalmstra:
Let me fix that!

baszalmstra:
Should be fixed in: https://github.com/conda/rattler/pull/1350

Ill do a patch release when its merged!

PR #1380, opened 2025-06-03
fix: add rust-src to py-rattler

benmoss:
### Description
Trying to use rust-analyzer with py-rattler and ran into the error 
> can't load standard library, try installing `rust-src` sysroot_path=/Volumes/workspace/rattler/py-rattler/.pixi/envs/test

I found [this thread](https://discord.com/channels/1082332781146800168/1331668634707431506/1331668875820929124) from @lucascolley where just adding rust-src to the environment was proposed, so here I'm actually gonna do it 😄

### DIFF
--- a/py-rattler/pixi.toml
+++ b/py-rattler/pixi.toml
@@ -36,6 +36,8 @@ pytest-xprocess = ">=0.23.0,<0.24"
 
 # used in examples
 typer = "*"
+# for rust-analyzer
+rust-src = ">=1.85.0,<2"
 
 [feature.test.pypi-dependencies]
 types-networkx = "*"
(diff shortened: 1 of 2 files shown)


py-rattler/pixi.toml
```diff
@@ -36,6 +36,8 @@ pytest-xprocess = ">=0.23.0,<0.24"
 
 # used in examples
 typer = "*"
+# for rust-analyzer
+rust-src = ">=1.85.0,<2"
```
baszalmstra:
```suggestion
rust-src = "~=1.85.0"
```

Lets use the same notion as used for rust itself

PR #1381, opened 2025-06-04
feat: make rattler_networking system integration optional

wolfv:
I was looking into the failing build in #1378 for Javascript and noticed that we might be able to slim down the dependency tree for the `js` build a bit. 

Most of the system integration does not make sense in WASM, and we can make it optional.

This PR makes `dirs`, `keyring`, and `netrc-rs` crates optional.

### DIFF
--- a/crates/rattler_networking/Cargo.toml
+++ b/crates/rattler_networking/Cargo.toml
@@ -16,9 +16,7 @@ native-tls = ["reqwest/native-tls"]
 rustls-tls = ["reqwest/rustls-tls"]
 gcs = ["google-cloud-auth"]
 s3 = ["aws-config", "aws-sdk-s3"]
-netrc = ["netrc-rs"]
-keyring = ["dep:keyring"]
-system-integration = ["keyring", "netrc", "dirs"]
+system-integration = ["keyring", "netrc-rs", "dirs"]
 
 [dependencies]
 anyhow = { workspace = true }
--- a/crates/rattler_networking/Cargo.toml
+++ b/crates/rattler_networking/Cargo.toml
@@ -11,31 +11,34 @@ license.workspace = true
 readme.workspace = true
 
 [features]
-default = ["rustls-tls"]
+default = ["rustls-tls", "system-integration"]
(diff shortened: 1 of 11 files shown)


crates/rattler_networking/Cargo.toml
```diff
@@ -11,31 +11,34 @@ license.workspace = true
...
 
 [features]
-default = ["rustls-tls"]
+default = ["rustls-tls", "system-integration"]
 native-tls = ["reqwest/native-tls"]
 rustls-tls = ["reqwest/rustls-tls"]
 gcs = ["google-cloud-auth"]
 s3 = ["aws-config", "aws-sdk-s3"]
+netrc = ["netrc-rs"]
+keyring = ["dep:keyring"]
```
baszalmstra: This is not needed. Adding an optional dependency also adds a feature by the same name.

crates/rattler_networking/Cargo.toml
```diff
@@ -11,31 +11,34 @@ license.workspace = true
 readme.workspace = true
 
 [features]
-default = ["rustls-tls"]
+default = ["rustls-tls", "system-integration"]
 native-tls = ["reqwest/native-tls"]
 rustls-tls = ["reqwest/rustls-tls"]
 gcs = ["google-cloud-auth"]
 s3 = ["aws-config", "aws-sdk-s3"]
+netrc = ["netrc-rs"]
```
baszalmstra: Same for this I guess?

baszalmstra (commented):
Nice! Do you know how many dependencies this safes?

wolfv:
I am not sure in terms of "actual numbers" because I suppose that the not all dependencies of the `Cargo.lock` file are actually going to be compiled into the WASM binary (e.g. the Windows crates are probably not going to be part of the final binary).

My main motivation is to get rid of the duplicate `getrandom` crate versions but unfortunately that will be difficult because I think `ring` and also `reqwest-retry` depend on the older version...

baszalmstra:
I applaud that effort! Thank you!

PR #1386, opened 2025-06-10
refactor: move upload from `rattler-build` to `rattler`

magentaqin:
### Description
Create new crate `rattler_upload` to solve issue [Align pixi upload with rattler-build upload](https://github.com/prefix-dev/pixi/issues/2302)

### DIFF
--- a/crates/rattler_upload/src/lib.rs
+++ b/crates/rattler_upload/src/lib.rs
@@ -8,12 +8,14 @@ use upload::opt::{
     AnacondaData, ArtifactoryData, CondaForgeData, PrefixData, QuetzData, ServerType, UploadOpts,
 };
 
-/// Upload.
+/// Upload package to different channels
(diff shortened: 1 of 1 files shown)


crates/rattler_upload/src/lib.rs
```diff
@@ -0,0 +1,70 @@
...
+pub(crate) mod utils;
+
+use crate::utils::tool_configuration;
+use miette::IntoDiagnostic;
+use rattler_conda_types::package::ArchiveType;
+use upload::opt::{
+    AnacondaData, ArtifactoryData, CondaForgeData, PrefixData, QuetzData, ServerType, UploadOpts,
+};
+
+/// Upload.
```
Hofer-Julian: Can you please add a more descriptive docstring here?

Doesn't have to be long, but one word is not enough :)

magentaqin: Got it!

PR #1388, opened 2025-06-11
fix: Removing skipping logic

AnastasiaSliusar:
### Description
This PR removes skipping resolving logic because when we need to resolve for the case when we need to remove packages, the solver does not run


wolfv:
@AnastasiaSliusar looks like some tests are not passing anymore.

baszalmstra:
Published as https://www.npmjs.com/package/@conda-org/rattler/v/0.3.1

PR #1389, opened 2025-06-12
feat: add `rattler_config` crate (derived from `pixi_config`)

wolfv:

### DIFF
--- a/crates/rattler_package_streaming/src/write.rs
+++ b/crates/rattler_package_streaming/src/write.rs
@@ -110,7 +110,8 @@ fn total_size(base_path: &Path, paths: &[PathBuf]) -> u64 {
 /// ```no_run
 /// use std::path::PathBuf;
 /// use std::fs::File;
-/// use rattler_package_streaming::write::{write_tar_bz2_package, CompressionLevel};
+/// use rattler_package_streaming::write::{write_tar_bz2_package};
+/// use rattler_conda_types::compression_level::CompressionLevel;
 ///
 /// let paths = vec![PathBuf::from("info/recipe/meta.yaml"), PathBuf::from("info/recipe/conda_build_config.yaml")];
 /// let mut file = File::create("test.tar.bz2").unwrap();


crates/rattler_config/src/config.rs
```diff
@@ -0,0 +1,297 @@
...
+        for path in paths {
+            let content = std::fs::read_to_string(path.as_ref())?;
+            let other: ConfigBase<T> = toml::from_str(&content)?;
+            config = config
+                .merge_config(&other)
+                .map_err(|e| LoadError::MergeError(e, path.as_ref().to_path_buf()))?;
+        }
+
+        // config.validate()?;
+        Ok(config)
```
pavelzw:
```suggestion
        config.validate()
```

crates/rattler_config/src/config.rs
```diff
@@ -0,0 +1,297 @@
...
+    #[error("Error parsing configuration file: {0}")]
+    ParseError(#[from] toml::de::Error),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct ConfigBase<T> {
+    #[serde(default)]
+    #[serde(alias = "default_channels")] // BREAK: remove to stop supporting snake_case alias
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub default_channels: Vec<NamedChannelOrUrl>,
```
pavelzw: don't we want to make this an `Option<Vec<NamedChannelOrUrl>>` instead? i don't like the 

```
            default_channels: if other.default_channels.is_empty() {
                self.default_channels.clone()
            } else {
                other.default_channels.clone()
            },
```

below

crates/rattler_config/src/config/concurreny.rs
```diff
@@ -0,0 +1,92 @@
...
+    // Needing to set this default next to the default of the full struct to avoid serde defaulting
+    // to 0 of partial struct was omitted.
+    #[serde(default = "default_max_concurrent_solves")]
+    pub solves: usize,
+
+    /// The maximum number of concurrent HTTP requests to make.
+    // Needing to set this default next to the default of the full struct to avoid serde defaulting
+    // to 0 of partial struct was omitted.
+    #[serde(default = "default_max_concurrent_downloads")]
+    pub downloads: usize,
```
pavelzw, line 36: this is great! we can also use this in rattler-index and conda-mirror 🚀

crates/rattler_config/src/edit.rs
```diff
@@ -0,0 +1,427 @@
...
+                        return Err(ConfigEditError::UnknownKey {
+                            key: key.to_string(),
+                            supported_keys: get_supported_keys(self),
+                        })
+                    }
+                }
+                Ok(())
+            }
+            _ => {
+                // We don't know this key, but possibly an extension does.
```
pavelzw, line 141:
```suggestion
                // TODO: We don't know this key, but possibly an extension does.
```

crates/rattler_config/Cargo.toml
```diff
@@ -0,0 +1,35 @@
...
+homepage.workspace = true
+repository.workspace = true
+license.workspace = true
+readme.workspace = true
+
+[features]
+# feature to get default configuration directories for pixi
+pixi = ["dirs"]
+edit = []
+default = ["pixi", "edit"]
```
baszalmstra: This seems not like something that should live here.

crates/rattler_config/src/config.rs
```diff
@@ -0,0 +1,317 @@
...
+    #[serde(alias = "tls_no_verify")] // BREAK: remove to stop supporting snake_case alias
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub tls_no_verify: Option<bool>,
+
+    #[serde(default)]
+    #[serde(skip_serializing_if = "IndexMap::is_empty")]
+    pub mirrors: IndexMap<Url, Vec<Url>>,
+
+    #[serde(default, skip_serializing_if = "BuildConfig::is_default")]
+    pub build: BuildConfig,
```
baszalmstra, line 91: Is this even something that belongs in the config? And shouldnt this just be part of rattler-build?

crates/rattler_config/src/config.rs
```diff
@@ -0,0 +1,317 @@
...
+    /// Run the post link scripts
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub run_post_link_scripts: Option<RunPostLinkScripts>,
+
+    #[serde(flatten)]
+    pub extensions: T,
+
+    #[serde(skip)]
+    #[serde(alias = "loaded_from")] // BREAK: remove to stop supporting snake_case alias
```
baszalmstra: THis is skipped some the alias is not needed?

crates/rattler_config/src/edit.rs
```diff
@@ -0,0 +1,427 @@
...
+                // self.extensions.set(key, value)
+                //     .into_diagnostic()
+                //     .wrap_err(format!("failed to set extension key '{}'", key))?;
```
baszalmstra: What would be the way forward for pixi to implement this?

crates/rattler_config/src/pixi.rs
```diff
@@ -0,0 +1,45 @@
...
+    .flatten()
+    .collect()
+}
+
+/// Returns the path to the system-level pixi config file.
+pub fn config_path_system() -> PathBuf {
+    // TODO: the base_path for Windows is currently hardcoded, it should be
+    // determined via the system API to support general volume label
+    #[cfg(target_os = "windows")]
+    let base_path = PathBuf::from("C:\\ProgramData");
```
baszalmstra: ProgramData doesnt make sense to me. Can we immediately implement the todo?

crates/rattler_config/src/config/s3.rs
baszalmstra, line 1: Can we add this behind a feature too?

crates/rattler_config/src/config/build.rs
baszalmstra, line 1: Perhaps we can have a feature to enable this?

crates/rattler_config/src/config.rs
```diff
@@ -0,0 +1,317 @@
...
+        Ok(())
+    }
+
+    fn keys(&self) -> Vec<String> {
+        vec![]
+    }
+}
+
+pub trait Config:
+    Serialize + for<'de> Deserialize<'de> + std::fmt::Debug + Clone + PartialEq + Eq + Default
```
baszalmstra, line 182: I dont think you need `Default`, you only need Default for `T` in `impl<T: Config + Default> Default for ConfigBase<T>`.

crates/rattler_config/src/edit.rs
```diff
@@ -0,0 +1,427 @@
...
+impl<T> ConfigBase<T>
+where
+    T: Config + DeserializeOwned,
+{
+    /// Modify this config with the given key and value
+    ///
+    /// # Note
+    ///
+    /// It is required to call `save()` to persist the changes.
+    pub fn set(&mut self, key: &str, value: Option<String>) -> Result<(), ConfigEditError> {
```
baszalmstra, line 70: This code does not seem very maintainable. Would it be possible to implement editing logic directly in the Config trait? instead of combing the logic for all of them here?

crates/rattler_config/src/config.rs
wolfv: Pixi needs it as well for `pixi-build` – was my thought process.

crates/rattler_config/src/edit.rs
wolfv: Pixi would implement a function just like this one, that cares about all the keys coming from the "extension". But this code should not have been commented, indeed.

crates/rattler_config/src/config.rs
wolfv: yep!

crates/rattler_config/Cargo.toml
wolfv: you mean the `pixi` option? I think it's useful for other projects to have access to the default storage location of pixi's configuration but I put it behind a feature flag as I agree.

However, we're actually not using this in `rattler-build` so we can also not add it.

crates/rattler_config/src/pixi.rs
wolfv: I copied this code straight from pixi.

crates/rattler_config/src/config/repodata_config.rs
```diff
@@ -0,0 +1,190 @@
...
+//     fn from(value: RepodataChannelConfig) -> Self {
+//         SourceConfig {
+//             jlap_enabled: !value.disable_jlap.unwrap_or(false),
+//             zstd_enabled: !value.disable_zstd.unwrap_or(false),
+//             bz2_enabled: !value.disable_bzip2.unwrap_or(false),
+//             sharded_enabled: !value.disable_sharded.unwrap_or(false),
+//             cache_action: Default::default(),
+//         }
+//     }
+// }
```
baszalmstra: Remove?

baszalmstra:
```suggestion
```

baszalmstra (approved):

crates/rattler_config/src/edit.rs
wolfv: I don't think it's a TODO ... we _don't_ know this key, but an extension might

PR #1465, opened 2025-06-24
fix: simplify validate env vars

wolfv:

### DIFF
--- a/crates/rattler_shell/src/shell/mod.rs
+++ b/crates/rattler_shell/src/shell/mod.rs
@@ -240,23 +240,10 @@ fn validate_env_var_name(name: &str) -> Result<(), ShellError> {
         ));
     }
 
-    // First character must be alphabetic or underscore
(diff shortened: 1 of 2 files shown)


crates/rattler_shell/src/shell/mod.rs
```diff
@@ -240,23 +240,10 @@ fn validate_env_var_name(name: &str) -> Result<(), ShellError> {
         ));
     }
 
-    // First character must be alphabetic or underscore
-    if !name
-        .chars()
-        .next()
-        .is_some_and(|c| c.is_ascii_alphabetic() || c == '_')
-    {
+    if name.contains(['=', '\n', '\r', '\t']) {
```
baszalmstra: I think we should also disallow control characters in environment variables.

wolfv: Done!

PR #1466, opened 2025-06-24
feat(rattler_index): Use rattler_config

### LINKED ISSUE #1153: Inconsistent definitions of `max_parallel` in rattler-index
https://github.com/conda/rattler/blob/c01c4601cdeeef0fe18ca5c99bc4bf3213769136/py-rattler/rattler/index/index.py#L15

https://github.com/conda/rattler/blob/c01c4601cdeeef0fe18ca5c99bc4bf3213769136/crates/rattler_index/src/main.rs#L35

we should probably make this an `Option` instead and just pass `None` from py-rattler

### LINKED ISSUE #1288: Use config file in rattler-index
In rattler-build, pixi-pack we can now use `--config` to specify a pixi config like `~/.pixi/config.toml` where we can now define things like mirror middlewares or s3 configuration.

depends on https://github.com/prefix-dev/pixi/issues/3691 or us building a `rattler_config` crate (which i would probably prefer in the long run)

pavelzw:
### Description

closes #1288
i decided to not include mirror config as this really doesn't make much sense in rattler-index (at the moment)

closes #1153

### DIFF
--- a/crates/rattler_index/src/main.rs
+++ b/crates/rattler_index/src/main.rs
@@ -1,6 +1,10 @@
+use std::path::PathBuf;
+
+use anyhow::Context;
 use clap::{arg, Parser, Subcommand};
 use clap_verbosity_flag::Verbosity;
 use rattler_conda_types::Platform;
+use rattler_config::config::concurreny::default_max_concurrent_solves;
 use rattler_index::{index_fs, index_s3, IndexFsConfig, IndexS3Config};
 use url::Url;
 
@@ -40,8 +44,8 @@ struct Cli {
 
     /// The maximum number of packages to process in-memory simultaneously.
     /// This is necessary to limit memory usage when indexing large channels.
-    #[arg(long, default_value = "32", global = true)]
-    max_parallel: usize,
+    #[arg(long, global = true)]
+    max_parallel: Option<usize>,
 
     /// A specific platform to index.
     /// Defaults to all platforms available in the channel.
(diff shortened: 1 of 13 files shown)


crates/rattler_index/src/main.rs
```diff
@@ -137,6 +155,20 @@ async fn main() -> anyhow::Result<()> {
...
+                .and_then(|config| config.s3_options.0.get(&bucket));
+            let region = region
+                .or(s3_config.map(|c| c.region.clone()))
+                .context("S3 region not provided")?;
+            let endpoint_url = endpoint_url
+                .or(s3_config.map(|c| c.endpoint_url.clone()))
+                .context("S3 endpoint url not provided")?;
+            let force_path_style = force_path_style
+                .or(s3_config.map(|c| c.force_path_style))
+                .context("S3 force-path-style not provided")?;
```
pavelzw, line 170: i removed the defaults in favor of throwing explicit error messages as this behavior would still be inconsistent with pixi:
in pixi/rattler-networking (we use opendal here), it relies on aws's magic to resolve those variables. we don't have (or want) this here so IMO we should just raise an error instead of using unsensible defaults

cc @delsner

delsner: this seems to be a breaking change; does this require a major release?

pavelzw: A minor version bump since we are 0.x.x
Hopefully this is handled automatically by release-plz

pavelzw: Since the public api changed, this should be a minor release change here automatically

baszalmstra (approved):

baszalmstra:
@delsner Ill merge when you approve!

delsner (approved):
Thanks, lgtm!

PR #1468, opened 2025-06-24
feat: Set default User-Agent if not provided

### LINKED ISSUE #1442: Missing User-Agent HTTP header
### Checklist

### What happened?

Hello, after some investigation, it looks like rattler doesn't set any User-Agent HTTP header while fetching repodata, etc.

Could it be possible to make sure that all HTTP requests generated by rattler have a proper User-Agent set please?

(And hopefully I'm not missing something)

Thanks!

### Additional Context

_No response_

JeanChristopheMorinPerso:
### Description

Fixes #1442 

This PR is an attempt at fixing #1442.

Changes on the rust side:

* Set a default user-agent in `GatewayBuilder.finish()` if no client is provided.

Changes on the Python side:
* Add a `headers` keyword argument to `rattler.Client` to allow setting HTTP headers.
* If no `headers` are provided to `rattler.Client`, set the default user-agent to `py-rattler/<version>`.
* A PyClient will be created if no client is passed to `rattler.Gateway`, ensuring that
  the default user-agent is `py-rattler/<version>`.

### DIFF
--- a/crates/rattler_networking/src/oci_middleware.rs
+++ b/crates/rattler_networking/src/oci_middleware.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use http::{
-    header::{ACCEPT, AUTHORIZATION, USER_AGENT},
+    header::{ACCEPT, AUTHORIZATION},
     Extensions,
 };
 use reqwest::{Request, Response};
@@ -14,8 +14,6 @@ use serde::Deserialize;
 use url::{ParseError, Url};
 
 use crate::mirror_middleware::create_404_response;
-
-static RATTLER_USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
 
 #[derive(thiserror::Error, Debug)]
 enum OciMiddlewareError {
--- a/crates/rattler_networking/src/oci_middleware.rs
+++ b/crates/rattler_networking/src/oci_middleware.rs
@@ -185,13 +185,6 @@ impl OCIUrl {
                 .expect("Could not parse token header"),
         );
 
-        req.headers_mut().insert(
-            USER_AGENT,
-            RATTLER_USER_AGENT
-                .parse()
-                .expect("Could not parse user-agent header"),
-        );
-
         // if we know the hash, we can pull the artifact directly
         // if we don't, we need to pull the manifest and then pull the artifact
         if let Some(expected_sha_hash) = req
@@ -208,7 +201,6 @@ impl OCIUrl {
                 .get(manifest_url)
                 .header(AUTHORIZATION, format!("Bearer {token}"))
                 .header(ACCEPT, "application/vnd.oci.image.manifest.v1+json")
-                .header(USER_AGENT, RATTLER_USER_AGENT)
                 .send()
                 .await?;
 
--- a/crates/rattler_networking/src/oci_middleware.rs
+++ b/crates/rattler_networking/src/oci_middleware.rs
@@ -5,7 +5,7 @@ use std::{
 };
 
 use http::{
-    header::{ACCEPT, AUTHORIZATION},
+    header::{ACCEPT, AUTHORIZATION, USER_AGENT},
     Extensions,
 };
 use reqwest::{Request, Response};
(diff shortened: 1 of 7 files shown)


crates/rattler_networking/src/oci_middleware.rs
```diff
@@ -183,6 +185,13 @@ impl OCIUrl {
                 .expect("Could not parse token header"),
         );
 
+        req.headers_mut().insert(
```
JeanChristopheMorinPerso: I'm not convinced I need to do this here... I ca remove this if it's not needed.

py-rattler/src/networking/client.rs
```diff
@@ -16,10 +18,28 @@ pub struct PyClientWithMiddleware {
...
+    pub fn new(
+        middlewares: Option<Vec<PyMiddleware>>,
+        headers: Option<HashMap<String, String>>,
+    ) -> PyResult<Self> {
         let middlewares = middlewares.unwrap_or_default();
-        let mut client = reqwest_middleware::ClientBuilder::new(reqwest::Client::new());
+
+        let mut client_builder = reqwest::Client::builder();
+
+        if let Some(headers) = headers {
```
JeanChristopheMorinPerso, line 32: I think I might need an else to handle the case where no headers are passed...

crates/rattler_networking/src/oci_middleware.rs
baszalmstra: Yeah I think this should already be handled by the client that handles the request.

py-rattler/src/networking/client.rs
baszalmstra: Yep! Otherwise there is still no default user agent!

JeanChristopheMorinPerso:
What should we do when client is None in https://github.com/conda/rattler/blob/main/py-rattler/src/repo_data/gateway.rs#L84-L86? With the code I wrote, it'll default to the user-agent set in the gateway, which will use the user-agent at https://github.com/conda/rattler/pull/1468/files#diff-495e40b4eb00406ac1c091ebae05e596eb8662a0eabdf52b0c7e74bb99461c7bR138... DO we want to set a default client on the python gateway too to get a consistent user-agent?

JeanChristopheMorinPerso:
@baszalmstra Opinions on ⬆️?

baszalmstra:
We could also opt to not make that an `Option` in the rust binding side and instead use a default parameter in python with the proper user agent set.

JeanChristopheMorinPerso:
So you are suggesting that we could change the type of `client` from `Option<ClientWithMiddleware>` to `ClientWithMiddleware` like in the rest of the codebase?

https://github.com/conda/rattler/blob/cfefadf4e91ee59656c531d79b06131157d1aec9/crates/rattler_repodata_gateway/src/gateway/builder.rs#L43

baszalmstra:
No I mean that in the python function we add a default argument value which is the "default" Client.

In the constructor of the python Client we add an argument for the user agent which we also assign a default value.

JeanChristopheMorinPerso:
Ok, thanks for the clarifications. `rattler.Gateway` already accepts a `client`. So what I did was construct a default `PyClientWithMiddleware` if none is provided. A user can customize the headers by passing a client.

crates/rattler_repodata_gateway/src/gateway/builder.rs
```diff
@@ -8,6 +8,8 @@ use reqwest_middleware::ClientWithMiddleware;
 
 use crate::{gateway::GatewayInner, ChannelConfig, Gateway};
 
+static USER_AGENT: &str = concat!(env!("CARGO_PKG_NAME"), "/", env!("CARGO_PKG_VERSION"),);
```
baszalmstra, line 11: I think this will resolve to the `rattler_repodata_gateway` crate and its version.  Not sure if thats what you want? On the other hand I also wouldnt know what would be better.

JeanChristopheMorinPerso: Yeah, I wasn't sure either. We could go with `rattler`, but AFAIK, rattler doesn't really have a central version? And even if it did, I don't know how we would get that version here.

baszalmstra: yeah its just something to be aware of I guess.

PR #1475, opened 2025-06-25
fix: recursively remove empty directories after uninstall

remimimimimi:
### Description

Fix of pixi issue prefix-dev/pixi#3945. Test case can be simplified, but I left it as related as possible to original issue for clarity.


crates/rattler/src/install/link.rs
```diff
@@ -422,9 +451,7 @@ fn hardlink_to_destination(
         match fs::hard_link(source_path, destination_path) {
             Ok(_) => return Ok(LinkMethod::Hardlink),
             Err(e) if e.kind() == ErrorKind::AlreadyExists => {
-                fs::remove_file(destination_path).map_err(|err| {
-                    LinkFileError::IoError(String::from("removing clobbered file"), err)
-                })?;
+                remove_file_or_empty_dirs(destination_path)?;
```
baszalmstra: I think this could cause a race condition. During linking of a single file we recursively construct directories and then install a file into it. However, this operation is not atomic so it could happen that before installing a file to a directory the directory is accidentally deleted by the removal of another file (and its directories).

wolfv:
I had some additional thoughts on this. IMO there is two ways to solve this:

- Do all `unlink` first (ie. remove all packages from the environment as a first step), and right after that (before installing new files) recursively remove any empty directories
- Handle this with the clobber registry and register if there are any empty directories that are not part of the installation anymore, or take care of it afterwards (figure out if the directory or the file should win).

Actually, we could probably already today use the clobber registry to determine that it is safe to delete the empty folder (by just looking at all paths registered in the paths registry).

baszalmstra:
Yeah unlinking everything first would be by far the easiest but it may slow down the overall progress if everything has to wait for uninstalling a large package like rust or pytorch. Perhaps we should benchmark that first?

wolfv:
I think as long as we start downloading new packages at the same time it should be fine, speed wise. But yes, benchmarking the before and after would definitely be good.

PR #1480, opened 2025-06-26
feat: Add glob support for package name in matchspec

pavelzw:
### Description

for `rattler::install`, `rattler_solve` and `rattler_repodata_gateway` i didn't support globbed matchspecs to keep this pr manageable.
instead, i behave exactly like when we had a nameless matchspec there (i.e. either skip or panic)
i don't think supporting them in `rattler_solve` makes much sense as there are no packages that have globbed packages as dependencies anyway.
for `rattler_repodata_gateway`, i do see some nice use cases (for example `pixi search jupyter*` which is currently handled in a custom manner). let's move this to another pr: #1723

### DIFF
--- a/py-rattler/rattler/repo_data/gateway.py
+++ b/py-rattler/rattler/repo_data/gateway.py
@@ -1,17 +1,16 @@
 from __future__ import annotations
 
 import os
-from typing import Optional, List, Iterable
 from dataclasses import dataclass
-
-from rattler.rattler import PyGateway, PySourceConfig, PyMatchSpec
+from typing import Iterable, List, Optional
 
 from rattler.channel import Channel
 from rattler.match_spec import MatchSpec
-from rattler.networking import Client, CacheAction
-from rattler.repo_data.record import RepoDataRecord
-from rattler.platform import Platform, PlatformLiteral
+from rattler.networking import CacheAction, Client
 from rattler.package.package_name import PackageName
+from rattler.platform import Platform, PlatformLiteral
+from rattler.rattler import PyGateway, PyMatchSpec, PySourceConfig
+from rattler.repo_data.record import RepoDataRecord
 
 
 @dataclass
@@ -175,7 +174,10 @@ class Gateway:
                 platform._inner if isinstance(platform, Platform) else Platform(platform)._inner
                 for platform in platforms
             ],
-            specs=[spec._match_spec if isinstance(spec, MatchSpec) else PyMatchSpec(str(spec), True) for spec in specs],
+            specs=[
+                spec._match_spec if isinstance(spec, MatchSpec) else PyMatchSpec(str(spec), True, True)
+                for spec in specs
+            ],
             recursive=recursive,
         )
 
--- a/py-rattler/rattler/solver/solver.py
+++ b/py-rattler/rattler/solver/solver.py
@@ -1,14 +1,13 @@
 from __future__ import annotations
+
 import datetime
-from typing import List, Optional, Literal, Sequence
+from typing import List, Literal, Optional, Sequence
 
-from rattler import Channel, Platform, VirtualPackage, SparseRepoData
-from rattler.match_spec.match_spec import MatchSpec
-
+from rattler import Channel, Platform, SparseRepoData, VirtualPackage
 from rattler.channel import ChannelPriority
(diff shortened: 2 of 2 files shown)


crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -260,8 +260,12 @@ impl SparseRepoData {
         let base_url = repo_data.info.as_ref().and_then(|i| i.base_url.as_deref());
         for (package_name, specs) in &spec.into_iter().chunk_by(|spec| spec.borrow().name.clone()) {
             let grouped_specs = specs.into_iter().collect::<Vec<_>>();
+            // TODO: support glob/regex package names
```
pavelzw, line 263: https://github.com/conda/rattler/issues/1723

pavelzw:
@baszalmstra ready for review! once this is merged, i can make the config for conda-mirror a bit easier to use!

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -252,18 +258,21 @@ impl MatchSpec {
     /// Returns whether the package is a virtual package.
     /// This is determined by the package name starting with `__`.
     /// Not having a package name is considered not virtual.
+    /// Matching both virtual and non-virtual packages is considered not virtual.
     pub fn is_virtual(&self) -> bool {
-        self.name
```
wolfv, line 270: This might deserve a little unit test.

crates/rattler_solve/tests/snapshots/backends__resolvo__issue_717.snap
```diff
@@ -4,4 +4,4 @@ expression: result.unwrap_err()
 ---
 Cannot solve the request because of: The following packages are incompatible
 └─ issue_717 * cannot be installed because there are no viable options:
-   └─ issue_717 2.1 is excluded because the constrains 'ray[default,data] >=2.9.0,<3.0.0' failed to parse: 'ray[default,data]' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or .
+   └─ issue_717 2.1 is excluded because the constrains 'ray[default,data] >=2.9.0,<3.0.0' failed to parse: invalid package name ray[default,data]: 'ray[default,data]' is not a valid package name. Package names can only contain 0-9, a-z, A-Z, -, _, or .
```
wolfv: this error message is a bit degraded. There are no quotes around the (double) `name` field.

pavelzw: [`54772d1` (#1480)](https://github.com/conda/rattler/pull/1480/commits/54772d1a07cc9cdd88223af1c21feceab1ea3290)

crates/rattler_conda_types/src/match_spec/mod.rs
pavelzw: [`54772d1` (#1480)](https://github.com/conda/rattler/pull/1480/commits/54772d1a07cc9cdd88223af1c21feceab1ea3290)

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -986,6 +986,20 @@ mod tests {
         let spec =
             MatchSpec::from_nameless(NamelessMatchSpec::from_str(">=12", Strict).unwrap(), None);
         assert!(!spec.is_virtual());
+
+        let spec = MatchSpec::from_str("__virtual_glob*", Strict).unwrap();
```
wolfv: Now that I see this I wonder if we should add a third mode for the parsing or a new constrcutor because I would like to validate at parsing in rattler-build that we get exact matchspecs. Maybe something like `from_str_wiht_matcher` or Strict, Lenient, StrictWithMatcher, LenientWithMatcher ...

pavelzw: decided on doing this:


```rs
pub struct ParseStrictnessWithNameMatcher {
    /// Whether to allow only `Exact` matchers for the package name or whether to also allow `Glob` or `Regex` matchers.
    pub exact_names_only: bool,
    /// Defines how strict a version parser should behave.
    pub parse_strictness: ParseStrictness,
}
```

pavelzw:
@baszalmstra @wolfv ready for review

wolfv (approved):

PR #1481, opened 2025-06-26
fix(ci): run pre-commit-run for all files

remimimimimi:
### Description

Before pre-commit-run didn't run all lints in CI, this PR should fix this and apply all suggested changes.


.pre-commit-config.yaml
```diff
@@ -1,4 +1,4 @@
-exclude: (^.pixi/|.snap)
+exclude: (^.pixi/|.snap|js-rattler/src/PackageName.test.ts|py-rattler/mkdocs.yml)
```
baszalmstra, line 4: Can you add a comment why these are excluded?

PR #1497, opened 2025-07-01
fix(clobber registry): directory and file clobbering

remimimimimi:
### Description

Introduce new path conflict resolver that would allow to gracefully resolve path conflicts between file/directory in addition of previously handled file/file conflicts. See tests in `package_path_resolver.rs` in order to understand how it will resolve different types of conflict.

Current state is draft since we don't do anything on-disk yet.


baszalmstra:
This is shaping up very nicely! Awesome work so far! Much clearer!

crates/rattler/src/install/package_path_resolver.rs
```diff
@@ -0,0 +1,741 @@
...
+        entries: &mut HashSet<Entry>,
+    ) {
+        let mut current = file_path.parent();
+        let mut dirs_to_create = Vec::with_capacity(file_path.components().count());
+
+        while let Some(parent) = current {
+            if parent == Path::new("") {
+                break;
+            }
+            dirs_to_create.push(parent);
```
wolfv: You could probably short circuit this here by creating the entry directly, and stopping the iteration when the entry already exists.

If the set already contained the value, `false` is returned: https://doc.rust-lang.org/std/collections/struct.HashSet.html#method.insert

remimimimimi:
I changed implementation of reprioritize method so now we have performance on par with previous implementation. I'm also much more confident that this function is correct.

image

crates/path_trie/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "path_trie"
```
baszalmstra: Does this name still make sense? It seems very much tailored towards unclobbering

crates/path_trie/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
```
baszalmstra: Missing description

crates/path_trie/src/lib.rs
```diff
@@ -0,0 +1,967 @@
+use std::{
```
baszalmstra: Please add 

```
#![deny(missing_docs)]
```

crates/path_trie/src/lib.rs
```diff
@@ -0,0 +1,967 @@
...
+                .sorted()
+                .collect::<Vec<PackageName>>();
+
+            let new_sorted_packages = new_order
+                .iter()
+                .cloned()
+                .sorted()
+                .collect::<Vec<PackageName>>();
+
+            assert_eq!(
```
baszalmstra: Maybe we can turn this into a `debug_assert_eq` to avoid comparing lots of sets?

crates/path_trie/src/lib.rs
```diff
@@ -0,0 +1,967 @@
...
+            &old_rank,
+            &new_rank,
+            &mut to_clobbers,
+            &mut from_clobbers,
+        );
+
+        self.packages.clear();
+        for pkg in new_order.iter().rev() {
+            self.packages.insert(pkg.clone());
+        }
```
baszalmstra: Lets avoid cloning and remove the for loop:

```suggestion
        self.packages.extend(new_order.into_iter().rev())
```

crates/path_trie/src/lib.rs
```diff
@@ -0,0 +1,967 @@
...
+    /// - For each `(p,pkg)` in `from_clobers`: `clobbers/pkg/p` → `base/p` if source exists and dest doesn’t.
+    pub fn sync_clobbers(
+        target_prefix: &Path,
+        clobbers_dir: &Path,
+        to_clobbers: &[(PathBuf, PackageName)],
+        from_clobbers: &[(PathBuf, PackageName)],
+    ) -> io::Result<()> {
+        fn mv(src: PathBuf, dst: PathBuf) -> io::Result<()> {
+            tracing::trace!("Moving from {} to {}", src.display(), dst.display());
+            if let Some(p) = dst.parent() {
```
baszalmstra: Is this actually needed? I assume since the file was previously clobbered the target directory already exists.

crates/path_trie/src/lib.rs
```diff
@@ -0,0 +1,967 @@
...
+
+        // 1) detect conflicts against the existing trie
+        for p in paths {
+            let p = p.as_ref();
+            let pbuf = p.to_path_buf();
+
+            // File vs File?
+            if let Some(n) = Self::get_node(&self.root, &pbuf) {
+                if !n.terminals.is_empty() {
+                    conflicts.insert(pbuf.clone());
```
baszalmstra: Is this clone required?

crates/path_trie/src/lib.rs
```diff
@@ -0,0 +1,967 @@
...
+                    continue;
+                }
+            }
+            // Directory vs File?
+            if let Some(n) = Self::get_node(&self.root, &pbuf) {
+                if !n.children.is_empty() {
+                    conflicts.insert(pbuf.clone());
+                    // Mark this as a directory-insert so we later propagate
+                    dir_inserts.push(pbuf.clone());
```
baszalmstra: I think this clone is also redundant.

crates/path_trie/src/lib.rs
```diff
@@ -0,0 +1,967 @@
...
+            }
+        }
+
+        let mut results = HashMap::new();
+        dfs(&self.root, Path::new(""), &self.packages, &mut results);
+        results
+    }
+}
+
+// TODO: Write property based tests.
```
baszalmstra: Is this still relevant?

crates/rattler/Cargo.toml
```diff
@@ -67,3 +68,5 @@ tokio = { workspace = true, features = ["macros", "rt-multi-thread"] }
 axum = { workspace = true }
 tower-http = { workspace = true, features = ["fs"] }
 tower = { workspace = true, features = ["util"] }
+proptest = "1.7.0"
```
baszalmstra: Please add a workspace dependency.

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -29,135 +29,70 @@ pub enum ClobberError {
...
 /// can be used to rename files that are already installed by another package.
 #[derive(Debug, Default, Clone)]
 pub struct ClobberRegistry {
-    /// A cache of package names
-    package_names: Vec<PackageName>,
-
-    /// The paths that exist in the prefix and the first package that touched
-    /// the file.
-    paths_registry: HashMap<PathBuf, PathState>,
+    package_name_map: HashMap<path_trie::PackageName, rattler_conda_types::PackageName>,
```
baszalmstra: Can you add a comment describing what does does?

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -29,135 +29,70 @@ pub enum ClobberError {
...
+                .paths
+                .iter()
+                .all(|pe| pe.original_path.is_none())
+        });
+
+        // I think we assume here that every conflict will occur
+        // independently of order in which we register prefix records
+        // if first record is previous winner or smth.
+        //
+        // We should test that
```
baszalmstra: Should we still test that? Or is this if statement said test?

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -568,36 +455,751 @@ mod tests {
...
+            ],
+        );
+
+        let prefix_records = PrefixRecord::collect_from_prefix(target_prefix.path()).unwrap();
+
+        let prefix_record_clobber_1 = find_prefix_record(&prefix_records, "clobber-1");
+        assert!(prefix_record_clobber_1.is_none());
+        let prefix_record_clobber_2 = find_prefix_record(&prefix_records, "clobber-2").unwrap();
+        assert_yaml_snapshot!(prefix_record_clobber_2.files);
+        assert_yaml_snapshot!(prefix_record_clobber_2.paths_data);
```
baszalmstra: Having multiple snapshots in a single test can be annoying because you run the test, get a snapshot change, update the snapshot and assume the test succeeds now. However, another snapshot might still fail. If possible I prefer to combine snapshots in a single snapshot to avoid this confusion.

crates/rattler/Cargo.toml
baszalmstra: Actually.. Is this still used?

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -568,36 +455,751 @@ mod tests {
...
+                            PathBuf::from("__clobbers__/clobber-3/clobber.txt")
+                        ]
+                    );
+                }
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn test_all_possible_orders_clobber_nested() {
```
baszalmstra, line 620: This test is huuuge! Would it be possible to automatically compute the permutations?

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -931,14 +1326,22 @@ mod tests {
...
-        assert_check_files(
+        assert_check_files!(
             target_prefix.path(),
+            // // We do things a little bit suboptimally, so we have different clobbers. This should be resolved eventually by better handling of updates.
+            // &[
+            //     "clobber.txt",
+            //     "another-clobber.txt",
+            //     "__clobbers__/clobber-1/clobber.txt",
+            //     "__clobbers__/clobber-3/clobber.txt",
+            //     "__clobbers__/clobber-3/another-clobber.txt",
```
baszalmstra: Can this be removed?

crates/rattler/src/install/mod.rs
```diff
@@ -353,6 +354,19 @@ pub async fn link_package(
                 break;
             }
         }
+
+        // Since we store clobbers in the separate directory
```
baszalmstra, line 358: So we also need to do this in the other function?

baszalmstra (changes requested):
I really like the structure and your approach! I left a few comments.

crates/path_trie/src/lib.rs
remimimimimi: But what if we update package? First every file is removed, then new package is in clobbers. It may be that directories do not exists when we move file from clobbers. And if dir exists, then this call has almost no cost https://doc.rust-lang.org/src/std/fs.rs.html#3139-3164.

crates/path_trie/src/lib.rs
remimimimimi: Yes, this is still relevant. I'm much more sure that this code behaves correctly, but I still may miss some cases. I spent this morning reading documentation of proptest crate and want to implement them today.

crates/rattler/Cargo.toml
remimimimimi: It isn't used. I tried to add some simple property-based test in rattler, but failed miserably because of current api, that's why I introduced new crate which handles main logic purely, which is much more suitable for testing.

crates/rattler/src/install/clobber_registry.rs
remimimimimi: Yes, we definitely should test that. I think this part is most likely to fail in some real-world scenarios. I'll prepare property test for it.

If it will fail then I plan to sort them based on the paths that are clobbered, trying to reconstruct original order. I'm not sure about details of this algorithm though.

crates/rattler/src/install/clobber_registry.rs
remimimimimi: See https://insta.rs/docs/cli/#test. I don't think that it would be easy or actually useful to combine snapshots.

If you want to use nextest runner, then you can call `cargo insta test --test-runner nextest --review`.

crates/rattler/src/install/clobber_registry.rs
remimimimimi: I needed manual assertions for something, but now that's unneeded. I changed it back to loop on permutations.

crates/rattler/src/install/clobber_registry.rs
remimimimimi: Yes.

crates/rattler/src/install/mod.rs
remimimimimi: Definitely, thanks for reminding!

crates/path_trie/Cargo.toml
remimimimimi: I changed name to path resolver.

crates/path_trie/src/lib.rs
remimimimimi: I that we should keep this assertion in release as well to catch potential bugs, at least for some time, then we can change it to debug. Without it we could get hard to debug issues.

I changed implementation to one with linear time and memory, so it should be fast now.

crates/rattler/src/install/clobber_registry.rs
baszalmstra: I dont particularly agree that people should just change their workflow, but this is fine for me.

PR #1528, opened 2025-07-16
fix: enable rattler_networking 's3' features and update tokio fs alias

magentaqin:
### Description
This PR is to solve error in release pipeline: https://github.com/conda/rattler/actions/runs/16263440455/job/45913720568
I did these three things:
1. turn on `s3` feature of `rattler_networking` 
2. update the alias of tokio_fs(as it's also used in other crates, we'd better keep the name the same)
3. update `fs-err` to be a workspace dependency


baszalmstra:
Thanks!

Would it be possible to make s3 a feature? It adds a bunch of heavy dependencies that we might not always need.

magentaqin:
> Thanks!
> 
> Would it be possible to make s3 a feature? It adds a bunch of heavy dependencies that we might not always need.
Gerne!

PR #1529, opened 2025-07-16
chore: update toml and other dependencies

wolfv:


rust-toolchain
```diff
@@ -1 +1 @@
-1.85.0
+1.86.0
```
baszalmstra: please also modify the pixi.toml

wolfv:
I think this is fine to be merged .

PR #1530, opened 2025-07-16
feat: solving with conditional dependencies

wolfv:
This parses `foobar; if python >=3.12`

However, the condition is parsed as a simple string for the time being. I wonder if should be parsed as `matchspec` again? Possibly with some constraint that it cannot contain another condition? Althouhg I believe we also want to support `or` and `and` (although I don't remember if one or both were implemented in `resolvo`.


wolfv:
This is superseded by #1545

PR #1542, opened 2025-07-21
feat: implement extras with conditional dependencies

baszalmstra:
Uses conditional dependency support in resolve to implement extras. Result of some late-night hacking with @wolfv . 

This also adds extras in a different way. Extras are completely optional now, so if a user depends on an extra that doesnt exist the solver will happily continue. Its up to users of the library to warn if an extra is selected that is not actually available for the selected package.

This is still in draft because conditional dependency support is not yet stabilized in resolvo.


crates/rattler_solve/tests/backends.rs~
```diff
@@ -0,0 +1,1446 @@
+use std::{collections::BTreeMap, str::FromStr, time::Instant};
```
wolfv: This whole file looks strange? ~crates? Remove?

test-data/channels/dummy-optional-dependencies/noarch/repodata.json
```diff
@@ -1,528 +1,55 @@
 {
-    "info": {
```
wolfv, line 2: Should we keep a few of these tests? Just to make sure the extras work as expected?

wolfv: OK, I saw that you have one functionality test. I guess that's fine.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -279,21 +282,25 @@ impl<'a> CondaDependencyProvider<'a> {
         strategy: SolveStrategy,
     ) -> Result<Self, SolveError> {
         let pool = Pool::default();
-        let mut records: HashMap<NameId, Candidates> = HashMap::default();
+        let mut records: HashMap<String, Candidates> = HashMap::default();
```
wolfv: I think switching to String here was a concious decision - is it because we create extras on the fly?

I guess it shouldn't have too much of a performance impact?

wolfv:
The benchmarks show a inconsistent difference of +-3% which seems fine to me!

crates/rattler_solve/src/resolvo/mod.rs
baszalmstra: Yeah Inthink we can change that back!

wolfv:
ugh, somehow the extras printing adds a single space after `]`? I checked and it doesn't seem to come from this code.

wolfv:
I think https://github.com/prefix-dev/resolvo/pull/164 should fix this.

baszalmstra:
I updated the resolvo dependencies, merged main, and fixed the thing you find. This looks good from my perspective @wolfv

wolfv:
Awesome stuff! I think now I need to rebase my PR in #1545

PR #1543, opened 2025-07-22
fix: use channel `subdir` instead of the one from the record

baszalmstra:
This changes the way urls are constructed for package records that incorrectly report their `subdir`. When constructing the url we now take the subdir from where the repodata was downloading instead of the subdir that is recorded in the package record.

Fixes https://github.com/prefix-dev/pixi/issues/4164


wolfv (approved):
Looks good to me!

PR #1545, opened 2025-07-22
add proper condition parsing and proper translation to resolvo

wolfv:


crates/rattler_conda_types/src/match_spec/condition.rs
```diff
@@ -0,0 +1,257 @@
...
+};
+use serde::Serialize;
+
+use crate::match_spec::parse::matchspec_parser;
+
+#[derive(Debug, Clone, PartialEq, Serialize)]
+pub struct Statement {
+    pub prefix: String,
+    pub condition: Option<MatchSpecCondition>,
+}
```
baszalmstra: I had a hard time understanding what a `Statement` is. Especially the `prefix` field. Maybe add some comments to clarify.

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -157,6 +159,8 @@ pub struct MatchSpec {
     pub url: Option<Url>,
     /// The license of the package
     pub license: Option<String>,
+    /// The condition under which this match spec applies.
+    pub condition: Option<MatchSpecCondition>,
```
baszalmstra, line 166: Shall we also add this to NamelessMatchSpec?

crates/rattler_conda_types/src/match_spec/condition.rs
```diff
@@ -0,0 +1,257 @@
...
+
+/// Represents a condition in a match spec, which can be a match spec itself or a logical combination
+#[derive(Debug, Clone, PartialEq, Serialize, Eq, Hash)]
+pub enum MatchSpecCondition {
```
baszalmstra, line 16: One slight issue I have with this approach is that some information is lost when converting from a string. E.g. converting back to a string loses the order and parenthesis.

Im wondering if we can structure this differently or add a wrapping struct that holds the original statement or something along those lines. Ideally we build a concrete syntax tree.

The reason this is important is that when we report the conditions in an unsat error you want the information to closely resemble the users input.

crates/rattler_conda_types/src/match_spec/condition.rs
```diff
@@ -0,0 +1,257 @@
...
+
+// Parse whitespace
+fn ws(input: &str) -> IResult<&str, &str> {
+    multispace0(input)
+}
+
+// Parse a matchspec by consuming until we hit a delimiter
+fn matchspec_token(input: &str) -> IResult<&str, &str> {
+    // Try to find the next delimiter
+    let delimiters = ["and", "or", ")", "("];
```
baszalmstra, line 43: Shall we make this case insensitive? E.g. also allow `"AND"` and `"OR"`.

crates/rattler_conda_types/src/match_spec/condition.rs
```diff
@@ -0,0 +1,257 @@
...
+            // Make sure it's a word boundary for "and"/"or"
+            if *delimiter == "and" || *delimiter == "or" {
+                // Check if it's preceded and followed by whitespace or start/end of string
+                let is_word_boundary = {
+                    let before_ok = pos == 0 || input.chars().nth(pos - 1).unwrap().is_whitespace();
+                    let after_ok = pos + delimiter.len() >= input.len()
+                        || input
+                            .chars()
+                            .nth(pos + delimiter.len())
+                            .unwrap()
```
baszalmstra: The `unwrap` here basically already checks the precondition that the position is within range. So this could be simplified to `map_or(true, char::is_whitespace)`.

crates/rattler_conda_types/src/match_spec/condition.rs
baszalmstra: This seems to only be used in the tests

crates/rattler_conda_types/src/match_spec/condition.rs
wolfv: It's not the case in Python. I don't have a super strong opinion but I am leaning towards no.

crates/rattler_conda_types/src/match_spec/mod.rs
wolfv: Yeah the main challenge is that we then have a MatchSpec in a NamelessMatchspec and MatchSpec doesn't have a `Deserialize` implementation. But I think we can annotate the Condition deserializer to deserialize / serialize the MatchSpec with ToString / FromStr.

crates/rattler_conda_types/src/match_spec/condition.rs
wolfv: Yeah this is indeed only used in the tests. I thought I moved it there.

wolfv:
@baszalmstra actually we have a few corner cases with our proposed syntax:

```
            "deep; if foobar >=1.23 *or* and(b >32.12,<=43 *and or(c and d))",
```

Or in other words, we cannot use `and` or `or` in build strings anymore.

Maybe it would be better to go for a "C-like" syntax, using `&&` and `||` as they are not found and hopefully forbidden in build strings? Looks slightly uglier though!

crates/rattler_conda_types/src/match_spec/mod.rs
baszalmstra: Yes that is what we do in more places!

Cargo.toml
```diff
@@ -126,7 +126,7 @@ regex = "1.11.1"
 reqwest = { version = "0.12.15", default-features = false }
 reqwest-middleware = "0.4.2"
 reqwest-retry = "0.7.0"
-resolvo = { version = "0.10.0" }
+resolvo = "0.10.0"
```
baszalmstra: `0.10.1` also seems to be available. Maybe we can bump to that in the lock-file.

crates/rattler_conda_types/src/match_spec/condition.rs
baszalmstra: We actually never parse the condition back to a string so this is not true.

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -157,6 +161,12 @@ pub struct MatchSpec {
     pub url: Option<Url>,
     /// The license of the package
     pub license: Option<String>,
+    /// The condition under which this match spec applies.
+    #[cfg(feature = "conditional_dependencies")]
+    pub condition: Option<MatchSpecCondition>,
+    /// The condition under which this match spec applies (disabled without conditional_dependencies feature).
+    #[cfg(not(feature = "conditional_dependencies"))]
```
baszalmstra: Does it make sense to keep this field here?

baszalmstra: No, I removed it.

wolfv:
I updated Cargo.lock and merged main :)

PR #1546, opened 2025-07-23
fix: indexing `extra-depends`

wolfv:
We currently do not properly index extras because we fill the field with a `default` hash map. This PR fixes it.


crates/rattler_conda_types/src/package/index.rs
```diff
@@ -35,6 +35,10 @@ pub struct IndexJson {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub depends: Vec<String>,
 
+    /// Extra dependency groups that can be selected using `foobar[extras=["scientific"]]`
+    #[serde(default, skip_serializing_if = "BTreeMap::is_empty")]
+    pub extras: BTreeMap<String, Vec<String>>,
```
wolfv: Let me change this to `extra_depends` for now.

baszalmstra: should we also mention the cep? and put this behind the same cargo flag?

PR #1551, opened 2025-07-23
fix: osx clobber directory creation bug

wolfv:
This fixes a critical issue in our new `clobber` directory logic.

On macOS we have an optimization where we `reflink` an entire directory (as it is a lot faster). However, we cannot reflink a directory that does not exist (which is the case for the `__clobber__` directory).

I added a `starts_with` check to make sure that we use the regular directory creation logic for clobber directories.

### DIFF
--- a/crates/rattler/src/install/clobber_registry.rs
+++ b/crates/rattler/src/install/clobber_registry.rs
@@ -12,7 +12,7 @@ use rattler_conda_types::{
     PackageName, PackageRecord, PrefixRecord,
 };
 
-const CLOBBERS_DIR_NAME: &str = "__clobbers__";
+pub const CLOBBERS_DIR_NAME: &str = "__clobbers__";
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ClobberedPath {
--- a/crates/rattler/src/install/mod.rs
+++ b/crates/rattler/src/install/mod.rs
@@ -69,7 +69,7 @@ pub use unlink::{empty_trash, unlink_package};
 
 pub use crate::install::entry_point::{get_windows_launcher, python_entry_point_template};
 use crate::install::{
-    clobber_registry::ClobberRegistry,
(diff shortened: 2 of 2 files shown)


crates/rattler/src/install/mod.rs
```diff
@@ -737,7 +737,11 @@ pub fn link_package_sync(
             continue;
         }
 
-        if allow_ref_links && cfg!(target_os = "macos") && !index_json.noarch.is_python() {
+        if allow_ref_links
+            && !directory.starts_with("__clobbers__")
+            && cfg!(target_os = "macos")
```
nichmor, line 741: should we put cfg!(target_os = "macos") first in the logical checks?

and maybe make __clobbers__ a const?

nichmor: maybe we can add a small test for this?

Hofer-Julian:
Could add a test for this? :)

crates/rattler/src/install/mod.rs
wolfv: I changed the test and confirmed that it would have found the issue by first running on main (using `link_package_sync` instead of `link_package`).

I also use the constant now.

The order of `cfg(macos)` should not matter. If the compiler finds a static "false" value it will definitely optimize out the entire `if`. I still moved it one place further to the front :)

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -12,7 +12,7 @@ use rattler_conda_types::{
     PackageName, PackageRecord, PrefixRecord,
 };
 
-const CLOBBERS_DIR_NAME: &str = "__clobbers__";
+pub const CLOBBERS_DIR_NAME: &str = "__clobbers__";
```
nichmor, line 15: maybe we can make it pub(crate)

nichmor: the clobber_registry is not `pub use ` so it's fine!

PR #1552, opened 2025-07-23
fix: allow variable expansion

fecet:
### Summary

This PR fixes an issue where environment variable values containing references like `${CONDA_PREFIX}` were not being expanded correctly in bash activation scripts per https://github.com/prefix-dev/pixi/issues/4122

### Problem

In recent versions, environment variable values with `$` symbols were quoted using single quotes. While this quoting ensures correctness for many shell values, it prevents variable expansion in bash:

```bash
export PATH='${CONDA_PREFIX}/bin:$PATH'  # Variables not expanded
```

This breaks expected behavior for users relying on dynamic paths such as `${CONDA_PREFIX}/bin`.

### Fix

Values containing variable references are now quoted using **double quotes** instead of single quotes in bash activation scripts. This preserves correct expansion while still protecting the value from word splitting or globbing.

```bash
export PATH="${CONDA_PREFIX}/bin:$PATH"  # Variables are expanded
```

### Impact

* Restores compatibility with tools relying on environment variable expansion (e.g., `pixi`)
* Affects only bash quoting behavior; other shells remain unchanged

---

Let me know if you'd like to include a test case or changelog entry as well.

### DIFF
--- a/crates/rattler_shell/src/shell/mod.rs
+++ b/crates/rattler_shell/src/shell/mod.rs
@@ -273,8 +273,18 @@ pub struct Bash;
 impl Shell for Bash {
     fn set_env_var(&self, f: &mut impl Write, env_var: &str, value: &str) -> ShellResult {
         validate_env_var_name(env_var)?;
-        let quoted_value = shlex::try_quote(value).unwrap_or_default();
-        Ok(writeln!(f, "export {env_var}={quoted_value}")?)
+
+        // Check if the value contains variable references ($)
+        // If so, use double quotes to allow variable expansion, otherwise use shlex quoting
+        if value.contains('$') {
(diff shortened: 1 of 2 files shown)


lucascolley:
Ahh, I wonder whether this explains something I was seeing with nushell and zsh also?? https://github.com/scipy/scipy/issues/23361#issuecomment-3109775680

fecet:
You can try my fork to test this https://github.com/fecet/pixi/releases/download/v0.50.0/pixi-x86_64-unknown-linux-musl

Hofer-Julian (approved):
Thanks for the fix, I tried to think of a more general approach, but it's just a bit of a messy situation and your solution is a good one :)

Pushed a commit to make the `shlex` error case a bit nicer

PR #1573, opened 2025-07-29
fix: improve hash mismatch warning to include package path

Hofer-Julian:
The error message gave no hint which package had a hash mismatch


wolfv (approved):

wolfv:
Did you run into this recently?

Hofer-Julian:
> Did you run into this recently?

Yes. 

However, after a discussion with @baszalmstra it wasn't the same problem that we pulled the latest Pixi release for.
If one builds a source dependency that is already cached, this warning always pops up.

PR #1575, opened 2025-07-31
feat: update `UploadOpts` and `get_auth_store`

magentaqin:
#### What changes I made
Let `UploadOpts` support `auth_store` field, which allows `pixi` to pass auth_store into `rattler_upload`. 

#### Background
In the original `get_auth_store` function, `rattler_networking::AuthenticationStorage::from_env_and_defaults()` only works in `rattler_build`, but in `pixi` it doesn't work. 

The reason behind this: 

1) auth_store logic is different in `pixi` and `rattler_build`:

In `pixi`, we can see it will read from Config. 
Screenshot 2025-07-31 at 11 35 12

But in `rattler_build`, it doesn't have this kind of logic. (`rattler_upload` crate is firstly migrated from `rattler_build`)
Screenshot 2025-07-31 at 11 41 08

2) the default key for `rattler_networking::AuthenticationStorage::from_env_and_defaults()` is hardcoded as "rattler".
Screenshot 2025-07-31 at 11 42 10

### DIFF
--- a/crates/rattler_upload/src/upload/opt.rs
+++ b/crates/rattler_upload/src/upload/opt.rs
@@ -3,13 +3,13 @@ use clap::{arg, Parser};
 use rattler_conda_types::utils::url_with_trailing_slash::UrlWithTrailingSlash;
 use rattler_conda_types::{NamedChannelOrUrl, Platform};
 use rattler_networking::mirror_middleware;
+use rattler_networking::{
+    authentication_storage::AuthenticationStorageError, AuthenticationStorage,
+};
 use rattler_solve::ChannelPriority;
 use std::{collections::HashMap, path::PathBuf, str::FromStr};
 use tracing::warn;
 use url::Url;
-use rattler_networking::{
-    authentication_storage::AuthenticationStorageError, AuthenticationStorage,
-};
 
 #[cfg(feature = "s3")]
 use rattler_networking::s3_middleware;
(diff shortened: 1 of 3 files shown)


crates/rattler_upload/src/upload/opt.rs
```diff
@@ -166,6 +169,19 @@ pub struct UploadOpts {
...
     pub common: CommonOpts,
+
+    #[clap(skip)]
+    pub auth_store: Option<Result<AuthenticationStorage, AuthenticationStorageError>>,
+}
+
+impl UploadOpts {
+    pub fn with_auth_store(
+        mut self,
+        auth_store: Result<AuthenticationStorage, AuthenticationStorageError>,
```
wolfv: I don't think this should be a `Result` here. The result / error should already have been handled by pixi or whoever calls this function.

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -166,6 +169,19 @@ pub struct UploadOpts {
     /// Common options.
     #[clap(flatten)]
     pub common: CommonOpts,
+
+    #[clap(skip)]
+    pub auth_store: Option<Result<AuthenticationStorage, AuthenticationStorageError>>,
```
wolfv: Same comment, this should not be a result. Just `AuthenticationStorage`.

crates/rattler_upload/src/upload/opt.rs
magentaqin: Agree!

crates/rattler_upload/src/upload/opt.rs
magentaqin: Solved!

wolfv (approved):

PR #1580, opened 2025-08-04
feat: add host-based upload support to `rattler_upload`

magentaqin:
### Description
This PR is to support uploading by host. It's based on my previous PR:  https://github.com/conda/rattler/pull/1575

### Why we need this host support?
`upload` function in `pixi` and `rattler_build` is different. In `pixi`, it only supports upload by url, but in `rattler_build`, it only supports upload by subcommands "prefix", "anaconda"...etc. We need to align these two.

### Code changes I made
1.support parsing url, match it to specific server type and extract base_url and channel from the host
2.align progress bar style: migrate `pixi_progress` and create `rattler_progress` that serves the uploading part.

### DIFF
--- a/crates/rattler_s3/src/lib.rs
+++ b/crates/rattler_s3/src/lib.rs
@@ -30,7 +30,7 @@ pub struct S3Credentials {
     pub endpoint_url: Url,
 
     /// The region of the S3 backend
-    pub region: Option<String>,
+    pub region: String,
 
     /// The addressing style to use for the bucket.
     #[cfg_attr(feature = "serde", serde(default))]
@@ -149,7 +149,7 @@ impl S3Credentials {
 
         Some(ResolvedS3Credentials {
             endpoint_url: self.endpoint_url,
-            region: self.region?,
+            region: self.region,
             access_key_id,
             secret_access_key,
             session_token,
--- a/crates/rattler_upload/src/lib.rs
+++ b/crates/rattler_upload/src/lib.rs
@@ -173,7 +173,6 @@ pub async fn upload_from_args(args: UploadOpts) -> miette::Result<()> {
                 s3_data.channel,
                 s3_data.credentials,
                 &args.package_files,
-                s3_data.region,
                 s3_data.force, // force parameter - using false as default
             )
             .await
--- a/crates/rattler_upload/src/upload/opt.rs
+++ b/crates/rattler_upload/src/upload/opt.rs
@@ -477,7 +477,11 @@ impl From<S3Opts> for S3Data {
                         .endpoint_url
                         .clone()
                         .expect("endpoint_url is required"),
-                    region: value.s3_credentials.region.clone(),
+                    region: value
+                        .s3_credentials
+                        .region
+                        .clone()
+                        .expect("region is required"),
                     addressing_style,
                     access_key_id: Some(access_key_id),
                     secret_access_key: Some(secret_access_key),
--- a/crates/rattler_s3/src/clap.rs
+++ b/crates/rattler_s3/src/clap.rs
@@ -71,7 +71,8 @@ impl From<S3CredentialsOpts> for Option<S3Credentials> {
             tracing::warn!("The `--force-path-style` option is deprecated, please use `--addressing-style=path` instead.");
             value.addressing_style = S3AddressingStyleOpts::Path;
         }
-        if let (Some(endpoint_url), Some(region)) = (value.endpoint_url, Some(value.region)) {
+
+        if let (Some(endpoint_url), Some(region)) = (value.endpoint_url, value.region) {
             Some(S3Credentials {
                 endpoint_url,
                 region,
--- a/crates/rattler_upload/src/upload/s3.rs
+++ b/crates/rattler_upload/src/upload/s3.rs
@@ -1,38 +1,30 @@
 use std::path::PathBuf;
 
-use indicatif::{MultiProgress, ProgressBar, ProgressStyle};
 use miette::IntoDiagnostic;
 use opendal::{services::S3Config, Configurator, ErrorKind, Operator};
+use rattler_digest::{HashingReader, Md5, Sha256};
 use rattler_networking::AuthenticationStorage;
 use rattler_s3::{ResolvedS3Credentials, S3Credentials};
+use tokio::io::{AsyncReadExt, AsyncSeekExt};
+use tokio_util::bytes::BytesMut;
 use url::Url;
 
 use crate::upload::package::ExtractedPackage;
 
-/// Uploads a package to a channel in an S3 bucket with progress tracking.
+const DESIRED_CHUNK_SIZE: usize = 1024 * 1024 * 10;
+
+/// Uploads a package to a channel in an S3 bucket.
 #[allow(clippy::too_many_arguments)]
 pub async fn upload_package_to_s3(
     auth_storage: &AuthenticationStorage,
     channel: Url,
     credentials: Option<S3Credentials>,
-    package_files: &[PathBuf],
-    region: Option<String>,
+    package_files: &Vec<PathBuf>,
     force: bool,
 ) -> miette::Result<()> {
     let bucket = channel
         .host_str()
         .ok_or(miette::miette!("No bucket in S3 URL"))?;
-
-    // Create progress tracking
-    let multi_progress = MultiProgress::new();
-    let main_progress = multi_progress.add(ProgressBar::new(package_files.len() as u64));
-    main_progress.set_style(
-        ProgressStyle::default_bar()
-            .template("{spinner:.green} [{elapsed_precise}] [{bar:40.cyan/blue}] {pos}/{len} {msg}")
-            .unwrap()
-            .progress_chars("█▉▊▋▌▍▎▏ "),
-    );
-    main_progress.set_message("Preparing upload...");
 
     // Create the S3 configuration for opendal.
     let mut s3_config = S3Config::default();
@@ -49,13 +41,8 @@ pub async fn upload_package_to_s3(
         }
     };
 
-    if let Some(r) = region {
-        s3_config.region = Some(r);
-    } else {
-        s3_config.region = Some(resolved_credentials.region);
-    }
-
     s3_config.endpoint = Some(resolved_credentials.endpoint_url.to_string());
+    s3_config.region = Some(resolved_credentials.region);
     s3_config.access_key_id = Some(resolved_credentials.access_key_id);
     s3_config.secret_access_key = Some(resolved_credentials.secret_access_key);
     s3_config.session_token = resolved_credentials.session_token;
@@ -65,9 +52,7 @@ pub async fn upload_package_to_s3(
     let builder = s3_config.into_builder();
     let op = Operator::new(builder).into_diagnostic()?.finish();
 
-    main_progress.set_message("Uploading packages...");
-
-    for (index, package_file) in package_files.iter().enumerate() {
+    for package_file in package_files {
         let package = ExtractedPackage::from_package_file(package_file)?;
         let subdir = package
             .subdir()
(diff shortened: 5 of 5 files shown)


crates/rattler_upload/src/utils/server_util.rs
```diff
@@ -0,0 +1,250 @@
...
+    artifactory: Regex,
+    #[cfg(feature = "s3")]
+    s3: Regex,
+    conda_forge: Regex,
+}
+
+impl ServerPatterns {
+    fn new() -> Self {
+        Self {
+            // Prefix.dev patterns
```
wolfv: I wonder if using `regex` is the best choice. We could probably also do this with regular URL parsing and some `contains` queries and similar, e.g.:

```
url.host_str.contains(prefix.dev)
url.path().ends_with(/conda-forge)
url.protocol() == "s3" ... 
```

crates/rattler_upload/src/utils/server_util.rs
```diff
@@ -0,0 +1,250 @@
...
+    } else {
+        Err("Invalid Quetz URL format".into())
+    }
+}
+
+// Extract Artifactory base_url and channel from host
+pub fn extract_artifactory_info(url: &Url) -> Result<(Url, String), Box<dyn std::error::Error>> {
+    let url_str = url.as_str();
+    let artifactory_pattern =
+        Regex::new(r"^(https?://[^/]+)/artifactory/([^/]+)(?:/(.+)/([^/]+))?").unwrap();
```
wolfv: I also think here just using regular URL parsing and then skipping components will be more robust than regex.

wolfv: For example:

```rust
let path_segments: Vec<&str> = url.path_segments()
    .ok_or("Invalid URL: cannot extract path segments")?
    .collect();

if path_segments.len() >= 2 && path_segments[0] == "artifactory" {
    // 1. Create base URL by removing path components
    let mut base_url = url.clone();
    base_url.set_path("");
    base_url.set_query(None);
    base_url.set_fragment(None);

    // 2. Extract channel
    let channel = path_segments[1].to_string();

    Ok((base_url, channel))
} else {
    Err("Invalid Artifactory URL format".into())
}
```

crates/rattler_upload/src/utils/server_util.rs
magentaqin: Good point! Indeed, Regex is fragile and has some maintainability issues...I'll switch to the URL parsing.

crates/rattler_upload/src/utils/server_util.rs
magentaqin: Got it!

magentaqin:
> Idk if a "generic" `rattler_progress` makes sense. The implementation is very specific. Id much rather have a slimmed down version of progress reporting in the rattler_upload code itself that can be enabled with a feature (we do the same for rattler install code).

Thanks @baszalmstra  for your advice! My original thought is not only for rattler_upload, but to align "progress" in `pixi` and `rattler-build`, so I directly moved the progress bar from `pixi` to `rattler`. Do you think it's necessary to have this `rattler_progress` crate?

crates/rattler_upload/src/lib.rs
```diff
@@ -29,8 +40,106 @@ pub async fn upload_from_args(args: UploadOpts) -> miette::Result<()> {
...
+                username: None,
+                password: None,
+                token: None,
+            })
+        }
+        SimpleServerType::Prefix => {
+            let host_url = args.host.as_ref().unwrap();
+            let (base_url, channel) =
+                extract_prefix_info(host_url).expect("Failed to parse Prefix URL");
+            ServerType::Prefix(PrefixOpts {
```
wolfv, line 92: Shouldn't we fill in e.g.. API key from the args here?

magentaqin: Hi, Wolf. There're three things I need to explain. 
1. According to pixi doc, upload doesn't have `api_key` arg. https://pixi.sh/latest/reference/cli/pixi/upload/#description 

2. The **common args** was designed like this(not having `api_key`) option.
Screenshot 2025-09-23 at 09 39 54

3. There're two ways to upload packages: URL and subcommand. 

Only when you specify the subcommand as "prefix", can you pass the arg `api_key`. But in our case, we upload package through URL(prefix pattern), `api_key` is not allowed to pass through the args. Users must pass the token by setting them locally.

crates/rattler_s3/src/lib.rs
```diff
@@ -23,14 +23,14 @@ pub enum S3AddressingStyle {
...
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct S3Credentials {
     /// The endpoint URL of the S3 backend
     pub endpoint_url: Url,
 
     /// The region of the S3 backend
-    pub region: String,
+    pub region: Option<String>,
```
baszalmstra: Why is this now optional?

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -426,19 +431,118 @@ fn parse_s3_url(value: &str) -> Result<Url, String> {
 #[cfg(feature = "s3")]
 #[derive(Clone, Debug, PartialEq, Parser)]
 pub struct S3Opts {
```
baszalmstra, line 433: You can reuse the S3CredentialsOpts from [crates/rattler_s3/src/clap.rs](https://github.com/conda/rattler/pull/1580/files#diff-4b9e488afedf21e4616031a0843fd61dba86f67044e972c199a428f4c6984e86) here to get a consistent CLI.

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -158,13 +160,16 @@ impl CommonData {
 /// Upload options.
 #[derive(Parser, Debug)]
 pub struct UploadOpts {
+    /// The host + channel (optional if the server type is provided)
+    pub host: Option<Url>,
```
baszalmstra, line 164: There are clap arguments here to make sure that if host is provide that channel must also be provide. Or that it conflicts with `server_type`.  Use those.

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -426,19 +431,118 @@ fn parse_s3_url(value: &str) -> Result<Url, String> {
...
+
+    /// The addressing style to use for the bucket.
+    #[arg(long, env = "S3_ADDRESSING_STYLE", default_value = "virtual-host")]
+    pub addressing_style: S3AddressingStyleOpts,
+}
+
+#[cfg(feature = "s3")]
+#[derive(Debug)]
+#[allow(missing_docs)]
+pub struct S3Data {
```
baszalmstra, line 454: What is the purpose of this struct? The name is quite ambigous. Also, doesnt the rattler_s3 crate provide similar functionality?

crates/rattler_upload/src/utils/server_util.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Determine server type from host URL
+///
+/// # Arguments
+/// * `host` - The host URL to analyze
+///
+/// # Returns
+/// * `SimpleServerType` - The detected server type or Unknown
+///
+/// ```
+pub fn check_server_type(host_url: &Url) -> SimpleServerType {
```
baszalmstra, line 25: Implement `From` or `SimpleServerType::from_url`.

crates/rattler_upload/src/utils/server_util.rs
```diff
@@ -0,0 +1,228 @@
...
+#[derive(Debug, Clone, PartialEq)]
+pub enum SimpleServerType {
+    Quetz,
+    Artifactory,
+    Prefix,
+    Anaconda,
+    #[cfg(feature = "s3")]
+    S3,
+    CondaForge,
+    Unknown,
```
baszalmstra, line 13: Cant we represent this as an `Option<SimpleServerType>`?

crates/rattler_upload/src/utils/server_util.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    // 2. Check Anaconda.org
+    if host.contains("anaconda.org") {
+        return SimpleServerType::Anaconda;
+    }
+
+    // 3. Check Conda-forge (GitHub)
+    if host.contains("conda-forge") {
+        return SimpleServerType::CondaForge;
```
baszalmstra, line 43: Is conda-forge a server?

crates/rattler_upload/src/utils/server_util.rs
baszalmstra, line 1: This file needs a lot of tests to verify that the SimpleServerType is correct for a number of URLs.

crates/rattler_upload/src/utils/server_util.rs
```diff
@@ -0,0 +1,228 @@
...
+/// Simplified server type without options
+#[derive(Debug, Clone, PartialEq)]
+pub enum SimpleServerType {
+    Quetz,
+    Artifactory,
+    Prefix,
+    Anaconda,
+    #[cfg(feature = "s3")]
+    S3,
+    CondaForge,
```
baszalmstra, line 12: What kind of server is this?

crates/rattler_upload/Cargo.toml
```diff
@@ -50,6 +50,7 @@ tokio = { workspace = true, features = [
   "rt-multi-thread",
   "process",
 ] }
+regex = "1.10"
```
baszalmstra: Is this used?

crates/rattler_s3/src/lib.rs
magentaqin: Because it supports upload by URL now, and the region can be written in the URL. For example, the user can upload like this " https://<bucket>.s3.<region>.amazonaws.com/my-channel"

crates/rattler_upload/src/upload/opt.rs
magentaqin: will do it! thanks for reminding!

crates/rattler_upload/Cargo.toml
magentaqin: I'll remove it. It was used before refactor.

crates/rattler_upload/src/utils/server_util.rs
magentaqin: It's originally from here: https://github.com/prefix-dev/rattler-build/blob/v0.44.0/src/upload/conda_forge.rs I just migrated it to rattler!

crates/rattler_upload/src/utils/server_util.rs
magentaqin: It's originally from here: https://github.com/prefix-dev/rattler-build/blob/v0.44.0/src/upload/conda_forge.rs I just migrated it to rattler! And for the URL part, I add the corresponding part.

crates/rattler_upload/src/upload/opt.rs
magentaqin: Hi, Bas! Nice catch! I mean the hostURL should consider "host" + "channel", and in `check_server_type`(https://github.com/magentaqin/rattler/blob/14a0e826791da57dbe5e02db24597b6cf5b8873a/crates/rattler_upload/src/utils/server_util.rs#L25), I will check whether the URL contains channel.
But you're right, I should use clap arguments to make "host" and "server_type" mutually exclusive.

crates/rattler_upload/src/upload/opt.rs
magentaqin: You're right, Bas. S3Data was migrated before we had `rattler_s3` crate. Let me see how to reuse the functionality!

wolfv:
Hey @magentaqin I'll take this PR over and get it merged! I'll remove some changes from the S3 upload because IMO progress bar shoudl go to a different PR.

magentaqin:
> Hey @magentaqin I'll take this PR over and get it merged! I'll remove some changes from the S3 upload because IMO progress bar shoudl go to a different PR.

Sounds good, thanks for taking it over! 🙌

baszalmstra:
Closing for now. @wolfv Or do you want to keep this open?

wolfv:
We have this in the publish code now. Can port it back later

PR #1596, opened 2025-08-12
feat: populate `requested_spec`

### LINKED ISSUE #1595: `rattler.install()` does not populate `requested_spec` in `conda-meta/*.json`
### Checklist

### What happened?

In the [given Quickstart example](https://conda.github.io/rattler/py-rattler/#quick-start), the requested specs are not annotated. `rattler.solve()` returns `RepoDataRecord()`s and `rattler.install()` converts them to `PrefixRecord()`s but `requested_spec` is not populated.

`RepoDataRecord()` does not know about this field, so I wonder how to solve this:

* Come up with a [`SolvedRecord`](https://github.com/conda/conda/pull/14821) subclass that augments `RepoDataRecord` with an optional `requested_spec` and have `rattler.solve()` populate and return that (all info is at hand).
* Add a new `requested_specs` argument to `rattler.install()` that will be used to populate the `requested_spec` field of the relevant records accordingly.

### Additional Context

_No response_

baszalmstra:
Fixes #1595

This PR attempts to add support for writing (and updating) the `requested_spec` field in `PrefixRecord`s when installing an environment.

This adds a breaking change where the `requested_spec` is changed from an `Option<String>` to an `Vec<String>`. I made this change because multiple matchspecs can be added for the same package. We could potentially also merge matchspecs but I dont think thats completely trivial. 

Another think we have to think about is whether the order of the matchspecs in the `requested_spec` is important. In theory their order could influence the solve but Im not sure whether or not that is important. At the moment the implementation retains the order of the matchspecs as specified in the input specs.

### DIFF
--- a/py-rattler/rattler/prefix/prefix_record.py
+++ b/py-rattler/rattler/prefix/prefix_record.py
@@ -180,8 +180,8 @@ class PrefixRecord(RepoDataRecord):
         >>> r = PrefixRecord.from_path(
         ...     "../test-data/conda-meta/requests-2.28.2-pyhd8ed1ab_0.json"
         ... )
-        >>> r.requested_spec
-        None
+        >>> r.requested_spec is None
+        True
         >>>
         ```
         """


crates/rattler/src/install/installer/mod.rs
```diff
@@ -680,3 +717,390 @@ async fn populate_cache(
...
+            // Clear the requested_spec if it's not in the mapping
+            let mut new_record = record.clone();
+            new_record.requested_spec = Vec::new();
+            updated_record = Some(new_record);
+        }
+
+        // Write the updated record back to disk if needed
+        if let Some(new_record) = updated_record {
+            let conda_meta_path = prefix.path().join("conda-meta");
+            let pkg_meta_path = format!(
```
wolfv: we should really make a function for this :)

crates/rattler/src/install/installer/mod.rs
```diff
@@ -680,3 +717,390 @@ async fn populate_cache(
...
+///
+/// This function takes existing records that weren't modified by the transaction
+/// and updates their `requested_spec` field based on the spec mapping or clears it if requested.
+/// The updated records are then written back to disk.
+fn update_existing_records(
+    existing_records: &[PrefixRecord],
+    spec_mapping: &HashMap<PackageName, Vec<String>>,
+    prefix: &Prefix,
+) -> Result<(), InstallerError> {
+    for record in existing_records {
```
wolfv: Should we use `rayon` to parallelize this loop?

wolfv:
I think I am mainly wondering: is `requested_spec` also a `list[str]` in Conda? Just want to make sure that we create "standards-compliant" JSON files here (and `requested_spec` sounds singular).

crates/rattler/src/install/installer/mod.rs
baszalmstra: Changed to use rayon!

baszalmstra:
> I think I am mainly wondering: is `requested_spec` also a `list[str]` in Conda? Just want to make sure that we create "standards-compliant" JSON files here (and `requested_spec` sounds singular).

@jaimergp and I discussed this in https://github.com/conda/rattler/issues/1595#issuecomment-3176710758 and decided the use a Vec. Note that it still serializes as string too if there is only one entry for compatibility.

Im not sure though if conda will be able to parse the record if the `requested_spec` becomes a list. Ill have to try.

wolfv:
We could also introduce a `requested_specs: [str]` variable, and fill the `requested_spec` with the first or last element of the second one.

baszalmstra:
That makes sense to me! @jaimergp thoughts? 

We could also maybe make a PR for conda to support a list?

baszalmstra:
I just tried this branch with conda to see if it works, it doesnt:

```shell
# Install an environment with conda in it.
$ cargo run --bin rattler -- create "conda" "conda>1" "conda<1000"

# Run conda in the environment 
$ conda run -p .prefix conda install python
...
    conda.auxlib.exceptions.ValidationError: Invalid value ['conda', 'conda >1', 'conda <1000'] for requested_spec
```

jaimergp:
We can patch this up in `conda` for sure. I would be ok with either solution:

- `requested_spec: str` + `requested_specs: list[str]`. No need to do anything in `conda` right away, but we should probably deprecate the singular form and prefer the plural one moving forward, which will require changes in existing clients.
- `requested_spec: str | list[str]`: Need to patch conda to avoid the issue observed above.

An early attempt at standardizing the `conda-meta/*.json` stuff is available at https://github.com/conda/ceps/pull/124/files#diff-bebfa3f1d70547b3695124927ba31659f76574696cb2ae65c971578b2069ff5fR152. In there you can see that `"None"` is the behavior I observed too, but it's really odd because it should either be absent or set to `null` or `""`, but not `"None"` (what if want to publish a package named None?!).

baszalmstra:
So maybe it would be best to go with the two fields approach for now and mark the singular field as deprecated. That way we stay compatible with with older versions of conda as well.

baszalmstra:
I changed the implementation, we now have two fields in `PrefixRecord`.

* `requested_spec` is just the original field. This field has been marked as deprecated.
* `requested_specs` is a new field of type `Vec<String>`.

I modified the `Installer` class to migrate the `requested_spec` to `requested_specs` and always update it if needed.

wolfv (approved):

PR #1599, opened 2025-08-15
feat: extend noarch key parser to support empty string

sevas:
### Description

This PR extends the noarchserde parser to accept empty strings.

### Summary of the problem addressed 
The reason is initially described in this issue: https://github.com/prefix-dev/pixi/issues/4344 

Essentially, package entries on conda channels may define `"noarch" : ""`, which are not decoded by rattler when e.g. doing a `pixi search pkg_name` . Searching with conda, mamba and micromamba on the same channel works fine. 

Since I observed this issue on private packages and repodata. I cannot share them or their recipe. I provided some test data that reproduces the issue. 

No package entry on conda-forge contain this field (I checked the full repodata.json), so presumably this did happened by having a faulty configuration in the recipe that was passed as-is to the package metadata by conda-build, at the time.

I say legacy because no package built in the last 4 years exhibits this issue. All packages with this field were published in 2019-2020. I could try to reproduce the issue that lead to that situation by building packages with an old version of conda-build to confirm, although this would take a bit more time. 

### Implementation
This PR contains changes as suggested by @baszalmstra on the linked ticket. However, I am not a rust expert, so do let me know if this needs to be reworked.


crates/rattler_conda_types/src/no_arch_type.rs
```diff
@@ -119,9 +120,11 @@ impl<'de> Deserialize<'de> for NoArchType {
     {
         #[derive(Clone, Debug, Deserialize)]
         #[serde(untagged)]
+        #[allow(clippy::enum_variant_names)]
```
sevas: this may be debatable. At this time clippy complains that all enum values are suffixed with `Format`

baszalmstra:
I took the liberty to cleanup the code slightly and add a comment explaining whats going on.

PR #1600, opened 2025-08-15
fix: fix login authentication

magentaqin:
### Description

This PR is to fix https://github.com/prefix-dev/pixi/issues/4152.


crates/rattler/src/cli/auth.rs
```diff
@@ -161,7 +178,50 @@ fn login(args: LoginArgs, storage: AuthenticationStorage) -> Result<(), Authenti
...
+            true => {
+                eprintln!("Error: Unauthorized or invalid token");
+                return Ok(());
+            }
+            false => {
+                // Store the authentication
+                storage.store(&host, &auth)?;
+                println!("✅ Token is valid. Storing it...");
+            }
+        }
```
baszalmstra: Lets move this into its own dedicated function.

crates/rattler/src/cli/auth.rs
```diff
@@ -161,7 +178,50 @@ fn login(args: LoginArgs, storage: AuthenticationStorage) -> Result<(), Authenti
...
+            Authentication::BearerToken(t) => t,
+            _ => return Err(AuthenticationCLIError::PrefixDevBadMethod),
+        };
+
+        // Validate whether the user exists
+        let client = Client::new();
+
+        // Allow override of API URL for testing
+        let url = std::env::var("PREFIX_DEV_API_URL")
+            .unwrap_or_else(|_| "https://prefix.dev/api/graphql".to_string());
```
baszalmstra: Can we construct this from the provided host? E.g. `https://{host}/api/graphql`?

crates/rattler/src/cli/auth.rs
```diff
@@ -161,7 +178,50 @@ fn login(args: LoginArgs, storage: AuthenticationStorage) -> Result<(), Authenti
...
+        let text = response.text()?;
+
+        // Parse JSON
+        let json: serde_json::Value = serde_json::from_str(&text)
+            .map_err(|e| AuthenticationCLIError::JsonParseError(e.to_string()))?;
+
+        match json["data"]["viewer"].is_null() {
+            true => {
+                eprintln!("Error: Unauthorized or invalid token");
+                return Ok(());
```
baszalmstra: Shouldnt we return an error in this case?

magentaqin:
🚀 All fixed! Have a re-check! @baszalmstra

crates/rattler/src/cli/auth.rs
```diff
@@ -186,45 +190,57 @@ fn login(args: LoginArgs, storage: AuthenticationStorage) -> Result<(), Authenti
...
+        .send()?;
+
+    let text = response.text()?;
+
+    // Parse JSON
+    let json: serde_json::Value = serde_json::from_str(&text)
+        .map_err(|e| AuthenticationCLIError::JsonParseError(e.to_string()))?;
+
+    // Return true if token is valid (viewer is not null), false otherwise
+    Ok(!json["data"]["viewer"].is_null())
```
wolfv: I think it would be nice to actually return `Username(String)` so that a user of this API could display: `Logged in as "foobar"`.

magentaqin: Will do!

baszalmstra:
@magentaqin Do you think it would be hard to expose this in the rattler binary as well?

magentaqin:
> @magentaqin Do you think it would be hard to expose this in the rattler binary as well?

Hi, @baszalmstra It's okay to expose, but I don't think it's necessary. `pixi` can already call it through "auth" subcommand.

baszalmstra:
I know but its easy to test auth if I can just test it with binaries in this repo.

magentaqin:
> I know but its easy to test auth if I can just test it with binaries in this repo.

oh, quite a good suggestion! I'll do that!

magentaqin:
@baszalmstra Hi, Bas! You could simply run `./build_and_run.sh` to test the binary functionality.

PR #1611, opened 2025-08-19
refactor: optimize path resolver

remimimimimi:
### Description

The optimized version is around 3x faster. Also adds benchmarks to make it easier to optimize later, although I think it would be hard to squeeze more performance.

### DIFF
--- a/crates/path_resolver/src/lib.rs
+++ b/crates/path_resolver/src/lib.rs
@@ -189,10 +189,10 @@ impl PathResolver {
             let mut has_conflict = false;
 
             // Single-pass approach: use peekable iterator to detect last component
-            let mut components = p.components().enumerate().peekable();
+            let mut components = p.components().peekable();
 
             // Navigate to the target node, checking for prefix conflicts along the way
-            while let Some((_i, component)) = components.next() {
+            while let Some(component) = components.next() {
                 let comp = component.as_os_str();
                 let is_last = components.peek().is_none();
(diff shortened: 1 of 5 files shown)


crates/path_resolver/src/lib.rs
```diff
@@ -3,18 +3,78 @@
...
 };
 
 use fs_err as fs;
+use fxhash::{FxHashMap as HashMap, FxHashSet as HashSet};
 use indexmap::IndexSet;
 use itertools::Itertools;
 
-/// Type to represent path owner.
-pub type PackageName = String;
+/// Type to represent path owner. Using Rc<String> to avoid cloning overhead while maintaining ownership semantics.
```
baszalmstra:
```suggestion
/// Type to represent path owner. Using Arc<String> to avoid cloning overhead while maintaining ownership semantics.
```

crates/path_resolver/src/lib.rs
```diff
@@ -3,18 +3,78 @@
...
+#[derive(Clone, Debug, PartialEq, Eq, Hash, PartialOrd, Ord)]
+pub struct PackageName(Arc<String>);
```
baszalmstra: Perhaps `Arc<str>` is even better?

baszalmstra (approved):

baszalmstra:
This look really really good! Thanks!

PR #1615, opened 2025-08-21
refactor: optimize transaction operations

remimimimimi:
### Description

Optimize `Transaction::from_current_and_desired` in case it is noop by manually parsing only "header" of `PrefixRecord` instead of parsing full `PrefixRecord`. If `Transaction` is not noop, we fallback to full parsing. 

This required reordering of fields of `PrefixRecord`, but that's not a breaking change. If manual parser is unable to find all fields before big ones, then it will parser file completely. 

To work directly either with `MinimalPrefixRecord` or `PrefixRecord` new type `InstallationResultRecord` is introduced which is simple enum. 

Another approach would be implement `SparsePrefixRecord`, which would parse fields on-demand, but there's quite a few design decisions to be made and it probably will be slower, so it is postponed.


crates/rattler/src/install/transaction.rs
```diff
@@ -46,6 +46,24 @@ pub enum TransactionOperation<Old, New> {
     Remove(Old),
 }
 
+impl<Old: Clone, New: Clone> TransactionOperation<&Old, &New> {
```
baszalmstra, line 55: Should we implement [`ToOwned`](https://doc.rust-lang.org/std/borrow/trait.ToOwned.html) instead?

crates/rattler_conda_types/src/minimal_prefix_record.rs
```diff
@@ -0,0 +1,298 @@
...
+use std::str::FromStr;
+
+use crate::{NoArchType, PackageName, PackageRecord, PrefixRecord, Version, VersionWithSource};
+use hex;
+use rattler_digest::{Md5Hash, Sha256Hash};
+
+/// A minimal version of `PrefixRecord` that only contains fields needed for transaction computation.
+/// This is much faster to parse than the full `PrefixRecord`.
+#[derive(Debug, Clone)]
+pub struct MinimalPrefixRecord {
```
baszalmstra, line 33: Looking at `Transaction::from_current_and_desired` you need waaaaay less fields! 

I think you only need:

- `name`, `version`, `build` but these can be derived from the name of the file. Because its `<name>-<version>-<build>.json`. 

```rust
// Filename is in the form of: <name>-<version>-<build>
let (build_string, version, name) = filename_without_ext.rsplitn(3, '-').next_tuple()?;
```

- `python_site_packages_path` only if the `name` is `"python"`.
- `sha256` 
- `md5` but only if there is no `sha256`
- `size` but only if there is no `sha256` and no `md5`.

baszalmstra: I think this warrants a specialized path for this particular case.

crates/rattler/src/install/transaction.rs
remimimimimi: I don't think it would be possible to implement

```rust
impl<Old: Clone, New:Clone > Borrow<TransactionOperation<&Old, &New>> for TransactionOperation<Old, New> {
    fn borrow(&self) -> &TransactionOperation<&Old, &New> {
        // ???
    }
}
```

since we would have to create a new object and then return a reference to it.

which is required to implement `ToOwned`.

crates/rattler/src/install/installer/mod.rs
```diff
@@ -361,27 +361,51 @@ impl Installer {
...
-            installed,
-            records.into_iter().collect::<Vec<_>>(),
-            self.reinstall_packages,
-            self.ignored_packages,
+        let desired_records: Vec<_> = records.into_iter().collect();
+        let mut transaction = Transaction::from_current_and_desired(
+            installed.iter(),
+            desired_records.iter(),
+            self.reinstall_packages.clone(),
+            self.ignored_packages.clone(),
```
baszalmstra: Can we pass this as a reference instead of cloning?

baszalmstra:
Did these last changes affect the performance? There is also a benchmark for loading these records in `crates/rattler_conda_types/benches/prefix_record_from_path.rs`. There you can also compare your implementation to loading just the PackageRecord from these files. I want to make sure this has a significant improvement over that because otherwise we could just do that because its simpler. :)

remimimimimi:
So the benchmark results are:
```
load_prefix_record_serially
                        time:   [6.0662 ms 6.0985 ms 6.1341 ms]
                        change: [+1.5327% +2.1690% +2.8411%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe

load_minimal_prefix_record_serially
                        time:   [2.5539 ms 2.5619 ms 2.5702 ms]
                        change: [−0.0458% +0.2866% +0.5686%] (p = 0.08 > 0.05)
                        No change in performance detected.

load_as_prefix_record   time:   [987.30 ns 989.50 ns 992.37 ns]
                        change: [−0.5248% −0.0532% +0.4301%] (p = 0.83 > 0.05)
                        No change in performance detected.
Found 23 outliers among 100 measurements (23.00%)
  2 (2.00%) high mild
  21 (21.00%) high severe

load_as_minimal_prefix_record
                        time:   [1.0047 µs 1.0109 µs 1.0174 µs]
                        change: [+0.7988% +1.3404% +1.8929%] (p = 0.00 < 0.05)
                        Change within noise threshold.

load_as_package_record  time:   [984.41 ns 988.57 ns 993.08 ns]
                        change: [−0.1965% +0.1378% +0.5429%] (p = 0.48 > 0.05)
                        No change in performance detected.
Found 19 outliers among 100 measurements (19.00%)
  4 (4.00%) high mild
  15 (15.00%) high severe

load_long_prefix_record time:   [730.62 µs 731.88 µs 733.35 µs]
                        change: [+0.2180% +0.4890% +0.7802%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 16 outliers among 100 measurements (16.00%)
  12 (12.00%) high mild
  4 (4.00%) high severe

load_long_minimal_prefix_record
                        time:   [287.09 µs 288.08 µs 289.15 µs]
                        change: [+1.9830% +2.3297% +2.7094%] (p = 0.00 < 0.05)
                        Performance has regressed.

load_long_package_record
                        time:   [185.93 µs 186.59 µs 187.35 µs]
                        change: [−0.6537% −0.3964% −0.1037%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 18 outliers among 100 measurements (18.00%)
  1 (1.00%) high mild
  17 (17.00%) high severe
```

And in pixi they are:

```
   ~/Projects/Mine/pixi on    main *2 ?2                                                            ✘ INT at   15:56:42
❯ hyperfine \                                                                     
    --prepare 'rm -f .pixi/envs/default/conda-meta/pixi' \
    --command-name 'pixi-minimal-record' \
    './pixi-minimal-record r echo "hello"' \
    --prepare 'rm -f .pixi/envs/default/conda-meta/pixi' \
    --command-name 'pixi-main' \
    './target/release/pixi r echo "hello"' \ 
    --warmup 3 \
    --min-runs 10
Benchmark 1: pixi-minimal-record
  Time (mean ± σ):     514.5 ms ±   3.7 ms    [User: 398.1 ms, System: 264.9 ms]
  Range (min … max):   509.3 ms … 519.5 ms    10 runs
 
Benchmark 2: pixi-main
  Time (mean ± σ):     654.3 ms ±   8.3 ms    [User: 509.4 ms, System: 292.0 ms]
  Range (min … max):   643.1 ms … 672.3 ms    10 runs
 
Summary
  pixi-minimal-record ran
    1.27 ± 0.02 times faster than pixi-main
```

baszalmstra:
But reading the benchmarks it seems that just loading as a `PackageRecord` actually performs better? Simply reading the PackageRecord is much simpler, so I wonder if the extra code to read using gjson is worth it.

I think you'll need to benchmark with a metadata file that is very large. This benchmark just takes microseconds. If you have an environment with Rust, there should be a *.json file of roughly 20 MBs.

Can you also add a `load_minimal_prefix_record_serially` but using parsing as `PackageRecord`?

baszalmstra:
Reading the PackageRecord simply uses serde. Have you tried using serde for your MinimalPackageRecord instead of gjson?

remimimimimi:
I've tried using serde for `MinimalPrefixRecord` and regression is around 40%.

So, here are results with gjson:

```
load_prefix_record_serially
                        time:   [30.464 ms 30.548 ms 30.631 ms]
                        change: [−0.3669% +0.0431% +0.4417%] (p = 0.85 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

load_minimal_prefix_record_serially
                        time:   [13.121 ms 13.141 ms 13.163 ms]
                        change: [+0.9238% +1.1798% +1.4569%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

load_package_record_serially
                        time:   [8.7965 ms 8.8408 ms 8.8902 ms]
                        change: [−0.0128% +0.5938% +1.1446%] (p = 0.04 < 0.05)
                        Change within noise threshold.
Found 11 outliers among 100 measurements (11.00%)
  11 (11.00%) high severe

load_as_prefix_record   time:   [1.0548 µs 1.0570 µs 1.0591 µs]
                        change: [−0.6574% −0.4392% −0.2027%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low severe
  3 (3.00%) low mild
  1 (1.00%) high mild

load_as_minimal_prefix_record
                        time:   [1.0724 µs 1.0746 µs 1.0767 µs]
                        change: [+0.8898% +1.0696% +1.2736%] (p = 0.00 < 0.05)
                        Change within noise threshold.

load_as_package_record  time:   [1.0495 µs 1.0520 µs 1.0545 µs]
                        change: [−0.3800% −0.0739% +0.2006%] (p = 0.61 > 0.05)
                        No change in performance detected.

load_long_prefix_record time:   [22.259 ms 22.306 ms 22.356 ms]
                        change: [−1.7617% −1.4193% −1.0801%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 10 outliers among 100 measurements (10.00%)
  10 (10.00%) high mild

load_long_minimal_prefix_record
                        time:   [8.7140 ms 8.7323 ms 8.7516 ms]
                        change: [−5.3794% −4.9184% −4.4269%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

load_long_package_record
                        time:   [5.3508 ms 5.3665 ms 5.3837 ms]
                        change: [−0.9978% −0.6685% −0.2692%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 8 outliers among 100 measurements (8.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
  4 (4.00%) high severe
```

And here are results with serde:

```
load_prefix_record_serially
                        time:   [31.859 ms 31.945 ms 32.034 ms]
                        change: [−2.0279% −1.5805% −1.1338%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

load_minimal_prefix_record_serially
                        time:   [18.123 ms 18.188 ms 18.270 ms]
                        change: [+37.862% +38.404% +39.076%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  2 (2.00%) high mild
  4 (4.00%) high severe

load_package_record_serially
                        time:   [8.3907 ms 8.4295 ms 8.4742 ms]
                        change: [−5.3520% −4.6523% −3.9326%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) high mild
  5 (5.00%) high severe

load_as_prefix_record   time:   [1.0620 µs 1.0656 µs 1.0689 µs]
                        change: [+0.2390% +0.4853% +0.7632%] (p = 0.00 < 0.05)
                        Change within noise threshold.

load_as_minimal_prefix_record
                        time:   [1.0573 µs 1.0605 µs 1.0636 µs]
                        change: [−1.5463% −1.2175% −0.8984%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe

load_as_package_record  time:   [1.0554 µs 1.0579 µs 1.0602 µs]
                        change: [−0.0200% +0.3011% +0.6353%] (p = 0.08 > 0.05)
                        No change in performance detected.
Found 5 outliers among 100 measurements (5.00%)
  2 (2.00%) low mild
  3 (3.00%) high mild

load_long_prefix_record time:   [23.452 ms 23.519 ms 23.586 ms]
                        change: [+5.0511% +5.4394% +5.8051%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) low mild
  1 (1.00%) high mild

load_long_minimal_prefix_record
                        time:   [12.552 ms 12.589 ms 12.629 ms]
                        change: [+43.630% +44.168% +44.709%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  3 (3.00%) high mild
  2 (2.00%) high severe

load_long_package_record
                        time:   [5.1495 ms 5.1612 ms 5.1731 ms]
                        change: [−4.1887% −3.8269% −3.4709%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  1 (1.00%) high mild
  1 (1.00%) high severe
```

crates/rattler_conda_types/Cargo.toml
```diff
@@ -54,6 +54,7 @@ indexmap = { workspace = true }
 dirs = { workspace = true }
 rayon = { workspace = true, optional = true }
 fs-err = { workspace = true }
+memmap2 = "0.9"
```
baszalmstra: Change to workspace dependency.

crates/rattler_conda_types/src/minimal_prefix_record.rs
```diff
@@ -0,0 +1,936 @@
...
+            track_features: Vec::new(),
+            python_site_packages_path: None,
+            experimental_extra_depends: std::collections::BTreeMap::new(),
+            legacy_bz2_md5: None,
+        }
+    }
+
+    /// Fast parsing using nom with memory mapping (64KB limit).
+    /// This is potentially much faster than the gjson-based parser for large files.
+    pub fn fast_from_path(path: &Path) -> Result<Self, io::Error> {
```
baszalmstra: What is the difference between this function and from_path?

crates/rattler_conda_types/src/minimal_prefix_record.rs
```diff
@@ -0,0 +1,936 @@
...
+                &mmap[..buffer_size]
+            } else {
+                &mmap[..] // Use entire file if smaller than buffer
+            };
+
+            let content = std::str::from_utf8(content_slice)
+                .map_err(|e| io::Error::other(format!("Invalid UTF-8: {e}")))?;
+
+            // Try to parse with current buffer size
+            match parse_minimal_json(content) {
```
baszalmstra: Nom can also function on bytes instead of str. Since only the content of strings can contain utf8 characters, maybe we can just parse using bytes? Then this akward loop is not required at all!

crates/rattler_conda_types/src/minimal_prefix_record.rs
```diff
@@ -0,0 +1,936 @@
...
+        if input.starts_with('}') {
+            return Ok((input, ()));
+        }
+
+        // Parse field name
+        let (rest, field_name) = parse_json_string(input)?;
+        input = rest;
+
+        // Check if we hit the "files" field - stop parsing here
+        if field_name == "files" {
```
baszalmstra: What happens if we are parsing an "old" record where the ordering is not changed? I think we should also cover that case. 

Perhaps checking if we already have the records we need would be safer.  I think the only optional field is the requested-specs field, perhaps we can just emit none or an empty array to ensure the field always exists?

crates/rattler_conda_types/src/minimal_prefix_record.rs
```diff
@@ -0,0 +1,936 @@
...
+        let (rest, _) = multispace0(input)?;
+        let (rest, _) = opt(char(',')).parse(rest)?;
+        let (rest, _) = multispace0(rest)?;
+        input = rest;
+    }
+}
+
+/// Parse JSON string value (handles escaping)
+fn parse_json_string(input: &str) -> IResult<&str, &str> {
+    delimited(char('"'), take_until("\""), char('"')).parse(input)
```
baszalmstra: Does this really handle escaping? I dont think it properly handles this case does it? `"\""`

crates/rattler_conda_types/src/minimal_prefix_record.rs
```diff
@@ -0,0 +1,936 @@
...
+        assert_eq!(gjson_result.requested_spec, None);
+        assert_eq!(fast_result.requested_spec, None);
+
+        // Requested specs should be empty
+        assert!(gjson_result.requested_specs.is_empty());
+        assert!(fast_result.requested_specs.is_empty());
+    }
+
+    #[test]
+    fn test_fast_parser_handles_legacy_requested_spec() {
```
baszalmstra: Maybe its better to test correctness against loading a PrefixRecord with serde.

baszalmstra (changes requested):
I think this would benefit from a few more edge case tests. What happens with invalid json, string escaping, etc.

crates/rattler_conda_types/src/minimal_prefix_record.rs
remimimimimi: Yes, I wanted to do this. Will do that!

crates/rattler_conda_types/src/minimal_prefix_record.rs
remimimimimi: Yes, this is incorrect behavior. I didn't implement the correct implementation, as it is quite complex. But will do that now.

crates/rattler_conda_types/src/minimal_prefix_record.rs
remimimimimi: No difference, just forgot to remove old `fast_from_path`.

crates/rattler_conda_types/src/minimal_prefix_record.rs
remimimimimi: That's the main question that I have. Now, if we parse "old" record, it should just put default values to `requested_specs` and `requested_spec` if it can't find them before `"files"`.

What we could do is to don't skip serialization of `requested_specs`. Then if we encounter it before `"files"` then we're sure that we have "new" record and so can stop parsing. If we don't see it, then we would have to parse rest of the file, which should still be faster than the current file. What do you think?

crates/rattler_conda_types/src/minimal_prefix_record.rs
remimimimimi: Good suggestion. Will do that.

remimimimimi:
Currently

1. I no longer use chunks of memory for parsing; this was my misunderstanding of how memory maps work.
2. Code is backward compatible. I changed PrefixRecord to serialize `requested_specs` always, so we can use it as a marker of the new version. If we can't find it before `"files"` field, then we parse till the end of the file, so it will work correctly in "outdated" environments. Even in this case, new parsing is 10 times faster that what we had before with `serde_json`.
3. I have tested new parser against `serde_json` and now it works correctly for all metadata files in `test-data/conda-meta`.
4. String unescaping is handled by `serde_json`.

I also don't think that we should be worried about proper error handling in it, as we expect all metadata files to be valid, since they are serialized by `serde_json`. But if even an erroneous case occurs, we will have a very modest error message.

crates/rattler_conda_types/src/prefix_record.rs
```diff
@@ -183,6 +183,20 @@ pub struct PrefixRecord {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub extracted_package_dir: Option<PathBuf>,
 
+    /// The spec that was used when this package was installed. Note that this
+    /// field is not updated if the currently another spec was used. Note:
+    /// conda seems to serialize a "None" string value instead of `null`.
+    ///
+    /// This field is deprecated. Use `requested_specs` instead.
+    #[deprecated(note = "Use `requested_specs` instead")]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
```
baszalmstra: Should we just remove the `skip_serializing_if` part here? Or do you use `requested_specs` as a marker?

remimimimimi: I use `requested_specs` as a marker

baszalmstra: In the tests the `requested_spec` is often `""`. Should we use `None` in those cases?

baszalmstra:
And would you be able to add a few more benchmark results of your latest improvements? Im mostly interested in: 

* How fast is serializing a complete set of *.json files when you deserialize them in parallel with `PrefixRecord`.
* How fast is that when deserializing them as `PackageRecord`s.
* How fast is your new approach when the layout of the json fields is still the "old" way.
* How fast is your new approach when the layout is also optimized.

baszalmstra:
Ideally, you benchmark this against the conda-meta files that include both small and very large files.

remimimimimi:
> And would you be able to add a few more benchmark results of your latest improvements? Im mostly interested in:
> 
>     * How fast is serializing a complete set of *.json files when you deserialize them in parallel with `PrefixRecord`.
> 
>     * How fast is that when deserializing them as `PackageRecord`s.
> 
>     * How fast is your new approach when the layout of the json fields is still the "old" way.
> 
>     * How fast is your new approach when the layout is also optimized.

I didn't really understand what you meant in the first two bullet points. But here are new benchmark results.

```
Serial Processing/PrefixRecord
                        time:   [32.080 ms 32.156 ms 32.233 ms]
                        change: [+0.6442% +1.0071% +1.3387%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Serial Processing/PackageRecord
                        time:   [8.6779 ms 8.7036 ms 8.7322 ms]
                        change: [+1.4258% +1.7703% +2.1722%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
Serial Processing/MinimalPrefixRecord
                        time:   [15.648 ms 15.706 ms 15.767 ms]
                        change: [+1.7209% +2.2364% +2.7141%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  3 (3.00%) high mild
Serial Processing/MinimalPrefixRecord new conda meta
                        time:   [365.44 µs 367.29 µs 369.07 µs]
                        change: [−2.9073% −2.2560% −1.5914%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) low mild
  2 (2.00%) high mild

Parallel Processing/PrefixRecord
                        time:   [26.937 ms 27.018 ms 27.105 ms]
                        change: [−0.3723% +0.0341% +0.4293%] (p = 0.88 > 0.05)
                        No change in performance detected.
Found 6 outliers among 100 measurements (6.00%)
  5 (5.00%) high mild
  1 (1.00%) high severe
Parallel Processing/PackageRecord
                        time:   [7.2178 ms 7.2585 ms 7.3013 ms]
                        change: [−0.6333% +0.1974% +0.9499%] (p = 0.65 > 0.05)
                        No change in performance detected.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
Parallel Processing/MinimalPrefixRecord
                        time:   [13.262 ms 13.337 ms 13.413 ms]
                        change: [+1.5392% +2.2419% +2.9316%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
Parallel Processing/MinimalPrefixRecord new conda meta
                        time:   [673.43 µs 679.37 µs 685.72 µs]
                        change: [+5.6996% +6.3861% +7.0771%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

Single File/PrefixRecord
                        time:   [25.771 ms 25.833 ms 25.900 ms]
                        change: [−0.1461% +0.2948% +0.7366%] (p = 0.20 > 0.05)
                        No change in performance detected.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high severe
Single File/PackageRecord
                        time:   [6.9277 ms 6.9539 ms 6.9827 ms]
                        change: [+1.0090% +1.4653% +1.9503%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  4 (4.00%) high mild
  1 (1.00%) high severe
Single File/MinimalPrefixRecord
                        time:   [12.625 ms 12.701 ms 12.780 ms]
                        change: [+2.4006% +3.2000% +4.0199%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild

Large File/PrefixRecord time:   [23.647 ms 23.744 ms 23.858 ms]
                        change: [−1.3644% −0.8764% −0.3271%] (p = 0.00 < 0.05)
                        Change within noise threshold.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high severe
Large File/PackageRecord
                        time:   [5.2343 ms 5.2611 ms 5.2912 ms]
                        change: [−3.4955% −2.8273% −2.1557%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 100 measurements (8.00%)
  6 (6.00%) high mild
  2 (2.00%) high severe
Large File/MinimalPrefixRecord
                        time:   [11.260 µs 11.321 µs 11.387 µs]
                        change: [+0.7341% +1.4493% +2.1814%] (p = 0.00 < 0.05)
                        Change within noise threshold.
```

I use the new format for 'Large file'. The old format is used everywhere if not stated otherwise.

remimimimimi:
> This look good to me! The low-level parser is really awesome!
> 
> One thing that would make this fantastic is if we would would not go from `MinimalRecord` -> `PackageRecord` but instead have a transaction of type `Transaction<MinimalRecord, ..>`. I think the fact that there is a comment now in transaction that says, if you change this logic you also have to change MinimalRecord is icky. Ideally we solve this completely with types instead.

Fixed this by introducing a new trait that describes all required fields.

baszalmstra:
The results looks very good, but I still wonder why serde with packagerecord in the old format is still twice as fast! Any idea?

Let me check the code one last time and then I would be happy to merge!

crates/rattler/src/install/installer/mod.rs
```diff
@@ -361,27 +361,51 @@ impl Installer {
         // Create a future to determine the currently installed packages. We
         // can start this in parallel with the other operations and resolve it
         // when we need it.
-        let installed: Vec<PrefixRecord> = if let Some(installed) = self.installed {
+        let installed_provided = self.installed.is_some();
+        let mut installed: Vec<PrefixRecord> = if let Some(installed) = self.installed {
```
baszalmstra: If I read this correctly we are still converting the minimalrecords to prefixrecords. So this part is still not typed.

Can you look into making this part typed as well? Where we have a transaction of Transaction<minimalprefixrecords, repodatarecors> ?

remimimimimi:
> The results looks very good, but I still wonder why serde with packagerecord in the old format is still twice as fast! Any idea?
> 
> Let me check the code one last time and then I would be happy to merge!

The only idea that I have is that for parsing `MinimalPrefixRecord` we use a memory map, which reads page by page, which could cause inefficient work with the disk.

crates/rattler/src/install/installer/mod.rs
remimimimimi: Yes, we're still converting, but now we're linking `PrefixRecord` with `MinimalPrefixRecord` fields using `ContentComparable`, so problem about which I wrote that note won't occur. I would love to express that we having `MinimalPrefixRecord` on type level, but can't think about easy way to do this. Code just below expects type of transaction to be with `PrefixRecord`. Proper implementation will be hard to understand, or require modification of a lot of code.

baszalmstra: I think you should implement two code paths, one where the prefixrecords are already provided and done where they arent.

crates/rattler_conda_types/src/minimal_prefix_record.rs
```diff
@@ -0,0 +1,876 @@
...
+        //             extracted_package_dir: None,
+        //             files: Vec::new(),
+        //             paths_data: crate::prefix_record::PrefixPaths::default(),
+        //             requested_spec: minimal.requested_spec.clone(),
+        //             requested_specs: minimal.requested_specs.clone(),
+        //             link: None,
+        //             installed_system_menus: Vec::new(),
+        //         }
+        //     })
+        //     .collect())
```
Hofer-Julian: Not sure if that commented out code is worth keeping around

crates/rattler/src/install/installer/mod.rs
```diff
@@ -48,6 +50,243 @@ pub struct LinkOptions {
...
+        python_info,
+        current_python_info,
+        platform,
+        unchanged,
+    })
+}
+
+/// Convert transaction to contain `InstallationResultRecord` instead of `PrefixRecord`.
+pub fn installation_result_record_transaction(
+    t: Transaction<PrefixRecord, RepoDataRecord>,
```
Hofer-Julian:
```suggestion
    transaction: Transaction<PrefixRecord, RepoDataRecord>,
```

And similar for other cases

crates/rattler/src/install/installer/mod.rs
```diff
@@ -48,6 +50,243 @@ pub struct LinkOptions {
     pub allow_ref_links: Option<bool>,
 }
 
+/// Special kind of record that either can be minimal or full.
```
baszalmstra: This could use some clarification. Without knowledge of the "optimization" it is unclear to the user what this entails.

crates/rattler/src/install/installer/mod.rs
```diff
@@ -48,6 +50,243 @@ pub struct LinkOptions {
...
+                    name = minimal_prefix_record.name.as_normalized()
+                );
+                let record_path = prefix.as_ref().join(record_name);
+                PrefixRecord::from_path(record_path)
+            }
+        }
+    }
+
+    /// Return reference to the underlying `requested_spec`.
+    pub fn requested_spec(&self) -> Option<&String> {
```
baszalmstra: Maybe we can make these functions pub crate? Since you dont provide accesors for other fields. Or what was your design decision?

crates/rattler/src/install/installer/mod.rs
```diff
@@ -48,6 +50,243 @@ pub struct LinkOptions {
...
+            }
+            InstallationResultRecord::Min(minimal_prefix_record) => {
+                minimal_prefix_record.python_site_packages_path()
+            }
+        }
+    }
+}
+
+/// Convert transaction to contain `PrefixRecord` instead of `InstallationResultRecord`.
+pub fn prefix_record_transaction(
```
baszalmstra: Maybe we can make this a function on Transaction itself?

crates/rattler/src/install/installer/mod.rs
```diff
@@ -48,6 +50,243 @@ pub struct LinkOptions {
...
+    /// Full record.
+    Max(PrefixRecord),
+    /// Minimal record.
+    Min(MinimalPrefixRecord),
+}
+
+#[allow(deprecated)]
+impl InstallationResultRecord {
+    /// Either just returns stored `PrefixRecord` or parses `PrefixRecord` from the given prefix.
+    pub fn prefix_record(self, prefix: impl AsRef<Path>) -> Result<PrefixRecord, io::Error> {
```
baszalmstra: Lets call this `into_prefix_record`

crates/rattler/src/install/installer/mod.rs
remimimimimi: Other fields are accessible through the `ContentComparable` trait.

baszalmstra: Is that trait exposed? It seems like an implementation detail to me.

remimimimimi: Now it should be better @baszalmstra. I made methods of `ContentComparable` part of the `InstallationResultRecord` and then used them to implement this trait. Also, marked mutable methods as `pub(crate)`, since they don't make much sense outside of the rattler.

crates/rattler/src/install/installer/result_record.rs
```diff
@@ -0,0 +1,337 @@
...
+                &mut minimal_prefix_record.requested_specs
+            }
+        }
+    }
+}
+
+/// Trait that defines the fields needed for package content comparison.
+/// This ensures type safety - if `describe_same_content` needs a new field,
+/// it must be added here and implemented for all types.
+pub trait ContentComparable {
```
baszalmstra, line 156: Shall we make this pub crate then too?

remimimimimi: Currently 
image
If we will make it `pub(crate)`, then we would not be able to use this method, since it will require to change this method visibility to `pub(crate)` as well. If we want to leave it, I would rather export `ContentComparable` to be able to see which types implement it. So what should I do? @baszalmstra

baszalmstra: Yeah makes sense, thanks for clarifying!

baszalmstra: Just leave pub 🙂

PR #1627, opened 2025-08-26
feat: derive default credentials from aws sdk

baszalmstra:
This PR adds support for getting S3 credentials directly from the AWS SDK. This adds support for using all the different ways you can sign in with AWS, e.g. keys, IAM Identity Center workforce, etc.

The goal of this PR is to enable only stating an s3 url (e.g. `s3://my-awesome-channel`) and let rattler figure out how to authenticate. The order of getting credentials is, 1) from the CLI (or environment variables), 2) from the default rattler authentication provider, 3) from the AWS SDK. This PR effectively adds option 3.

1. `rattler_networking` already supported this behavior but I found a few cases where error messages where very unclear. Especially around using invalid regions. I improved the error messages.
2. `rattler_index` required making some CLI arguments optional and using the AWS SDK to derive these values. When the system is properly configured `rattler-index s3 s3://my-awesome-channel` should "just work". 
3. `rattler_upload` required making the same changes. I also added a few preconditions to not upload shards if they already exist, and I added some other small changes. The goal is that `rattler upload s3 --channel s3://my-awesome-channel <packages>` should just work.

I still want to add an e2e test where we set up a minio instance, upload a package, index the channel and create an environment from the channel without any additional configuration.


baszalmstra:
Recreated in https://github.com/conda/rattler/pull/1629 . I cant acquire OIDC credentials through a fork.

PR #1628, opened 2025-08-26
fix(menuinst): improve linux

Hofer-Julian:
This fixes a couple of problems with our linux menuinst implementation:

- `bash -c` is only need if pre-command is used to chain with `&&`, I dropped it since it makes it harder for launchers to parse `%F` and similar
- The pre-command is supposed to be used *before* the environment is activated, so I fixed that
- fix quoting
- clean up code a bit

I tested it locally with `pixi global` and Zed and it works for me. Also added a couple of tests.

### DIFF
--- a/crates/rattler_menuinst/src/linux.rs
+++ b/crates/rattler_menuinst/src/linux.rs
@@ -661,7 +661,7 @@ mod tests {
     };
     use tempfile::TempDir;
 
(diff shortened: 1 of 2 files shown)


Hofer-Julian:
@lucascolley

lucascolley (approved):
changes seem good to me!

PR #1629, opened 2025-08-26
feat: derive default credentials from aws sdk

baszalmstra:
This PR adds support for getting S3 credentials directly from the AWS SDK. This adds support for using all the different ways you can sign in with AWS, e.g. keys, IAM Identity Center workforce, etc.

The goal of this PR is to enable only stating an s3 url (e.g. `s3://my-awesome-channel`) and let rattler figure out how to authenticate. The order of getting credentials is, 1) from the CLI (or environment variables), 2) from the default rattler authentication provider, 3) from the AWS SDK. This PR effectively adds option 3.

1. `rattler_networking` already supported this behavior but I found a few cases where error messages where very unclear. Especially around using invalid regions. I improved the error messages.
2. `rattler_index` required making some CLI arguments optional and using the AWS SDK to derive these values. When the system is properly configured `rattler-index s3 s3://my-awesome-channel` should "just work". 
3. `rattler_upload` required making the same changes. I also added a few preconditions to not upload shards if they already exist, and I added some other small changes. The goal is that `rattler upload s3 --channel s3://my-awesome-channel <packages>` should just work.

I still want to add an e2e test where we set up a minio instance, upload a package, index the channel and create an environment from the channel without any additional configuration.

Duplicate of #1627, created this one to allow creating a PR on this repository to allow for OIDC to work.

### DIFF
--- a/crates/rattler_s3/src/lib.rs
+++ b/crates/rattler_s3/src/lib.rs
@@ -117,9 +117,9 @@ impl ResolvedS3Credentials {
         Ok(Self {
             endpoint_url,
             region,
+            addressing_style,
             access_key_id,
             secret_access_key,
-            addressing_style,
             session_token,
         })
     }


.github/workflows/e2e-s3-tests.yml
```diff
@@ -0,0 +1,218 @@
...
+          BIN_DIR="$RUNNER_TEMP"
+          DATA_DIR="$RUNNER_TEMP/minio-data"
+          LOG_FILE="$RUNNER_TEMP/minio.log"
+          PID_FILE="$RUNNER_TEMP/minio.pid"
+          
+          mkdir -p "$BIN_DIR" "$DATA_DIR"
+          
+          # Download MinIO server and client without sudo; add to PATH
+          curl -fsSL -o "$BIN_DIR/minio" https://dl.min.io/server/minio/release/linux-amd64/minio
+          curl -fsSL -o "$BIN_DIR/mc" https://dl.min.io/client/mc/release/linux-amd64/mc
```
pavelzw: technically we could also add minio to pixi.toml as a dev dependency which might help for reproducibility. no strong opinion though

.github/workflows/e2e-s3-tests.yml
```diff
@@ -0,0 +1,218 @@
...
+        uses: aws-actions/configure-aws-credentials@7474bc4690e29a8392af63c5b98e7449536d5c3a  # v4.3.1
+        with:
+          aws-region: ${{ env.AWS_REGION }}
+          role-to-assume: arn:aws:iam::239378270001:role/conda-rattler-e2e-test
+
+      - name: Generate unique bucket name
+        id: bucket
+        run: |
+          # Create a unique bucket name using repository owner, name, run ID, and attempt number
+          aws s3api create-bucket \
```
pavelzw: is this version of awscli coming from `aws-actions/configure-aws-credentials` or `ubuntu-latest`? Might be nice to install it via setup-pixi and use `activate-environment: true`

Cargo.toml
```diff
@@ -72,17 +73,8 @@ generic-array = "0.14.7"
...
-  "sso",
-  "credentials-process",
-] }
-aws-sdk-s3 = { version = "1.85.0", default-features = false, features = [
-  "rt-tokio",
-  "rustls",
-  "sigv4a",
-] }
+aws-config = { version = "=1.5.18", default-features = false }
+aws-sdk-s3 = { version = "1.85.0", default-features = false }
```
pavelzw, line 77: why was the `sso` feature removed? Does it still work with the `aws sso login` workflow described in https://pixi.sh/latest/deployment/s3/#using-aws-configuration?

is it not necessary anymore because we use the feature downstream in rattler-networking etc?

py-rattler/rattler/index/index.py
```diff
@@ -46,8 +46,8 @@ async def index_fs(
 
 async def index_s3(
     channel_url: str,
-    region: str,
-    endpoint_url: str,
+    region: Optional[str] = None,
+    endpoint_url: Optional[str] = None,
     force_path_style: bool = False,
```
pavelzw: shouldn't this be `bool | None` as well?

py-rattler/rattler/index/index.py
```diff
@@ -46,8 +46,8 @@ async def index_fs(
 
 async def index_s3(
     channel_url: str,
-    region: str,
-    endpoint_url: str,
+    region: Optional[str] = None,
+    endpoint_url: Optional[str] = None,
```
pavelzw:
```suggestion
    region: str | None = None,
    endpoint_url: str | None = None,
```

or is the minimally supported python version _that_ old?

crates/rattler_index/src/lib.rs
```diff
@@ -521,7 +525,19 @@ pub async fn write_repodata(
             let future = async move || {
                 let shard_path = format!("{subdir}/shards/{digest:x}.msgpack.zst");
                 tracing::trace!("Writing repodata shard to {shard_path}");
-                op.write(&shard_path, encoded_shard).await
+                match op
+                    .write_with(&shard_path, encoded_shard)
+                    .if_not_exists(true)
+                    .cache_control("public, max-age=31536000, immutable")
```
pavelzw: should we put this into a const?

crates/rattler_index/src/lib.rs
```diff
@@ -594,17 +611,25 @@ pub async fn index_fs(
...
+    /// detected from the endpoint URL.
+    pub region: Option<String>,
+    /// The endpoint URL of the S3 bucket. If not set:
+    /// - The endpoint will be loaded from `AWS_ENDPOINT_URL` environment
+    ///   variable.
+    /// - If that is not set, the default endpoint will be used (`https://s3.amazonaws.com`).
+    pub endpoint_url: Option<Url>,
     /// Whether to force path style for the S3 bucket.
-    pub force_path_style: bool,
+    pub force_path_style: Option<bool>,
```
pavelzw: so far, we did the following in rattler:

either: provide everything at once, no implicit defaults
or: let aws-sdk handle everything

this is also done so in pixi's configuration and i think this makes most sense as otherwise you might get some weird behavior when only specifying one of the three options.
wdyt of putting all of this in a struct which can then be made optional? This would then match the behavior in pixi's configuration

crates/rattler_index/src/lib.rs
```diff
@@ -594,17 +611,25 @@ pub async fn index_fs(
...
     /// The access key ID for the S3 bucket.
-    /// If not set, the authentication storage will be queried.
+    /// If not set, the access key will be loaded from the environment, if the
+    /// access key was still not loaded, the authentication storage will be
+    /// queried.
```
pavelzw: we should specify what "environment" means. it also contains stuff like sso credentials store

pavelzw: so far, we only read from the "aws environment" (i.e. env vars plus things like sso, sts) when we were in "aws mode", i.e. not specifying region, endpoint-url and force-path-style manually.
are you sure we want to change this behavior? i find it a bit weird that like this, it could happen that you read the credentials for your bucket from some sso store while you originally want to read from something that has nothing to do with AWS S3 like cloudflare R2 or minio

pavelzw: this doesn't match the implementation below. below is first function args, then authentication storage and then aws defaults

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -111,20 +121,17 @@ impl S3 {
...
+                ) => aws_sdk_s3::config::Builder::from(sdk_config)
+                    .endpoint_url(endpoint_url)
+                    .region(aws_sdk_s3::config::Region::new(region))
+                    .force_path_style(force_path_style)
+                    .credentials_provider(aws_sdk_s3::config::Credentials::new(
+                        access_key_id,
+                        secret_access_key,
+                        session_token,
+                        None,
+                        "pixi",
```
pavelzw: what was `provider_name` used for again? Shouldn't it be `rattler` instead?

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -135,12 +142,15 @@ impl S3 {
...
+
             // Infer if we expect path-style addressing from the endpoint URL.
             if let Some(endpoint_url) = sdk_config.endpoint_url() {
-                // If the endpoint URL is localhost, we probably have to use path-style addressing.
-                // xref: https://github.com/awslabs/aws-sdk-rust/issues/1230
+                // If the endpoint URL is localhost, we probably have to use path-style
+                // addressing. xref: https://github.com/awslabs/aws-sdk-rust/issues/1230
                 if endpoint_url.starts_with("http://localhost") {
                     s3_config_builder = s3_config_builder.force_path_style(true);
                 }
```
pavelzw, line 157: this issue is basically marked as not planned. i think we just added this for some integration tests iirc. we should probably get rid of this

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -166,19 +178,31 @@ impl S3 {
...
-        .map_err(reqwest_middleware::Error::middleware)
+                .context("failed to presign S3 HEAD request")?,
+            Method::POST => client
+                .put_object()
+                .bucket(bucket_name)
+                .key(key)
+                .presigned(presign_config)
+                .await
+                .context("failed to presign S3 PUT request")?,
+            _ => client
```
pavelzw: why is this not `Method::GET` and then on `_` we panic with not implemented?

crates/rattler_repodata_gateway/src/fetch/with_cache.rs
```diff
@@ -432,6 +432,17 @@ pub async fn fetch_repo_data(
             });
         }
 
+        // Fail if the status code is not a success
+        if !response.status().is_success() {
+            return Err(FetchRepoDataError::HttpError(
+                reqwest_middleware::Error::Middleware(anyhow::format_err!(
+                    "received unexpected status code ({}) when fetching {}",
```
pavelzw: i'm wondering whether we should also add the response text. this would have been helpful in some scenarios for me

crates/rattler_repodata_gateway/src/fetch/with_cache.rs
```diff
@@ -432,6 +432,17 @@ pub async fn fetch_repo_data(
...
+                    response.status(),
+                    repo_data_url.redact(),
```
pavelzw, line 443: do you have an example on how this would look like?

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/tokio/index.rs
```diff
@@ -45,6 +46,16 @@ pub async fn fetch_index(
         permit: Option<tokio::sync::SemaphorePermit<'_>>,
     ) -> Result<ShardedRepodata, GatewayError> {
         let response = response.error_for_status()?;
+        if !response.status().is_success() {
+            let mut url = response.url().clone().redact();
+            url.set_query(None);
+            url.set_fragment(None);
+            return Err(GatewayError::ReqwestMiddlewareError(anyhow::format_err!(
+                "received unexpected status code ({}) when fetching {}",
```
pavelzw: s.a.

crates/rattler_upload/src/upload/s3.rs
```diff
@@ -0,0 +1,143 @@
...
+use crate::upload::package::ExtractedPackage;
+
+/// Uploads a package to a channel in an S3 bucket.
+#[allow(clippy::too_many_arguments)]
+pub async fn upload_package_to_s3(
+    auth_storage: &AuthenticationStorage,
+    channel: Url,
+    endpoint_url: Option<Url>,
+    region: Option<String>,
+    force_path_style: Option<bool>,
```
pavelzw: same comment about merging into `Option` struct as above

crates/rattler_upload/src/upload/s3.rs
```diff
@@ -0,0 +1,143 @@
...
+        };
+        let credentials: Credentials = credentials_provider
+            .provide_credentials()
+            .await
+            .into_diagnostic()
+            .context("failed to determine AWS credentials")?;
+        s3_config.access_key_id = Some(credentials.access_key_id().to_string());
+        s3_config.secret_access_key = Some(credentials.secret_access_key().to_string());
+        s3_config.session_token = credentials.session_token().map(ToString::to_string);
+    }
```
pavelzw: this is basically the same between rattler-index and rattler-upload. in pixi-pack i would also like to do the same and can imagine this might be useful for conda-mirror as well. how about moving this shared code in a separate crate?

crates/rattler_upload/src/upload/s3.rs
```diff
@@ -0,0 +1,143 @@
...
+#[allow(clippy::too_many_arguments)]
+pub async fn upload_package_to_s3(
+    auth_storage: &AuthenticationStorage,
+    channel: Url,
+    endpoint_url: Option<Url>,
+    region: Option<String>,
+    force_path_style: Option<bool>,
+    access_key_id: Option<String>,
+    secret_access_key: Option<String>,
+    session_token: Option<String>,
```
pavelzw: same comments about these three as in rattler-index

pavelzw (commented):
thanks for the work! some comments

.github/workflows/e2e-s3-tests.yml
pavelzw, line 1: while you're at it, what about adding cloudflare r2 integration tests as well in this file? @delsner and i definitely wrote some at some point, not sure whether in this repo or in pixi/rattler-build only.

crates/rattler_index/src/lib.rs
baszalmstra: I like that a lot, Ill set it up that way!

crates/rattler_networking/src/s3_middleware.rs
baszalmstra: Good idea lets change that.

crates/rattler_repodata_gateway/src/fetch/with_cache.rs
baszalmstra: The response text in the s3 case will be an xml file. Ill add it to message!

crates/rattler_repodata_gateway/src/fetch/with_cache.rs
baszalmstra: It happens when you specify the wrong region for a bucket. Youll get a redirect notice without the headers that will actually redirect you. Ill add an example tomorrow.

crates/rattler_upload/src/upload/s3.rs
baszalmstra: Ill add a crate for this. `rattler-s3`?

py-rattler/rattler/index/index.py
baszalmstra: The minimum version is 3.8.

Cargo.toml
baszalmstra: I like to keep features in the crates that use the dependencies to ensure you only build the features that are actually used.

I think I added sso to all the individual crates. Or did I miss one?

pavelzw: Looks good

crates/rattler_upload/src/upload/s3.rs
pavelzw: Sounds good 👍🏻

.github/workflows/e2e-s3-tests.yml
delsner: Yes, we could either use the cf build of `minio-client` and `minio-server` or create a `docker-compose.yml` file with `minio` included so that users could simply run `docker compose up minio --wait` to be able to run the tests locally.

.github/workflows/e2e-s3-tests.yml
delsner: If we decide to use `minio-client` here, we could also use [`mc mb`](https://docs.min.io/community/minio-object-store/reference/minio-mc/mc-mb.html) to create the bucket.

.github/workflows/e2e-s3-tests.yml
delsner: I think it was in this repo: https://github.com/conda/rattler/blob/b3eff3438a28e39923c011414dff17320f4f8132/crates/rattler_networking/tests/s3_integration_test.rs#L18-L39

baszalmstra:
@pavelzw Not done yet, but I streamlined the use of s3 arguments, and made the e2e test run fully with pixi. Please have a look if you have time. 

Im not planning on adding cloudflare in this same PR. We can probably do that in another PR.

crates/rattler_networking/src/s3_middleware.rs
baszalmstra: Gonna leave it in place, I dont feel this PR should change that.

crates/rattler_networking/src/s3_middleware.rs
baszalmstra: Add a `provider_name` to the `S3` object which allows configuring this. By default its `rattler`.

crates/rattler_repodata_gateway/src/fetch/with_cache.rs
baszalmstra: I added the body and this is now the result:

```
received unexpected status code (301 Moved Permanently) when fetching https://bas-rattler-s3-test.s3.eu-west-
1.amazonaws.com/win-64/repodata_shards.msgpack.zst.

Body:
<?xml version="1.0" encoding="UTF-8"?>
<Error><Code>PermanentRedirect</Code><Message>The bucket you are attempting to access must be addressed using
the specified endpoint. Please send all future requests to this endpoint.</Message><Endpoint>bas-rattler-s3-
test.s3.eu-central-1.amazonaws.com</Endpoint><Bucket>bas-rattler-s3-test</Bucket><RequestId>T0SMBENQDSDGPX0T</
RequestId><HostId>EU92iaHh4guioVWmjQjonZcC6i6KDeNuglOrA4Sho3v2hDT8d1lD2+csJjMjU/Db5cy8RHNA4/5x0oMaC0F3tQ==</
HostId></Error>
```

Not great, but a lot better than nothing!

.github/workflows/e2e-s3-tests.yml
```diff
@@ -0,0 +1,86 @@
...
+        with:
+          submodules: recursive
+
+      - uses: prefix-dev/setup-pixi@fef5c9568ca6c4ff7707bf840ab0692ba3f08293 # v0.9.0
+        with:
+          environments: minio
+
+      - run: pixi run e2e-s3-minio
+
+  e2e-aws-s3-test:
```
pavelzw, line 49:
```suggestion
  # TODO: add cloudflare R2 integration tests here as well
  e2e-aws-s3-test:
```

crates/rattler_s3/src/clap.rs
```diff
@@ -0,0 +1,46 @@
...
+/// See [`super::S3Credentials`] for details on how these credentials are used.
+#[derive(Clone, Debug, PartialEq, Parser)]
+pub struct S3CredentialsOpts {
+    /// The endpoint URL of the S3 backend
+    #[arg(long, env = "S3_ENDPOINT_URL", requires_all = ["region"], help_heading = "S3 Credentials")]
+    pub endpoint_url: Option<Url>,
+
+    /// The region of the S3 backend
+    #[arg(long, env = "S3_REGION", requires_all = ["endpoint_url"], help_heading = "S3 Credentials")]
+    pub region: Option<String>,
```
pavelzw, line 45: shouldn't we add force-path-style here as well?

crates/rattler_s3/src/lib.rs
```diff
@@ -0,0 +1,154 @@
...
+            {
+                (access_key_id, secret_access_key, self.session_token)
+            } else if let Some((access_key_id, secret_access_key, session_token)) =
+                load_s3_credentials_from_auth_storage(auth_storage, bucket_url.clone())
+            {
+                // Use the credentials from the authentication storage if they are available.
+                (access_key_id, secret_access_key, session_token)
+            } else {
+                return None;
+            };
```
pavelzw, line 148: is this the order we want to commit to? I personally would prefer first using rattler's native auth storage and then checking aws credentials (with proper logging what happens ideally).

crates/rattler_index/src/lib.rs
```diff
@@ -594,21 +610,10 @@ pub async fn index_fs(
...
-    /// The access key ID for the S3 bucket.
-    /// If not set, the authentication storage will be queried.
-    pub access_key_id: Option<String>,
-    /// The secret access key for the S3 bucket.
-    /// If not set, the authentication storage will be queried.
-    pub secret_access_key: Option<String>,
-    /// The session token for the S3 bucket.
-    /// If not set, the authentication storage will be queried.
-    pub session_token: Option<String>,
+    pub force_path_style: Option<bool>,
```
pavelzw: should belong in resolveds3credentials imo

crates/rattler_index/src/main.rs
```diff
@@ -81,33 +84,17 @@ enum Commands {
...
-
-        /// The region of the S3 backend
-        #[arg(long, env = "S3_REGION")]
-        region: Option<String>,
+        #[clap(flatten)]
+        credentials: rattler_s3::clap::S3CredentialsOpts,
 
         /// Whether to use path-style S3 URLs
         #[arg(long, env = "S3_FORCE_PATH_STYLE")]
         force_path_style: Option<bool>,
```
pavelzw: s.a.

crates/rattler_index/src/main.rs
```diff
@@ -81,33 +84,17 @@ enum Commands {
...
 
-        /// The endpoint URL of the S3 backend
-        #[arg(long, env = "S3_ENDPOINT_URL")]
-        endpoint_url: Option<Url>,
-
-        /// The region of the S3 backend
-        #[arg(long, env = "S3_REGION")]
-        region: Option<String>,
+        #[clap(flatten)]
+        credentials: rattler_s3::clap::S3CredentialsOpts,
```
pavelzw, line 88: the CLI is still sort-of the same as before, right?

crates/rattler_networking/src/s3_middleware.rs
```diff
@@ -80,15 +84,30 @@ impl S3 {
             config,
             auth_storage,
             expiration: std::time::Duration::from_secs(300),
+            default_client: Arc::new(OnceCell::new()),
+            provider_name: "rattler",
         }
     }
+
+    /// Sets the provider name that is used by the s3 client.
```
pavelzw: maybe add a small comment on what this does? it's not immediate to me what this is used for

crates/rattler_networking/src/s3_middleware.rs
pavelzw: #1648

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -422,37 +424,21 @@ fn parse_s3_url(value: &str) -> Result<Url, String> {
...
-    #[arg(long, env = "S3_REGION", default_value = "eu-central-1")]
-    pub region: String,
+    #[clap(flatten)]
+    pub credentials: rattler_s3::clap::S3CredentialsOpts,
 
     /// Whether to use path-style S3 URLs
-    #[arg(long, env = "S3_FORCE_PATH_STYLE", default_value = "false")]
-    pub force_path_style: bool,
+    #[arg(long, env = "S3_FORCE_PATH_STYLE")]
+    pub force_path_style: Option<bool>,
```
pavelzw: s.a.

py-rattler/rattler/index/index.py
```diff
@@ -1,12 +1,33 @@
...
+from dataclasses import dataclass
 import os
 from typing import Optional
 
 from rattler.platform import Platform
 from rattler.rattler import py_index_fs, py_index_s3
 
 
+@dataclass
+class S3Credentials:
```
pavelzw, line 12: we should add force-path-style here, s.a.

py-rattler/rattler/index/index.py
```diff
@@ -46,12 +67,8 @@ async def index_fs(
...
 async def index_s3(
     channel_url: str,
-    region: str,
-    endpoint_url: str,
-    force_path_style: bool = False,
-    access_key_id: Optional[str] = None,
-    secret_access_key: Optional[str] = None,
-    session_token: Optional[str] = None,
+    credentials: Optional[S3Credentials] = None,
+    force_path_style: Optional[bool] = False,
```
pavelzw: s.a.

py-rattler/src/index.rs
```diff
@@ -41,35 +44,44 @@ pub fn py_index_fs(
...
+    py: Python<'py>,
     channel_url: String,
-    region: String,
-    endpoint_url: String,
-    force_path_style: bool,
-    access_key_id: Option<String>,
-    secret_access_key: Option<String>,
-    session_token: Option<String>,
+    credentials: Option<Bound<'py, PyAny>>,
+    force_path_style: Option<bool>,
```
pavelzw: s.a.

pavelzw (commented):
cool! mostly looks good but i think we should move `force-path-style` in s3credentials as well

crates/rattler_s3/src/lib.rs
baszalmstra: The order is now:

If anything is manually specified, everything must be manually specified. When you add the endpoint or region or access_key yourself you opt out of any implicit behavior except for getting the access key id from rattler's native authentication storage.

If you don't specify anything (either as arguments (e.g. `credentials=None`), or CLI options), everything will be take from the AWS sdk.

baszalmstra:
I left the `force_path_style` out because you can also use this option _together_ with the AWS SDK. E.g. you can configure your AWS SDK credentials to target a minio server which requires `force_path_style`.

crates/rattler_index/src/main.rs
baszalmstra: Yes, except that when you specify any of these arguments, you must specify them all (except for the access key). E.g.

```
❯ rattler-index s3 --endpoint-url http://localhost s3://my-channel
error: the following required arguments were not provided:
  --region <REGION>
  
❯ rattler-index s3 --access-key-id http://localhost s3://my-channel
error: the following required arguments were not provided:
  --endpoint-url <ENDPOINT_URL>
  --region <REGION>
  --secret-access-key <SECRET_ACCESS_KEY>
```

baszalmstra:
And this is not exposed through the AWS SDK: https://github.com/awslabs/aws-sdk-rust/issues/1230

pavelzw:
> I left the force_path_style out because you can also use this option together with the AWS SDK. E.g. you can configure your AWS SDK credentials to target a minio server which requires force_path_style.
> And this is not exposed through the AWS SDK: https://github.com/awslabs/aws-sdk-rust/issues/1230

but is this really a use case we want to support? the main benefit of using the aws-sdk style is that it just works in combination with things like aws sso, aws sts, oidc in gha, awscli, etc.

i don't see a reason why one would want to configure minio through aws variables as it's not related to aws products at all and you don't gain any of the before-mentioned benefits. also, that it's not exposed through aws sdk is in my opinion a bug/issue there where we shouldn't build weird workarounds in our API for unpopular use cases.

~also, people would still be able to use env variables `S3_REGION`, `S3_FORCE_PATH_STYLE`, ... to configure their rattler-based tools if we move them to `S3CredentialsOpts`. i intentionally switched the `AWS_*` with `S3_*` variables as that has nothing to do with aws anymore~ just noticed that this is currently not the case in pixi but only rattler upload and rattler index

baszalmstra:
Ok, that makes sense. Ill move it into the credentials.

crates/rattler_networking/src/s3_middleware.rs
baszalmstra: I think it was only used to identify where certain credentials come from. I hardcoded this from rattler and removed the ability to configure it.

crates/rattler_s3/src/lib.rs
pavelzw: yes, i just misunderstood this code. this is exactly as i imagine it

crates/rattler_s3/src/lib.rs
```diff
@@ -0,0 +1,179 @@
...
+    pub access_key_id: String,
+
+    /// The secret access key for the S3 bucket.
+    pub secret_access_key: String,
+
+    /// How to address the S3 buckets.
+    pub addressing_style: S3AddressingStyle,
+
+    /// The session token for the S3 bucket.
+    pub session_token: Option<String>,
```
pavelzw:
```suggestion
    /// The endpoint URL of the S3 backend
    pub endpoint_url: Url,

    /// The region of the S3 backend
    pub region: String,

    /// How to address the S3 buckets.
    pub addressing_style: S3AddressingStyle,

    /// The access key ID for the S3 bucket.
    pub access_key_id: String,

    /// The secret access key for the S3 bucket.
    pub secret_access_key: String,

    /// The session token for the S3 bucket.
    pub session_token: Option<String>,
```

PR #1650, opened 2025-09-03
feat: lock package build source

remimimimimi:
### Description

We need this to make pixi build packages reproducible.


crates/rattler_lock/src/conda.rs
```diff
@@ -369,3 +428,18 @@ fn merge_package_record<'a>(
...
+    if left == right {
+        Cow::Borrowed(left)
+    } else if let Some(right_source) = right {
+        // New data takes precedence
+        Cow::Owned(Some(right_source.clone()))
+    } else {
+        // Right is None, keep left unchanged
+        Cow::Borrowed(left)
+    }
+}
```
remimimimimi, line 445: Does this even makes sense to do?

baszalmstra: 🤷 Lets just keep it. :)

crates/rattler_lock/src/parse/models/v6/conda_package_data.rs
```diff
@@ -113,6 +113,10 @@ pub(crate) struct CondaPackageDataModel<'a> {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub input: Option<InputHash<'a>>,
 
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde_as(as = "Option<PackageBuildSourceSerializer>")]
+    pub package_build_source: Option<PackageBuildSource>,
```
baszalmstra, line 118: Maybe we can think of a better name for this. Just `source` or `build_source` perhaps? We know its already a package.

crates/rattler_lock/src/conda.rs
```diff
@@ -140,6 +140,55 @@ impl CondaBinaryData {
...
+/// This allows detecting when the source specification has changed (e.g., from
+/// branch `main` to branch `dev`) even if the current commit hash is the same.
+/// Without this information, we wouldn't know if the lock file needs to be
+/// updated when the requested branch/tag changes.
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+pub enum GitShallowSpec {
+    /// A git branch reference (e.g., "main", "develop")
+    Branch(String),
+    /// A git tag reference (e.g., "v1.0.0", "release-2023")
+    Tag(String),
```
baszalmstra, line 154: There is also git `ref`.

baszalmstra (commented):
A few small nits.

Can you explain whether this change is backwards and forward compatible?

crates/rattler_lock/src/parse/models/v6/conda_package_data.rs
remimimimimi: I first used `source`, then used `build_source`, and then migrated to `package_build_source` to avoid confusion with `sources`. I also thought about naming it `pixi_build_source`, but didn't want to make it pixi build specific.

crates/rattler_lock/src/conda.rs
remimimimimi: Check out the doc comment of this enum. This was deliberate because 'PackageBuildSource' stores revision.

remimimimimi:
> A few small nits.
> 
> Can you explain whether this change is backwards and forward compatible?

It is backwards compatible because the field that we added is optional to parse and to store. It is forward compatible because the parser will just ignore unknown fields. I just confirmed this by commenting out code I wrote and rerunning new tests.

PR #1656, opened 2025-09-04
fix: update correct desktop database when installing packages with shortcuts

Hofer-Julian:
Implement it on rattler as discussed in https://github.com/prefix-dev/pixi/pull/4510

I went with a much simpler approach in the end.
We already called `update-desktop-database` on all search paths on every menuinst install which takes about `~64 ms`. Unfortunately the local application dir isn't in the search path (at least on Fedora).

This changes this to only search in our application dir. This takes only `~5 ms` on my machine and actually works.

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -5351,54 +5351,54 @@ packages:
     - Cargo.toml
 - conda: crates/rattler_index
   name: rattler_index
-  version: 0.24.12
+  version: 0.25.0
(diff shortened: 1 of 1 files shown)


wolfv:
BTW the rattler executable exposes a cli command to install/remove a menu entry.

baszalmstra (approved):

PR #1674, opened 2025-09-10
chore: Replace fxhash with rustc-hash

haecker-felix:
The `fxhash` crate is no longer maintained. See https://rustsec.org/advisories/RUSTSEC-2025-0057.html

The `rustc-hash` is pretty much a drop-in replacement for `fxhash`. It has the same type names, but uses a different hashing algorithm under the hood:

> The original hash algorithm provided by this crate was one taken from Firefox, hence the hasher it provides is called FxHasher. This name is kept for backwards compatibility, but the underlying hash has since been replaced. The current design for the hasher is a polynomial hash finished with a single bit rotation, together with a wyhash-inspired compression function for strings/slices, both designed by Orson Peters.


baszalmstra:
If we are switching the hash algorithm we could also use ahash instead!

haecker-felix:
Now using the standard library’s `HashMap` and `HashSet` with `ahash::RandomState`.

Unlike `FxHashMap` and `FxHashSet`, the standard library versions no longer guarantee deterministic ordering of elements. I had to update some tests to account for this. 

Not sure whether there may be other unintended side effects that the tests haven’t caught...

crates/rattler_conda_types/src/repo_data/topological_sort.rs
```diff
@@ -27,11 +30,16 @@ pub fn sort_topologically<T: AsRef<PackageRecord> + Clone>(packages: Vec<T>) ->
...
-fn find_all_cycles<T: AsRef<PackageRecord>>(packages: &FxHashMap<String, T>) -> Vec<Vec<String>> {
+fn find_all_cycles<T: AsRef<PackageRecord>>(
+    packages: &HashMap<String, T, RandomState>,
+) -> Vec<Vec<String>> {
     let mut all_cycles = Vec::new();
-    let mut visited = FxHashSet::default();
+    let mut visited = HashSet::default();
+
+    let mut package_names: Vec<_> = packages.keys().collect();
+    package_names.sort();
```
haecker-felix, line 36: Another thing I had to change because of the no longer deterministic ordering of the `HashSet`.

crates/path_resolver/src/lib.rs
```diff
@@ -99,11 +99,11 @@ pub struct ClobberedPath {
 #[derive(Default, Debug, Clone, PartialEq, Eq)]
 struct PathTrieNode {
     /// All tags that touch this prefix *or* any descendant.
-    prefixes: HashSet<PackageName>,
+    prefixes: HashSet<PackageName, RandomState>,
```
baszalmstra: I think we should either use: 

```suggestion
    prefixes: HashSet<PackageName, ahash::RandomState>,
```

or preferably:

```suggestion
    prefixes: ahash::HashSet<PackageName>,
```

That makes it clear that we are using `ahash`.

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -51,16 +51,16 @@ pub struct RepoData {
         rename = "packages.conda",
         serialize_with = "sort_map_alphabetically"
     )]
-    pub conda_packages: FxHashMap<String, PackageRecord>,
+    pub conda_packages: HashMap<String, PackageRecord, RandomState>,
```
baszalmstra: If the order matters, which it does in this case, I think we should use a different map. We can also use an `OrderMap<String, PackageRecord, ahash::RandomState>`.

baszalmstra: We can also remove the `serialize_with` part if the insertion order is kept.

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -280,6 +280,8 @@ impl RepoData {
                 file_name: filename,
             });
         }
+        // Sort to keep the resulting list deterministic across hash implementations.
+        records.sort_by(|a, b| a.file_name.cmp(&b.file_name));
```
baszalmstra: This should not be needed.

crates/path_resolver/src/lib.rs
haecker-felix: Done

crates/rattler_conda_types/src/repo_data/mod.rs
haecker-felix: I assumed you mean `IndexMap` - done.

crates/rattler_conda_types/src/repo_data/mod.rs
haecker-felix: Yup, not needed anymore with `IndexMap`. Dropped.

crates/rattler_conda_types/src/repo_data/mod.rs
baszalmstra: `IndexMap` is not hashable but if the top level struct is not hashable thats fine.

haecker-felix: So it should be fine in this case, as `RepoData` isn't hashable.

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -40,27 +42,27 @@ pub struct RepoData {
     pub info: Option<ChannelInfo>,
 
     /// The tar.bz2 packages contained in the repodata.json file
-    #[serde(default, serialize_with = "sort_map_alphabetically")]
-    pub packages: FxHashMap<String, PackageRecord>,
+    #[serde(default, serialize_with = "sort_indexmap_alphabetically")]
```
baszalmstra:
```suggestion
    #[serde(default)]
```

We can remove the serialize_with code since the insertion order is what we want.

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -40,27 +42,27 @@ pub struct RepoData {
...
+    pub packages: IndexMap<String, PackageRecord, ahash::RandomState>,
 
     /// The conda packages contained in the repodata.json file (under a
     /// different key for backwards compatibility with previous conda
     /// versions)
     #[serde(
         default,
         rename = "packages.conda",
-        serialize_with = "sort_map_alphabetically"
+        serialize_with = "sort_indexmap_alphabetically"
```
baszalmstra:
```suggestion
```

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -430,14 +432,14 @@ pub struct SubdirRunExportsJson {
     info: Option<ChannelInfo>,
 
     #[serde(default, serialize_with = "sort_map_alphabetically")]
```
baszalmstra:
```suggestion
    #[serde(default)]
```

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -430,14 +432,14 @@ pub struct SubdirRunExportsJson {
...
-    packages: FxHashMap<String, PackageRunExports>,
+    packages: ahash::HashMap<String, PackageRunExports>,
 
     #[serde(
         default,
         rename = "packages.conda",
         serialize_with = "sort_map_alphabetically"
```
baszalmstra:
```suggestion
```

crates/rattler_conda_types/src/repo_data/topological_sort.rs
baszalmstra: I prefer we use `IndexMap` for these types to make sure they are deterministic instead of reordering which is more expensive.

crates/rattler_lock/src/lib.rs
```diff
@@ -219,18 +218,18 @@ impl LockFile {
...
-                (
-                    name.as_str(),
-                    Environment {
-                        lock_file: self,
-                        index: *index,
-                    },
-                )
-            })
+        let mut environments: Vec<_> = self.inner.environment_lookup.iter().collect();
+        environments.sort_by(|(a, _), (b, _)| a.cmp(b));
```
baszalmstra: This is not strictly needed for the function to work correctly. Rather it is something we want in the test. I think we should modify the tests instead.

baszalmstra (changes requested):

crates/rattler_conda_types/src/repo_data/topological_sort.rs
haecker-felix:
```
FAIL [   0.040s] rattler_conda_types repo_data::topological_sort::tests::test_topological_sort::case_6
  stdout ───

    running 1 test
    test repo_data::topological_sort::tests::test_topological_sort::case_6 ... FAILED

    failures:

    failures:
        repo_data::topological_sort::tests::test_topological_sort::case_6

    test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 324 filtered out; finished in 0.03s
    
  stderr ───

    thread 'repo_data::topological_sort::tests::test_topological_sort::case_6' panicked at crates/rattler_conda_types/src/repo_data/topological_sort.rs:292:21:
    attempting to install panel (package 336 of 339) but dependency holoviews is not yet installed
    note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

@baszalmstra Even with `IndexMap` / `IndexSet` that particular test keeps failing if I don't sort the package names manually. 

Could it be possible, that there's a bug which didn't get triggered with `FxHashMap` as it had a different random ordering - but what happened to be accidentally  the correct order for the test to be successful?

Hofer-Julian:
@baszalmstra I think it would be good if you could quickly glance at Felix' comment

@haecker-felix could you please solve the merge conflicts?

haecker-felix:
Rebased, merge conflicts resolved, and tests are passing now (windows test is an unrelated network error, should be fine when running again). 

Looks like my issue in https://github.com/conda/rattler/pull/1674#discussion_r2405989183 has been resolved by another change through rebasing.

baszalmstra:
Ah Im sorry I didnt didnt find the time to properly comment here. I was still noodling what was bothering me and why we now sometimes need to resort the entries. 

The current PR changes the functionality of the library in unexpected way, as fields are no longer sorted before serialization. I think we should add back the serialize_with code to ensure that when writing the repodata files they are properly sorted.

baszalmstra:
I think we should fix that before we release because it might break repodata indexing.

PR #1678, opened 2025-09-17
fix(menuinst): activate menuinst per default

Hofer-Julian:
According to the docs the original implementation also defaults to true
@lucascolley


lucascolley (approved):
thanks Julian!

PR #1685, opened 2025-09-25
fix: Rattler not copying back symbolic links from the clobbers directory.

andrejvelichkovski:
### Description

This change should resolve a small issue in rattler on how it moves things around in the clobbers directory.

Namely, if we have to move a symbolic link back to the original location from the clobbers directory, we don't perform this and end up in a state where in the original location we miss the file.

This happens because we check whether file exists using `.exists()` which follows symbolic links which at the time they are in the `__clobbers__` path they might be invalid. This PR replaces the validation with either this or the `.is_symlink()` which checks whether the file is a symlink (without validating it).

The issue here happens only if the packages need to be repriotized and moved around, and this is a bit tricky to simulate in the automated tests, so we generate test packages with dependency between them to force the topological sort to be different from the new_priorities which sort by the package name.

Please let me know if I should add anything else in this PR.

### DIFF
--- a/crates/rattler/src/install/clobber_registry.rs
+++ b/crates/rattler/src/install/clobber_registry.rs
@@ -1444,6 +1444,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(not(target_os = "windows"))]
     async fn test_clobber_with_symlinks_and_reprioritization() {
         let repodata_record_with_symlink1 = get_repodata_record(
             get_test_data_dir().join("clobber/clobber-with-symlink-a-0.1.0-h4616a5c_0.conda"),


baszalmstra:
Maybe we can use fs::symlink_metadata to avoid having to do two syscalls.

Otherwise looks good to me! Thanks for the extensive tests as well!

andrejvelichkovski:
> Maybe we can use fs::symlink_metadata to avoid having to do two syscalls.
> 
> Otherwise looks good to me! Thanks for the extensive tests as well!

Thanks for the suggestion @baszalmstra, I now replaced it to use the `symlink_metadata.is_ok()` check instead. Would you prefer to add this into a new helper function: `check_file_exists` as I added in the PR, or should I move the same call to every path instead without adding a new helper function?

crates/rattler/src/install/clobber_registry.rs
```diff
@@ -1442,4 +1442,48 @@ mod tests {
             ],
         );
     }
+
+    #[tokio::test]
```
baszalmstra, line 1446: Can you make this only run on non-windows platforms? Symlinks are not enabled by default on windows.

baszalmstra (commented):
This looks good to me!

crates/rattler/src/install/clobber_registry.rs
andrejvelichkovski: I have now added a configuration to not run this test on windows platform.

baszalmstra:
Thanks!

wolfv:
Thanks, this is great!

PR #1694, opened 2025-09-30
feat: move variant parsing from rattler-build

Hofer-Julian:
This adds a new crate called `rattler_variants`.
Most of it is directly moved from rattler-build.
The only bigger duplication is `VariantValue`.
In `rattler-build` we reused `Variable` that is used for variables in the `recipe.yaml`. Therefore, we have to add conversion functions on the rattler-build side.

### DIFF
--- a/crates/rattler_variants/Cargo.toml
+++ b/crates/rattler_variants/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "rattler_variants"
+version = "0.1.0"
+edition = "2021"
+authors = ["Prefix.dev"]
+license = "BSD-3-Clause"
+description = "Utilities for parsing conda variant configuration files"
+
+[lib]
+path = "src/lib.rs"
+
+[dependencies]
+miette = { workspace = true }
+fs-err = { workspace = true }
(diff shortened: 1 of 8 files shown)


baszalmstra:
Does this also properly parse `variants.yaml`? Or only `conda_build_config.yaml`?

Hofer-Julian:
> Does this also properly parse `variants.yaml`? Or only `conda_build_config.yaml`?

Both. You can find the relevant code here:

https://github.com/conda/rattler/blob/8b59d795dc23d9f6c263f67f05eef4bd7c9022eb/crates/rattler_variants/src/parser.rs#L35-L39

crates/rattler_variants/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "rattler_variants"
+version = "0.1.0"
+edition = "2021"
+authors = ["Prefix.dev"]
```
wolfv, line 5: I think this should include at least the `hi@prefix.dev` email?

wolfv:
Can we add a few snapshot tests? :)

Issue #1696, opened 2025-09-30
perf: dont run shell if no activation scripts


baszalmstra:
When there are no activation _scripts_, the `Activation::run_activation` function will still start a shell to evaluate all the environment variables. I think we can forgo this step and just return the activation variables directly. 

https://github.com/conda/rattler/blob/498ab0cb06a537e29139bacadc8a377b3c75d5b3/crates/rattler_shell/src/activation.rs#L617

LED-0102:
Hi! Can I work on this issue?

baszalmstra:
Go for it!

LED-0102:
Hi @baszalmstra,

I drafted this early-return for empty activation scripts:

```
let ret = variables.current_env.clone();
let activation_script = self.activation(variables)?.script;

if activation_script.contents()?.trim().is_empty() {
    return Ok(ret);
}
```
This would skip the shell when the script is empty and return the current activation variables.

Does this match the optimization you had in mind?

baszalmstra:
Jep! Just needs some tests!

baszalmstra:
actually its not quite it because the activation script will also set variables. So it will never be empty. You just have to check if the activation script would source others.

LED-0102:
Hi @baszalmstra,

Thanks for clarifying! From reading through `activation()`, it looks like the script only “sources” others in two cases:

1. When `variables.conda_prefix` is set and we need to run deactivation scripts.

2. When `self.activation_scripts` is non-empty.

If both are empty, the script just sets variables like `CONDA_PREFIX`, `CONDA_SHLVL`, `$PATH`, and static env vars.

In that case, should I skip the shell and deterministically return those variable changes as the diff (instead of just returning `variables.current_env`)?

LED-0102:
Hi @baszalmstra,

I’m wondering about the “before” values of environment variables like `CONDA_SHLVL`, `CONDA_PREFIX`, `PATH`, etc. I think normally these are captured by the shell, so without running it, I’m not sure how we can know what they actually were.

In this situation, I’m not sure what the correct approach is: should we just return `variables.current_env`, or should we try to return `variables.current_env` plus the deterministic changes applied by `activation()` even if they might be identical to the “before” values?

I think some guidance here would help clarify the intended behavior.

KartikDua1504:
Hello @baszalmstra ,
Could I be allowed to work on this issue ?? Since, this has been inactive for a bit.

teambugbusters00:
[...]

PR #1700, opened 2025-10-03
draft: perf: skip shell if no activation scripts or conda prefix (#1696)

LED-0102:
### Description

This is a draft PR for discussion of #1696. 

For now, `run_activation` simply returns `variables.current_env` directly 
when there is no conda prefix and no activation scripts, skipping the shell.

Opening this PR to discuss whether this approach is acceptable and how 
we might handle deterministic environment diffs in these cases.

### DIFF
--- a/crates/rattler_shell/src/activation.rs
+++ b/crates/rattler_shell/src/activation.rs
@@ -619,6 +619,10 @@ impl<T: Shell + Clone> Activator<T> {
         variables: ActivationVariables,
         environment: Option<HashMap<&OsStr, &OsStr>>,
     ) -> Result<HashMap<String, String>, ActivationError> {
+        if variables.conda_prefix.is_none() && self.activation_scripts.is_empty() {
+            return Ok(variables.current_env);
+        }
+
         let activation_script = self.activation(variables)?.script;
 
         // Create a script that starts by emitting all environment variables, then runs


wolfv:
I think the idea is solid but I would like to double check whether this takes into account the environment variables from `env_dir.d` which are stored in JSON files?

wolfv:
Also I would put this into a new function and make sure it's well tested etc.

LED-0102:
Hi @iamthebot,

I'll look into it today if that's okay

LED-0102:
Hey all,

I’ve implemented the fast path in `run_activation` so that if there’s no `conda_prefix` and no activation scripts, we skip spawning a shell entirely. The environment variables are still computed deterministically using `env_vars` and `post_activation_env_vars`, so everything from `env_dirs.d` and `conda-meta/state` is included.

The snapshot for the fast path currently looks like this (from the new test case [test_run_activation_fast_path](https://github.com/conda/rattler/pull/1700/files#diff-31e0caeac39ff3fb1305155334fbe0c08d87ec07de85fa77dbe343e8455d70fbR1100)):
```
CONDA_SHLVL: "1"
PKG1: "Hello, world!"
PKG2: "Hello, world!"
STATE: "Hello, world!"
```

I’d love some eyes on this to confirm:

1. The snapshot looks correct
2. The testing approach makes sense
3. The functionality matches what we want for simple activations

After this review, I can clean up the block into a separate function, add comments, etc.

Thanks!

LED-0102:
Hi all. I'm just checking in on this PR.
Let me know if there’s anything else needed from my side, or if someone would like me to help move this toward merge. Thanks!

crates/rattler_shell/src/activation.rs
```diff
@@ -619,6 +673,12 @@ impl<T: Shell + Clone> Activator<T> {
         variables: ActivationVariables,
         environment: Option<HashMap<&OsStr, &OsStr>>,
     ) -> Result<HashMap<String, String>, ActivationError> {
+        if variables.conda_prefix.is_none() && self.activation_scripts.is_empty() {
+            println!("Fast track");
+            return Ok(self.run_activation_fast_path(&variables, environment.as_ref()));
+        }
+        println!("Slow track");
```
baszalmstra: Woops!

crates/rattler_shell/src/activation.rs
```diff
@@ -1050,6 +1110,55 @@ mod tests {
         insta::assert_yaml_snapshot!("after_activation", env_diff);
     }
 
+    #[allow(dead_code)]
+    fn test_run_activation_fast_path(shell: ShellEnum, with_unicode: bool) {
+        let environment_dir = tempfile::TempDir::new().unwrap();
+
+        let env = if with_unicode {
```
baszalmstra, line 1128: with_unicode seems to never be set to true?

crates/rattler_shell/src/activation.rs
```diff
@@ -609,6 +609,60 @@ impl<T: Shell + Clone> Activator<T> {
         })
     }
 
+    /// Fast path activation when there are no activation scripts and no conda prefix to deactivate.
+    /// This avoids spawning a shell by directly computing the environment variable changes.
+    fn run_activation_fast_path(
```
baszalmstra: I feel like there is a lot of duplication in this function. Cant we reuse/abstract some other code to set these environment variables properly?

crates/rattler_shell/src/activation.rs
baszalmstra: Maybe split the rest of the function into a `run_activation_slow_path` so you can easily test them without interference?

crates/rattler_shell/src/activation.rs
```diff
@@ -1402,4 +1517,157 @@ mod tests {
...
+        let fast_path_result = activator.run_activation_fast_path(&variables, None);
+        let normal_result = activator.run_activation(variables, None).unwrap();
+
+        assert_eq!(fast_path_result, normal_result);
+    }
+
+    /// Test that `run_activation_fast_path` and `run_activation` produce equivalent results
+    /// with post-activation environment variables.
+    #[test]
+    fn test_fast_path_vs_normal_activation_with_post_env_vars() {
```
baszalmstra: The setup of this function and the one below is virtually the same. Maybe they can be combined?

LED-0102:
Hi @baszalmstra,

I’m currently unable to resolve the CI failure related to the `pixi.lock `file.

The workflow is failing with `lock-file not up-to-date with the workspace` during `pixi install -e lint --locked`, even after regenerating and committing the lockfile locally. I’ve verified that `pixi install -e lint` succeeds on my machine, but CI continues to fail.

I would appreciate any guidance on what I might be missing, particularly if there are CI-specific constraints or platform considerations involved.

RaunakSeksaria:
Hi, since this had been inactive for a while, thought that I would take this up slightly. Also decided to sortof review and contribute here itself, since there seems to have been significant work already done.

While looking at the code, however, I saw that in the `compute_activation_env` function, the paths are being computed implicitly equivalent to the `Prepend` behavior, instead of allowing for different behavior if `variables.path_modification_behavior` have the `Replace` and `Append` values. 

Since a new shell is not being spawned, ideally the functionality of `set_path`in trait `Shell` should be implemented in the `compute_activation_env` function.

Do let me know your thoughts on this!

PR #1705, opened 2025-10-06
feat(upload): attach hashes to uploaded packages

baszalmstra:
Adds custom metadata with the sha256 and md5 hash of a package during uploading, which will allow us to not have to stream the entire file when we index. This PR also changes the way we upload to a streaming approach, which should reduce the memory requirements of uploading very large packages.


wolfv:
This looks great!

PR #1709, opened 2025-10-07
feat: add py-rattler-types crate

wolfv:
This adds a new crate "py-rattler-types" that we can use from py-rattler as well as py-rattler-build. 

The idea is that we can seamlessly use `rattler` Python types from the Python bindings in `rattler-build`.


wolfv:
Turns out sharing types from one Python binding to another is impossible with PyO3, sadly. So this wouldn't really help us.

PR #1716, opened 2025-10-09
feat: detect concurrent s3 indexing

baszalmstra:
This PR adds support for safe concurrent indexing operations by implementing conditional writes with ETags and automatic retry logic. Multiple processes can now safely index the same channel simultaneously without corrupting repodata files.

###  Before

  Previously, the indexing process had no protection against concurrent modifications. If multiple processes attempted to index the same channel simultaneously:

  - The last writer would silently overwrite changes from other processes
  - No detection of concurrent modifications
  - Risk of data loss or corrupted repodata files
  - Required external locking or coordination to prevent race conditions

  The flow was:
  1. Read existing repodata.json (if exists)
  2. Index packages
  3. Write new repodata.json unconditionally

###  After

  The indexing process now uses conditional writes based on file metadata (ETags/last-modified timestamps):

  1. Collect metadata (ETags) for all critical files before reading (`repodata.json`, `repodata.json.zst`)
  2. Read with validation - verify files haven't changed since metadata collection
  3. Index packages
  4. Write with conditional checks - fail if files were modified by another process.
  5. Automatic retry - if race condition detected, retry entire operation with exponential backoff.

### Testing

  This PR includes a comprehensive test-only backend implementation because:

  1. Testing requires ETag support: The standard opendal memory backend doesn't support ETags or conditional requests. Real S3 would be slow, expensive, and non-deterministic for CI.
  2. Need deterministic race conditions: The backend includes a hook system that allows tests to inject synchronization barriers at specific points (before/after stat/read/write). This enables creating predictable race conditions for testing.
  3. Complete implementation required: To faithfully test S3 behavior, we need:
    - ETag generation (MD5 hashing like S3)
    - Conditional request semantics (if-match, if-none-match, if-unmodified-since)
    - Proper error handling (ConditionNotMatch errors)
    - Full opendal Access trait implementation (stat, read, write, list, delete, create_dir)
    - Per-file locking to simulate realistic concurrency
    
### Future work

1. We can use the hashes from uploaded by rattler-upload (https://github.com/conda/rattler/pull/1705) to avoid having to download the entire file.

### AI Disclosure

Most of `etag_memory_backend.rs` was written by claude. Including the `test_concurrent_index_with_race_condition_and_retry` test.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2156,28 +2156,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "futures-retry-policies"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "6deb4087bf085042b95438ac07d98b85976f22549c4ae8dee6b44a1ff094b049"
-dependencies = [
- "chrono",
- "futures-retry-policies-core",
- "retry-policies 0.2.1",
- "tokio",
- "tracing",
-]
-
-[[package]]
-name = "futures-retry-policies-core"
-version = "0.1.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "222776451f4c0381948cc6ab590f674e0216e99738c436e4f09bc9754d7f9933"
-dependencies = [
- "pin-project",
-]
-
-[[package]]
 name = "futures-sink"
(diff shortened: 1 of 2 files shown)


wolfv:
Looks good to me!

PR #1719, opened 2025-10-09
feat(lock): add path to package_build_source

remimimimimi:
### Description

Add path to `package_build_source`.

### DIFF
--- a/crates/rattler_lock/src/conda.rs
+++ b/crates/rattler_lock/src/conda.rs
@@ -183,7 +183,7 @@ pub enum PackageBuildSource {
         url: Url,
         /// The SHA256 hash of the archive content.
         sha256: Sha256Hash,
-        /// Subdirectory on which focus on.
+        /// Subdirectory to use.
         subdir: Option<Utf8TypedPathBuf>,
     },
     /// Source is some local path.


crates/rattler_lock/src/parse/models/v6/source_data.rs
```diff
@@ -301,14 +326,14 @@ impl<'a> TryFrom<PackageBuildSourceData<'a>> for PackageBuildSource {
...
-                },
-                subdirectory,
+            Ok(PackageBuildSource::Git {
+                url: git,
+                spec,
+                rev,
+                subdir,
             })
+        } else if let Some(path) = path {
+            Ok(PackageBuildSource::Path { path })
```
baszalmstra, line 342: The description says that subdir and path are joined but looking at the code that doesnt seem to be the case. I would also not expect them to be joined. I would expect an error if both subdir and path are specified.

crates/rattler_lock/src/conda.rs
```diff
@@ -157,50 +157,40 @@ pub enum GitShallowSpec {
...
-    /// URL-based archive source with content hash
+    /// URL-based archive source with content hash.
     Url {
-        /// The URL to the archive
+        /// The URL to the archive.
         url: Url,
-        /// The SHA256 hash of the archive content
+        /// The SHA256 hash of the archive content.
         sha256: Sha256Hash,
+        /// Subdirectory on which focus on.
```
Hofer-Julian:
```suggestion
        /// Subdirectory to use.
```

PR #1720, opened 2025-10-10
feat: js-rattler include checksum in the output

martinRenou:
### Description

Including the checksums in the output of js-rattler's `simpleSolve`

### DIFF
--- a/js-rattler/crate/solve.rs
+++ b/js-rattler/crate/solve.rs
@@ -4,9 +4,9 @@ use std::str::FromStr;
 use crate::{error::JsError, platform::JsPlatform};
 use rattler_conda_types::{
     Channel, ChannelConfig, MatchSpec, NoArchType, PackageName, PackageRecord, ParseChannelError,
-    ParseStrictness::Lenient, RepoDataRecord, Version
+    ParseStrictness::Lenient, RepoDataRecord, Version,
 };
-use rattler_digest::{parse_digest_from_hex, Sha256, Md5};
+use rattler_digest::{parse_digest_from_hex, Md5, Sha256};
 use rattler_repodata_gateway::{Gateway, SourceConfig};
 use rattler_solve::{SolverImpl, SolverTask};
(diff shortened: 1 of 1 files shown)


baszalmstra:
Let me know when this is ready! :)

martinRenou:
My first rust contribution, have to admit I don't yet understand what I'm doing (did some copy from another place where you seemed to be converting the hash into string). But the CI seems to be happy now so I guess it's ready! 

I fail to test it properly on my side though.

js-rattler/crate/solve.rs
baszalmstra: Should we maybe also include it here?

martinRenou: I assumed this wasn't important? Is is really used in the end?

baszalmstra: If it is not used then we should remove the feature entirely. But as long as it is in the API I think when we add hashes it should also work for that case.

martinRenou: I meant that the installed packages feature may be only checking for the package name/version/buildnumber but the md5/sha256 wouldn't be important?

I've tried briefly setting the md5 and sha256, it seems I need to turn the `string` back into the `GenericArray`, would you have pointers on how I can do that?

martinRenou: I just pushed something, please tell me if that looks ok :D

baszalmstra: It is important because then the roundtrip wont include the hashes anymore. But you changes look good! :)

martinRenou: ah got it, thanks!

martinRenou:
@baszalmstra may I kindly ask for a release of the npm package? :D

baszalmstra:
Published as v0.3.2: https://www.npmjs.com/package/@conda-org/rattler/v/0.3.2

martinRenou:
Thanks!

PR #1732, opened 2025-10-13
fix: Prevent noarch installer from installing non-noarch packages

### LINKED ISSUE #1712: [bug] Install can be run with `noarch` as a platform and non-noarch packages. Produces broken OSX binaries.
### Checklist

### What happened?

At least with `py-rattler` you can run `install` with packages that are OS/arch specific but pass `noarch` as a platform. The packages will get linked, etc. but you end up without codesigning of the mach O files on OSX (or a bad signature). At least for the Python interpreter.

It seems like the desired behavior would be to fail fast if non-noarch packages are to be linked but only the `noarch` platform was passed to install?

### Additional Context

_No response_

iamthebot:
### Description
Closes #1712 

If an installer platform is `noarch` then non-noarch packages should cause it to fail early. Not doing this leads to bad binaries and on OSX no codesigning. See the original issue for more details.

### AI Disclosure
Used sonnet 4.5 1M w/ codecompanion.nvim to bootstrap the unit tests.


baszalmstra:
You changed the `libsolv` submodule which causes the build failure. If you just revert that, it should be good.

crates/rattler/src/install/installer/mod.rs
```diff
@@ -413,6 +413,30 @@ impl Installer {
...
+            let non_noarch_packages: Vec<String> = transaction
+                .installed_packages()
+                .filter(|record| record.package_record.subdir != "noarch")
+                .map(|record| {
+                    format!(
+                        "{}-{}-{} (subdir: {})",
+                        record.package_record.name.as_normalized(),
+                        record.package_record.version,
+                        record.package_record.build,
+                        record.package_record.subdir
```
baszalmstra:
```suggestion
                        "{}/{}-{}-{}",
                        record.package_record.subdir
                        record.package_record.name.as_normalized(),
                        record.package_record.version,
                        record.package_record.build,
```

We usually format packages like this.

PR #1743, opened 2025-10-14
feat: Add progress bar for gateway query/names

### LINKED ISSUE #1710: feat: Progress bar for repodata fetch when using gateway
### Checklist

### What is the idea?

Currently, py-rattler exposes a progress bar for the `install` method. However, `solve` doesn't have a toggleable progress bar when it's passed a gateway object and repodata is fetched.

My thinking is the gateway object could have a setting that enables/disables the progress bar when it fetches repodata. This aligns it with `install` that has an option to enable/disable a progress bar.

### Why is this needed?

Cleaner reporting when using py-rattler in interactive applications

### What should happen?

_No response_

### Additional Context

_No response_

iamthebot:
### Description
Closes #1710 

Adds a toggleable progress bar for repodata fetches mimicking the progress bar available for `install`. Plumbed it through to the python bindings as well.

Tested it functionally (since it's hard to really unit test a progress bar) and indeed the progress bar works through the python bindings.

### AI Disclosure

Most of the code is hand written though I used Sonnet 4.5 1M via claude code using the ACP integration in Neovim's codecompanion.nvim to help with finding existing patterns in using indicatif and throwing together some functional tests.


crates/rattler_repodata_gateway/Cargo.toml
```diff
@@ -96,6 +97,7 @@ native-tls = ['reqwest/native-tls', 'reqwest/native-tls-alpn', 'rattler_networki
 rustls-tls = ['reqwest/rustls-tls', 'rattler_networking/rustls-tls', 'rattler_cache/rustls-tls', 'rattler_redaction/rustls-tls']
 sparse = ["rattler_conda_types", "memmap2", "self_cell", "superslice", "itertools", "serde_json/raw_value"]
 gateway = ["sparse", "http", "http-cache-semantics", "parking_lot", "async-trait", "rattler_package_streaming"]
+indicatif = ["dep:indicatif", "gateway"]
```
baszalmstra: I would not expect the gateway to be enabled when indicatif is enabled. I think you can remove this feature entirely. `optional = true` packages already implicitly add a feature.

PR #1747, opened 2025-10-15
fix: js-rattler: Do not set a user-agent

martinRenou:
They lead to CORS issues in Firefox. It's also not supposed to be allowed to set this header in the browser

### DIFF
--- a/js-rattler/Cargo.toml
+++ b/js-rattler/Cargo.toml
@@ -43,7 +43,7 @@ url = "2.5.4"
 
 chrono = { version = "0.4.40", features = ["wasmbind"] }
 reqwest = { version = "0.12.24", default-features = false }
-reqwest-middleware = "0.4.2"
+reqwest-middleware = { version = "0.4.2", default-features = false }
 
 # We have transient dependency on getrandom 0.2.16 which needs to have the `js` feature
 # enabled.


baszalmstra (commented):
Can you add a comment in the code to explain why a default is created?

martinRenou:
Done!

js-rattler/Cargo.toml
```diff
@@ -42,6 +42,8 @@ rattler_solve = { path = "../crates/rattler_solve", default-features = false, fe
 url = "2.5.4"
 
 chrono = { version = "0.4.40", features = ["wasmbind"] }
+reqwest = "0.12.24"
```
wolfv: I think we should use `default-features = false` to avoid OpenSSL to be pulled in.

js-rattler/Cargo.toml
```diff
@@ -42,7 +42,7 @@ rattler_solve = { path = "../crates/rattler_solve", default-features = false, fe
 url = "2.5.4"
 
 chrono = { version = "0.4.40", features = ["wasmbind"] }
-reqwest = "0.12.24"
+reqwest = { version = "0.12.24", default-features = false }
 reqwest-middleware = "0.4.2"
```
wolfv: same for this one probably wrt `default-features`. Does it then remove openssl from the lockfile again when re-locking?

martinRenou:
Done! Thanks for the reviews. 

If you're happy with it, I'd love a release with that patch included if you don't mind :D

wolfv:
Looks good to me!

baszalmstra:
0.3.3 is cooking: https://github.com/conda/rattler/actions/runs/18525436983

martinRenou:
Thanks both!

PR #1748, opened 2025-10-15
fix: Allow to create a Gateway without setting a user_agent

martinRenou:
Follow-up of #1747


baszalmstra:
I dont understand, why doesnt your previous PR work? If I read the code correctly, if you manually set a client a new one shouldnt be created (with a user agent).

martinRenou:
I'm a bit in the dark as well. I still see requests with `User-Agent: rattler_repodata_gateway/0.24.9` instead of the browser-set user-agent for some reason.

martinRenou:
Also this PR doesn't seem to work anyway as well.

martinRenou:
Closing in favor of https://github.com/conda/rattler/pull/1750

PR #1752, opened 2025-10-16
chore(js): bump to 0.3.4

wolfv:
So that we can release, finally :)


baszalmstra:
Published as 0.3.4: https://www.npmjs.com/package/@conda-org/rattler/v/0.3.4

martinRenou:
Thanks both!

PR #1753, opened 2025-10-16
fix: sign dylibs

baszalmstra:
This PR fixes an issue where dylibs without executable bits were not properly signed.


wolfv (approved):
Works on my machine!

PR #1755, opened 2025-10-17
fix: use byte string instead of array for hash encoding

baszalmstra:
@dholth identified that we encode our hashes as arrays of bytes instead of storing them as a byte string. When encoding them naively this is a known issue: https://docs.rs/rmp-serde/latest/rmp_serde/#efficient-storage-of-u8-types .

This PR changes that behavior by using `serde_bytes` to encode the hashes as a byte string while keeping backwards compatibility when deserializing. I also added several tests that verify this is now using the correct behavior.

### AI Disclosure

I wrote the implementation myself but used Claude Haiku 4.5 to write the tests and update the documentation.


wolfv (commented):
Looks great!

wolfv (approved):
Looks great!

PR #1758, opened 2025-10-19
perf: Improve performance of bash shell activation on Windows

mwiebe:
### Description

The PATH variable handling for bash on Windows called the cygpath command once for each path in the list. By using cygpath's option to transform a full PATH variable at once (--path), the number of subprocess calls is much smaller, and in a test case performance of `pixi shell-hook` improved by a factor of 4 from ~0.4s to ~0.1s

```
$ time ../pixi/pixi-release-unmodified.exe shell-hook
export PATH="/c/...
...
export PS1="(proj) ${PS1:-}"

real    0m0.433s
user    0m0.000s
sys     0m0.015s
```
compared to

```
$ time ../pixi/pixi-release-updated.exe shell-hook
export PATH="/c...
...
export PS1="(proj) ${PS1:-}"

real    0m0.104s
user    0m0.000s
sys     0m0.016s
```

### DIFF
--- a/crates/rattler_shell/src/shell/mod.rs
+++ b/crates/rattler_shell/src/shell/mod.rs
@@ -198,7 +198,10 @@ pub trait Shell {
     }
 }
 
-/// Convert a native PATH on Windows to a Unix style path using cygpath.
+/// Convert a native PATH on Windows to a Unix style path using cygpath. This uses
+/// the `--path` option of cygpath, which converts a search PATH like
+/// "C:\path1;D:\path2" to "/c/path1:/d/path2" or the equivalent for the specific
+/// build of cygpath. Batching multiple paths together is better for performance.
 pub(crate) fn native_path_to_unix(path: &str) -> Result<String, std::io::Error> {
     // call cygpath on Windows to convert paths to Unix style
     let output = Command::new("cygpath")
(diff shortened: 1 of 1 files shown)


crates/rattler_shell/src/shell/mod.rs
```diff
@@ -304,49 +307,44 @@ impl Shell for Bash {
...
+            paths_vec.join(self.path_separator(platform))
+        };
+        // Replace, Append, or Prepend the path variable to the paths.
+        let posix_separator = self.path_separator(&Platform::Linux64);
         let path_var = self.path_var(platform);
-        let paths_str = paths_string.as_str();
+        let combined_paths_string: String = match modification_behavior {
+            PathModificationBehavior::Replace => paths_string,
+            PathModificationBehavior::Prepend => {
+                paths_string + posix_separator + &self.format_env_var(path_var)
```
baszalmstra:
```suggestion
                format!("{paths_string}{posix_separator}{}", &self.format_env_var(path_var))
```

crates/rattler_shell/src/shell/mod.rs
```diff
@@ -304,49 +307,44 @@ impl Shell for Bash {
...
         // would correctly escape ${PATH} so that it literally is in the result.
-        Ok(writeln!(f, "export {path_var}=\"{paths_str}\"")?)
-    }
-
-    /// For Bash, the separator in the path variable is always ":", even on
-    /// Windows
-    fn path_separator(&self, _platform: &Platform) -> &str {
-        ":"
+        Ok(writeln!(f, "export {path_var}=\"{combined_paths_str}\"")?)
     }
```
baszalmstra, line 350: Was this incorrect?

crates/rattler_shell/src/shell/mod.rs
```diff
@@ -304,49 +307,44 @@ impl Shell for Bash {
...
+        let posix_separator = self.path_separator(&Platform::Linux64);
         let path_var = self.path_var(platform);
-        let paths_str = paths_string.as_str();
+        let combined_paths_string: String = match modification_behavior {
+            PathModificationBehavior::Replace => paths_string,
+            PathModificationBehavior::Prepend => {
+                paths_string + posix_separator + &self.format_env_var(path_var)
+            }
+            PathModificationBehavior::Append => {
+                self.format_env_var(path_var) + posix_separator + &paths_string
```
baszalmstra:
```suggestion
                format!("{}{posix_separator}{paths_string}", self.format_env_var(path_var))
```

crates/rattler_shell/src/shell/mod.rs
```diff
@@ -304,49 +307,44 @@ impl Shell for Bash {
...
-        match modification_behavior {
-            PathModificationBehavior::Replace => (),
-            PathModificationBehavior::Prepend => paths_vec.push(self.format_env_var(path_var)),
-            PathModificationBehavior::Append => paths_vec.insert(0, self.format_env_var(path_var)),
-        }
         // Create the shell specific list of paths.
-        let paths_string = paths_vec.join(self.path_separator(platform));
-
+        let paths_string = if cfg!(windows) {
+            let joined_path = paths_vec.join(self.path_separator(platform));
```
baszalmstra: Here `path_seperator` is used with `platform` but below its always `Linux64`. Is that correct?

crates/rattler_shell/src/shell/mod.rs
```diff
@@ -1254,7 +1252,11 @@ mod tests {
             )
             .unwrap();
         assert!(script.contents.contains("/foo:/bar"));
+    }
 
+    #[cfg(windows)]
+    #[test]
+    fn test_path_separator_windowws() {
```
baszalmstra:
```suggestion
    fn test_path_separator_windows() {
```

crates/rattler_shell/src/shell/mod.rs
mwiebe: When calling `cygpath`, the input separator should be ";" and the output is ":". Here's an example:

```
$ cygpath --unix --path 'C:\Path\first;D:\other;C:\Program Files;D:\Path2'
/c/Path/first:/d/other:/c/Program Files:/d/Path2
```

It feels more natural to me to directly use ";" and ":", but I'm trying to follow the interface patterns that were set up here before.

crates/rattler_shell/src/shell/mod.rs
mwiebe: Yeah, there is no `Posix` so I picked a posix OS. It could be just ":" as well.

crates/rattler_shell/src/shell/mod.rs
baszalmstra: That makes sense, thanks!

crates/rattler_shell/src/shell/mod.rs
baszalmstra: I think its more clear to just use the literals (‘:’) in that case. Could you change that?

mwiebe: I removed the usage of the path_separator function and switched to ":" and ";" literals. The result looks better to me.

baszalmstra (approved):
Very nice! Thanks for the contribution!

PR #1760, opened 2025-10-19
feat: Relative path support in `LockFile`

hunger:
### Description

Make rattler handle relative paths to pypi dependencies in the `LockFile`. This helps when checking in `pixi.lock` files that reference local python packages. Instead of ending up with a machine specific path in the lock file we have a relative path that should work for all developers.

This is the rattler part of prefix-dev/pixi#4680

The PR consists of several changes:

* Bump the LockFile version: The relative paths will break when somebody goes back to a pixi version not supporting this
* Enable `non-pep508_extensions` in the `pep508_rs?` crate. This is needed for it to parse relative URLs in the lock file correctly
* Add an API to pass in the base directory used -- needed by `pep508_rs` to parse a relative path.
* Keep the raw input for (some) path and urls as read from the lock file around. `pep508_rs` does turn everything into absolute paths right away, we need to keep that info somewhere.

Closes: #1977

### DIFF
--- a/crates/rattler_lock/src/lib.rs
+++ b/crates/rattler_lock/src/lib.rs
@@ -93,6 +93,7 @@ mod pypi_indexes;
 pub mod source;
 mod url_or_path;
 mod utils;
+mod verbatim;
 
 pub use builder::{LockFileBuilder, LockedPackage};
 pub use channel::Channel;
@@ -108,6 +109,7 @@ pub use pypi::{PypiPackageData, PypiPackageEnvironmentData, PypiSourceTreeHashab
 pub use pypi_indexes::{FindLinksUrlOrPath, PypiIndexes};
 pub use rattler_conda_types::Matches;
 pub use url_or_path::UrlOrPath;
+pub use verbatim::Verbatim;
 
 /// The name of the default environment in a [`LockFile`]. This is the
 /// environment name that is used when no explicit environment name is
--- a/crates/rattler_lock/src/parse/deserialize.rs
+++ b/crates/rattler_lock/src/parse/deserialize.rs
@@ -23,7 +23,7 @@ use crate::{
     },
     Channel, CondaPackageData, EnvironmentData, EnvironmentPackageData, LockFile, LockFileInner,
     PackageHashes, ParseCondaLockError, PypiIndexes, PypiPackageData, PypiPackageEnvironmentData,
-    SolveOptions, UrlOrPath,
+    SolveOptions, UrlOrPath, Verbatim,
 };
 
 #[serde_as]
@@ -47,7 +47,7 @@ pub struct PypiPackageDataRaw {
     pub version: pep440_rs::Version,
 
     /// The location of the package. This can be a URL or a path.
-    pub location: UrlOrPath,
+    pub location: Verbatim<UrlOrPath>,
 
     /// Hashes of the file pointed to by `url`.
     pub hash: Option<PackageHashes>,
@@ -171,7 +171,7 @@ enum DeserializablePackageSelector {
         subdir: Option<String>,
     },
     Pypi {
-        pypi: UrlOrPath,
+        pypi: Verbatim<UrlOrPath>,
         #[serde(flatten)]
         runtime: DeserializablePypiPackageEnvironmentData,
     },
@@ -213,6 +213,7 @@ pub fn parse_from_document_v6_and_v7(
 }
 
 fn convert_raw_pypi_package(
+    file_version: FileFormatVersion,
     raw_package: PypiPackageDataRaw,
     base_dir: Option<&Path>,
 ) -> Result<PypiPackageData, ParseCondaLockError> {
@@ -230,10 +231,16 @@ fn convert_raw_pypi_package(
         })
         .collect::<Result<Vec<_>, _>>()?;
 
+    let location = if file_version < FileFormatVersion::V7 {
+        Verbatim::new(raw_package.location.take())
+    } else {
+        raw_package.location
+    };
+
     Ok(PypiPackageData {
         name: raw_package.name,
         version: raw_package.version,
-        location: raw_package.location,
+        location,
         hash: raw_package.hash,
         requires_dist,
         requires_python: raw_package.requires_python,
@@ -249,11 +256,12 @@ fn parse_from_lock<P>(
     // Split the packages into conda and pypi packages.
     let mut conda_packages = Vec::new();
     let mut pypi_packages = Vec::new();
+
     for package in raw.packages {
         match package {
             PackageData::Conda(p) => conda_packages.push(p),
             PackageData::Pypi(p) => {
-                pypi_packages.push(convert_raw_pypi_package(p, base_dir)?);
+                pypi_packages.push(convert_raw_pypi_package(file_version, p, base_dir)?);
             }
         }
     }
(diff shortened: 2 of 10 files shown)


hunger:
I do not think the failed CI job is related to my change: It's about running out of space while extracting the cache.

crates/rattler_lock/src/lib.rs
```diff
@@ -565,6 +573,17 @@ mod test {
...
+    #[case::v7_conda_source_path("v7/conda-path-lock.yml")]
+    #[case::v7_derived_channel("v7/derived-channel-lock.yml")]
+    #[case::v7_sources("v7/sources-lock.yml")]
+    #[case::v7_options("v7/options-lock.yml")]
+    #[case::v7_pixi_build_pinned_source("v7/pixi-build-pinned-source-lock.yml")]
+    #[case::v7_pixi_build_url_source("v7/pixi-build-url-source-lock.yml")]
+    #[case::v7_pixi_build_git_tag_source("v7/pixi-build-git-tag-source-lock.yml")]
+    #[case::v7_pixi_build_git_rev_only_source("v7/pixi-build-git-rev-only-source-lock.yml")]
+    #[case::v7_pypi_relative_url("v7/pypi_absolute_url.yml")]
+    #[case::v7_pypi_relative_url("v7/pypi_relative_url.yml")]
```
tdejager: Might be better top give these unique names to dinstinguish them.

crates/rattler_lock/src/parse/mod.rs
```diff
@@ -42,38 +45,37 @@ pub enum ParseCondaLockError {
...
-
-    fn from_str(s: &str) -> Result<Self, Self::Err> {
-        // First parse the document to a `serde_yaml::Value`.
-        let document: Value = serde_yaml::from_str(s).map_err(ParseCondaLockError::ParseError)?;
-
-        // Read the version number from the document
-        let version: FileFormatVersion = document
-            .get("version")
-            .ok_or_else(|| {
+pub(crate) fn from_str_with_base_directory(
```
tdejager, line 48: This would as be a API break :) cc @baszalmstra

crates/rattler_lock/src/parse/mod.rs
```diff
@@ -31,6 +31,9 @@ pub enum ParseCondaLockError {
     #[error("environment {0} and platform {1} refers to a package that does not exist: {2}")]
     MissingPackage(String, Platform, UrlOrPath),
 
+    #[error("workdir required when parsing relative pypi requirements")]
```
tdejager: I think the error message does not match the inner message

tdejager:
Couple of questions for @baszalmstra and @hunger: 

* @baszalmstra This PR removes the `from_str` because of the relative directory requirements, any way around that you think?
* I like that this does not introduce additional fields but keep everything in memory. Everyone agreed?
* @hunger do we still catch parsing errors now that requires_dist is a string, @baszalmstra you reckon we do need a special type for that?
* @hunger why do we need `Verbatim` in the conda part of the locking code?

hunger:
@tdejager:

* Yes that is the troubling part, but wrapping some `UrlOrPath` into a `Verbatim` is a breaking change, too. We could in theory add `FromStr` back and make that use `None` base directory. That will of course break as soon as somebody uses a relative path into some `LockFile` with a parse error. That does not feel safe.
* I was tempted to drop the bump to the file version when I saw how little the files changed. But I thing we need that version bump anyway: We do not want to break the relative paths just because somebody ran an older `pixi` version.
* requires_dist is a string only in a first step. It gets processed, so it should continue to raise the parse errors. I'll double check tonight.
* Good catch, it does not really make sense in the conda part. I'll remove that again.

crates/rattler_lock/src/lib.rs
hunger: Good catch, I'll fix that:-)

crates/rattler_lock/src/parse/mod.rs
hunger: So is wrapping some of the returned `UrlOrPath` into a `Verbatim`.

crates/rattler_lock/src/parse/mod.rs
hunger: I'll fix that tonight.

tdejager:
@baszalmstra any Idea why that gatekeeper is failing I tried to re-run it a bunch of times.

baszalmstra:
The merge gatekeeper is just a job that fails if any other job fails. I used that to make "required" jobs based on conditions. It looks like the Rust / Linux_86_64 is failing, which is what causes Gatekeeper to fail.

The linux job seems to be running out of disk space when extracting the cache (crazy..). I just removed those caches, lets see if that helps.

crates/rattler_lock/src/lib.rs
```diff
@@ -182,13 +184,22 @@ impl LockFile {
     pub fn from_reader(mut reader: impl Read) -> Result<Self, ParseCondaLockError> {
         let mut str = String::new();
         reader.read_to_string(&mut str)?;
-        Self::from_str(&str)
+        parse::from_str_with_base_directory(&str, None)
     }
```
baszalmstra: Maybe we should either get rid of this or also add a base path to it.

crates/rattler_lock/src/parse/mod.rs
baszalmstra: Jup I think thats fine.

baszalmstra:
> * @baszalmstra This PR removes the `from_str` because of the relative directory requirements, any way around that you think?

No I think it makes total sense to add the base directory to the API.

crates/rattler_lock/src/parse/mod.rs
tdejager: Yeah one is a user-facing API break the other is a programmatic one, so we should bump the version in the Cargo.toml as well. :)

crates/rattler_lock/src/lib.rs
tdejager: Makes sense to me!

crates/rattler_lock/src/lib.rs
hunger: This will be in the next push.

crates/rattler_lock/src/parse/mod.rs
hunger: This will be fixed in the next push.

crates/rattler_lock/src/lib.rs
hunger: I'll change `from_reader(...)` to take a base_dir and pass that on.

Will be in the next push.

hunger:
Is there anything I can do to push this forward? I do not think I can fix the Linux CI to not run out of space :-(

baszalmstra:
What would really help is if you could try to integrate this in pixi! If we can show there that it solves the issue that would be great!

Im also working on a lock-file update that requires a lockfile bump, Id love to combine our work.

tdejager:
@baszalmstra isn't it integrated here? https://github.com/prefix-dev/pixi/pull/4788

baszalmstra:
Ah yes! I missed that!

tdejager:
We are going to merge this in with the other lock-file break, we'll close this branch after, don't be suprised!

hunger:
I updated this PR and rebased it on top of the current `main` branch.

It contains the lockfile version bump, so we probably want this one in first :-)

PR #1774, opened 2025-10-23
chore: zip crate to 6.0

wolfv:
Wanted to make sure that we use `zip = 6.0.0`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -802,9 +802,9 @@ dependencies = [
 
 [[package]]
 name = "backon"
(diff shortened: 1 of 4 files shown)


baszalmstra:
Unless we really need a feature from a new version (like in the case of zip@6) I prefer we not bump the versions in the Cargo.toml. That way we stay more compatible for downstream libraries.

wolfv:
OK, only updating zip - mainly to slim down dependency trees in other projects :)

PR #1775, opened 2025-10-23
feat(lock): upgrade lock file to v7

baszalmstra:
The big changes are:
* Additions of variants to the lock file:
```
python_min: '3.9'
target_platform: noarch
```
* Making `version` optional (currently filled in for non-mutable source like git and unfilled for path etc.) This will help with dynamic versioning in the future
* Adding support for relative file-url's locked for pyproject.toml and PyPI. PR's are already:
   - https://github.com/prefix-dev/pixi/pull/4788
   - These changes will need to be pulled in *first* before we merge.
* `dev` which marks packages as only install its host, build, run and not build these themselves.


crates/rattler_lock/src/parse/models/v7/deserialize.rs
```diff
@@ -0,0 +1,240 @@
...
+pub(crate) struct CondaSelectorV7 {
+    /// URL or path to the conda artifact referenced by the selector.
+    #[serde(rename = "conda")]
+    pub(crate) conda: UrlOrPath,
+    /// Optional package name used to disambiguate source artifacts.
+    pub(crate) name: Option<PackageName>,
+    /// Variants recorded in the lock file to disambiguate source packages.
+    #[serde(default)]
+    pub(crate) variants: BTreeMap<String, crate::VariantValue>,
+    /// Whether this is a dev package (only for source packages).
```
tdejager, line 89: Guess this can use a bit better comment

PR #1777, opened 2025-10-27
feat: make precondition checks configurable

wolfv:
We recently added precondition checks when indexing channels on S3.

However, it was not possible to disable them.
Also, we got a report that we should retry on `409` errors from AWS S3 as well.


crates/rattler_libsolv_c/libsolv
baszalmstra, line 1: This is a woopsie! Revert please.

PR #1779, opened 2025-10-27
fix: sort packages when serializing repodata

baszalmstra:
Fixes https://github.com/conda/rattler/pull/1674#issuecomment-3451187309


wolfv:
Why wouldn't we want to use a BTreeMap in the first place?

PR #1791, opened 2025-10-29
fix: `Version::starts_with` for versions with differing non-final component sizes

schmelczer:
### Description

Users can be quite inventive when it comes to versioning schemes for their libraries. Fortunately, rattler is robust against a wide-range of versioning approaches.

However, rattler exhibits a confusing behaviour: depending on `lib=1.0.0_version_2` would result in a solve to `1.0.0_version1`. Using an `StrictVersion` dependency specification would still work (`lib 1.0.0_version_2`) but not a `StrictRangeOperator::StartsWith` (`lib=1.0.0_version_2`)

I've added a test that proves the issue with the preexisting behaviour of `Version::start_with`.

The bug came from only allowing segments with multiple components to be considered matching even if only their prefix matched. This is the right behaviour for segments at the very end of the version but not for middle ones.

### DIFF
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -535,8 +535,7 @@ fn segments_starts_with<
     a: A,
     b: B, // the prefix we're looking for in 'a'
 ) -> bool {
-    let mut zipped = a.zip_longest(b).peekable();
-    while let Some(ranges) = zipped.next() {
+    for ranges in a.zip_longest(b) {
         let (left, right) = match ranges {
             EitherOrBoth::Both(left, right) => (left, right),
             EitherOrBoth::Left(_) => return true,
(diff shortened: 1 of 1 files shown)


baszalmstra:
Im at a conference right now, I will take a closer look in the next few days. 

But in general I want to stay as close as possible to the behavior of conda. As I consider that the standard implementation. 

Have you tried what happens with conda?

baszalmstra:
I just tried this with the following conda script:

```python
from conda.models.version import VersionSpec
VersionSpec("1.0.0_version_2*").match("1.0.0_version1")
> False
```

With `py-rattler`:

```python
from rattler import Version
Version("1.0.0_version1").starts_with(Version("1.0.0_version_2"))
> True
```

which indeed seems very wrong.

crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -552,8 +553,15 @@ fn segments_starts_with<
...
             if !match values {
                 EitherOrBoth::Both(a, b) => a == b,
-                EitherOrBoth::Left(_) => return true,
-                EitherOrBoth::Right(_) => return false,
+                // 'a' is allowed to have extra components at the end of a segment
+                // if the corresponding 'b' segment is the last one
+                EitherOrBoth::Left(_) => {
+                    let next_ranges = zipped.peek();
+                    return matches!(next_ranges, None | Some(EitherOrBoth::Left(_)));
+                }
```
baszalmstra: I think the logic should be slightly different. I think if there are no more right components and there are left components we should consider the right component to be `0`. You can see that this logic works when testing against conda.

```python
In [10]: VersionSpec("1.0.0_version1*").match("1.0.0_version")
Out[10]: False

In [11]: VersionSpec("1.0.0_version0*").match("1.0.0_version")
Out[11]: True

In [12]: VersionSpec("1.0.0_version0foo*").match("1.0.0_version")
Out[12]: False
```

schmelczer: Thanks for the flag, it's a very good point! I've added your examples as a separate test and updated the logic based on your suggestion. Thanks again for the quick response!

Issue #1795, opened 2025-11-03
rattler-index: Set cache control headers in s3 metadata


pavelzw:
as discussed in https://discord.com/channels/1082332781146800168/1433513350150619146/1433708155535097916

wolfv:
We are doing this now!

PR #1796, opened 2025-11-03
feat: add cache control for repodata.json / zst and shard index

wolfv:
This should add cache control (5 minutes) to repodata.json, repodata.json.zst and the shard index file. Shards themselves remain immutable.

Should it be 5 minutes for repodata.json or longer?

Written by Claude


wolfv:
Should fix #1795 

@pavelzw do you have a way to test this easily in a real-world bucket?

pavelzw:
@christianrothqc might have one that's not aws s3. i can try out with s3 as well though

baszalmstra:
Nice!

PR #1810, opened 2025-11-05
perf: remove `sync_all`

baszalmstra:
Perfomance optimization by removing the `sync_all` function call (its not needed), writing non-pretty files in conda-meta, using the default buffer size for the `BufWriter` to optimize for smaller files (most of the files are tiny).

### AI Disclosure

None


baszalmstra:
@wolfv Can you check if this is also faster on your machine?

wolfv:
Is a small buffer really better for small files? Becasue we have to allocate less? I was thinking that it would be equally fast but the drawback is more memory usage and slower writes to disk (different tradeoff in other use cases).

PR #1813, opened 2025-11-07
feat: optimize syscalls by only copying executable bits

### LINKED ISSUE #486: Consider not copying permissions when reflinking on macOS
In order to keep maximum performance we might want to chekc why the behavior is different between linux & macOS and not do the permission adjustment when reflinking on macOS.

baszalmstra:
Only copy the executable bit permissions when the source file is executable AND the destination doesn't already have the correct executable bits. This drastically reduces syscalls for non-executable files and executable files that already have correct permissions.

Changes:
- Created `copy_executable_permissions()` helper function to encapsulate the optimization logic
- Applied optimization to reflink operations on Linux
- Applied optimization to patched file operations (prefix replacement)
- Only copy executable bits (0o111), not all permission bits
- Skip permission updates if bits already match

Since the vast majority of package files are not executable, and many executable files already have correct permissions after creation/reflink, this provides a significant performance improvement during package installation.

The optimization only affects Unix systems. On Windows, permissions are handled differently and this optimization is not applicable.

Fixes #486

## AI Disclosure

Written by Claude Code Web

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -220,12 +220,12 @@ pub fn link_file(
         // We no longer need the file.
         drop(file);
 
-        // Copy over filesystem permissions. We do this to ensure that the destination file has the
-        // same permissions as the source file.
+        // Copy over executable bit permissions from the source file if needed.
+        // We only copy the executable bits, not all permissions, to reduce syscalls.
         let metadata = fs::symlink_metadata(&source_path)
             .map_err(LinkFileError::FailedToReadSourceFileMetadata)?;
-        fs::set_permissions(&destination_path, metadata.permissions())
-            .map_err(LinkFileError::FailedToUpdateDestinationFilePermissions)?;
+        #[cfg(unix)]
+        copy_executable_permissions(&metadata.permissions(), &destination_path)?;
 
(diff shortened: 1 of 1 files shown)


baszalmstra:
@wolfv This also no longer copies the other permission bits (e.g. read/write). Im not sure if that would cause issues. I assume that either these are already copied because the underlying filesystem works like that _or_ they are not copied but we get the "default" permissions of the user, which might arguably be more correct anyway. 

WDYT?

baszalmstra:
Running this locally on my WSL Ubuntu 22.04 (which doesn't support reflinking) the performance impact is small (~1%). Which is expected. I would be mostly interested to see if this has an impact on macOS.

tdejager:
@baszalmstra how can I measure the impact?

baszalmstra:
Build `rattler-bin` twice, once on `main` and once using this PR then run:

```
hyperfine -w 1 \
  -p "rm -rf .prefix" "./rattler-main create jupyterlab -c https://prefix.dev/conda-forge" \
  -p "rm -rf .prefix" "./target/release/rattler create jupyterlab -c https://prefix.dev/conda-forge"
```

I expect the biggest difference will be one linux with a filesystem that allows reflinks.

PR #1818, opened 2025-11-07
perf: use global cache lock to reduce per-package lock overhead

baszalmstra:
This PR removes locking of individual package lock files in favor of a single lock for the entire package cache. This reduces the number of flock syscalls.

The impact is small, but it's a nice improvement.

## AI Disclosure

Claude Code implemented this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -192,9 +192,9 @@ dependencies = [
 
 [[package]]
 name = "async-compression"
-version = "0.4.32"
+version = "0.4.33"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5a89bce6054c720275ac2432fbba080a66a2106a44a1b804553930ca6909f4e0"
+checksum = "93c1f86859c1af3d514fa19e8323147ff10ea98684e6c7b307912509f50e67b2"
 dependencies = [
  "compression-codecs",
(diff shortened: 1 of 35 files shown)


wolfv:
Nice! Do you think we can merge this without breaking compatibility with older clients?

wolfv:
We could also start a v2 cache using content addressing etc

baszalmstra:
This cache is still compatible but any locks by older clients are not respected. So this might likely break concurrent access with older clients.

I think most of the time users will use the same client to access the cache concurrently so I think the impact will be low.

We should also measure if this has a noticeable impact on the pixi tests where multiple tests do access the same cache often but with different packages. Ill try that next week.

wolfv:
I am putting together a Github workflow to build rattler as a release-build so we can test in Github Codespaces where it was noticeably "slower" than I would hope. I'll also make a branch where I merge all your performance related PRs to that we can measure the difference :)

Looking forward to some more benchmarking!

baszalmstra:
One thing that makes quite an impact is using another memory allocator like we do in pixi. You might also want to incorporate that!

baszalmstra:
I verified in pixi that this doesnt not regress parallel execution times: prefix-dev/pixi#4908

PR #1819, opened 2025-11-07
perf: add skip validation mode to minimize cache validation overhead

baszalmstra:
Changes:
  - Added `ValidationMode::Skip` enum variant
  - Updated `validate_package_directory` to only check `index.json` and `paths.json` exists in `Skip` mode
  - Changed package cache to use `Skip` instead of `Fast` validation.

Impact: Reducing loads of `stat` syscalls because the validation no longer checks if all files exist.

## AI Disclosure

Made by Claude Code

### DIFF
--- a/crates/rattler_cache/src/package_cache/mod.rs
+++ b/crates/rattler_cache/src/package_cache/mod.rs
@@ -595,7 +595,7 @@ where
 
             // Validate the package directory.
             let validation_result = tokio::task::spawn_blocking(move || {
-                validate_package_directory(&path_inner, ValidationMode::Fast)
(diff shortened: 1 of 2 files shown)


wolfv:
Looks good to me! Do you remember if we do atomic extraction & rename? E.g. extract the folder to some temporary name and only rename it after finishing the download/extraction? I think that would be a nice safe guard against corrupted downloads (e.g. cancelled halfway through).

baszalmstra:
I dont think we do at this point. But that would indeed be a good addition here.

baszalmstra:
Superseded by #1972

PR #1821, opened 2025-11-10
feat: allow relative path parsing for matchspecs

tdejager:
This adds logic to look at a file-path during a matchspec parse to see if its relative and in that case handle it. This is needed to solve the issue: https://github.com/prefix-dev/pixi/issues/4867. Where we cannot pass relative paths into exec, tested this with a patched pixi version and seems to work well.

## AI Disclosure
* Mostly codex for the generation of the code, did a manual review and touch-ups after.

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/parse.rs
+++ b/crates/rattler_conda_types/src/match_spec/parse.rs
@@ -106,6 +106,12 @@ pub enum ParseMatchSpecError {
     #[error("found multiple values for: {0}")]
     MultipleValueForKey(String),
 
+    /// Relative paths cannot be resolved without knowing the current directory
+    #[error(
+        "relative package path '{0}' cannot be resolved because the current directory is unavailable"
+    )]
+    RelativePathRequiresCurrentDir(String),
+
     /// More than one semicolon in match spec
     #[error("more than one semicolon in match spec")]
     MoreThanOneSemicolon,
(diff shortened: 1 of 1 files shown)


crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -395,9 +395,49 @@ pub fn parse_url_like(input: &str) -> Result<Option<Url>, ParseMatchSpecError> {
             .map(Some)
             .map_err(|_err| ParseMatchSpecError::InvalidPackagePathOrUrl);
     }
+
+    // If it looks like a relative path use the cwd to resolve
+    if looks_like_relative_path(input) {
+        let cwd = typed_path::utils::utf8_current_dir()
```
baszalmstra: Are we passing the channel config into the parser? If so, we can use the root path from that.

tdejager:
Hmm thinking how to solve this from a wasm standpoint :)

crates/rattler_conda_types/src/match_spec/parse.rs
tdejager: Don't think so we use `from_str` for nameless matchespec I think.

tdejager: I decided to just throw an error in the wasm context for now when detecting a relative path. There could be other options but I think all of them would break the API. Let me know if you agree.

baszalmstra: Yeah that makes sense!

baszalmstra: > Don't think so we use `from_str` for nameless matchespec I think.

Mm I really dont like using currentdir in this api. Cant we keep it relative and resolve it later when we do have a channelconfig?

tdejager: I get it yeah. Or do you want to pass in the channel config and break the API?

tdejager: I don't mind either way :)

PR #1832, opened 2025-11-11
feat: add `rattler extract` utility to the rattler-bin crate

ruben-arts:
### Description

Since it's not easy to unpack a conda file on a macbook I added this utility function to the `rattler-bin`.

The `rattler extract` command extracts conda packages (.tar.bz2 or .conda files) from local paths or URLs:

It looks like this in usage:
```shell
# extract a local package:
rattler extract python-3.11.0-h1234567_0.conda
# Extracts to ./python-3.11.0-h1234567_0/

# extract from a URL:
rattler extract https://conda.anaconda.org/conda-forge/linux-64/python-3.11.0-h1234567_0.conda
# Extracts to ./python-3.11.0-h1234567_0/

# Specify custom destination:
rattler extract python-3.11.0-h1234567_0.conda -d my-python
# Extracts to ./my-python/
```
  Output:
```
Extracting python-3.11.0-h1234567_0.conda to python-3.11.0-h1234567_0
✓ Successfully extracted package
  Destination: python-3.11.0-h1234567_0
  SHA256: a1b2c3d4...
  MD5: e5f6g7h8...
  Size: 12345678 bytes
```
 
You can already install it with `pixi global install --path crates/rattler-bin`

# AI disclosure
Most code written by Claude


crates/rattler-bin/src/commands/extract.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    /// Destination directory where the package will be extracted
+    /// If not specified, extracts to a directory with the same name as the package
+    #[clap(short, long)]
+    destination: Option<PathBuf>,
+}
+
+pub async fn unpack(opt: Opt) -> miette::Result<()> {
+    // Try to parse as URL, otherwise treat as file path
+    let is_url = Url::parse(&opt.package).is_ok();
```
baszalmstra: Instead of parsing into a boolean, use a match statement.

crates/rattler-bin/src/commands/extract.rs
```diff
@@ -0,0 +1,106 @@
...
+    // Extract the package
+    let result = if is_url {
+        // Create HTTP client with authentication middleware
+        let download_client = Client::builder()
+            .no_gzip()
+            .build()
+            .into_diagnostic()
+            .context("Failed to create HTTP client")?;
+
+        let authentication_storage = AuthenticationStorage::empty();
```
baszalmstra: Use `AuthenticationStorage::from_env_and_defaults`

crates/rattler-bin/src/commands/extract.rs
ruben-arts: I've used an if let statement now, is that what you wanted? I first tried a `match` but that was not passing the linter.

crates/rattler-bin/src/commands/extract.rs
ruben-arts: Thanks, using that now.

crates/rattler-bin/src/commands/extract.rs
```diff
@@ -0,0 +1,112 @@
...
+
+    /// Destination directory where the package will be extracted
+    /// If not specified, extracts to a directory with the same name as the package
+    #[clap(short, long)]
+    destination: Option<PathBuf>,
+}
+
+pub async fn extract(opt: Opt) -> miette::Result<()> {
+    // Try to parse as URL, otherwise treat as file path
+    let (destination, result) = if let Ok(url) = Url::parse(&opt.package) {
```
baszalmstra: This if statement now creates rather deep nesting. Perhaps split this into two seperate functions if they share nothing?

tdejager:
I think an extract library function or seperate crate would be a lot more useful, then people can potentially use it in pixi extensions, we can have a small shim in the rattler-bin then. @baszalmstra?

baszalmstra:
That's already what this is though. It just wraps `rattler_package_streaming`s [`extract`](https://docs.rs/rattler_package_streaming/latest/rattler_package_streaming/tokio/fs/fn.extract.html) function and adds some nice CLI usage.

tdejager:
Ah right I see now, you are right! I still think I'd rather have the code in a pixi extension or pixi itself, as most people do not have the rattler binary. However, those things are not mutually exclusive :)

ruben-arts:
@tdejager yeah this was basically a prototype for a pixi extension for me. Maybe even embedded in pixi itself at some point.

crates/rattler-bin/src/commands/extract.rs
ruben-arts: Right, I let claudy split it up. This does look nicer to me. What do you think?

crates/rattler-bin/src/commands/extract.rs
```diff
@@ -0,0 +1,147 @@
...
+        .build();
+
+    Ok(client)
+}
+
+/// Determines the destination directory for URL-based extraction
+fn determine_url_destination(url: &Url, destination: Option<PathBuf>) -> miette::Result<PathBuf> {
+    if let Some(dest) = destination {
+        return Ok(dest);
+    }
```
baszalmstra: Why not do `destination.or_else(|| determine_destination_from_url(url)`? 

https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html#Push-Ifs-Up

crates/rattler-bin/src/commands/extract.rs
```diff
@@ -0,0 +1,147 @@
...
+}
+
+/// Determines the destination directory for file-based extraction
+fn determine_path_destination(
+    package_path: &str,
+    destination: Option<PathBuf>,
+) -> miette::Result<PathBuf> {
+    if let Some(dest) = destination {
+        return Ok(dest);
+    }
```
baszalmstra: same as above. also consider rename to `determine_destination_from_path`

crates/rattler-bin/src/commands/extract.rs
ruben-arts: Changed to that 👍

crates/rattler-bin/src/commands/extract.rs
ruben-arts: Done and renamed

PR #1816, opened 2025-11-07
refactor: make `extras` and `conditionals` runtime-configurable

baszalmstra:
Replace compile-time feature flags with runtime parsing options for experimental extras and conditionals syntax in matchspecs. Conditions and extras are not yet parsed from repodata.json.

## Core Changes

- Replace `ParseStrictness` parameter with `ParseMatchSpecOptions` struct
- Add `ParseMatchSpecOptions` with:
  - `::strict()` and `::lenient()` constructors
  - `allow_experimental_extras` and `allow_experimental_conditionals` fields
- Remove `experimental_extras` and `experimental_conditionals` cargo features
- Remove all `cfg` feature guards from `extras` and conditionals code
- Extras and conditionals are now always available but controlled via options

### Rattler Binary

- Enable experimental features by default in rattler-bin create command

### Python Bindings

- Add optional `experimental_extras` and `experimental_conditionals` parameters to `MatchSpec` and `NamelessMatchSpec` constructors (default: `False`)
- Expose `extras` and `condition` fields via getter methods
- Add comprehensive test suite for experimental features

## AI Disclosure

Written by Claude Code Web

PR #1836, opened 2025-11-11
feat: add condition dependency and `extras` support for libsolv

baszalmstra:
This PR adds conditional dependency support for libsolv alongside resolvo. 

libsolv already has support for this, this PR just ties it together. 

I also extended the test suite to run the same conditional-dependency tests for libsolv and resolvo.

### DIFF
--- a/crates/rattler_solve/src/libsolv_c/input.rs
+++ b/crates/rattler_solve/src/libsolv_c/input.rs
@@ -159,7 +159,7 @@ pub fn add_repodata_records<'a>(
                     // Create the dependency without the condition
                     let mut dep_spec = match_spec.clone();
                     dep_spec.condition = None;
-                    let dep_id = pool.conda_matchspec(&c_string(&dep_spec.to_string()));
+                    let dep_id = pool.conda_matchspec(&c_string(dep_spec.to_string()));
 
                     // Parse the condition
                     let condition_id = parse_condition(condition, pool);


baszalmstra:
Id like to merge https://github.com/conda/rattler/pull/1816 before continuing on this.

crates/rattler_solve/src/libsolv_c/input.rs
```diff
@@ -121,11 +149,35 @@ pub fn add_repodata_records<'a>(
...
-            let match_spec_id = pool.conda_matchspec(&c_string(match_spec));
+        for match_spec_str in record.depends.iter() {
+            // Parse the match spec to check for conditions
+            if let Ok(match_spec) = MatchSpec::from_str(
+                match_spec_str,
+                ParseMatchSpecOptions::lenient().with_experimental_conditionals(true),
+            ) {
+                if let Some(condition) = match_spec.condition.as_ref() {
+                    // Create the dependency without the condition
+                    let dep_str = match_spec_str
```
wolfv: this string spltting shouldn't be necessary, ideally.

crates/rattler_solve/tests/snapshots/backends__resolvo__solve_dummy_repo_extra_depends_recursive_feature.snap
```diff
@@ -1,28 +0,0 @@
----
```
wolfv, line 1: Just clarifying that these were not forgotten, but deleted on purpose.

PR #1837, opened 2025-11-11
feat: cache and reuse `paths.json` and `index.json` from package validation

baszalmstra:
During package verification, paths.json and index.json are read but the data was previously dropped. This PR modifies the code to cache this data in CacheMetadata and reuse it during installation, avoiding redundant file reads.

This optimization reduces I/O operations by eliminating duplicate reads of paths.json and index.json during the package installation workflow.

## AI Disclosure

Written by Claude Code Web


wolfv:
Amazing!

wolfv:
Maybe we should also skip decoding SHA hashes from paths json as we don't need them?

PR #1808, opened 2025-11-05
refactor: replace async `.tar.bz2` with `astral-tokio-tar`.

baszalmstra:
This PR replaces the spawn_blocking-based tar.bz2 extraction implementation with a fully async approach using `astral-tokio-tar` and `async-compression`. The new implementation eliminates the need for bridging between async and sync contexts, resulting in significant performance improvements for realistic package installation workloads.

**Performance improvements:**
- **42-47% faster** for mixed cache/network workloads at typical concurrency (8-16)
- **18-77% faster** for download+extract operations (package installation scenario)
- Better concurrency scaling and lower per-package latency for network-bound operations

## What changed

**Core implementation:**
- `crates/rattler_package_streaming/src/tokio/async_read.rs`
  - Replaced synchronous `tar::Archive` with `tokio_tar::Archive` from astral-tokio-tar
  - Replaced synchronous `bzip2::read::BzDecoder` with `async_compression::tokio::bufread::BzDecoder`
  - Removed `SyncIoBridge` bridging logic and `spawn_blocking` calls
  - Native async I/O throughout the extraction pipeline

**Dependencies:**
- Added `astral-tokio-tar = "0.5.6"` (forked tokio-tar with async compression support)
- Added `async-compression` with bzip2 feature
- Added `tokio-util` with io-util features

**Benchmark:**
- Added `benches/extraction_benchmark.rs` - comprehensive benchmark suite testing extraction performance across multiple scenarios and concurrency levels

## AI Disclosure

Written by Claude Code

PR #1855, opened 2025-11-18
refactor: replace async `conda` with `astral-async-zip`.

baszalmstra:
This PR replaces the spawn_blocking-based .conda extraction implementation with a fully async approach using astral-async-zip and async-compression. The new implementation eliminates the need for bridging between async and sync contexts, resulting in significant performance improvements for realistic package installation workloads.

This is build on top of #1808 and provides a similar performance gain in download and extraction speeds.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -206,6 +206,21 @@ dependencies = [
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


wolfv:
great!

PR #1857, opened 2025-11-19
feat: add accessor methods to PackageNameMatcher

baszalmstra:
## Summary
- Added `as_exact()` method to return `Option<&PackageName>` for exact matches
- Added `as_glob()` method to return `Option<&glob::Pattern>` for glob matches
- Added `as_regex()` method to return `Option<&regex::Regex>` for regex matches

These accessor methods provide convenient access to the inner values of each `PackageNameMatcher` variant.


wolfv:
yep I missed them, too! :)

PR #1859, opened 2025-11-20
refactor: Make MatchSpec deserializable

haecker-felix:
`MatchSpec` is currently only serializable. I have a use case which requires that `MatchSpec` is deserializable as well, I see no reason why it shouldn't be.


haecker-felix:
Oh, nevermind me, it already is `Deserialize` - I just had a outdated checkout.

PR #1860, opened 2025-11-20
feat: clean cache, also on disk

wolfv:
This adds a new mode to clean cached repodata not only from memory, but also from the disk cache.

@baszalmstra lmk if you prefer to also keep the old function signature around for backwards compat.

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/mod.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/mod.rs
@@ -203,6 +203,14 @@ impl Gateway {
     /// * `subdirs` - Which subdirectories to clear (all or specific platforms)
     /// * `mode` - Whether to clear only in-memory cache or both in-memory and on-disk cache
     ///
+    /// # Behavior
+    ///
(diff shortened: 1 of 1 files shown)


crates/rattler_repodata_gateway/src/gateway/mod.rs
```diff
@@ -181,15 +193,159 @@ impl Gateway {
...
+                                tracing::debug!("Deleted cache file: {:?}", file_path);
+                            }
+                        }
+                    }
+
+                    // Drop the lock explicitly before deleting the lock file itself
+                    drop(_lock);
+
+                    // Now delete the lock file
+                    if lock_file_path.exists() {
```
wolfv: @baszalmstra should we try to delete the `.lock` file or keep it around?

baszalmstra: Nope we need to keep it in case another process get exclusive access to the lock

wolfv: Keeping it!

PR #1863, opened 2025-11-20
feat: add `__cuda_arch` virtual package

baszalmstra:
This PR implements the `__cuda_arch` virtual package based on the [nvidia-virtual-packages](https://github.com/conda-incubator/nvidia-virtual-packages) reference implementation. The package detects the **minimum CUDA compute capability** (SM version) across all CUDA devices on the system, enabling hardware-specific package constraints and multi-variant builds.

@carterbox I wanted to get a head start with this to also make it available in rattler-build so we can experiment with this more easily. 

### AI Disclosure

Written by Claude Sonnet 4.5, edited and reviewed by me.

### DIFF
--- a/crates/rattler_virtual_packages/src/cuda.rs
+++ b/crates/rattler_virtual_packages/src/cuda.rs
@@ -1,29 +1,378 @@
-//! Provides functionality to detect the CUDA version present on the current system.
+//! Provides functionality to detect CUDA information present on the current system.
 //!
-//! Two methods are provided:
+//! This module detects two types of CUDA information:
 //!
-//! * [`detect_cuda_version_via_nvml`]
-//! * [`detect_cuda_version_via_libcuda`]
+//! ## CUDA Driver Version (`__cuda`)
 //!
-//! Both will detect the current supported CUDA version but the first method has less edge cases.
-//! See the function documentation for more information.
+//! The CUDA driver version represents the maximum CUDA version supported by the installed
+//! NVIDIA drivers. This is detected via:
+//!
+//! * CUDA driver library (libcuda) - Standard method
+//! * nvidia-smi command - Fallback on musl systems where dynamic library loading is not supported
+//!
+//! ## CUDA Compute Capability (`__cuda_arch`)
+//!
+//! The CUDA compute capability (also known as SM version or architecture version) represents
+//! the **minimum** compute capability of all CUDA devices detected on the system.
 
-use libloading::Symbol;
+use libloading::{Library, Symbol};
 use once_cell::sync::OnceCell;
 use rattler_conda_types::Version;
 use std::process::Command;
 use std::{
+    ffi::CStr,
     mem::MaybeUninit,
     os::raw::{c_int, c_uint, c_ulong},
     str::FromStr,
 };
 
-/// Returns the maximum Cuda version available on the current platform.
+/// Maximum length for device names in build strings to comply with CEP-26.
+const MAX_BUILD_STRING_LEN: usize = 64;
+
+/// Checks if a character is valid in a conda build string according to CEP-26.
+///
+/// Valid characters are: alphanumeric (a-z, A-Z, 0-9), underscore (_), period (.), and plus (+).
+fn is_valid_build_string_char(c: char) -> bool {
+    c.is_ascii_alphanumeric() || c == '_' || c == '.' || c == '+'
+}
+
+/// Strips a prefix from a string in a case-insensitive manner.
+///
+/// Returns the remainder of the string if it starts with the prefix (ignoring case),
+/// otherwise returns `None`.
+fn strip_prefix_ci<'a>(s: &'a str, prefix: &str) -> Option<&'a str> {
+    if s.len() >= prefix.len() && s[..prefix.len()].eq_ignore_ascii_case(prefix) {
+        Some(&s[prefix.len()..])
+    } else {
+        None
+    }
+}
+
+/// Sanitizes a device name to comply with CEP-26 build string requirements.
+///
+/// This function:
+/// 1. Removes "NVIDIA" prefix (case-insensitive) to save space
+/// 2. Filters out any characters not allowed in build strings
+/// 3. Truncates to maximum 64 characters
+///
+/// Returns the sanitized device name.
+pub(crate) fn sanitize_device_name(name: &str) -> String {
+    // Remove NVIDIA prefix (case-insensitive) and trim whitespace
+    let name = strip_prefix_ci(name, "NVIDIA").unwrap_or(name).trim();
+
+    // Filter valid characters and truncate
+    name.chars()
+        .filter(|c| is_valid_build_string_char(*c))
+        .take(MAX_BUILD_STRING_LEN)
+        .collect()
+}
+
+/// Validates that a string is in the format "major.minor" where both parts are digits.
+///
+/// Returns `true` if the format is valid for CUDA compute capability.
+pub(crate) fn is_valid_cuda_version_format(s: &str) -> bool {
+    let mut parts = s.split('.');
+    match (parts.next(), parts.next(), parts.next()) {
+        (Some(major), Some(minor), None) => {
+            !major.is_empty()
+                && major.chars().all(|c| c.is_ascii_digit())
+                && !minor.is_empty()
+                && minor.chars().all(|c| c.is_ascii_digit())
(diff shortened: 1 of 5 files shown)


crates/rattler_virtual_packages/src/cuda.rs
```diff
@@ -1,29 +1,321 @@
...
-//! * [`detect_cuda_version_via_libcuda`]
+//! ## CUDA Driver Version (`__cuda`)
 //!
-//! Both will detect the current supported CUDA version but the first method has less edge cases.
-//! See the function documentation for more information.
+//! The CUDA driver version represents the maximum CUDA version supported by the installed
+//! NVIDIA drivers. This is detected via:
+//!
+//! * CUDA driver library (libcuda) - Standard method
+//! * nvidia-smi command - Fallback on musl systems where dynamic library loading is not supported
```
carterbox: !? I am not familiar with musl libc. Is this a platform restriction or a rust restriction?

crates/rattler_virtual_packages/src/cuda.rs
```diff
@@ -1,29 +1,321 @@
...
+                )
+            } == 0
+            {
+                // Convert C string to Rust string
+                if let Ok(cstr) = CStr::from_bytes_until_nul(&name_buffer) {
+                    if let Ok(device_name) = cstr.to_str() {
+                        min_arch = Some(CudaArchInfo {
+                            major: cc_major,
+                            minor: cc_minor,
+                            device_name: device_name.to_string(),
```
carterbox: https://github.com/conda-incubator/nvidia-virtual-packages/pull/5

AFAICT, this MR is missing some post-processing that I do to the device string. I strip the substring "NVIDIA" from the device name to make it shorter. Also, today I started some work on making the plugin CEP-26 compliant. This CEP limits the length of a build string and states which chars are allowed.

crates/rattler_virtual_packages/src/cuda.rs
```diff
@@ -1,29 +1,321 @@
...
+    /// The maximum CUDA version supported by the installed driver.
+    ///
+    /// This corresponds to the `__cuda` virtual package.
+    pub version: Option<Version>,
+
+    /// Information about the minimum compute capability across all detected devices.
+    ///
+    /// This corresponds to the `__cuda_arch` virtual package. Returns `None` if
+    /// no CUDA devices are detected or if device enumeration fails.
+    pub arch_info: Option<CudaArchInfo>,
```
carterbox, line 79: We will need to have a discussion about which default values make sense. I currently use `__cuda_arch=0=None` as the fallback value, which you should probably match here.

However, I haven't determined what the correct behavior is if the system has no devices. Use the default values or just don't expose the virtual package?

Also, CEP-26 requires that you print a warning if default/fallback values are used.

carterbox (changes requested):

crates/rattler_virtual_packages/src/cuda.rs
baszalmstra: musl is an alternative libc implementation that doesnt allow dynamic library loading.

crates/rattler_virtual_packages/src/cuda.rs
baszalmstra: I will follow your implementation

crates/rattler_virtual_packages/src/cuda.rs
baszalmstra: Makes sense, Ill implement that as well

baszalmstra:
Thanks for taking a look at this @carterbox I highly appreciate it. 

Ill try to follow your implementation closely.

crates/rattler_virtual_packages/src/cuda.rs
baszalmstra: Currently, `__cuda` is not present if there are no CUDA drivers. The same goes for `__win`, `__unix`, etc. What would be the advantage of always exposing the virtual package? Knowing that the installer has the capability of exposing it?

carterbox: I wasn't sure of the existing behavior of `__cuda` if the driver is not detected. While I think there is value in knowing that the installer has the capability of detecting a virtual package but didn't find anything. I think it's better that we follow existing conventions and hide the virtual package when a device is not detected or fails to detect.

I suppose there could be a warning message to notify the user that the installer is capable of `__cuda` and `__cuda_arch`, but didn't detect anything? OR maybe that would be annoying to users without NVIDIA hardware.

ruben-arts:
Ran this on a Jetson Nano Orin:
```
> pixi run cargo run -p rattler-bin -- virtual-packages
__unix=0=0
__linux=5.15.148=0
__glibc=2.35=0
__cuda=12.6=0
__cuda_arch=8.7=Orin
__archspec=1=cortex_a72
```

carterbox:
Things to fix:

- [x] `__cuda_arch` is not coupled to `__cuda`'s presence. When `__cuda` is absent. `__cuda_arch` must also be absent even when overridden by environment variable. i.e. A user must override both on a system with no driver.
- [x] `PyVirtualPackageOverrides` lacks a `cuda_arch` getter/setter, so Python callers cannot programmatically override `__cuda_arch`.
- [ ] Invalid `CONDA_OVERRIDE_CUDA_ARCH` values silently fall back to `"0.0"` rather than rejecting. The reference implementation rejects with a warning message rather than falling back.

According to me + Claude

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -592,6 +606,92 @@ impl From<Cuda> for VirtualPackage {
...
+        // Parse format: "major.minor" (e.g., "8.6")
+        // Per CEP specification, only the version is needed
+
+        // Validate version format: must be "major.minor" where both are digits
+        if !cuda::is_valid_cuda_version_format(env_var_value) {
+            tracing::warn!(
+                "Invalid CUDA compute capability format '{}'. Expected format: 'major.minor' (e.g., '8.6'). \
+                 The default capability of '0.0' will be used instead.",
+                env_var_value
+            );
```
carterbox: Fallback instead of disabling the virtual package.

https://github.com/conda-incubator/nvidia-virtual-packages/blob/13e7a698f962bb01b3202e4539f7917f2f25de8c/nvidia_virtual_packages/cuda/arch.py#L163-L169

baszalmstra: Just to clarify, should this error? Or warn and disable the virtual package?

baszalmstra: I'm gonna go with error because:
- That's what the other virtual packages also do when they get malformed input
- When the user overrides the virtual package, they do so intentionally; simply warning and ignoring the environment variable does not seem to align with the user's intent.

baszalmstra:
I think I incorporated your review comments!

crates/rattler_virtual_packages/src/lib.rs
carterbox: If you raise an error, the it would crash the package manager any time `conda info` was called or an environment was updated? I don't know if conda allows plugins to raise errors.

carterbox: You should do whatever's appropriate for rattler as long as the user is notified in some way and we don't fallback to a default value.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -592,6 +614,84 @@ impl From<Cuda> for VirtualPackage {
...
+
+impl EnvOverride for CudaArch {
+    fn parse_version(env_var_value: &str) -> Result<Self, ParseVersionError> {
+        // CEP requires exactly "major.minor" format where both are digits
+        if !cuda::is_valid_cuda_version_format(env_var_value) {
+            return Err(ParseVersionError::new(
+                env_var_value,
+                ParseVersionErrorKind::ExpectedComponent,
+            ));
+        }
```
carterbox: My only remaining feedback is that raising this error may not be super helpful because it doesn't tell the user that the CONDA_OVERRIDE_CUDA_ARCH is the source of the error and may not accurately explain what values are accepted.

baszalmstra: Yes I agree, let me refactor that.

wolfv (approved):
Looks good to me!

PR #1881, opened 2025-11-25
feat(gateway): allow cleaning disk cache

baszalmstra:
This PR supersedes https://github.com/conda/rattler/pull/1860.

This adds a new mode to clean cached repodata not only from memory, but also from the disk cache.

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/sharded_subdir/tokio/mod.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/sharded_subdir/tokio/mod.rs
@@ -23,7 +23,8 @@ use rattler_networking::LazyClient;
 use simple_spawn_blocking::tokio::run_blocking_task;
 use url::Url;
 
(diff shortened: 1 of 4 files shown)


crates/rattler_repodata_gateway/src/gateway/sharded_subdir/tokio/mod.rs
```diff
@@ -103,6 +111,49 @@ impl ShardedSubdir {
...
+        let index_base_url = channel
+            .base_url
+            .url()
+            .join(&format!("{}/", platform.as_str()))
+            .expect("invalid subdir url");
+        let canonical_shards_url = index_base_url
+            .join(REPODATA_SHARDS_FILENAME)
+            .expect("invalid shard base url");
+        let cache_path = cache_dir.join(format!(
+            "{}.shards-cache-v1",
```
wolfv: should we replace this with a constant?

PR #1890, opened 2025-11-26
feat(solver): add minimum package age filter for security

baszalmstra:
Add support for filtering packages based on their "minimum age", similar to pnpm's `minimumReleaseAge` feature. This helps reduce the risk of installing compromised packages by delaying the installation of newly published versions.

In most cases, malicious releases are discovered and removed from channels within a short time window (often within an hour). By requiring packages to have been published for a minimum duration, users can give the community time to identify and report malicious packages before they are installed.

Mostly written by Claude Opus 4.5.

I actually want to merge #1836 first so that we can use the much nicer test framework from that PR.

### DIFF
--- a/crates/rattler_solve/tests/backends/min_age_tests.rs
+++ b/crates/rattler_solve/tests/backends/min_age_tests.rs
@@ -36,7 +36,7 @@ fn create_timestamped_repo() -> Vec<rattler_conda_types::RepoDataRecord> {
     ]
 }
 
-/// Test that min_age filters out packages that are too new.
+/// Test that `min_age` filters out packages that are too new.
 pub fn solve_min_age_filters_new_packages<T: SolverImpl + Default>() {
     let repo = create_timestamped_repo();
 
@@ -53,7 +53,7 @@ pub fn solve_min_age_filters_new_packages<T: SolverImpl + Default>() {
         .run::<T>();
 }
 
-/// Test that packages can be exempted from min_age filtering.
+/// Test that packages can be exempted from `min_age` filtering.
 pub fn solve_min_age_with_exemption<T: SolverImpl + Default>() {
     let repo = create_timestamped_repo();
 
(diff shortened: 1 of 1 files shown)


crates/rattler_solve/src/libsolv_c/input.rs
```diff
@@ -108,6 +111,10 @@ pub fn add_repodata_records<'a>(
     // details)
     let data = repo.add_repodata();
 
+    // Compute the cutoff time for min_age.
+    // Packages published after this time will be excluded (unless exempt).
+    let min_age_cutoff = min_age.map(super::super::MinimumAgeConfig::cutoff);
```
wolfv: We are importing this, right? Do we need to use super super like this?

crates/rattler_solve/src/libsolv_c/input.rs
```diff
@@ -119,6 +126,17 @@ pub fn add_repodata_records<'a>(
             _ => {}
         }
 
+        // Skip packages that haven't been published long enough (unless exempt)
+        if let (Some(config), Some(cutoff), Some(timestamp)) = (
```
wolfv: What do we do with packages without timestamp?! Always forbid if there is a config?

baszalmstra: Good question, should we add that to the options?

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -336,11 +341,23 @@ impl<'a> CondaDependencyProvider<'a> {
...
+                // Determine if this record will be excluded by exclude_newer.
+                let excluded_by_newer = matches!((&exclude_newer, &record.package_record.timestamp),
                     (Some(exclude_newer), Some(record_timestamp))
                         if record_timestamp > exclude_newer);
 
+                // Determine if this record will be excluded by min_age.
+                let excluded_by_age =
+                    match (&min_age, &min_age_cutoff, &record.package_record.timestamp) {
+                        (Some(config), Some(cutoff), Some(timestamp)) => {
+                            // Exclude if published after cutoff and not exempt
```
wolfv, line 353: Same question as in libsolv code - what's with packages withotu timestamp?

crates/rattler_solve/src/libsolv_c/input.rs
wolfv: Hmm either add it to options, or actually always forbid them if a config exists I would say. Otherwise it would be a way to circumvent this, right?

PR #1897, opened 2025-11-28
feat: integrate sigstore-sign for automatic signing of attestations

wolfv:
Instead of using cosign, we can now use `sigstore-sign` :)


crates/rattler_upload/src/upload/attestation.rs
```diff
@@ -48,35 +30,67 @@ pub struct AttestationConfig {
...
+    let digest_hex = format!("{digest:x}");
+
+    // Step 3: Get package filename
+    let filename = package_path
+        .file_name()
+        .and_then(|n| n.to_str())
+        .ok_or_else(|| miette::miette!("Invalid package file name"))?;
+
+    // Step 4: Create attestation with in-toto statement
+    let predicate = CondaV1Predicate {
```
baszalmstra, line 64: Shall we put this type in a crate? rattler-attestation?

wolfv: hmm, yeah, maybe after we validate that this works? I'll test this code from my fork later :)

baszalmstra:
Did you end up testing this?

wolfv:
I am working on a test right now. Will tell you in a bit! :)

.github/workflows/build-and-upload.yml
```diff
@@ -0,0 +1,97 @@
...
+  RUST_LOG: info
+  RUST_BACKTRACE: 1
+  CARGO_TERM_COLOR: always
+
+jobs:
+  build-and-upload:
+    name: Build and Upload with Attestation
+    runs-on: ubuntu-22.04
+    # Only run on wolfv's fork
+    if: github.repository_owner == 'wolfv'
```
baszalmstra: This is a bit of a red flag. Should we remove this workflows?

wolfv: Yeah this is more a "demo"-workflow to be honest. You need to have trusted publishing configured as well, in order to make this work...

crates/rattler_upload/Cargo.toml
```diff
@@ -12,6 +12,7 @@ readme.workspace = true
 
 [features]
 s3 = ["rattler_networking/s3", "rattler_s3"]
+sigstore = ["dep:sigstore-sign"]
```
baszalmstra: This feature is the same as just the dependency. Maybe we can just remove this feature if favor of the dependency?

wolfv: We might add `sigstore-verify` soon, as a optional dependency for the client-side verification. But we can also turn them both into separate optional features.

baszalmstra: Yeah I think it makes sense to have them as different optional features!

.github/workflows/build-and-upload.yml
baszalmstra: Maybe we should add this to some documentation then? This is also not really findable.

wolfv: It's primary use case is for testing :) Would it be fine if I add a big comment on the workflow itself?

.github/test-recipe/recipe.yaml
baszalmstra, line 1: Is this used somewhere?

crates/rattler_upload/src/upload/prefix.rs
```diff
@@ -109,32 +110,57 @@ pub async fn upload_package_to_prefix(
         AttestationSource::GenerateAttestation
     );
 
+    // Check if attestation generation is requested but sigstore feature is not enabled
+    #[cfg(not(feature = "sigstore-sign"))]
+    if wants_generate {
+        return Err(miette::miette!(
+            "Attestation generation was requested, but the 'sigstore' feature is not enabled.\n\
+             Please rebuild with the 'sigstore' feature enabled."
```
baszalmstra:
```suggestion
            "Attestation generation was requested, but the 'sigstore-sign' feature is not enabled.\n\
             Please rebuild with the 'sigstore-sign' feature enabled."
```

.github/test-recipe/recipe.yaml
wolfv: It was used by the workflow I deleted. Let me delete this file, too.

PR #1901, opened 2025-12-03
fix: Tune heavy py-rattler imports + lazy import networkx

iamthebot:
### Description
Prevent some cascading imports where not needed (eg; from unnecessary top-level imports) + lazy-import networkx. This alone speeds up the worst offenders of import time for py-rattler significantly. Obviously for networkx you incur that if you use to_graph but only then.

For PackageRecord alone this this drops import time by 90% from networkx alone. Verified with an import profile + `tuna`.


baszalmstra:
Thanks!

PR #1905, opened 2025-12-03
refactor: replace `chrono` with `jiff`

baszalmstra:
Migrate from `chrono` to `jiff` which seems to be taking over.

~The current `opendal` version still depends on `chrono`~

### DIFF
--- a/crates/rattler_solve/tests/backends/snapshots/backends__backends__resolvo__exclude_newer_error.snap
+++ b/crates/rattler_solve/tests/backends/snapshots/backends__backends__resolvo__exclude_newer_error.snap
@@ -4,4 +4,4 @@ expression: result.unwrap_err()
 ---
 Cannot solve the request because of: The following packages are incompatible
 └─ foo >=4 cannot be installed because there are no viable options:
-   └─ foo 4.0.2 is excluded because the package is uploaded after the cutoff date of 2021-12-12 12:12:12 UTC
+   └─ foo 4.0.2 is excluded because the package is uploaded after the cutoff date of 2021-12-12T12:12:12Z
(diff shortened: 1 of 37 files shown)


wolfv:
OpenDAL seems to be using `jiff` as well, to some extent. But I also saw `chrono` somewhere there.

crates/rattler_solve/tests/backends/snapshots/backends__backends__resolvo__exclude_newer_error.snap
```diff
@@ -4,4 +4,4 @@ expression: result.unwrap_err()
 ---
 Cannot solve the request because of: The following packages are incompatible
 └─ foo >=4 cannot be installed because there are no viable options:
-   └─ foo 4.0.2 is excluded because the package is uploaded after the cutoff date of 2021-12-12 12:12:12 UTC
+   └─ foo 4.0.2 is excluded because the package is uploaded after the cutoff date of 2021-12-12T12:12:12Z
```
wolfv: I think this is fine but wondering if we shoudl display thsi in user timezone, a bit more human friendly? But alos fine to merge as is (and I guess it maps better to the input value).

wolfv (approved):

crates/rattler_solve/tests/backends/snapshots/backends__backends__resolvo__exclude_newer_error.snap
baszalmstra: Let me ask clanker!

baszalmstra: @wolfv Done!

PR #1907, opened 2025-12-04
feat: add `--force` to prefix.dev upload and improve function signatures

wolfv:
Adds `--force` for `prefix` upload, and wraps a few naked bools.
Also adds an enum for the attestation setting because that was unclear.

### DIFF
--- a/crates/rattler_upload/src/upload/prefix.rs
+++ b/crates/rattler_upload/src/upload/prefix.rs
@@ -161,7 +161,7 @@ pub async fn upload_package_to_prefix(
             .into_diagnostic()?;
 
         if prefix_data.force.is_enabled() {
-            url.set_query(Some("force=true"));
+            url.query_pairs_mut().append_pair("force", "true");
         }
 
         // Determine attestation path based on attestation source


crates/rattler_upload/src/upload/prefix.rs
```diff
@@ -152,12 +155,16 @@ pub async fn upload_package_to_prefix(
...
         let file_size = package_file.metadata().into_diagnostic()?.len();
-        let url = prefix_data
+        let mut url = prefix_data
             .url
             .join(&format!("api/v1/upload/{}", prefix_data.channel))
             .into_diagnostic()?;
 
-        // Generate attestation if we're using trusted publishing and it was requested
+        if prefix_data.force.is_enabled() {
+            url.set_query(Some("force=true"));
```
baszalmstra: Can we use [`query_pairs_mut`](https://docs.rs/url/latest/url/struct.Url.html#method.query_pairs_mut) and insert that way?

wolfv: Done, thanks!

baszalmstra (approved):

PR #1918, opened 2025-12-09
experimental: add support for wheel packages to repodata parsing

travishathaway:
### Description

This pull request is directly related to this work:

- https://github.com/conda-incubator/conda-pypi/issues/168

Please consider this just one idea for how it could be implemented, and a way to start a conversation. This is not meant to be merged in its current state.

### DIFF
--- a/crates/rattler_conda_types/src/minimal_prefix_record.rs
+++ b/crates/rattler_conda_types/src/minimal_prefix_record.rs
@@ -235,6 +235,7 @@ impl MinimalPrefixRecord {
             depends: Vec::new(),
             constrains: Vec::new(),
             features: None,
+            filename: None,
             legacy_bz2_size: None,
             license: None,
             license_family: None,
--- a/crates/rattler_conda_types/src/repo_data/mod.rs
+++ b/crates/rattler_conda_types/src/repo_data/mod.rs
@@ -130,6 +130,11 @@ pub struct PackageRecord {
     /// mutually exclusive features.
     pub features: Option<String>,
 
+    /// Optionally the filename of the package
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "fn")]
+    pub filename: Option<String>,
+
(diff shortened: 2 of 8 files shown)


baszalmstra:
Thanks for taking the time to contribute this. Overall this particular part looks fine to me!

But before merging this though I would like to understand much better what this does. Is there a CEP (even draft) that goes over the details?

Are the filenames still conda filenames? They follow a strict pattern that is relied on in this code.

Can you also add some test repodata to both show what it would look like and to have some form of regression data.

travishathaway:
Hi @baszalmstra,

Thanks for the quick response. Just to give you some more background on this, we are currently implementing a conda plugin that will allow us to install wheel files by performing a solve against a `repodata.json` file that contains a top-level `packages.whl` key. For more information about how this repodata file will look, you can check out an example here:

- https://github.com/conda-incubator/conda-pypi/issues/168#issuecomment-3612939126

> Are the filenames still conda filenames? They follow a strict pattern that is relied on in this code.

No, they are not. They are `.whl` files. I'll be happy to add necessary edits elsewhere in the repository to make sure it's fully supported. As well as adding the extra tests too.

Yes, a CEP would be great here to define the exact format of this new section option section in `repodata.json`. I would be happy to write this.

travishathaway:
Hey @baszalmstra,

I just wanted to post this here for visibility:

- https://github.com/conda-incubator/conda-pypi/issues/190#issuecomment-3636363242

This pull request is far from complete (i.e. the many `TODO`s I had to add today 🙈), but it's nice to know that it kind of works! 😅

The thing that makes me the uneasiest about it is the way I setup repodata file type merging. I think it's okay to assume that the `.conda` file type is always going to be the preferred file type to choose by default. I'm just not sure I've handled the `both` case correctly. Now, that there's three, it also feels weird calling it "both".

Anyways, these are things to be spelled out and debated in a proper CEP 🤝 🔏 , but I just wanted to give my thoughts on it so far here too.

PR #1922, opened 2025-12-09
feat: enable sharded repodata by default

baszalmstra:
Enable sharded repodata by default.


wolfv (approved):

baszalmstra:
Let me check if we should do something similar in the python bindings before we merge this.

PR #1931, opened 2025-12-12
feat: add verify subcommand for rattler-bin

wolfv:
This adds a sigstore verification subcommand to rattler, backed by the new `rattler_sigstore` crate (now on **sigstore 0.9**, matching `main`).

## Running the example

Build/run with the `sigstore` feature and point it at a signed package; the signatures are fetched from `<package-url>.v0.sigs` by default:

```console
$ cargo run --bin rattler --features sigstore -- \
    verify-package https://prefix.dev/github-releases/osx-arm64/actionlint-1.7.12-h60d57d3_0.conda

Downloading package from https://prefix.dev/github-releases/osx-arm64/actionlint-1.7.12-h60d57d3_0.conda...
Package: actionlint-1.7.12-h60d57d3_0.conda
Verifying signatures...
✓ Package signature verified successfully!
  Identity: https://github.com/hunger/octoconda/.github/workflows/octoconda.yaml@refs/heads/main
  Issuer:   https://token.actions.githubusercontent.com
  Signed:   2026-03-31T02:58:31Z
```

You can also verify a local file (a signatures URL is required in that case), and optionally pin the expected signer:

```console
$ cargo run --bin rattler --features sigstore -- \
    verify-package ./actionlint-1.7.12-h60d57d3_0.conda \
    --signatures-url https://prefix.dev/github-releases/osx-arm64/actionlint-1.7.12-h60d57d3_0.conda.v0.sigs \
    --identity 'https://github.com/hunger/octoconda/.github/workflows/octoconda.yaml@refs/heads/main' \
    --issuer   'https://token.actions.githubusercontent.com'
```

Verification runs in `Require` mode, so a missing/invalid signature (or a signer that does not match `--identity`/`--issuer`) exits non-zero.

## Notes

- Merged latest `main` and upgraded to **sigstore 0.9** (`sigstore-sign`/`sigstore-trust-root`/`sigstore-verify`), in line with the bump on `main`.
- Migrated the new code off `chrono` onto `jiff` (the `Signed:` timestamp above), since `main` dropped `chrono`.
- The `verify-package` command no longer reimplements verification; it delegates to `rattler_sigstore::verify_package`. The cargo feature was renamed `sigstore-verify` → `sigstore` for consistency with `rattler_cache`.


crates/rattler_sigstore/src/verify.rs
```diff
@@ -0,0 +1,298 @@
...
+}
+
+/// Verify a package using pre-computed SHA256 digest instead of full bytes.
+///
+/// This is more efficient for large packages where we already have the digest.
+/// The digest should be a hex-encoded SHA256 hash string.
+pub async fn verify_package_by_digest(
+    policy: &VerificationPolicy,
+    package_url: &Url,
+    package_sha256: &str,
```
baszalmstra: Can we use the already parsed version here?

baszalmstra (changes requested):
Thanks, this looks like a good direction, but I think the cache integration needs a few fixes before merge.

`VerificationPolicy::Require` can silently skip verification if `CacheMetadata.sha256` is `None`. `get_or_fetch_from_url_with_retry` only verifies inside `if let Some(sha256) = &result.sha256`; there is no required-mode error otherwise. Please fail closed when verification is required and no digest is available.

Verification failure currently removes `result.path`, but `get_or_fetch` may have returned an existing cache hit rather than a freshly downloaded package. That means a transient signature/trust-root/network failure can delete an existing cache entry. Please track whether this call fetched the package, or verify before committing the temp extraction into the final cache path.

`VerificationConfig::signatures_url` builds the default from `package_url.as_str()`, so package URLs with query strings produce the wrong `.v0.sigs` URL. It also silently falls back if an explicit signatures URL is invalid. I think this should return `Result<Url>` and append `.v0.sigs` to the URL path, not the full serialized URL.

The docs say unmapped channels are skipped in `Warn` mode when no default publishers are configured, but the implementation treats `publishers == None` as unconstrained and accepts any valid signer. Please either change the behavior to match the docs or make the unconstrained behavior explicit.

Channel publisher matching is raw `starts_with` on URLs. That is too easy to misconfigure, especially without a trailing slash. Please normalize channel URLs or compare URL path segment boundaries.

TrustedRoot::production().await is called per package verification. In PackageCache, that means one trusted-root load per package. That may be expensive in installs with many packages.

I would also add tests for: required mode with no sha256, cached package not deleted on verification failure, package URLs with query strings, warn mode with no publisher config, and invalid explicit `--signatures-url`.

pavelzw:
this is outdated. there is now `rattler_sigstore` #2813

PR #1932, opened 2025-12-15
fix: always emit depends in index.json

baszalmstra:
See https://github.com/conda-forge/tzdata-feedstock/issues/36


wolfv:
Ahhhh the snapshots :)

PR #1934, opened 2025-12-15
chore: Use lefthook no_auto_install

pavelzw:
### Description

I hate when tools install something without my consent, especially if it makes my lazygit workflow slower. 

---

# Explicit dependencies

|Dependency|Before|After|Change|Environments|
|-|-|-|-|-|
|[lefthook](https://prefix.dev/channels/conda-forge/packages/lefthook)|2.0.9|2.0.11|Patch Upgrade|*all*|

# Implicit dependencies

|Dependency|Before|After|Change|Environments|
|-|-|-|-|-|

[^1]: **Bold** means explicit dependency.
[^2]: Dependency got downgraded.


baszalmstra:
Thanks!

PR #1935, opened 2025-12-15
feat: Add support for downloading info files via range requests

pavelzw:
### Description

needed for https://github.com/pavelzw/pixi-diff-to-markdown/issues/82

inspired by https://github.com/conda/conda-package-streaming/blob/d8dc1c62ee01688011ceeea8c87a11e5dbe479cf/conda_package_streaming/lazy_wheel.py#L294

done with claude code and codex

### DIFF
--- a/crates/rattler_package_streaming/src/reqwest/snapshots/rattler_package_streaming__reqwest__sparse__tests__fetch_package_file_sparse-2.snap
+++ b/crates/rattler_package_streaming/src/reqwest/snapshots/rattler_package_streaming__reqwest__sparse__tests__fetch_package_file_sparse-2.snap
@@ -1,0 +1,6 @@
+---
+source: crates/rattler_package_streaming/src/reqwest/sparse.rs
+expression: about_json
+---
+channels:
+  - "https://conda.anaconda.org/conda-forge/"


baszalmstra:
Nice! 

No offence but this looks pretty AI generated so Ill take a very close look tomorrow. 

Have you considered using astral-async-zip in combination with async_http_range_reader reader?

You should consider adding a number of tests to verify the number of requests performed by the library.

pavelzw:
> No offence but this looks pretty AI generated so Ill take a very close look tomorrow

got me there ^^

> Have you considered using astral-async-zip in combination with async_http_range_reader reader?

not yet, will look into it 

> You should consider adding a number of tests to verify the number of requests performed by the library

👍🏻

Cargo.toml
```diff
@@ -37,7 +37,7 @@ async-fd-lock = "0.2.0"
 fs4 = "0.13.1"
 async-once-cell = "0.5.4"
 async-trait = "0.1.88"
-astral_async_zip = { version = "0.0.17", default-features = false }
+astral_async_zip = { git = "https://github.com/pavelzw/rs-async-zip", branch = "pub", default-features = false }
```
pavelzw: once the proposed changes look good to you, i'll open a PR over at astral-async-zip.

https://github.com/astral-sh/rs-async-zip/compare/main...pavelzw:rs-async-zip:pub

crates/rattler_package_streaming/src/reqwest/range.rs
```diff
@@ -0,0 +1,768 @@
...
+                .error_for_status()
+                .expect_err("non-success status should error");
+            Err(ExtractError::ReqwestError(error.into()))
+        }
+    }
+}
+
+/// Builds a `Range` header value for a byte segment.
+fn build_range_header(start: u64, len: u64) -> Option<String> {
+    let end = start.checked_add(len)?.checked_sub(1)?;
```
pavelzw: we could also panic using `expect` in these cases as it won't happen anyway in "normal" use. any opinions on this?

crates/rattler_package_streaming/src/reqwest/fetch.rs
pavelzw, line 1: we might eventually want to build an `AsyncRead` + `AsyncSeek` HTTP range reader and delegate almost all ZIP logic to `async_zip::base::read::file` / `seek::ZipFileReader`. Codex says this is feasible, but a larger change.

crates/rattler_package_streaming/tests/extract.rs
```diff
@@ -309,12 +308,8 @@ fn test_extract_data_descriptor_package_fails_streaming_and_uses_buffering() {
...
         extract_conda_via_buffering(File::open(package_path).unwrap(), &target_dir).unwrap();
 
-    let combined_result = json!({
-        "sha256": format!("{:x}", new_result.sha256),
-        "md5": format!("{:x}", new_result.md5),
-    });
-
-    insta::assert_snapshot!(combined_result, @r###"{"sha256":"6a5d6d8a1a7552dbf8c617312ef951a77d2dac09f2aeaba661deebce603a7a97","md5":"a1d1adb5a5dc516dfb3dccc7b9b574a9"}"###);
+    insta::assert_snapshot!(format!("{:x}", new_result.sha256), @r"6a5d6d8a1a7552dbf8c617312ef951a77d2dac09f2aeaba661deebce603a7a97");
+    insta::assert_snapshot!(format!("{:x}", new_result.md5), @r"a1d1adb5a5dc516dfb3dccc7b9b574a9");
```
baszalmstra: Having two snapshots in a single test is annoying because you need to trigger both individually if you simply use `cargo test`. I prefer the old behavior.

crates/rattler_package_streaming/src/reqwest/range.rs
baszalmstra: No this look good to me!

crates/rattler_package_streaming/src/reqwest/range.rs
```diff
@@ -0,0 +1,768 @@
...
+    {
+        debug!("central directory found in tail bytes");
+        bytes
+    } else {
+        // CD is not (fully) in our tail, need to fetch it
+        debug!("central directory not in tail, fetching separately");
+        match fetch_range_bytes(&client, &url, cd_start_in_file, cd_size).await? {
+            Some(bytes) => bytes,
+            None => {
+                return fetch_package_file_full_download(&client, &url, CondaArchiveType::Conda)
```
baszalmstra: Should we maybe add a debug statement here as well?

crates/rattler_package_streaming/src/reqwest/fetch.rs
baszalmstra: Yes that is possible because we can use https://github.com/prefix-dev/async_http_range_reader. You can see how that is done in `uv` here: https://github.com/astral-sh/uv/blob/02e92aed3b2d06383e4f78ed31b961080f168848/crates/uv-client/src/remote_metadata.rs . I feel like the current approach requires quite a bit finicking with zip internals which I feel is not strictly needed.

crates/rattler_package_streaming/src/reqwest/fetch.rs
```diff
@@ -0,0 +1,164 @@
...
+        let client = reqwest_middleware::ClientWithMiddleware::from(reqwest::Client::new());
+        let url = Url::parse(
+            "https://conda.anaconda.org/conda-forge/noarch/tzdata-2024b-hc8b5060_0.conda",
+        )
+        .unwrap();
+
+        let about_json: AboutJson = fetch_package_file_from_url(client, url).await.unwrap();
+
+        // tzdata package should have license info
+        assert!(about_json.license.is_some());
```
baszalmstra: This is a bit of a week test. Given that the package will likely not change, maybe we can just assert the expected contents. Or the hash of the contents maybe?

crates/rattler_package_streaming/src/reqwest/fetch.rs
```diff
@@ -0,0 +1,164 @@
...
+    P::from_str(&String::from_utf8_lossy(&content))
+        .map_err(|e| ExtractError::ArchiveMemberParseError(P::package_path().to_owned(), e))
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[tokio::test]
+    async fn test_fetch_index_json_from_conda_forge() {
```
baszalmstra: We can also setup a local server to download the file from, I think our mock implementation also supports range requests and otherwise it would not be hard to add I think. That makes the tests much faster and way less flaky.

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,222 @@
...
+            .iter()
+            .enumerate()
+            .find(|(_, e)| {
+                e.filename()
+                    .as_str()
+                    .map(|f| f.starts_with("info-") && f.ends_with(".tar.zst"))
+                    .unwrap_or(false)
+            })
+            .ok_or(ExtractError::MissingComponent)?;
+
```
baszalmstra: You should make sure that the bytes of the info are prefetched. If the initial 64k was not enough that ensures that the all the required missing bytes are read in a single request. You can do that on the `AsyncHttpRangeReader` itself but its a bit of a hassle to get mutable access to it. 

This is how uv does that: https://github.com/astral-sh/uv/blob/02e92aed3b2d06383e4f78ed31b961080f168848/crates/uv-client/src/remote_metadata.rs#L90

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,222 @@
...
+/// let client = ClientWithMiddleware::from(Client::new());
+/// let url = Url::parse("https://conda.anaconda.org/conda-forge/noarch/tzdata-2024b-hc8b5060_0.conda").unwrap();
+///
+/// let archive = SparseRemoteArchive::new(client, url).await.unwrap();
+/// let index_json: IndexJson = archive.read().unwrap();
+/// println!("Package: {}", index_json.name.as_normalized());
+/// # }
+/// ```
+pub struct SparseRemoteArchive {
+    files: HashMap<PathBuf, Vec<u8>>,
```
baszalmstra: Ah its too bad that we still have to read and decompress the entire info section here. Maybe I was wrong and we should make this stateless regardless. We should stream the info section bytes and stop streaming when we have located the file requested.

baszalmstra: `info` sections also often include the `info/recipe` folder which can still be relatively large.

crates/rattler_package_streaming/src/reqwest/fetch.rs
pavelzw: [`700b8fa` (this PR)](https://github.com/conda/rattler/pull/1935/commits/700b8faec99cc91251dc3e3e31a4e8aff51b1f60)

crates/rattler_package_streaming/src/reqwest/fetch.rs
pavelzw: [`20f5023` (this PR)](https://github.com/conda/rattler/pull/1935/commits/20f5023bab04358aecbfbd80b4bf07b81f02a5b2) adds some boilerplate ai-generated code to do so

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: done in [`80f20e3` (this PR)](https://github.com/conda/rattler/pull/1935/commits/80f20e37dc72d940fffafc0e5f7304ba07a1fbda) and [`6a7cf86` (this PR)](https://github.com/conda/rattler/pull/1935/commits/6a7cf86af7577bc7b7b25536041412f2eb4425fe).

i verified that it doesn't fetch the full info part by serving a file with a very large (1gb) recipe folder and running `cargo run --bin rattler -- inspect http://localhost:80/osx-arm64/my-package-0.0.1-ha646cdb_1.conda` (this was pretty fast)

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: [`41ceb25` (this PR)](https://github.com/conda/rattler/pull/1935/commits/41ceb2540d8ead792021c315c07dbf66e3fa3b6c)

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,222 @@
...
+/// Fetch the raw bytes of a single file from a remote `.conda` package's info
+/// section using HTTP range requests.
+///
+/// Streams the zstd data directly from the server through an async decompressor
+/// and tar reader, stopping as soon as the target file is found. Only the bytes
+/// needed to reach the target file are downloaded and decompressed.
+///
+/// Returns an error if the URL does not point to a `.conda` archive, the server
+/// does not support range requests, or the file is not found.
+pub async fn fetch_file_from_remote_conda(
```
baszalmstra, line 62: This also only fetches data from the `info/` section. Maybe we can make this generic enough that you can request any file?

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,222 @@
...
+        if path.as_ref() == target_path {
+            let size = entry.header().size().map_err(ExtractError::IoError)?;
+            let mut buf = Vec::with_capacity(size as usize);
+            entry
+                .read_to_end(&mut buf)
+                .await
+                .map_err(ExtractError::IoError)?;
+            return Ok(buf);
+        }
+    }
```
baszalmstra: If you extract this into a function you can use the same in the download full example probably, especially when directly streaming a tar.bz2.

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,222 @@
...
+/// and tar reader, stopping as soon as the target file is found. Only the bytes
+/// needed to reach the target file are downloaded and decompressed.
+///
+/// Returns an error if the URL does not point to a `.conda` archive, the server
+/// does not support range requests, or the file is not found.
+pub async fn fetch_file_from_remote_conda(
+    client: ClientWithMiddleware,
+    url: Url,
+    target_path: &Path,
+) -> Result<Vec<u8>, ExtractError> {
```
baszalmstra: Maybe we should return `Result<Option<Vec<u8>>, ExtractError>` where `None` means the file is not contained in the package?

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,222 @@
...
+/// let index_json: IndexJson = fetch_package_file_sparse(client, url).await.unwrap();
+/// println!("Package: {}", index_json.name.as_normalized());
+/// # }
+/// ```
+pub async fn fetch_package_file_sparse<P: PackageFile>(
+    client: ClientWithMiddleware,
+    url: Url,
+) -> Result<P, ExtractError> {
+    let bytes = fetch_file_from_remote_conda(client, url, P::package_path()).await?;
+    P::from_str(&String::from_utf8_lossy(&bytes))
```
baszalmstra: Maybe we can add from_slice to PackageFile?

crates/rattler_package_streaming/src/reqwest/test_server.rs
```diff
@@ -0,0 +1,88 @@
...
+/// size, which is exactly what `async_http_range_reader` sends on its initial
+/// probe for small files.
+pub async fn serve_file(file_path: impl AsRef<Path>) -> Url {
+    let file_path = file_path.as_ref();
+    let file_name = file_path.file_name().unwrap().to_string_lossy().to_string();
+    let data = std::fs::read(file_path).expect("failed to read test file");
+
+    let app = axum::Router::new().fallback(move |req: Request| {
+        let data = data.clone();
+        async move { serve_with_ranges(&data, req.headers()) }
```
baszalmstra: You can use use `tower_http::services::ServeDir` to serve a file over http with support for range requests.

crates/rattler_package_streaming/src/reqwest/fetch.rs
```diff
@@ -0,0 +1,161 @@
...
+
+    let response = client
+        .get(url.clone())
+        .send()
+        .await
+        .map_err(ExtractError::ReqwestError)?
+        .error_for_status()
+        .map_err(|e| ExtractError::ReqwestError(e.into()))?;
+
+    let bytes = response
```
baszalmstra: This will now read the entire package to memory which is a bit wasteful. I think we could also streaming read the tar.bz2 and stop streaming when we find the file (same as what you do with the sparse reading). For the .conda case you can also stream the files in the zip file (without seeking), stream to the right tar archive, and then use the same trick to stream that and stopping when you found the right file. I think that can be a huge safer in terms of how much data needs to be read.

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: [`fa2a769` (this PR)](https://github.com/conda/rattler/pull/1935/commits/fa2a7698d02f96ae3dfd73e3bcb5ce46fdbe0721)

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: [`271d89f` (this PR)](https://github.com/conda/rattler/pull/1935/commits/271d89f971a9f0f3d03cec797bc7ddc99ac9bb2d)

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: [`71a0a95` (this PR)](https://github.com/conda/rattler/pull/1935/commits/71a0a950b52b02a766ddcbd48096c40f65881996)

crates/rattler_package_streaming/src/reqwest/test_server.rs
pavelzw: [`23ef968` (this PR)](https://github.com/conda/rattler/pull/1935/commits/23ef96893f091c6a50c0a68e6c0a69272214c6f0)

crates/rattler_package_streaming/src/reqwest/fetch.rs
pavelzw: [`df4a3c1` (this PR)](https://github.com/conda/rattler/pull/1935/commits/df4a3c136d2e3adc8aa2ae1d797517edf8bd4516)

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: [`be2bae3` (this PR)](https://github.com/conda/rattler/pull/1935/commits/be2bae3dbb91c221d77dc6620120eb501100dd54)

crates/rattler_conda_types/src/package/mod.rs
```diff
@@ -52,6 +52,14 @@ pub trait PackageFile: Sized {
     /// error.
     fn from_str(str: &str) -> Result<Self, std::io::Error>;
 
+    /// Parses the object from a byte slice, using a format appropriate for the file type.
+    ///
+    /// The bytes are interpreted as a UTF-8 string (with lossy conversion for
+    /// invalid sequences) and then parsed via [`Self::from_str`].
```
baszalmstra:
```suggestion
```

I think we should remove this since it might be implementation specific. 

Also we should implement this for all the current implementations using [`serde_json::from_slice`](https://docs.rs/serde_json/latest/serde_json/fn.from_slice.html)

crates/rattler_package_streaming/src/tokio/async_read.rs
```diff
@@ -222,3 +223,41 @@ pub async fn extract_conda_via_buffering(
         total_size,
     })
 }
+
+/// Returns the ZIP entry prefix (`"info-"` or `"pkg-"`) that contains the
+/// given `target_path` inside a `.conda` archive.
+#[cfg(feature = "reqwest")]
+pub(crate) fn conda_entry_prefix(target_path: &Path) -> &'static str {
+    if target_path.starts_with("info") {
```
baszalmstra, line 231:
```suggestion
    if target_path.starts_with("info/") {
```

Otherwise `info-custom.txt` would not be found.

baszalmstra: Maybe also `info\\`? Not sure how that works on windows.

baszalmstra: Although https://doc.rust-lang.org/std/path/struct.Path.html#method.starts_with says it only takes into consideration entire path components.. Maybe you can add a unit test to verify that this works as expected?

crates/rattler_package_streaming/src/reqwest/test_server.rs
```diff
@@ -0,0 +1,61 @@
...
+
+    format!("http://{}:{}/{file_name}", addr.ip(), addr.port())
+        .parse()
+        .unwrap()
+}
+
+/// Clamp suffix ranges (`bytes=-N`) that exceed the file size so `ServeDir`
+/// doesn't return 416. Per RFC 9110 §14.1.2, a suffix range exceeding the
+/// representation length should select the entire representation.
+async fn clamp_suffix_range(
```
baszalmstra, line 41: Doesnt servedir do this correctly?

baszalmstra: Maybe mention that in the comment if thats the case.

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,236 @@
...
+        })
+        .ok_or(ExtractError::MissingComponent)?;
+
+    // Prefetch the entire info entry in a single HTTP request so the
+    // streaming pipeline doesn't trigger many small range requests.
+    let entry = &zip_reader.file().entries()[index];
+    let offset = entry.header_offset();
+    let size = entry.header_size() + entry.compressed_size();
+    let buffer_size: u64 = 8192;
+    let size = size.div_ceil(buffer_size) * buffer_size;
```
baszalmstra: This might require a comment. Why only in chunks of 8192 bytes?

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,236 @@
...
+        .prefetch(offset..offset + size)
+        .await;
+
+    // Get a streaming reader for the ZIP entry (futures::io::AsyncRead).
+    // This does NOT buffer the entire entry — bytes are fetched on demand
+    // via HTTP range requests as the downstream decompressor/tar reader
+    // consumes them.
+    //
+    // The pipeline borrows zip_reader, so we scope it in a block to release
+    // the borrow before accessing the inner reader for debug logging.
```
baszalmstra: This comment doesnt make sense.

crates/rattler_package_streaming/src/reqwest/sparse.rs
```diff
@@ -0,0 +1,236 @@
...
+///
+/// Returns `Ok(None)` if the file is not found in the archive.
+/// Returns an error if the URL does not point to a `.conda` archive or the
+/// server does not support range requests.
+pub async fn fetch_file_from_remote_conda(
+    client: ClientWithMiddleware,
+    url: Url,
+    target_path: &Path,
+) -> Result<Option<Vec<u8>>, ExtractError> {
+    debug!("fetching {:?} from remote archive {}", target_path, url);
```
baszalmstra: You can also instrument the function, then the context is propagated to other debug statements as well. See https://docs.rs/tracing/latest/tracing/attr.instrument.html

baszalmstra: You can also use rattler_redact to redact any credentials from the url.

crates/rattler_package_streaming/src/reqwest/test_server.rs
pavelzw: nope it doesn't unfortunately

pavelzw: > so `ServeDir` doesn't return 416.

isn't this basically done through that?

crates/rattler_conda_types/src/package/mod.rs
pavelzw: [`b4f87fd` (this PR)](https://github.com/conda/rattler/pull/1935/commits/b4f87fdabe6354500883c5ac47d86cf81d1e533d)

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: i think we can actually just remove this... [`466135f` (this PR)](https://github.com/conda/rattler/pull/1935/commits/466135f768d8b4334c3db885e34bf27b45a38fa9)

crates/rattler_package_streaming/src/tokio/async_read.rs
pavelzw: i'll add a unit test that checks `info-custom.txt` as well 👍🏻

pavelzw: [`2a0ea11` (this PR)](https://github.com/conda/rattler/pull/1935/commits/2a0ea11204295b038c30b68e900610a2200725f9)

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: seems like the agent didn't do their refactoring job fully 😅 removed in [`6e52878` (this PR)](https://github.com/conda/rattler/pull/1935/commits/6e52878ce0f000d7e20749bcf95ff8dce5f464d1)

crates/rattler_package_streaming/src/tokio/async_read.rs
baszalmstra: I think a unit test for the function would be more appropriate.

crates/rattler_package_streaming/src/reqwest/sparse.rs
pavelzw: good idea 👍🏻 

[`18a55e8` (this PR)](https://github.com/conda/rattler/pull/1935/commits/18a55e83c09959c268ec79ab18e742ee5958d341)

crates/rattler_package_streaming/src/tokio/async_read.rs
pavelzw: [`3ec648a` (this PR)](https://github.com/conda/rattler/pull/1935/commits/3ec648a1f6a70eefb72ee66f30ece07499197c58)

test-data/package-with-info/with-info-0.1.0-h4616a5c_0.conda
pavelzw, line 1: do you want me to revert this now that we have the unit tests?

baszalmstra: Yeah lets.

pavelzw: [`339a683` (this PR)](https://github.com/conda/rattler/pull/1935/commits/339a683b581621d466b692163e3e455f3ce586d2)

PR #1938, opened 2025-12-17
fix: validate new prefix length against old prefix

maurosilber:
### Description

If the target prefix is larger than the placeholder prefix, it would ~overwrite into other regions of the binary~ truncate the prefix. It would likely lead to broken binaries: https://discord.com/channels/1082332781146800168/1212673297347518464/1450859607072182374

### Possible issue outside rattler

Being a public function, this function might be being used outside rattler in some valid ways that would now return an error. For example, relocating an environment:

1. On installation, the new prefix is not longer than the old prefix (the build prefix placeholder) -> ✅
2. On relocation, the new prefix is longer than the old prefix (but not longer that the build prefix placeholder) -> returns an error although it would be correct.

### Alternatives

It might be more "rustic" to encode this restriction into a type such as:

```rust
struct PlaceholderChange {
    old_placeholder: &str,
    new_placeholder: &str,
}

pub fn copy_and_replace_cstring_placeholder(
    mut source_bytes: &[u8],
    mut destination: impl Write,
    placeholder_chance: PlaceholderChange,
) -> Result<(), std::io::Error>
```

instead of accepting any placeholders:

```rust
pub fn copy_and_replace_cstring_placeholder(
    mut source_bytes: &[u8],
    mut destination: impl Write,
    prefix_placeholder: &str,
    target_prefix: &str,
) -> Result<(), std::io::Error>
```

but would be a breaking change in a public function.

Also, it would not solve the relocation use case (for which I have an idea that would ~imply changing the `info/has_prefix` file, which would be an even bigger breaking change in the conda ecosystem~ not work after thinking about it some more).

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -696,6 +696,13 @@ pub fn copy_and_replace_cstring_placeholder(
     let old_prefix = prefix_placeholder.as_bytes();
     let new_prefix = target_prefix.as_bytes();
 
+    if new_prefix.len() > old_prefix.len() {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::InvalidData,
+            "target prefix cannot be longer than the placeholder prefix",
+        ));
+    }
+
     let finder = memchr::memmem::Finder::new(old_prefix);
 
     loop {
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

wolfv:
Thank you @maurosilber!

PR #1949, opened 2025-12-25
chore: Add support for riscv64gc-unknown-linux-musl

pavelzw:
### Description

with riscv support on conda-forge gaining traction (https://github.com/conda-forge/conda-forge.github.io/issues/1744), we should also support it in rattler-based tooling.

xref https://doc.rust-lang.org/rustc/platform-support.html

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -4421,7 +4421,7 @@ packages:
   license: BSD-3-Clause
 - conda: crates/rattler_index
   name: rattler_index
-  version: 0.27.7
+  version: 0.27.8
   build: h1ec88cb_0
   subdir: osx-arm64
   variants:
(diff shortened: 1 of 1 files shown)


baszalmstra:
Nice! Do we also support this platform in the `Platform` type? Especially in the `Platform::current `function?

pavelzw:
We already have this: https://github.com/conda/rattler/blob/5e19ffc74d11e1861e278de9b21ebf783f0bb2b5/crates/rattler_conda_types/src/platform.rs#L127-L128

If it wasn't supported there should be a compile error

https://github.com/conda/rattler/blob/5e19ffc74d11e1861e278de9b21ebf783f0bb2b5/crates/rattler_conda_types/src/platform.rs#L142

baszalmstra (approved):

baszalmstra (approved):

PR #1950, opened 2025-12-31
fix: preserve entitlements when codesigning on macOS

XhstormR:
Fix: Preserve entitlements when codesigning on macOS

This fixes an issue on macOS where codesigning did not preserve entitlements, preventing programs requiring special permissions (such as virtualization tools) from running.

Key Changes:

1. Added the `--preserve-metadata=entitlements` parameter to the `codesign` command.

2. Expanded signing support to all macOS platforms (Intel and Apple Silicon).

3. Updated related documentation comments.

Root Cause:

- Hard links, copying, etc., invalidated existing signatures.

- The old implementation only re-signed on ARM64; signatures were invalidated on Intel Macs but not re-signed.

- The old implementation did not preserve entitlements, causing programs requiring special permissions to fail to run.

Fixed Effects:

- All macOS binaries can now be signed correctly.

- Existing entitlements (such as virtualization and network permissions) are preserved.

- Tools requiring special permissions, such as lima-vm, work correctly.

Fixes: https://github.com/prefix-dev/pixi/issues/4990

### DIFF
--- a/crates/rattler/src/install/apple_codesign.rs
+++ b/crates/rattler/src/install/apple_codesign.rs
@@ -4,7 +4,7 @@ use super::LinkFileError;
 use std::path::Path;
 
 /// Controls the behavior of the [`super::link_package`] function when it encounters a binary that needs
-/// to be signed on macOS ARM64 (Apple Silicon).
+/// to be signed on macOS (both Intel and Apple Silicon).
 #[derive(Debug, Clone, Copy, PartialEq, Default)]
 pub enum AppleCodeSignBehavior {
     /// Do nothing (do not attempt to sign any binary)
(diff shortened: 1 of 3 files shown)


baszalmstra:
Im not sure if we also require signing on intel macos. Can you clarify?

XhstormR:
After fresh install the sign become invalid:
```
└──❯ codesign -vvv /Users/user/.pixi/envs/lima-vm/bin/limactl
/Users/user/.pixi/envs/lima-vm/bin/limactl: invalid signature (code or signature have been modified)
In architecture: x86_64
```

Current Signing Logic

- See link.rs:245. Signing is only triggered when the following conditions are met simultaneously:

- The file has a prefix_placeholder (path replacement is required)

- The content has actually been modified (content_changed = true)

- The file is executable or a Mach-O binary

- The target platform is macOS

- The file mode is Binary

Key point: Re-signing is only performed when prefix replacement causes content changes

Why does Intel Mac need re-signing?

- Because prefix replacement modifies binary content:

- Many conda packages embed the installation path in their binaries (e.g., programs compiled with Go or Rust)

- limactl is likely to have a prefix_placeholder

- Installation path replacement → Content modification → Signature invalidation

- Old code only re-signed on ARM64 → Signature invalidated on Intel Mac without re-signing → Program cannot run

baszalmstra:
I appreciate the AI summary but this has always worked fine on intel macs, so I dont understand why we need this now?

(apart from the entitlements)

XhstormR:
When you say "work fine," do you mean the signature is valid? In my actual testing on intel macos, after a clean install of lima-vm, the signature is invalid.

baszalmstra:
Ah so this also already caused issues on intel macos? edit: I read the corresponding issue, I see the problem!

XhstormR:
Comments have been updated

baszalmstra (approved):

baszalmstra:
Thanks!!

PR #1954, opened 2026-01-06
fix: correctly handle channel_alias with path segments (#1953)

### LINKED ISSUE #1953: [BUG] Parsing channel names with channel_alias containing path segments
### Checklist

### What happened?

When using something like Artifactory, the channel alias would be `https://example.jfrog.io/artifactory/api/conda/`.  I'd then expect that these two channels should be identical

```
let channel_alias = "https://example.jfrog.io/artifactory/api/conda/"
let channel_config = ChannelConfig {
    channel_alias: Url::from_str(channel_alias)?,
    root_dir: std::env::current_dur()?,
};
let channel_from_name = Channel::from_name("conda_example", &channel_config);
let channel_from_str = Channel::from_str(!format("{channel_alias}conda_example"), &channel_config);
```

They end up being different though because the later is parsed with `name: Some("artifactory/api/conda/conda_example")`

### Additional Context

_No response_

DCchoudhury15:
### Description

This PR fixes **#1953**, where channel names were incorrectly parsed when the `channel_alias` (base URL) contained path segments. This is a common scenario in Artifactory or JFrog environments (e.g., `https://example.com/artifactory/api/conda/`).

Previously, the parser assumed the channel name started immediately after the domain/host, which resulted in names like `artifactory/api/conda/my-channel` instead of the expected `my-channel`.

### Key Changes

* **Idiomatic Prefix Stripping**: Updated `strip_channel_alias` in `ChannelConfig` to use the Rust-idiomatic `strip_prefix`. This handles the full URL string (including path segments) safely and avoids manual index slicing that Clippy flagged as a warning.
* **Prioritized Alias Matching**: Modified `Channel::from_str` to check if a provided URL is a sub-path of the configured `channel_alias` using the updated stripping logic. This ensures that if a URL matches the alias prefix, the remaining suffix is correctly identified as the channel name.
* **Regression Testing**: Added a specific test case `test_issue_1953_artifactory_path_parsing` in `crates/rattler_conda_types/src/channel/mod.rs` using the Artifactory URL structure provided in the bug report.

### Verification

I have verified these changes using the following steps:
1.  **Regression Test**: Ran the new test case specifically:
    `pixi run test -p rattler_conda_types --lib channel::tests::test_issue_1953_artifactory_path_parsing` (Passed)
2.  **Full Crate Testing**: Ran all 780 tests in `rattler_conda_types` to ensure no regressions in standard conda-forge or file-path parsing. (Passed)
3.  **Linting & Formatting**: 
    - Verified `pixi run cargo clippy -p rattler_conda_types -- -D warnings` passes.
    - Verified `pixi run cargo fmt --all` was applied.

Fixes #1953

### DIFF
--- a/crates/rattler_conda_types/src/channel/mod.rs
+++ b/crates/rattler_conda_types/src/channel/mod.rs
@@ -63,20 +63,13 @@ impl ChannelConfig {
     /// `https://conda.anaconda.org/conda-forge` when the channel alias is
     /// `https://conda.anaconda.org`).
     pub fn strip_channel_alias(&self, base_url: &Url) -> Option<String> {
-        let base_str = base_url.as_str();
-        let alias_str = self.channel_alias.as_str();
-
-        // Clippy fix: Use strip_prefix instead of starts_with + manual slicing
-        if let Some(res) = base_str.strip_prefix(alias_str) {
-            let name = res.trim_matches('/');
-            return if name.is_empty() {
-                None
-            } else {
-                Some(name.to_string())
-            };
-        }
-        None
+        base_url
+            .as_str()
+            .strip_prefix(self.channel_alias.as_str())
+            .map(|s| s.trim_matches('/').to_string())
+            .filter(|s| !s.is_empty())
     }
+
     /// Returns the canonical name of a channel with the given base url.
     pub fn canonical_name(&self, base_url: &Url) -> String {
         if let Some(stripped) = self.strip_channel_alias(base_url) {
@@ -220,17 +213,13 @@ impl Channel {
 
         let channel = if parse_scheme(channel).is_some() {
             let url = Url::parse(channel)?;
-            if let Some(name) = config.strip_channel_alias(&url) {
-                Channel {
-                    platforms,
-                    base_url: url.into(),
-                    name: Some(name),
-                }
-            } else {
-                Channel {
-                    platforms,
-                    ..Channel::from_url(url)
-                }
+            let name = config.strip_channel_alias(&url);
+            let base_channel = Channel::from_url(url);
+
+            Channel {
+                name: name.or(base_channel.name),
+                platforms,
+                ..base_channel
             }
         } else if is_path(channel) {
             #[cfg(target_arch = "wasm32")]


crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -58,22 +58,29 @@ impl ChannelConfig {
...
     /// This returns the name of the channel (for example "conda-forge" for
     /// `https://conda.anaconda.org/conda-forge` when the channel alias is
     /// `https://conda.anaconda.org`).
     pub fn strip_channel_alias(&self, base_url: &Url) -> Option<String> {
-        base_url
-            .as_str()
-            .strip_prefix(self.channel_alias.as_str())
-            .map(|s| s.trim_end_matches('/').to_string())
-    }
+        let base_str = base_url.as_str();
```
baszalmstra: Instead of rewriting the entire function we could just do something along these lines: 

```rust
base_url
    .as_str()
    .strip_prefix(self.channel_alias.as_str())
    .map(|s| s.trim_end_matches('/').to_string())
    .take_if(|s| !s.is_empty()) // your change
```

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -213,9 +220,17 @@ impl Channel {
...
         let channel = if parse_scheme(channel).is_some() {
             let url = Url::parse(channel)?;
-            Channel {
-                platforms,
-                ..Channel::from_url(url)
+            if let Some(name) = config.strip_channel_alias(&url) {
+                Channel {
+                    platforms,
+                    base_url: url.into(),
+                    name: Some(name),
```
baszalmstra: Shouldnt we use `..Channel::from_url` in this case too? But overwrite the name if we have one?

DCchoudhury15: @baszalmstra i have made the changes

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -213,9 +213,17 @@ impl Channel {
...
+                    platforms,
+                    ..Channel::from_url(url)
+                }
+            } else {
+                Channel {
+                    platforms,
+                    ..Channel::from_url(url)
+                }
             }
```
baszalmstra: These branches are now very similar. Maybe you can simplify?

DCchoudhury15:
@baszalmstra thanks for suggesting the corrections patiently , i am relatively new to the conda org and am still learning ,  will make better code quality prs further down the line.

baszalmstra:
No worries, this was great! Thanks for contributing!

PR #1534, opened 2025-07-18
feat: add support for loong64 platform to rattler

wszqkzqk:
### Description

[LoongArch](https://docs.kernel.org/arch/loongarch/introduction.html) is a new RISC ISA developed by loongson. There are already a lot of [community support and testing](https://www.phoronix.com/search/LoongArch) about it. This PR adds loong64 support to rattler.

PS. [debian names this architecture `loong64`](https://wiki.debian.org/Ports/loong64) so I followed this custom, but the gnu and rust toolchain still uses `loongarch64`. It's not a typo.

PR #1957, opened 2026-01-07
refactor: rename `loong64` to `loongarch64`

baszalmstra:
Unfortunately we had to rename `loong64` to `loongarch64` because the latter is used by both mamba and conda.

Support was added in #1534 with an explicit note that `loong64` follows debian standards. Unfortunately, conda already went another way so to stay compatible we have to change this. 

@wszqkzqk This might affect you. Are you already using this in the wild?


wszqkzqk:
I think that's fine, but I'm curious why so many architectures in projects use short names.

PR #1965, opened 2026-01-09
refactor: Make s3 optional for rattler_index

pavelzw:
### Description

We shouldn't include aws dependencies if we are not planning on using s3. With this change, we get the following:

```
❯ cargo tree -i aws-config --no-default-features --features native-tls
aws-config v1.8.12
└── rattler_s3 v0.1.18 (/Users/pavel/projects/rattler/crates/rattler_s3)
```

### DIFF
--- a/crates/rattler_s3/Cargo.toml
+++ b/crates/rattler_s3/Cargo.toml
@@ -9,6 +9,14 @@ license.workspace = true
 edition.workspace = true
 readme.workspace = true
 
+[features]
+default = ["rustls-tls"]
+# Note: aws-smithy-http-client does not support native-tls, only rustls variants.
+# This feature is provided for API compatibility with other rattler crates but
+# will use rustls under the hood.
+native-tls = ["rustls-tls"]
+rustls-tls = ["aws-smithy-http-client/rustls-ring"]
+
 [dependencies]
 clap = { workspace = true, optional = true }
 thiserror = { workspace = true }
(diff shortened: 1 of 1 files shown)


crates/rattler-bin/Cargo.toml
```diff
@@ -17,9 +17,24 @@ name = "rattler"
 path = "src/main.rs"
 
 [features]
-default = ["rustls-tls"]
-native-tls = ["reqwest/native-tls", "rattler/native-tls", "rattler_repodata_gateway/native-tls", "rattler_networking/native-tls", "rattler_cache/native-tls"]
-rustls-tls = ["reqwest/rustls-tls", "reqwest/rustls-tls-native-roots", "rattler/rustls-tls", "rattler_repodata_gateway/rustls-tls", "rattler_networking/rustls-tls", "rattler_cache/rustls-tls"]
+default = ["rustls-tls", "s3", "gcs"]
```
pavelzw, line 20: i made s3 optional here as well to make it easier to spot accidental aws crate leakages by running `cargo tree -i aws-config --no-default-features --features native-tls` in the project root

crates/rattler_index/src/lib.rs
pavelzw, line 1: we might want to split the s3 part off to a separate file which i can do in a follow-up pr.
then, we could also make the fs feature optional as well for people who want rattler-index only for s3 channels

PR #1969, opened 2026-01-11
fix(lock): omit channel: null for source packages in v6 lockfile

DCchoudhury15:
### Description

This PR fixes an issue where `channel: null` was explicitly serialized into the lockfile for packages that do not have a channel (specifically source/git dependencies). This aligns the V6 lockfile serialization with the expected behavior of omitting `null` fields to keep the file clean.

**Changes:**
- Added a `skip_channel` helper function in `crates/rattler_lock/src/parse/models/v6/conda_package_data.rs`.
- Applied `#[serde(skip_serializing_if = "skip_channel")]` to the `channel` field.
- Updated test snapshots to remove the `channel: null` line.
- Ran `cargo fmt` and `cargo clippy`.

**Reason for change:**
This addresses a bug reported in `prefix-dev/pixi` (see [prefix-dev/pixi#5180](https://github.com/prefix-dev/pixi/issues/5180)).

### DIFF
--- a/crates/rattler_lock/src/builder.rs
+++ b/crates/rattler_lock/src/builder.rs
@@ -426,7 +426,9 @@ mod test {
     };
     use url::Url;
 
-    use crate::{CondaBinaryData, LockFile, PypiPrereleaseMode};
+    use crate::{
+        CondaBinaryData, CondaPackageData, CondaSourceData, LockFile, PypiPrereleaseMode, UrlOrPath,
+    };
 
     #[test]
     fn test_merge_records_and_purls() {
@@ -599,5 +601,88 @@ mod test {
                 Some(mode)
             );
         }
+    }
+
+    #[test]
+    fn test_channel_null_serialization_roundtrip() {
+        use std::str::FromStr;
+
+        // 1. Setup Binary Package
+        let binary_pkg_record = PackageRecord::new(
+            PackageName::new_unchecked("binary-pkg"),
+            Version::from_str("1.0.0").unwrap(),
+            "build_string".to_string(),
+        );
+
+        let binary_data = CondaBinaryData {
+            package_record: binary_pkg_record,
+            location: UrlOrPath::from_str(
+                "https://conda.anaconda.org/conda-forge/linux-64/binary-pkg-1.0.0.tar.bz2",
+            )
+            .unwrap(),
+            // Updated to use the correct type from main
+            file_name: "binary-pkg-1.0.0.tar.bz2"
+                .parse::<DistArchiveIdentifier>()
+                .unwrap(),
+            channel: None,
+        };
+
+        // 2. Setup Source Package
+        let source_pkg_record = PackageRecord::new(
+            PackageName::new_unchecked("source-pkg"),
+            Version::from_str("1.0.0").unwrap(),
+            "build_string".to_string(),
+        );
+
+        let source_data = CondaSourceData {
+            package_record: source_pkg_record,
+            location: UrlOrPath::from_str("git+https://github.com/org/repo").unwrap(),
+            variants: None,
+            package_build_source: None,
+            input: None,
+            sources: std::collections::BTreeMap::default(),
+        };
+
+        // 3. Build & Serialize
+        let lock_file = LockFile::builder()
+            .with_conda_package(
+                "default",
+                Platform::Linux64,
+                CondaPackageData::Binary(binary_data),
+            )
+            .with_conda_package(
+                "default",
+                Platform::Linux64,
+                CondaPackageData::Source(source_data),
+            )
+            .finish();
+
+        let serialized = lock_file.render_to_string().unwrap();
+
+        // 4. Assertions
+        assert_eq!(
+            serialized.matches("channel: null").count(),
+            1,
+            "Should have exactly one 'channel: null'"
+        );
+
+        let parsed = LockFile::from_str(&serialized).unwrap();
+        let packages: Vec<_> = parsed
+            .environment("default")
+            .unwrap()
+            .packages(Platform::Linux64)
+            .unwrap()
+            .collect();
+
+        let binary_pkg = packages
+            .iter()
+            .find(|p| p.name() == "binary-pkg")
+            .unwrap()
+            .as_binary_conda()
+            .unwrap();
+        assert!(
+            binary_pkg.channel.is_none(),
+            "Binary package channel should be None"
+        );
     }
 }
--- a/crates/rattler_lock/src/parse/models/v6/conda_package_data.rs
+++ b/crates/rattler_lock/src/parse/models/v6/conda_package_data.rs
@@ -15,11 +15,11 @@ use rattler_conda_types::package::{
     ArchiveIdentifier, CondaArchiveType, DistArchiveIdentifier, DistArchiveType,
 };
 use rattler_conda_types::{
-    package::CondaArchiveIdentifier, utils::TimestampMs, BuildNumber, ChannelUrl, NoArchType,
-    PackageName, PackageRecord, PackageUrl, VersionWithSource,
+    utils::TimestampMs, BuildNumber, ChannelUrl, NoArchType, PackageName, PackageRecord,
+    PackageUrl, VersionWithSource,
 };
 use rattler_digest::{serde::SerializableHash, Md5Hash, Sha256Hash};
-use serde::{Deserialize, Serialize};
+use serde::{Deserialize, Deserializer, Serialize};
 use serde_with::serde_as;
 use url::Url;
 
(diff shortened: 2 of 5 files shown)


baszalmstra:
Can you add a unit test that shows that roundtrip serialization works for a number of cases?  

Especially in case there is a binary packages which explictly had its channel set to None?

DCchoudhury15:
hi @baszalmstra 
I've added the roundtrip unit test to `crates/rattler_lock/src/builder.rs` as requested! 

It verifies that `channel: null` is omitted from the serialized output and that the lockfile correctly roundtrips for both source and binary packages.

For full transparency: I used AI assistance to help construct the initial boilerplate for this test case, but I have manually verified the logic and ensured all linting/formatting checks pass.

crates/rattler_lock/src/builder.rs
```diff
@@ -588,4 +588,92 @@ mod test {
...
+        let packages: Vec<_> = env
+            .packages(Platform::Linux64)
+            .expect("Packages for linux-64 should exist")
+            .collect();
+
+        // ASSERTION 2: Verify we have 2 packages
+        assert_eq!(packages.len(), 2, "Should have 2 packages after roundtrip");
+
+        // ASSERTION 3: Verify specific packages are present
+        assert!(
```
baszalmstra: This test does not actually verify that the channel fields is correctly roundtripped.

baszalmstra (changes requested):

DCchoudhury15:
hi @baszalmstra ,
I've updated the test to explicitly verify that `binary_pkg.channel` remains `None` after roundtripping.
I also switched the test fixture to use a local file path (`/tmp/...`) instead of a URL. This prevents the parser from automatically inferring the channel (which happens with standard URLs), ensuring we are truly testing that the `channel: None` state is preserved.

baszalmstra:
But that means it doesnt roundtrip when you dont use a local path right?

Maybe we should infer the channel to be null for git based urls instead?

DCchoudhury15:
My understanding is that for standard Anaconda URLs (like `conda.anaconda.org`), the parser is designed to automatically infer the channel upon deserialization. So even if we save it as `null`, it comes back as `Some(...)` because of that inference logic. Is that correct?
That was the main reason I switched to a local path for the Binary test case—I wanted to bypass that auto-inference to verify that the serialization logic itself is correctly omitting the field when it's supposed to.

Regarding the **Git-based URLs**:
If i am not wrong `git+https` URLs don't trigger that inference mechanism, so they do essentially "roundtrip" as `None` (which seems to solve the original issue where `channel: null` was being written out explicitly).
Am i correct or did i miss something in the solution?

baszalmstra:
Yes thats what the inference logic is there for. But if we have a reford where inference does not produce the desired value for a field we overwrite it. Thats why we have to store `channel: null` if the `channel` field is `None` while inference would yield a different value.

In the case of a git package the code does store `channel: null`, which indicates to me that inference would compute something else.

Does that make sense?

DCchoudhury15:
Your point about using `null` to override incorrect inference! That definitely holds true for Binary packages.
However, I noticed a key structural difference for Git packages:
* `CondaBinaryData` has a `pub channel: Option<Url>` field.
* `CondaSourceData` does not seem to have a `channel` field at all (I couldn't set it in the struct initialization in the test).
Since the Source struct doesn't have a slot to store a channel, it seems physically impossible for the parser to "infer and store" a value there. This suggests that for Git packages, the `channel: null` in the lockfile was purely a serialization artifact (generic model field) rather than a functional override.
Does that match your understanding of the `CondaSourceData` struct? If so, then removing `null` should be perfectly safe for these packages, isnt it?

baszalmstra:
Yeah correct. So it seems that just for source dependencies we dont set the channel field correctly during serialization?

DCchoudhury15:
Yes, exactly!
Since source packages don't store a channel, the V6 model wrapper was defaulting to writing `channel: null`. My fix ensures we just skip that field when it's `None` (which is always true for source packages), treating it as a clean omission rather than an explicit `null`.
So this PR purely fixes that serialization artifact.

baszalmstra:
But the fix is incomplete because for normal http binary packages, if they dont have a channel, they are incorrectly roundtripped. This is clear since you had to use a local package to get the test to work 🙃

DCchoudhury15:
hi @baszalmstra ,
You were right regarding the regression—I missed that `channel: null` is functionally required for binary packages to prevent the parser from re-inferring the channel from the URL.
I have reworked the implementation to handle the two cases separately:
1.  **Binary Packages:** If `channel` is `None` (but inference would find one), we now correctly serialize it as `null`. This acts as an explicit override.
2.  **Source Packages:** We strictly omit the field, ensuring the lockfile remains clean for git dependencies.
To support this, I also added a `deserialize_channel_field` helper. This ensures that an explicit `null` in the file is deserialized as `Some(None)` (blocking inference), whereas a missing field is deserialized as `None` (allowing inference).
Tests have been updated to cover both scenarios and now pass successfully.

crates/rattler_lock/src/parse/models/v6/conda_package_data.rs
```diff
@@ -329,3 +336,23 @@ fn strip_trailing_slash(url: &Url) -> Cow<'_, Url> {
         Cow::Borrowed(url)
     }
 }
+
+/// Helper to deserialize the channel field, distinguishing "null" from "missing".
+fn deserialize_channel_field<'de, 'a, D>(
```
baszalmstra, line 344: I think we can remove this function now right?

crates/rattler_lock/src/parse/models/v6/conda_package_data.rs
```diff
@@ -290,8 +294,11 @@ impl<'a> From<&'a CondaPackageData> for CondaPackageDataModel<'a> {
             noarch: (package_record.noarch != derived_noarch)
                 .then_some(Cow::Borrowed(&package_record.noarch)),
             variants: variants.map(Cow::Borrowed),
-            channel: (channel != derived.channel.as_ref())
-                .then_some(Cow::Owned(normalized_channel)),
+            channel: if value.as_binary().is_some() {
+                (channel != derived.channel.as_ref()).then_some(Cow::Owned(normalized_channel))
+            } else {
+                None
+            },
```
baszalmstra:
```suggestion
            channel: if value.as_binary().is_some() && channel != derived.channel.as_ref() {
                Some(Cow::Owned(normalized_channel))
            } else {
                None
            },
```

crates/rattler_lock/src/builder.rs
```diff
@@ -588,4 +588,106 @@ mod test {
...
+        // 1. Setup a Binary Package that explicitly has NO channel (None)
+        let binary_pkg_record = PackageRecord::new(
+            PackageName::new_unchecked("binary-pkg"),
+            Version::from_str("1.0.0").unwrap(),
+            "build_string".to_string(),
+        );
+
+        let binary_data = CondaBinaryData {
+            package_record: binary_pkg_record,
+            // REVERTED: Use a standard URL to prove the "channel: null" override works!
```
baszalmstra: This comment only makes sense in the context of this PR, but once merged it looses its context.

crates/rattler_lock/src/builder.rs
```diff
@@ -588,4 +588,106 @@ mod test {
...
+            location: UrlOrPath::from_str(
+                "https://conda.anaconda.org/conda-forge/linux-64/binary-pkg-1.0.0.tar.bz2",
+            )
+            .unwrap(),
+            file_name: "binary-pkg-1.0.0.tar.bz2".to_string(),
+            channel: None,
+        };
+
+        // 2. Setup a Source Package (which should naturally have None)
```
baszalmstra:
```suggestion
        // 2. Setup a Source Package (which should naturally have no channel)
```

DCchoudhury15:
I tested removing `deserialize_channel_field` as suggested, but it breaks the roundtrip test (`test_channel_null_serialization_roundtrip` fails).

The reason i guess is Standard Serde deserialization treats explicit `null` and "missing field" exactly the same (both become `None`).
* If `channel` becomes `None` (missing), the model falls back to auto-inference (deriving `conda-forge` from the URL).
* We need `null` to deserialize as `Some(None)` to strictly block that inference.

I've applied your other suggestions (cleaner `if` logic and comment fixes) in the latest commit!

DCchoudhury15:
@baszalmstra I have resolved the merge conflicts with main (adapting to the new DistArchiveIdentifier types) and fixed the regression where source paths were incorrectly identified as binaries. All ci tests are passing. up for review.

DCchoudhury15:
@baszalmstra up for review

baszalmstra:
We are currently in a big refactor for lockfile v7. Ill hold of on merging this to see if this is still relevant when thats merged.

baszalmstra:
I think this is pretty much outdated with the new lockfile version.

Issue #1966, opened 2026-01-09
Add support for `.whl` archives to rattler


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

The following pull request contains an incomplete implementation of wheel support for rattler:

- https://github.com/conda/rattler/pull/1918/

We want to add proper support for wheel archive files though and that pull request should be seen as a starting point. For this issue, we just want to focus on adding the new `ArchiveType` support. Adding support for parsing this new repodata format will be handled in a separate issue.

### Why is this needed?

This work is being done so that rattler can properly implement the [Repodata Wheel Support CEP](https://github.com/conda/ceps/pull/145) when/if it is passed.

### What should happen?

- [ ] Look for all the places where the existing archive file types are referenced (e.g. `tar.bz` and `.conda`) and add a `.whl` entry
- [ ] Examine existing tests for these new formats and add new ones where it makes sense
- [ ] Update `py-rattler` to account for this new archive type (`py-rattler/package_streaming`) (add new tests for this there too)

### Additional Context

I understand that this conversation is still in flux and the CEP is yet to be formally adopted, so please take all of this with a grain a salt and acknowledge that it may change.

danyeaw:
Closing since initial support for wheel as an ArchiveType is now merged.

PR #1970, opened 2026-01-12
feat: add support for `.whl` package types to rattler

travishathaway:
### Description

fixes: #1966 

I decided to close [this pull request](https://github.com/conda/rattler/pull/1918) and open a new one to keep the changes limited.

### DIFF
--- a/crates/rattler_conda_types/src/package/archive_type.rs
+++ b/crates/rattler_conda_types/src/package/archive_type.rs
@@ -23,7 +23,7 @@ impl ArchiveType {
             .map(|(_, archive_type)| archive_type)
     }
 
-    /// Tries to determine the type of a Conda archive from its magic bytes.
+    /// Tries to determine the type of archive from its magic bytes.
     pub fn try_from_magic_bytes<T: AsRef<[u8]>>(bytes: T) -> Option<ArchiveType> {
         // https://en.wikipedia.org/wiki/List_of_file_signatures
         let bytes = bytes.as_ref();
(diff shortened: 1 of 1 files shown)


baszalmstra:
Thanks for working on this! Let me know if you need a review or other support!

travishathaway:
Hey @baszalmstra,

So, one thing we could use some advice on is adding support for wheels in `rattler_package_streaming`. I initially thought that `.conda` and `.whl` files were both using zip compression, but that's not the case because `.conda` actually uses `.zstd`.

How difficult do you think it would be to add support for the `.zip` format to that crate? Or, am I still getting this wrong somehow?

Seeing that this isn't strictly necessary for adding support for the solver and what we want to do, we could potentially just skip this, but if it isn't too difficult to add, I would like to add it now, so we don't have to do it later.

Thanks!

baszalmstra:
`.conda` packages are actually zip files! But they contain two `.tar.zst` encoded/compressed archives. 

I think adding extraction of `.whl` files would definitely be nice! I think it should be pretty straightforward to also add support for extracting these.

I would also be ok if we first add a `todo!` or `unimplemented!` and an issue, and merge this PR without.

crates/rattler_index/src/lib.rs
```diff
@@ -279,6 +279,9 @@ fn parse_package_buffer(buffer: opendal::Buffer, filename: &str) -> std::io::Res
     match archive_type {
         ArchiveType::TarBz2 => package_record_from_tar_bz2_reader(reader),
         ArchiveType::Conda => package_record_from_conda_reader(reader),
+        ArchiveType::Whl => Err(std::io::Error::other(
+            "Package type \".whl\" not yet supported.",
+        )),
```
travishathaway, line 284: Not sure if this is the best way to raise this error. Please advise if there is a better way.

travishathaway:
Hi @baszalmstra,

I'm working together on this PR with @ryanskeith. We were able to come up with an implementation for extracting wheel files. Ryan's still working on adding a few more test, but we're both feeling pretty good about the PR and think it's ready for a review.

crates/rattler_conda_types/src/package/archive_identifier.rs
```diff
@@ -15,12 +15,14 @@ pub struct ArchiveIdentifier {
...
     /// The build string of the package.
     pub build_string: String,
-    /// The archive type of the package (tar.bz2 or conda)
+    /// The archive type of the package (tar.bz2, conda or whl)
     pub archive_type: ArchiveType,
 }
 
 impl ArchiveIdentifier {
     /// Converts the archive identifier into a filename for a Conda package.
+    /// TODO: This doesn't handle ArchiveType:Whl correctly because we don't
```
baszalmstra: This is problematic because that should essentially always work, its the contract of the type. 

Instead I think we have to be a bit more rigorous. How about we split the enum into a `CondaArchiveType` and a `DistArchiveType`?  `CondaArchiveType` is solely focused on conda archives and `DistArchiveType` on anything "installable". 

We also have to implement a `CondaArchiveIdentifier` and a `WheelArchiveIdentifier`. A `DistArchiveIdentifier` would then be an enum wrapping the other two.

crates/rattler_config/src/config/build.rs
```diff
@@ -62,10 +62,12 @@ impl FromStr for PackageFormatAndCompression {
             number if number.parse::<i32>().is_ok() => {
                 let number = number.parse::<i32>().unwrap_or_default();
                 match archive_type {
-                    ArchiveType::TarBz2 => {
+                    ArchiveType::TarBz2 | ArchiveType::Whl => {
```
baszalmstra, line 65: If we split the ArchiveType into conda and non conda, this also makes more sense.

crates/rattler_index/src/lib.rs
baszalmstra: Splitting the enum type into distinct types will solve this problem.

crates/rattler_package_streaming/src/tokio/async_read.rs
```diff
@@ -183,6 +183,15 @@ pub async fn extract_conda_via_buffering(
     .await
 }
 
+/// Extracts the contents of a `.whl` package archive.
+/// This will perform on-the-fly decompression by streaming the reader.
+pub async fn extract_whl(
+    reader: impl AsyncRead + Send + 'static,
+    destination: &Path,
+) -> Result<ExtractResult, ExtractError> {
+    extract_conda_internal(reader, destination, crate::read::extract_whl_via_streaming).await
```
baszalmstra, line 192: We can also do this at a later stage but I think we can take a page out of uv for this implementation: https://github.com/astral-sh/uv/blob/0c08cba0c414ba6e5cb06608624993a9fba0a179/crates/uv-extract/src/stream.rs#L46

Note that we also have https://github.com/conda/rattler/pull/1855 which implements streaming zip extraction.

crates/rattler_conda_types/src/package/archive_identifier.rs
travishathaway: I think this is a good idea. This was probably the most awkward part about defining the archives this way.

Just wondering, are you saying we should remove `ArchiveType` entirely? Wouldn't that risk breaking external dependencies?

baszalmstra: Yeah it will be a breaking change, but thats what we have semver for. The automations in this repo will take care of the right version bumps. 

You can also leave the current names the same and come up with new ones for the other types.

travishathaway: Glad that semver allows us to make break changes here.

I'm a little confused on your suggestion though. Are you saying that `ArchiveIdentifier` and `ArchiveType` stay the same and I just need to add `DistArchiveType` and `WheelArchiveIdentifier`? Or Do I need to rename `ArchiveIdentifier` to `CondaArchiveIdentifier` and `ArchiveType` to `CondaArchiveType`?

Thanks!

baszalmstra:
With https://github.com/conda/rattler/pull/1996 merged, can we close this?

PR #1971, opened 2026-01-13
chore: Update feature-list used for tests in pixi.toml

hunger:
## Description

Update feature list in the pixi `test` task.

`experimental_extras` nop longer exists, so `pixi run test` fails right now.


hunger:
Sorry, I rebased this from on top of my change to `origin/main` and `jj` fixed up the change -- which apparently is not needed anymore in `main`. Sorry for the noise.

PR #1972, opened 2026-01-13
feat: allow skipping validation completely and add atomic extraction

baszalmstra:
Implements the `Skip` validation logic which skips validating packages in the cache completely. 

This also implements first downloading and extracting to a temporary directory, followed by renaming the directory to avoid corrupted downloads being considered valid.

Closes #1819

### DIFF
--- a/crates/rattler_cache/src/package_cache/mod.rs
+++ b/crates/rattler_cache/src/package_cache/mod.rs
@@ -693,20 +693,76 @@ where
             .write_revision_and_sha(new_revision, given_sha)
             .await?;
 
-        // Fetch the package.
-        fetch_fn(path.clone())
-            .await
-            .map_err(|e| PackageCacheLayerError::FetchError(Arc::new(e)))?;
+        // Create a temporary directory in the same parent folder for atomic extraction.
+        // Using the same parent ensures the rename operation is atomic (same filesystem).
+        let parent_dir = path.parent().ok_or_else(|| {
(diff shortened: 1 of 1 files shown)


crates/rattler_cache/src/package_cache/mod.rs
```diff
@@ -693,20 +693,64 @@ where
...
-            sha256: given_sha.copied(),
-            path,
-            index_json: None,
-            paths_json: None,
-        })
+        // Create a temporary directory in the same parent folder for atomic extraction.
+        // Using the same parent ensures the rename operation is atomic (same filesystem).
+        let temp_dir = {
+            let mut temp_path = path.as_os_str().to_owned();
+            temp_path.push(&format!(".tmp-{}", std::process::id()));
```
wolfv: we could also use `tempfile::TemporaryDirectory::new_in(..)` or something along this lines? At least for temporary files an API like thsi exists.

baszalmstra: I made the chance!

wolfv:
I think this is good. The only code path I am worried about is that deletion might take a long time.

So we could hypothetically run into some weird concurrency issues where:

A finishes writing foobar
B deletes foobar
C finishes writing foobar
B wants to write foobar

or something like this, if people run lots of pixi's in parallel and our file locking doesn't work - although I expect that we still have that and it should guard against most of these!

PR #1982, opened 2026-01-14
chore: Remove `editable` flag from Pypi packages

### LINKED ISSUE #1975: rattler-lock v7: remove editable from pypi records
We want to remove `editable` from the pypi records. This field is not used.

hunger:
Fixes #1975


hunger:
I forgot to remove getters for `editable` in `py-rattler`. Let's try again.

Issue #1980, opened 2026-01-13
rattler-lock v7: allow configurable platforms


baszalmstra:
It is not possible to associate any data with the "platforms" stored in the lock-file. This is something that would be needed for https://github.com/prefix-dev/pixi/issues/4458 .

Currently we store platforms as:

```yaml
environments :
    <env-name>: 
        packages:
            <platform1>: []
            <platform2>: []
```

I think for maximum compatibility we would need to do something like:

```yaml
platforms: 
  - name: my-win-64-with-cuda:
    subdir: win-64
    virtual-packages:
    - __cuda=12.0
  - my-win-64:
    subdir: win-64    
environments:
  <env-name>:
    packages:
      my-win-64-with-cuda: []
      my-win-64: []
```

Note that the order of platforms is important. When selecting the "best" compatible platform the first one wins.

Ideas welcome.

hunger:
Do I understand correctly that this is a new abstraction over our current platforms and system-requirements. This new abstraction happens to have the same name as the platform we use today -- which is basically the `subdir` in this new layer?

for rattler-lock, you hit the right level of abstraction I think, but for pixi this is way too low-level though.

hunger:
This is now merged into feature/lockfile-v7

PR #1991, opened 2026-01-19
feat: Add platform section into lock file

hunger:
This adds a `platforms` section into the lock file, as described in #1980

The py-rattler change was done with heavy help from opus 4.5.

Fixes: #1980


hunger:
Questions I still have after writing this PR:

 * Why does the `v7/multiple-source-packages-with-variants-lock.yml` claim to be version 6 and why does it fail when I update that to version 7? This wass OK before I rebased on top of todays feature/lockfile-v7 branch.
 * The API for adding platforms (and then packages to a env/platform) is not idea. Better ideas welcome :-)

crates/rattler_lock/src/parse/deserialize.rs
```diff
@@ -592,12 +637,65 @@ fn parse_from_lock_legacy<P>(
...
+        .map(|(_, subdir)| -> Result<_, ParseCondaLockError> {
+            let name = crate::platform::PlatformName::try_from(subdir.as_str())?;
+
+            Ok(crate::platform::PlatformData {
+                name,
+                subdir: *subdir,
+                virtual_packages: Vec::new(),
+            })
+        })
+        .collect::<Result<Vec<_>, _>>()
```
baszalmstra, line 665:
```suggestion
        .collect()
```

crates/rattler_lock/src/parse/deserialize.rs
```diff
@@ -592,12 +637,65 @@ fn parse_from_lock_legacy<P>(
...
+                    Ok(platform)
+                } else {
+                    Err(ParseCondaLockError::DuplicatePlatformName(
+                        platform.name.to_string(),
+                    ))
+                }
+            }
+            Err(e) => Err(e.into()),
+        })
+        .collect::<Result<Vec<_>, _>>()
```
baszalmstra, line 688:
```suggestion
        .collect()
```

crates/rattler_lock/src/parse/mod.rs
```diff
@@ -28,18 +27,37 @@ pub enum ParseCondaLockError {
...
+    #[error("Duplicate platform name `{0}` found")]
+    DuplicatePlatformName(String),
+
     #[error("missing field `{0}` for package {1}")]
     MissingField(String, UrlOrPath),
 
+    #[error("`platforms` were not supported in lockfile version {0}")]
+    UnexpectedPlatforms(FileFormatVersion),
+
+    #[error("Environment `{environment}` is using an unregistered platform `{platform}`")]
```
baszalmstra:
```suggestion
    #[error("Environment `{environment}` is using an unknown platform `{platform}`")]
```

Just a nitpick but "unregistered" sounds like you would have to do something.

crates/rattler_lock/src/platform.rs
```diff
@@ -0,0 +1,219 @@
+/// An error that can occur when parsing a platform from a string.
+#[derive(Debug, Clone, thiserror::Error, Eq, PartialEq)]
+#[allow(missing_docs)]
+pub enum ParsePlatformError {
+    #[error("Failed to parse '{0}' as a PlatformName")]
```
baszalmstra:
```suggestion
    #[error("failed to parse '{0}' as a PlatformName")]
```

We try to start sentences with a lower-case so you can easily chain them in a sentence. E.g. `"blabla failed because {0}"`

crates/rattler_lock/src/platform.rs
```diff
@@ -0,0 +1,219 @@
...
+        let value = value.to_string();
+        if value
+            .chars()
+            .all(|c| c.is_alphanumeric() || c == '_' || c == '-')
+        {
+            Ok(Self(value))
+        } else {
+            Err(ParsePlatformError::ParsePlatformNameError(value))
+        }
+    }
```
baszalmstra, line 56: Should this use the From<String> impl to avoid duplication?

crates/rattler_lock/src/platform.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// Represents a package with platform-specific information.
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct PlatformData {
+    /// The name of the platform.
+    pub name: PlatformName,
+    /// The subdir of the platform.
+    pub subdir: rattler_conda_types::Platform,
+    /// The list of virtual conda packages.
+    pub virtual_packages: Vec<String>,
```
baszalmstra, line 166: We could also consider storing this as a `GenericVirtualPackage` but it would add a dependency on `rattler_virtual_package`.. Maybe we could vendor that type here?

py-rattler/examples/install_to_prefix.py
```diff
@@ -24,7 +24,14 @@ async def _install(
...
         raise ValueError(f"Environment {environment_name} not found in lock file {lock_file_path}")
-    records = environment.conda_repodata_records_for_platform(platform)
+    # Find the lock platform matching the requested platform
+    lock_platform = next(
+        (p for p in environment.platforms() if p.name == str(platform)),
+        None,
+    )
+    if lock_platform is None:
+        raise ValueError(f"Platform {platform} not found in environment {environment_name}")
+    records = environment.conda_repodata_records_for_platform(lock_platform)
```
baszalmstra, line 34: Maybe we should add a method to find the first "appropriate" platform?

crates/rattler_lock/src/parse/deserialize.rs
hunger: I don't think I can leave this one out.

crates/rattler_lock/src/parse/mod.rs
hunger: But you are supposed to register the platform:-) I'll change it.

crates/rattler_lock/src/platform.rs
hunger: Of course we should!

PR #1995, opened 2026-01-20
refactor: make package record optional for source dependencies

baszalmstra:
### Description

Make the package record data optional for source records. 

- [x] Still need to update the python bindings
- [x] Still need to create a stable hash for source packages


baszalmstra:
The source identifier hash now also roundtrips, this ensures that if you have a lock-file, you load it, dont modify it and recreate it, the source identifier is kept.

Issue #1967, opened 2026-01-09
Add support for parsing repodata with wheel support


travishathaway:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

The following pull request began to explore how this could be done:

- https://github.com/conda/rattler/pull/1918

This issue should be seen as a proper implementation of this feature taking into account that experimental pull request and everything currently being discussed in [this CEP](https://github.com/conda/ceps/pull/145).

### Why is this needed?

Adds support for repodata with wheel support.

### What should happen?

- [ ] Update the `rattler_repodata_gateway` crate so that it can properly parse repodata containing a `packages.whl` section
- [ ] Update the `py-rattler` bindings so this is possible from Python
- [ ] Add tests in both `rattler_repodata_gateway` and `py-rattler` to ensure this works
  - For example repodata with this field, checkout https://github.com/conda-incubator/conda-pypi/pull/186 

### Additional Context

The following issue should be completed before this issue:

- https://github.com/conda/rattler/issues/1966

Also, I acknowledge that the discussions surrounding the CEP are not completed, so this issue is liable to change to reflect any updates in decisions about the standard.

PR #1996, opened 2026-01-20
feat: add support for `packages.whl` to repodata parsing

travishathaway:
### Description

fixes: #1967

This pull request adds support for parsing `packages.whl` sections in `repodata.json`. The biggest difference in this section of repodata is keys for individual packages can no longer be assumed to be filenames. This is why we now rely on the presence of a `fn` field in individual `PackageRecord`s.

I've been testing this manually with the repodata from this pull request in [conda-pypi](https://github.com/conda-incubator/conda-pypi):

- https://github.com/conda-incubator/conda-pypi/pull/207

Below is an example py-rattler program to show how these new additions can be used:

```python
import asyncio
from pprint import pprint
from rich.console import Console
from rich.table import Table

from rattler import Channel, ChannelConfig, SparseRepoData
from rattler import GenericVirtualPackage
from rattler import MatchSpec
from rattler import PackageName, Version
from rattler import PackageFormatSelection
from rattler import solve_with_sparse_repodata
from rattler import PackageFormatSelection


def solve(deps):
    # Create the mock channel - value isn't relevant
    chn = Channel("http://localhost")

    # Repodata with wheels
    sparse = SparseRepoData(
        chn, 
        "noarch",
        "conda-pypi/tests/conda_local_channel/noarch/repodata.json"
    )

    # Repodata with conda packages
    sparse_conda = SparseRepoData(
        chn, 
        "osx-arm64",
        "/Users/travishathaway/dev/conda-pypi/tests/conda_local_channel/osx-arm64/repodata.json"
    )

    # Setup mock virtual packages so solve works
    virtpkg_unix = GenericVirtualPackage(PackageName("__unix"), Version("15"), "0")
    virtpkg = GenericVirtualPackage(PackageName("__osx"), Version("15"), "0")

    solution = asyncio.run(
        solve_with_sparse_repodata(
            deps,
            [sparse, sparse_conda],
            virtual_packages=[virtpkg, virtpkg_unix],
            package_format_selection=PackageFormatSelection.PREFER_CONDA_WITH_WHL
        )
    )

    table = Table(title="Dependencies")

    table.add_column("Arch", justify="right", style="cyan")
    table.add_column("Package", style="magenta")
    table.add_column("File type", style="green")

    for package in solution:
        table.add_row(
            package.subdir,
            f"{package.name.normalized}-{package.version}",
            package.file_name.split('.')[-1]
        )

    console = Console()
    console.print(table)
```

### DIFF
--- a/py-rattler/rattler/solver/solver.py
+++ b/py-rattler/rattler/solver/solver.py
@@ -178,7 +178,7 @@ async def solve_with_sparse_repodata(
         constraints: Additional constraints that should be satisfied by the solver.
             Packages included in the `constraints` are not necessarily installed,
             but they must be satisfied by the solution.
-        package_format_selection: If defined, will override `use_only_tar_bz2` and use the desired package selection.
+        package_format_selection: Defines which package formats are selected
 
     Returns:
         Resolved list of `RepoDataRecord`s.


crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -138,6 +138,11 @@ pub struct PackageRecord {
     /// mutually exclusive features.
     pub features: Option<String>,
 
+    /// Run exports that are specified in the package.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde(rename = "fn")]
+    pub filename: Option<String>,
```
baszalmstra: If we are going to add a field `filename`, I think it would make much more sense to add a `url` field that is either an absolute URL or a relative URL (relative to the current subdir). That is much more versatile.

baszalmstra: Regardless of the above, lets prefix this with `experimental_` to indicate that this field is still experimental.

baszalmstra: The `RepoDataRecord` also contains a `filename` and a `url` I think we should take this new field into account in at least the `url`.

crates/rattler_repodata_gateway/src/sparse/mod.rs
baszalmstra, line 1: We need a number of unit tests for the new functionality in this file.

py-rattler/rattler/solver/solver.py
```diff
@@ -179,10 +180,18 @@ async def solve_with_sparse_repodata(
             Packages included in the `constraints` are not necessarily installed,
             but they must be satisfied by the solution.
         use_only_tar_bz2: If `True` only `.tar.bz2` packages are used. If `False` `.conda` packages are preferred.
```
baszalmstra: Lets just remove this field and add a `package_format_selection` argument with a default value.

baszalmstra (commented):
I left a few comments. Since @travishathaway is now (hopefully) sleeping I will take care of my own comments. 👍

py-rattler/rattler/solver/solver.py
travishathaway: Totally agree. Was just thinking about backwards compatibility.

py-rattler/rattler/solver/solver.py
```diff
@@ -178,12 +178,11 @@ async def solve_with_sparse_repodata(
         constraints: Additional constraints that should be satisfied by the solver.
             Packages included in the `constraints` are not necessarily installed,
             but they must be satisfied by the solution.
-        use_only_tar_bz2: If `True` only `.tar.bz2` packages are used. If `False` `.conda` packages are preferred.
+        package_format_selection: If defined, will override `use_only_tar_bz2` and use the desired package selection.
```
baszalmstra:
```suggestion
        package_format_selection: Defines which package formats are selected
```

baszalmstra:
I introduced a `WhlPackageRecord` which is a `PackageRecord` and a `url` field. I removed `filename` as discussed on slack. The `url` field is either a relative path which is added to the base url or a complete url. I added a number of tests to the SparseRepoData and change the python code to replace `use_only_tar_bz2` with `package_format_selection`. 

There is still more work to do to integrate this into more subsystems (the Gateway for instance) but for now this should suffice.

PR #1997, opened 2026-01-21
feat: add soft requirements to solvertask

nichmor:
### Description

Add API to expose soft_requirements from resolvo to rattler_solve itself


baszalmstra (commented):
I see now that `soft_requirements` in resolvo require `Solvables` instead of matchspecs. I misunderstood that. That makes the feature far less valuable because it doesnt allow for alternatives.

The way you implemented this is probably also not very performant when there are loads of variants for a single matchspec.

Maybe we should first update resolvo to use requirements for soft_requirements instead of solvables.

baszalmstra:
Closing for now. Please reopen when you want to continue. :)

PR #1999, opened 2026-01-21
refactor: add type for archive file name

baszalmstra:
This PR refactors how we deal with `fn` in `RepoDataRecord` and `PrefixRecord`. Both types also have a `url` which is used to download the package itself. The `fn` field was only used as a unique identifier within a subdir and, therefore, as a key. Both for reading indexing the repodata, and for creating repodata.json (or shards) from existing records. The name of the field, however, does not properly reflect this. 

With the introduction of `packages.whl` this became more obvious. There, the key in the repodata dictionary does not actually represent the filename of the file which makes the name very ambiguous. 

To facilitate/disambiguate this, I changed the field name to `identifier`. I also changed the type from a string to a `DistArchiveIdentifier`, which holds the identifying parts of the identifier (`name`, `version`, `build_string`) and the archive type (`.conda`, `.tar.bz2`, `.whl`).

### DIFF
--- a/py-rattler/tests/unit/test_prefix_record.py
+++ b/py-rattler/tests/unit/test_prefix_record.py
@@ -80,14 +80,14 @@ def test_create_prefix_record() -> None:
 
     repodata_record = RepoDataRecord(
         package_record,
-        file_name="foobar.tar.bz2",
-        url="https://foobar.com/foobar.tar.bz2",
+        file_name="foobar-1.0-abc.tar.bz2",
+        url="https://foobar.com/foobar-1.0-abc.tar.bz2",
         channel="https://foobar.com/win-64",
     )
 
-    assert repodata_record.url == "https://foobar.com/foobar.tar.bz2"
+    assert repodata_record.url == "https://foobar.com/foobar-1.0-abc.tar.bz2"
     assert repodata_record.channel == "https://foobar.com/win-64"
-    assert repodata_record.file_name == "foobar.tar.bz2"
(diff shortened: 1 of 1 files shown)


crates/rattler_conda_types/src/package/archive_identifier.rs
```diff
@@ -1,21 +1,59 @@
-use super::CondaArchiveType;
```
baszalmstra, line 1: This file introduces new types.

crates/rattler_conda_types/src/repo_data_record.rs
```diff
@@ -18,7 +18,7 @@ pub struct RepoDataRecord {
 
     /// The filename of the package
     #[serde(rename = "fn")]
-    pub file_name: String,
+    pub identifier: DistArchiveIdentifier,
```
baszalmstra, line 21: This changes the field name and type in RepoDataRecord.

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -41,7 +42,7 @@ pub struct RepoData {
 
     /// The tar.bz2 packages contained in the repodata.json file
     #[serde(default, serialize_with = "sort_index_map_alphabetically")]
-    pub packages: IndexMap<String, PackageRecord, ahash::RandomState>,
+    pub packages: IndexMap<DistArchiveIdentifier, PackageRecord, ahash::RandomState>,
```
baszalmstra, line 45: These changes make the filenames typed.

wolfv (approved):

Issue #1998, opened 2026-01-21
Construct `RepoData` instances from `list[RepoDataRecord]` directly?


jaimergp:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

I'd like to initialize `RepoData` objects from a collection of `PrefixRecord` objects I have already loaded, and then use those as channels in the `solve()` function (or the sparse variant).

### Why is this needed?

I'm building some integrations for shards in `conda-rattler-solver` and this would save a trip to disk and back. Pseudocode is something like:

```python
records: dict[str, Any] = load_shards(channel, input_specs)
dump(records, temp_json)
repo = SparseRepoData(channel, subdir, temp_json)
```

### What should happen?

Instead, I would love to be able to do this:

```python
records: dict[str, Any] = load_shards(channel, input_specs)
rattler_records = [dict_to_prefix_record(r) for r in records]
repo = SparseRepoData.from_records(channel, subdir, rattler_records)
```

### Additional Context

See context in https://github.com/conda-incubator/conda-rattler-solver/pull/37#discussion_r2709889778

baszalmstra:
I think we can accommodate that through another `solve_` alternative.

baszalmstra:
Implemented in #2000, let me know if thats what you're looking for @jaimergp!

jaimergp:
I think that's an useful API, but thinking about it, the current approach of a family of `solve()` functions requires all channel sources to be homogeneous; you can't mix them. For this particular case (loading a collection of repodata shards from memory), we may run into the case where a user is mixing `conda-forge` with a personal channel with e.g. development builds. conda-forge has shards, but in anaconda.org the other channels don't. This means that we would either have to fallback to monolithic repodata.json for conda-forge or error out.

Instead, would it be feasible to have an abstraction around channels that presents the same interface to the solver, no matter if it has to be fetched remotely (`Gateway`), the full JSON is available locally (`SparseRepoData`) or it's a categorised collection of records?

baszalmstra:
Thats essentially what the Gateway is.. It has the responsibility to load the records using different protocols. 

It only doesnt support passing in custom records..

jaimergp:
It would be awesome if it could support a callable interface with a signature like:

```python
def get_records(channel, subdir, package_name) -> Iterable[RepoDataRecord]:
    ...
```

Then I would just define that callable and use the regular `solve()` function.

[...]

PR #2000, opened 2026-01-22
feat(py): add `solve_with_records` to solve with preparsed records

baszalmstra:
Add a new solver alternative that accepts a list of lists of `RepoDataRecord` types directly. This allows users to provide pre-loaded records organized by channel, without needing to use a `Gateway` or `SparseRepoData`.

The new function `solve_with_records` complements the existing:
- `solve` - fetches records on the fly via Gateway
- `solve_with_sparse_repodata` - uses memory-mapped SparseRepoData

Closes https://github.com/conda/rattler/issues/1998


wolfv:
Looks good to me!

baszalmstra:
Closing as this is not the correct solution for the problem.

PR #2007, opened 2026-01-23
refactor: use `when` syntax instead of `; if`

baszalmstra:
This PR refactors how we parse conditionals. The previous syntax was `foo; if bar`, but in the CEP we landed on `foo[when="bar"]`. This PR implements parsing of the new syntax.

### DIFF
--- a/crates/rattler_conda_types/src/package_name.rs
+++ b/crates/rattler_conda_types/src/package_name.rs
@@ -142,7 +142,7 @@ impl PackageName {
 /// Returns `true` if the byte is a matchspec delimiter (whitespace or version
 /// constraint character: `>`, `<`, `=`, `!`, `~`, `;`).
 fn is_matchspec_delimiter(b: u8) -> bool {
-    b.is_ascii_whitespace() || matches!(b, b'>' | b'<' | b'=' | b'!' | b'~' | b';')
+    b.is_ascii_whitespace() || matches!(b, b'>' | b'<' | b'=' | b'!' | b'~' | b';' | b'[')
 }
 
 /// Scans a matchspec string to find the package name boundary and whether it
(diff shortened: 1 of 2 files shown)


crates/rattler_conda_types/src/match_spec/condition.rs
```diff
@@ -149,71 +181,39 @@ pub(crate) fn parse_condition(input: &str) -> IResult<&str, MatchSpecCondition>
...
-            "mixed; if (alpha and beta) or (gamma and (delta or epsilon))",
+            "foobar or bizbaz",
+            "python >=3.12 or foobar [version='3.12.*', url='https://foobar.com/bla.tar.bz2']",
+            "foobar and (bizbaz or blabla)",
+            "single_condition",
+            "a and b or c",
+            "(a or b) and (c or d)",
+            "a and (b or (c and d))",
+            "a and(b or(c and d))",
+            "foobar >=1.23 *or* and(b >32.12,<=43 *and or(c and d))",
```
wolfv, line 208: this looks pretty weird grammar wise. Could maybe add one more "regularly' formatted test case.

crates/rattler_conda_types/src/match_spec/parse.rs
```diff
@@ -108,6 +108,10 @@ pub enum ParseMatchSpecError {
     #[error("more than one semicolon in match spec")]
     MoreThanOneSemicolon,
 
+    /// Deprecated `; if` syntax used
+    #[error("the '; if' syntax for conditional dependencies is deprecated, use '[when=\"...\"]' bracket syntax instead")]
+    DeprecatedIfSyntax,
```
wolfv, line 113: Should we also deprecate the ` .. if ..` "syntax" that conda silently strips? Or will that anyways trigger an error with our implementation?

baszalmstra: We didnt really do that before so Im inclined to leave that as is.

PR #2010, opened 2026-01-25
feat: Add support for partial path matching authentication

benmoss:
### Description

To fix/add support for https://github.com/prefix-dev/pixi/issues/4524, so that users can have different authentication for different channels on the same domain.

### DIFF
--- a/crates/rattler_networking/src/authentication_storage/storage.rs
+++ b/crates/rattler_networking/src/authentication_storage/storage.rs
@@ -143,13 +143,87 @@ impl AuthenticationStorage {
         Ok(None)
     }
 
-    /// Retrieve the authentication information for the given URL
-    /// (including the authentication information for the wildcard
-    /// host if no credentials are found for the given host)
+    /// Try to find credentials by matching host + path prefix (longest match first).
     ///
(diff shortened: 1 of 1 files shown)


baszalmstra (commented):
Looks good! I want to test whether this doesnt break previously stored tokens but looking at the code I dont think it will.

wolfv:
I wanted to double check whether this has a negative performance impact.

benmoss:
> I wanted to double check whether this has a negative performance impact.

Yeah very good point, if this is done per-package and the cost of backend `get` calls is high, this could be expensive since it's now O(n) where `n` is the number of path and domain segments

benmoss:
We could optimize for host lookups, so basically make the current implementation the first attempt instead of starting from a full URL match first. In other words:

```
# lookup example.com/org/repo/file.json
get("example.com")
get("example.com/org/repo/file.json")
get("example.com/org/repo")
get("example.com/org")
get("*.com")
```

The downside of this is we get unintuitive behavior where `example.com` overrides longer path matches if you have both

benmoss:
We could also do this as a runtime configuration so that we could expose the algorithm to users in pixi settings

baszalmstra:
@wolfv Did you have time to double check?

PR #2012, opened 2026-01-28
feat: add rattler_glob as a crate

tdejager:
### Description

This adds a new crate called `rattler_glob` that allows globbing and hash calculations for sets of globs. We have been using this in pixi but are thinking of also using it in `rattler_build`. So I copied the crate and refactored it a bit with help from Claude to clean it up. I did not really change the API.

I think it's okay to go, but I put it in draft so we can take a good look at it.

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/mod.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/mod.rs
@@ -12,6 +12,7 @@ mod remote_subdir;
 mod repo_data;
 mod run_exports_extractor;
 mod sharded_subdir;
+mod source;
 mod subdir;
 mod subdir_builder;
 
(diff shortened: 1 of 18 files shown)


crates/rattler_glob/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+license.workspace = true
+name = "rattler_glob"
+readme.workspace = true
+repository.workspace = true
+version = "0.1.0"
+
+[dependencies]
+dashmap = { workspace = true }
+fs-err = { workspace = true }
+ignore = "0.4"
```
baszalmstra: Workspace dependency please.

tdejager: @copilot can you change this to be a workspace dependency instead?

tdejager:
Lets close and open if we need it.

PR #2014, opened 2026-01-29
fix: issue when run_exports.json is not in sync with the repodata

wolfv:
@traversaro noticed that an "assimp" run export went missing in one of the builds. We had slightly buggy code where we filled in run_exports from the separate `run_exports.json` file, but we did not properly react when the package was not (yet) present in the `run_exports.json` file.

This might happen due to issues updating the files synchronously, or just faulty generation etc.

In this case we should rather fall back to the package, and fill in the run exports from there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -66,6 +66,20 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "683d7910e743518b0e34f1186f92494becacb047c7b6bf616c96772180fef923"
 
 [[package]]
+name = "ambient-id"
+version = "0.0.7"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "b8cad022ed72ad2176498be1c097bb46e598193e92f3491ea0766980edeee168"
+dependencies = [
+ "reqwest",
+ "reqwest-middleware",
+ "secrecy",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.18",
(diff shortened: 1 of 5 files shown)


crates/rattler_repodata_gateway/src/gateway/run_exports_extractor.rs
```diff
@@ -377,6 +385,12 @@ impl RunExportExtractor {
         match rattler_package_streaming::seek::read_package_file(path) {
             Ok(run_exports_json) => Ok(Some(run_exports_json)),
             Err(rattler_package_streaming::ExtractError::MissingComponent) => Ok(None),
+            Err(rattler_package_streaming::ExtractError::IoError(err))
+                if err.kind() == std::io::ErrorKind::NotFound =>
+            {
+                // File doesn't exist - treat as no run_exports available
+                Ok(None)
+            }
```
baszalmstra: If the file is not there its an error.

crates/rattler_repodata_gateway/src/gateway/mod.rs
```diff
@@ -1181,7 +1220,11 @@ mod test {
             .unwrap();
 
         let total_records: usize = records.iter().map(RepoData::len).sum();
-        assert_eq!(total_records, 16);
```
baszalmstra: Lets revert all changes to this test and change this to 19 ..

PR #2016, opened 2026-01-29
feat: custom channel sources in repodata gateway

### LINKED ISSUE #1998: Construct `RepoData` instances from `list[RepoDataRecord]` directly?

baszalmstra:
This PR exposes the `RepoDataSource` trait to Python, allowing users to provide custom repodata sources (databases,
  APIs, in-memory caches, etc.) that can be used alongside traditional conda channels.

### Changes

  - Add RepoDataSource Protocol in Python with fetch_package_records (async) and package_names methods
  - Add Rust adapter (PyRepoDataSource) to bridge Python implementations to the Rust trait
  - Rename channels parameter to sources in Gateway.query(), Gateway.names(), and solve() (breaking change)
  - Add comprehensive tests including examples with SparseRepoData

### Example
```python
  from rattler import (
      Gateway,
      Platform,
      PackageName,
      RepoDataRecord,
      RepoDataSource,
      solve,
  )

  class MyCustomSource(RepoDataSource):
      """Custom source backed by a database or API."""

      async def fetch_package_records(
          self, platform: Platform, name: PackageName
      ) -> list[RepoDataRecord]:
          # Fetch records from your custom backend
          return await my_database.get_records(str(platform), name.normalized)

      def package_names(self, platform: Platform) -> list[str]:
          # Return available package names
          return my_database.list_packages(str(platform))

  # Use with Gateway.query()
  gateway = Gateway()
  records = await gateway.query(
      sources=["conda-forge", MyCustomSource()],  # Mix channels and custom sources
      platforms=["linux-64"],
      specs=["numpy"],
  )

  # Use with solve()
  solved = await solve(
      sources=[MyCustomSource()],
      specs=["numpy"],
      platforms=["linux-64", "noarch"],
  )
```

### Notes

  - Custom sources are not cached by the gateway (caching must be implemented in the source)
  - Custom sources are slower than channels due to Python/Rust data marshalling

Fixes #1998

### DIFF
--- a/py-rattler/tests/unit/test_custom_source.py
+++ b/py-rattler/tests/unit/test_custom_source.py
@@ -237,9 +237,8 @@ async def test_custom_source_backed_by_sparse_repodata() -> None:
     """Test a custom RepoDataSource that wraps SparseRepoData.
 
     This demonstrates how to create a custom source that loads records
-    from local repodata files for multiple platforms (linux-64 and noarch),
-    which can be useful for offline scenarios or when you want to
-    filter/transform repodata before solving.
+    from local repodata files, which can be useful for offline scenarios
+    or when you want to filter/transform repodata before solving.
     """
     import os
 
@@ -247,8 +246,7 @@ async def test_custom_source_backed_by_sparse_repodata() -> None:
 
     # Load sparse repodata from the test data directory
     data_dir = os.path.join(os.path.dirname(__file__), "../../../test-data/")
-    linux64_path = os.path.join(data_dir, "channels/conda-forge/linux-64/repodata.json")
-    noarch_path = os.path.join(data_dir, "channels/conda-forge/noarch/repodata.json")
+    linux64_path = os.path.join(data_dir, "channels/dummy/linux-64/repodata.json")
 
     class SparseRepoDataSource(RepoDataSource):
         """A custom source that wraps SparseRepoData files."""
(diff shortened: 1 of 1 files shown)


py-rattler/tests/unit/test_custom_source.py
```diff
@@ -0,0 +1,304 @@
...
+        specs=["my-package"],
+        platforms=["linux-64"],
+    )
+
+    assert len(solved) == 1
+    assert record_snapshot(solved[0]) == "my-package=1.0.0=test_0"
+
+
+@pytest.mark.asyncio
+async def test_custom_source_backed_by_sparse_repodata() -> None:
```
baszalmstra, line 236: @jaimergp you probably want to do something like this

jaimergp:
FYI we tried this in https://github.com/conda-incubator/conda-rattler-solver/pull/42 and the overheads are significant for complex solve that require "moving" a lot of records back and forth.

flying-sheep:
This broke ASV: https://github.com/airspeed-velocity/asv/issues/1577

wolfv:
Sorry @flying-sheep - we will be much more careful about Python API changes in the future and add backwards compatible keyword arguments. 

Would it help if we make a release that re-introduces the previous keyword or is it easier if you change the argument?

flying-sheep:
I don’t know how fast the ASV people are to fix thinks, that fluctuated in the past.

I think going forward, it helps to add both `/` and `*` to your APIs, and add them early (both time and position wise).

I usually:
1. put short mandatory “tag” parameters first if they exist (e.g. `html.node("div", …)`)
2. then come other mandatory parameters.
   - if there is 1, make it positional-only using `/`
   - same if there multiple order independent ones, such as in `__add__(a, b, /)`
   - if there are 2, it’s OK to make them positional-or-keyword and let the user decide how to call things
   - if there are 3 or more mandatory parameters, make 0–2 of them positional-only (based on which ones you consider most important), and the rest keyword-only
5. If there is one really common optional parameter, it can be positonal-or-keyword; other optional parameters are keyword-only
6. exceptions can be made if it’s obvious, e.g. `def rgba(r: float, g: float, b: float, a: float = 1.0)` makes sense because the name spells out the parameter order.

Following these rules would have led to this API for me, which would have indicated that the name of `channels` is part of the API, and therefore needs some backwards compat treatment.

Positional-only parameters can of course be renamed whenever you want.

```py
async def query(
    self,
    specs: Iterable[MatchSpec | PackageName | str],   # I’d consider this the most important parameter
    /,
    *,
    channels: Iterable[Channel | str],                # these two could be optional in a different …
    platforms: Iterable[Platform | PlatformLiteral],  # … world where inferring them makes sense
    recursive: bool = True,                           # boolean flags are always positional-only
) -> List[List[RepoDataRecord]]: ...
```

PR #2020, opened 2026-01-30
fix: Also enforce io concurrency during package linking

iamthebot:
### Description
In investigating prefix-dev/rattler-build#1638 I found that even if we did pass io_concurrency_limit to rattler from rattler-build, the link_package method that's *actually* used doesn't use the concurrency limit at all. So you'd get unbounded concurrency when spawning the linking tasks.


baszalmstra (commented):
Would be nice to test if this causes a performance regression.

gabrielsimoes:
Is think this might also be a fix for: https://github.com/conda/rattler/issues/2021

crates/rattler/src/install/installer/mod.rs
```diff
@@ -722,6 +722,12 @@ async fn link_package(
     let cached_package_dir = cached_package_dir.to_path_buf();
     let clobber_registry = driver.clobber_registry.clone();
 
+    let _io_permit = driver.acquire_io_permit().await.map_err(|_| {
```
gabrielsimoes: Will this make it so io_concurrency has to be at least 2, given another permit will be acquired for file linking?

iamthebot: Yes. But that seems reasonable, no? I can add enforcement where we first capture the io_concurrency to not make this a surprise for users.

Or is there a specific reason (the change here, aside) the lower bound must be one instead?

baszalmstra:
Are you also aware that linking is also already limited by rayon? We perform the linking in the rayon threadpool where there is at most one thread per core. 

Could also be interesting to try to remove rayon and soley rely on the ionconcurrency semaphore.

I have played with all these things a lot but your analysis seems much more capable.

crates/rattler/src/install/installer/mod.rs
gabrielsimoes: No, makes sense, just curious!

wolfv:
@baszalmstra PR is green. Should we merge or run more analysis?

baszalmstra:
Im not sure what the conclusion of the measurements are. Does this indeed limit concurrency? Since we are using rayon, which is a thread-per-core scheduler the concurrency will already effectively be limited to the number of cores anyway.

PR #2029, opened 2026-02-03
fix: Ignore CONDA_SHLVL in tests

hunger:
The tests run scripts assuming an empty environment in the real shell environment. So if CONDA_SHLVL is set already, it may result in the variable getting removed from the output, which triggers the tests to fail.


hunger:
The failed tests look like networking issues... I'll try rerunning them again in a bit.

baszalmstra:
Looks like a service interruption. I asked here: https://conda.zulipchat.com/#narrow/channel/487124-infrastructure/topic/Service.20interruption.3F/with/571664466

Issue #1383, opened 2025-06-05
Content addressed cache


jaimergp:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

Deduplicate sha256-identical files in the `pkgs/` cache.

Right now we have something like this:

```bash
pkgs/
  package-1.0-0/
     file1
     file2
     file3
  package-2.0-0/
     file1
     file2
     file3
     file4
```

Across releases, many packages will keep identical files. Even more across variant builds. If the file "identity" (path) is instead its sha256, we can restructure the cache as such

```bash
pkgs/
  content-addressed/  # probably with some more subdirectory levels per hash prefix
    48/
      4819c16dc70219ac16888407d77e97a7774d5f877efec3998272248eed40b4b9
    d0/
      d0229d43699e06e8f4efc5de64bdffa2b41d92f1aeb4bd27981f46f322eb1847
    14/
      146e0dafbc5327c347f29ceffe904132a37c0b72b26819827d91832dfb8ec263
    ba/
      baed639523840f550125dc6f7ed12cccdbb5ee15f5d1ccf957155b13eb8debad
  package-1.0-0/
     file1 -> ../content-addressed/48/4819c16dc70219ac16888407d77e97a7774d5f877efec3998272248eed40b4b9
     file2 -> ../content-addressed/d0/d0229d43699e06e8f4efc5de64bdffa2b41d92f1aeb4bd27981f46f322eb1847
     file3 -> ../content-addressed/14/146e0dafbc5327c347f29ceffe904132a37c0b72b26819827d91832dfb8ec263
  package-2.0-0/
     file1 -> ../content-addressed/48/4819c16dc70219ac16888407d77e97a7774d5f877efec3998272248eed40b4b9
     file2 -> ../content-addressed/d0/d0229d43699e06e8f4efc5de64bdffa2b41d92f1aeb4bd27981f46f322eb1847
     file3 -> ../content-addressed/14/146e0dafbc5327c347f29ceffe904132a37c0b72b26819827d91832dfb8ec263
     file4 -> ../content-addressed/ba/baed639523840f550125dc6f7ed12cccdbb5ee15f5d1ccf957155b13eb8debad
```

with `->` denoting a hardlink.

There might be even more cache hits if we standardize the prefix placeholder to the path of the cache plus the necessary padding.

### Why is this needed?

My preliminary analysis show substantial storage savings:

- Global size 28.8GiB
[...]

PR #2031, opened 2026-02-03
feat: implement package file extraction deduplication with a CAS

baszalmstra:
With this PR an optional CAS location can be specified when extracting files. During extraction, hashes of the files are computed and the files are placed in a content addressable cache. Once in the cache the files are hardlinked into the directory structure described by the package itself. This ensures that package files are deduplicated. 

Initial results suggest that this can reduce the size of the package cache by 10-40%, which is quite a lot! The downside is that the extra filesystem operations required can be a bottleneck. On my Windows machine, where IO is expensive, the total extraction duration for extracting Rust was doubled while only saving 2% disk space.

The `rattler extract` command now takes an additional `--cas` argument which allows you to test this feature locally. 

I think we could still merge this feature as using the CAS is disabled by default. 

- I choose to recompute the hashes of the files because I cannot trust the sha256 hashes from the `paths.json` file. 
- I used a SHA256 hash because I cannot trust the files that Im hashing so a cryptographic hash is more appropriate (over for example XXH3).

Closes https://github.com/conda/rattler/issues/1383

### DIFF
--- a/crates/rattler_cas/src/lib.rs
+++ b/crates/rattler_cas/src/lib.rs
@@ -57,11 +57,11 @@ mod sync_writer;
 mod writer;
 
 use std::{
-    io::Read,
+    io::{Read, Write},
     path::{Path, PathBuf},
 };
 
-use rattler_digest::Sha256Hash;
+use rattler_digest::{Sha256, Sha256Hash};
 pub use sync_writer::SyncWriter;
 pub use writer::Writer;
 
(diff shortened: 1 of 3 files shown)


baszalmstra:
So it turns out that this has a significant performance impact on extraction times. This is because most likely this introduces a lot more syscalls. Every file now needs at least two/three: extracting into the CAS, (moving into the right place in the cas), and hardlinking from the cas.

I think we can still merge this in, but simply not enable it by default.

PR #2032, opened 2026-02-04
feat: Implementation offsets in paths.json

Dagmar-Dinjens:
### Description
An implementation for a new field within Paths.json which will improve the linking of a package by including the offsets (and possibly cstring end offsets) for the files which need some form of prefix replacements.
This would improve linking time and with look onto a VFS 

### What happened
Within this PR I added support to handle a paths.json with offsets supplied. 
 
### What should still be done/ next steps
A next step would be the implementation of paths.json when building a package with rattler-build & a CEP to get this up and running


crates/rattler/src/install/link.rs
```diff
@@ -203,16 +205,33 @@ pub fn link_file(
...
+        match offsets {
+            Some(offsets) => {
+                copy_and_replace_placeholders_with_offsets(
+                    source.as_ref(),
+                    &mut destination_writer,
+                    placeholder,
+                    &target_prefix,
+                    &target_platform,
+                    *file_mode,
+                    offsets.to_vec()
```
baszalmstra: Cant we keep this as a slice instead of allocating on the heap?

crates/rattler_cache/src/validation.rs
```diff
@@ -46,6 +46,10 @@ pub enum PackageValidationError {
     #[error("neither a 'paths.json' or a deprecated 'files' file was found")]
     MetadataMissing,
 
+    /// An error occurred with the version of the `path.json` file
+    #[error("")]
```
baszalmstra: Could you improve this error?

crates/rattler_conda_types/src/package/paths.rs
```diff
@@ -213,15 +221,20 @@ pub struct PathsEntry {
...
     /// The size of the file in bytes
-    /// This entry is only present in version 1 of the paths.json file.
+    /// This entry is present in version 1 and up of the paths.json file.
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub size_in_bytes: Option<u64>,
+
+    /// When a file is executable this will be true
+    /// This entry is only present in version 2 of the paths.json file
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub executable: Option<bool>,
```
baszalmstra: You did not implement this yet right?

baszalmstra (commented):
I think this is going in the right direction. 

I do think we can make this change backwards compatible. Now it isnt because the `version` in `paths.json` is bumped. This means that other installers will most likely fail with these packages. That will hamper adoption.

Instead perhaps we can keep the version the same. The `offsets` field is already backwards compatible. But the `executable` flag is not because we dont know if the absence of the field means nothing is executable or it was not written. Maybe instead we can add another field to the top-level keys to indicate that the `executable` field should be taken into account.

crates/rattler_conda_types/src/package/paths.rs
Dagmar-Dinjens: I did not really understand what improvements were the goal in adding the executability to the paths, because of this I didn't yet implement them no

baszalmstra: Makes sense. Let me explain. At the moment the code has to acquire the metadata in the cache to determine if the file has the executable permissions set. If so, then the file also needs to have it set in the environment. However, this metadata call performs a syscall which adds a bit of overhead. Using the executable flag in the paths.json we can skip this metadata check.

crates/rattler/src/install/link.rs
```diff
@@ -787,6 +905,80 @@ pub fn copy_and_replace_cstring_placeholder(
...
+        let mut end = offset + old_prefix.len();
+        while end < source_bytes.len() && source_bytes[end] != b'\0' {
+            end += 1;
+        }
+
+        let segment = &source_bytes[offset..end];
+        let old_len = segment.len();
+
+        let mut out = Vec::with_capacity(old_len);
+        let mut remaining = segment;
```
baszalmstra: I feel like this inner loop could be simplified. The code know exactly where the offsets are and how long they are. I dont think we need the finder or the `Vec`. This seems very much like the code from the previous implementation, I encourage you to rethink it.

crates/rattler/src/install/link.rs
```diff
@@ -787,6 +905,80 @@ pub fn copy_and_replace_cstring_placeholder(
...
+    if new_prefix.len() > old_prefix.len() {
+        return Err(std::io::Error::new(
+            std::io::ErrorKind::InvalidData,
+            "target prefix cannot be longer than the placeholder prefix",
+        ));
+    }
+
+    if offsets.is_empty() {
+        return destination.write_all(source_bytes);
+    }
```
baszalmstra: This seems redundant. If the offsets is empty, the loop will have no iterations and at the end of the function the rest of the bytes will be written in a similar fashion.

Dagmar-Dinjens: You're right!

crates/rattler/src/install/link.rs
Dagmar-Dinjens: I did indeed not push that which I had been working on, in the new commit I did.

crates/rattler_conda_types/src/package/paths.rs
Dagmar-Dinjens: makes a lot more sense now! I think I now implemented this with the new commit

crates/rattler/src/install/link.rs
Dagmar-Dinjens: Yes! that seems like a way prettier solution indeed

Dagmar-Dinjens:
> Instead perhaps we can keep the version the same. The `offsets` field is already backwards compatible. But the `executable` flag is not because we dont know if the absence of the field means nothing is executable or it was not written. Maybe instead we can add another field to the top-level keys to indicate that the `executable` field should be taken into account.

The absence of the field would inherently mean it was not written otherwise it would read as true or false, I guess you assumed that it would, like the prefix_placeholder, be left empty if false. 

this currently isn't the case but something like the following would be the preference?

```
"paths": [
                {
                    "_path": "bin/example",
                    "no_link": false,
                    "path_type": "hardlink",
                    "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855",
                    "size_in_bytes": 1024,
                    "executable": true,
                    "file_mode": "binary",
                    "prefix_placeholder": "/opt/conda",
                    "offsets": [100, 200, 300]
                },
                {
                    "_path": "lib/library.so",
                    "no_link": false,
                    "path_type": "hardlink",
                    "sha256": "d7a8fbb307d7809469ca9abcb0082e4f8d5651e46d3cdb762d02d0bf37c9e592",
                    "size_in_bytes": 2048
                },
                {
                    "_path": "share/doc/readme.txt",
                    "no_link": false,
                    "path_type": "hardlink",
                    "sha256": "a665a45920422f9d417e4867efdc4fb8a04a1f3fff1fa07e998e86f7f7a27ae3",
                    "size_in_bytes": 256,
                    "file_mode": "text",
                    "prefix_placeholder": "/home/builder/conda",
                    "offsets": [10, 45]
                },
                {
                    "_path": "bin/symlink-example",
                    "no_link": false,
                    "path_type": "softlink",
                    "executable": true
                }
            ],
            "has_executable": true,
            "paths_version": 1
            }"
```

crates/rattler_cache/src/validation.rs
Dagmar-Dinjens: I did improve it but think I am not using it anywhere, Depending on whatever leads out of the v1/v2 comment you left before I might even remove it

crates/rattler_conda_types/src/package/paths.rs
baszalmstra: Not entirely. What we try to avoid is the have to query the file metadata all together. You currently implementation still requests this regardless of whether `executable` was set. 

But what Im trying to avoid is having to bump the `version` field in `paths.json`. Because doing so will break installers that aren't compatible with this version (e.g. conda and mamba). We need to make sure that adding the `executable` field is a backwards compatible change. Simply adding a field is backwards compatible but then if the field is optional we dont know if the absence of the fields means either the value is false or that the value was not computed in the first place.

To fix that we could add another top-level key (next to `version`) that indicates that the absence of the `executable` field means it should be considered `false`. Then, if that top-level field is present we no longer have to read the metadata from disk and we can rely on the per-file `executable` field.

Dagmar-Dinjens: Ah correct, I hope that the current adaptation is what you meant! 
I was still retrieving the metadata because it seems to be needed for the file permissions and timestamps, 
should there be another solution for retrieving this metadata or does this mean we keep the current approach?

crates/rattler/src/install/link.rs
```diff
@@ -787,6 +912,74 @@ pub fn copy_and_replace_cstring_placeholder(
...
+
+        if end == next_offset {
+            unfinished_changes += 1;
+            last_pos = end;
+            continue;
+        };
+
+        let padding = (unfinished_changes + 1) * length_change;
+        if padding > 0 {
+            destination.write_all(&vec![0; padding])?;
```
baszalmstra: Instead of allocating a `Vec` here you can use a const:

```rs
fn write_zeros<W: Write>(mut w: W, mut n: usize) -> Result<(), std::io::Error> {
    const ZEROS: [u8; 256] = [0; 256];
    while n > 0 {
        let k = n.min(ZEROS.len());
        w.write_all(&ZEROS[..k])?;
        n -= k;
    }
    Ok(())
}
```

crates/rattler/src/install/link.rs
```diff
@@ -787,6 +912,74 @@ pub fn copy_and_replace_cstring_placeholder(
...
 }
 
+/// Given the contents & offsets of a file, copies it to the `destination` and in the process replace
+/// any binary c-style string that contains the text `prefix_placeholder` with a binary compatible
+/// c-string where the `prefix_placeholder` text is replaced with the `target_prefix` text.
+///
+/// The length of the input will match the output.
+///
+/// This function replaces binary c-style strings using pre-computed offsets for better performance.
+pub fn copy_and_replace_cstring_placeholder_offsets(
```
baszalmstra, line 1019: I took the liberty to rewrite this function a little in a sense that makes more sense to me. I think the current implementation is a bit confusing. We can also use `memchr` to search for the `NUL` terminators quicker:

```rs
pub fn copy_and_replace_cstring_placeholder_offsets(
    source_bytes: &[u8],
    mut destination: impl Write,
    prefix_placeholder: &str,
    target_prefix: &str,
    offsets: &[usize],
) -> Result<(), std::io::Error> {
    // Bytes of the placeholder prefix to be replaced.
    let old_prefix = prefix_placeholder.as_bytes();
    // Bytes of the replacement prefix.
    let new_prefix = target_prefix.as_bytes();

    // Compute how many bytes each replacement shrinks the string by.
    // Fail if the replacement is longer than the placeholder.
    let shrink = old_prefix
        .len()
        .checked_sub(new_prefix.len())
        .ok_or_else(|| {
            std::io::Error::new(
                std::io::ErrorKind::InvalidData,
                "target prefix cannot be longer than the placeholder prefix",
            )
        })?;

    // Last byte position written to `destination`.
    let mut last_pos = 0usize;
    // Accumulated padding bytes for the current C-string (flushed on NUL).
    let mut pending_pad_bytes = 0usize;

    // Iterator over the "next offset" for each placeholder, with EOF as sentinel.
    let nexts = offsets
        .iter()
        .skip(1)
        .copied()
        .chain(std::iter::once(source_bytes.len()));

    for (&offset, next_offset) in offsets.iter().zip(nexts) {
        // Write bytes before the placeholder.
        destination.write_all(&source_bytes[last_pos..offset])?;

        // Write the replacement prefix.
        destination.write_all(new_prefix)?;

        // Start of the C-string tail after the replaced prefix.
        let cstr_tail_start = offset + old_prefix.len();
        // Bytes until the next placeholder (or EOF).
        let cstr_tail_slice = &source_bytes[cstr_tail_start..next_offset];

        // Length of the C-string tail until the NUL terminator (or next placeholder).
        let cstr_tail_len =
            memchr(0, cstr_tail_slice).unwrap_or(cstr_tail_slice.len());

        // Absolute end position of this C-string segment.
        let cstr_end = cstr_tail_start + cstr_tail_len;

        // Write the remainder of the C-string after the replaced prefix.
        destination.write_all(&source_bytes[cstr_tail_start..cstr_end])?;

        // Accumulate how many bytes we need to pad for this C-string so far.
        pending_pad_bytes += shrink;

        // If we reached the NUL terminator (end of this C-string),
        // flush all accumulated padding bytes.
        if cstr_end != next_offset {
            if pending_pad_bytes != 0 {
                write_zeros(&mut destination, pending_pad_bytes)?;
            }
            pending_pad_bytes = 0;
        }

        last_pos = cstr_end;
    }

    // Write any remaining bytes after the last replacement.
    if last_pos < source_bytes.len() {
        destination.write_all(&source_bytes[last_pos..])?;
    }

    Ok(())
}
```

Feel free to check if this makes sense to you too!

baszalmstra, line 1019: This code also works for an edge case where a single C-string contains the prefix twice. Could you add a test for that as well so we dont break it in the future. E.g.:

```rs
"foo__PREFIX__bar__PREFIX__baz\0"
```

crates/rattler_cache/src/validation.rs
```diff
@@ -46,6 +46,10 @@ pub enum PackageValidationError {
     #[error("neither a 'paths.json' or a deprecated 'files' file was found")]
     MetadataMissing,
 
+    /// An error occurred with the version of the `path.json` file
+    #[error("The version of the 'paths.json' did not correspond to the included fields")]
+    VersionError(),
```
baszalmstra: This is now completely unused right?

baszalmstra (changes requested):
Sorry for the late reply. 

I have been looking at the code and I think it would be better to split the PR into two. The `has_executable` bit needs some more work because as it stands now there is not really a benefit because you are still reading the metadata from disk anyway. This has to happen because we also need to get the `mtime` so the initial premise was already wrong. I apologize for steering you in the wrong direction.

baszalmstra:
I apologize for pitching you another idea! But here goes: 😄 

The current code still has to search through the source bytes. I wonder if we could change this by storing more information in the `offsets`. Like the number of bytes that follow the prefix. I assume that could make the whole thing even faster.

crates/rattler/src/install/mod.rs
```diff
@@ -411,6 +411,8 @@ pub async fn link_package(
         }
     }
 
+    // let has_executable = paths_json.has_executable.unwrap_or(false);
```
baszalmstra: Remove?

crates/rattler_conda_types/src/package/paths.rs
```diff
@@ -187,6 +189,14 @@ pub struct PrefixPlaceholder {
     /// was build.
     #[serde(rename = "prefix_placeholder")]
     pub placeholder: String,
+
+    /// The offsets on which the placeholders are found in the file
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub offsets: Option<Vec<usize>>,
+
+    /// The offsets on which the placeholders are found in the file
```
baszalmstra: Im not quite clear what this entails. The comment is the same as the field above it. 

Also is `None` the same as an empty vector or is there a difference?

crates/rattler/src/install/link.rs
```diff
@@ -214,16 +216,38 @@ pub fn link_file(
...
-            source.as_ref(),
-            &mut destination_writer,
-            placeholder,
-            &target_prefix,
-            &target_platform,
-            *file_mode,
-        )
-        .map_err(|err| LinkFileError::IoError(String::from("replacing placeholders"), err))?;
+        // depending on the availability of the offsets
+        match offsets {
```
baszalmstra, line 223: Does it matter if null_offsets is still None?

crates/rattler/src/install/link.rs
```diff
@@ -852,6 +990,80 @@ pub fn copy_and_replace_cstring_placeholder(
...
+    let mut null_idx = 0usize;
+
+    let nexts = offsets
+        .iter()
+        .skip(1)
+        .copied()
+        .chain(std::iter::once(source_bytes.len()));
+
+    for (&offset, next_offset) in offsets.iter().zip(nexts) {
+        let cstr_end = null_offsets[null_idx];
```
baszalmstra: This will panic if null_offsets is an empty slice.

crates/rattler_conda_types/src/package/paths.rs
baszalmstra: After reviewing the algorithm I _feel_ like there is a relation between these two fields that we could also encapsulate in a type. But TBH Im not entirely sure how.

baszalmstra (changes requested):
Would you also be able to update the description to explain why we want this and how your implementation solves the problem we are facing?

crates/rattler_conda_types/src/package/paths.rs
Dagmar-Dinjens: What could be an option: (de)serialising it from the current proposed PrefixPlaceholder (with the two lists in the Json) into when applicable:
 - CStringReplacement 
 ```
pub struct Offsets {
    pub offset: usize,
    pub cstr_end: Option<usize>,
}
```
 - the list of just offsets when dealing with a text document

I think this could be a step in the right direction?

File: crates/rattler/src/install/link.rs (hunk 11 above)
Comment (Dagmar-Dinjens) [reply crates/rattler/src/install/link.rs:223]: changed where I check if the null_offsets is none to an empty vector, If null_offsets is None it could be because of 
1) The file is a text file 
2) There was no found end? 
Both should be handled better than in the current approach, possibly with the change to the Offsets struct or within this match statement

File: crates/rattler/src/install/link.rs [hunk 13]
```diff
@@ -214,16 +219,45 @@ pub fn link_file(
...
+        // depending on the availability of the offsets
+        match offsets {
+            Some(offsets) => {
+                let null_offsets = match null_offsets {
+                    Some(null_offsets) => null_offsets,
+                    None => {
+                        // There are offsets but no null_offsets, meaning it is either a binary file or short file where the eof is the 
+                        &EMPTY_VEC
+                    }
+                };
```
baszalmstra, line 231:
```suggestion
                let null_offsets = null_offsets.unwrap_or(&EMPTY_VEC);
```

crates/rattler/src/install/link.rs
```diff
@@ -871,6 +1008,87 @@ pub fn copy_and_replace_cstring_placeholder(
...
+        .iter()
+        .skip(1)
+        .copied()
+        .chain(std::iter::once(source_bytes.len()));
+
+    for (&offset, next_offset) in offsets.iter().zip(nexts) {
+        let cstr_end = if null_offsets.is_empty(){
+            // there were no stored null offsets, fall back to the length of the source bytes
+            source_bytes.len()
+        } else { null_offsets[null_idx] };
```
baszalmstra, line 1054: There is no guarantee that `null_idx` is a valid index here. Im wondering if nulls are tied to specific offsets. Do you maybe want to do:

```suggestion
        let cstr_end = null_offsets.get(null_idx).unwrap_or_else(|| source_bytes.len());
```

crates/rattler_conda_types/src/package/paths.rs
```diff
@@ -187,6 +189,16 @@ pub struct PrefixPlaceholder {
     /// was build.
     #[serde(rename = "prefix_placeholder")]
     pub placeholder: String,
+
+    ///
```
wolfv, line 193: Empty lines?

wolfv:
I just had a look! I am not sure if we would get the design through as a CEP - my main critique is that we have two lists that are not easy to "correlate" but - if I understand correctly - should be ready together. 

I am thinking it would be better to stroe a list of lists if we want `(start, end)` markers, e.g.:

```
offsets: [ [100, 302], [405, 1000], ...]
```

Bas also reminded me that a given region can contain the placeholder more than once (e.g. in a embedded PATH variable it could be something like `$PLACEHOLDER/bin:$PLACEHOLDER/sbin:/some/stuff\0`. Then we could store a list like `[405, 505, 700]` where the first numbers would be each occurence and the last the `\0`.

Dagmar-Dinjens:
Superseded by #2565 ("feat: Implementation offsets in paths.json"), which is the up-to-date continuation of this work and already references "Closes #2032". Closing this duplicate in favour of #2565.

PR #2034, opened 2026-02-04
chore: Let renovate update macos in GitHub Actions

pavelzw:
### Description

macos-15-intel won't be there forever. we should use cross-compilation instead


baszalmstra:
I think we can keep this in and let renovate handle updates for us. We should probably just remove:

https://github.com/conda/rattler/blob/c34aa544739e4a07941979ab48dea2375e8c8845/.github/renovate.json5#L27C5-L32C7

PR #2036, opened 2026-02-04
feat: support glob and regex patterns in repodata queries

### LINKED ISSUE #1723: Add support for glob/regex in repodata queries
follow-up of #1480

pavelzw:
Enable QueryExecutor to handle glob patterns (e.g., `foo*`) and regex patterns (e.g., `^numpy.*$`) in package names. Previously, these specs were silently dropped.

The implementation stores pattern specs separately and expands them by matching against all package names from the subdirs before fetching records.

closes #1723

### Description

This is part 1 of improving pixi search: Xref https://github.com/prefix-dev/pixi/issues/378

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/mod.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/mod.rs
@@ -1690,10 +1690,7 @@ mod test {
 
         let records = gateway
             .query(
-                vec![
-                    super::Source::Custom(src_a),
-                    super::Source::Custom(src_b),
-                ],
+                vec![super::Source::Custom(src_a), super::Source::Custom(src_b)],
                 vec![Platform::Linux64, Platform::NoArch],
                 vec![matchspec].into_iter(),
             )


crates/rattler_repodata_gateway/src/gateway/error.rs
```diff
@@ -49,6 +49,9 @@ pub enum GatewayError {
     #[error("the match spec '{0}' does not specify an exact name")]
     MatchSpecWithoutExactName(Box<MatchSpec>),
 
+    #[error("the match spec '{0}' does not have a package name")]
+    MatchSpecWithoutName(Box<MatchSpec>),
```
pavelzw, line 54: i'm not sure whether we should make `*` a special case matchspec (i.e. name = None). how about we just make it a glob matchspec with `*` as the glob matcher? then we can probably simplify some things where we check for this.

wdyt @baszalmstra?

pavelzw: would do this in a separate PR, though

crates/rattler-bin/src/commands/search.rs
```diff
@@ -0,0 +1,232 @@
...
+pub async fn search(opt: Opt) -> miette::Result<()> {
+    let channel_config =
+        ChannelConfig::default_with_root_dir(env::current_dir().into_diagnostic()?);
+
+    println!("Searching for '{}' on {}", opt.pattern, opt.platform);
+
+    // Parse the pattern as a matchspec with glob/regex support
+    let matchspec = MatchSpec::from_str(
+        &opt.pattern,
+        ParseStrictnessWithNameMatcher {
```
baszalmstra: Should we also enable extras?

crates/rattler_repodata_gateway/src/gateway/error.rs
baszalmstra: Yeah, using a glob makes sense to me!

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -391,10 +409,58 @@ impl QueryExecutor {
...
+        }
+    }
+
     /// Run the main event loop.
     async fn run(mut self) -> Result<Vec<RepoData>, GatewayError> {
         self.spawn_direct_url_fetches()?;
 
+        // If we have pattern specs, we need to wait for all subdirs to be loaded
+        // first so we can get all package names to match against
+        if !self.pending_pattern_specs.is_empty() {
```
baszalmstra: I think it would be nicer if we can do this concurrently. This now blocks everything else meaning that if you have a channel that takes much longer we are essentially wasting time.

When a subdir resolves we query its names for matching packages. 

I can also give this a go if that helps?

py-rattler/rattler/repo_data/gateway.py
```diff
@@ -199,6 +199,20 @@ async def query(
...
         >>>
         ```
+
+        Glob and regex patterns are also supported using MatchSpec with `exact_names_only=False`:
+        ```python
+        >>> import asyncio
+        >>> from rattler import Gateway, MatchSpec
+        >>> gateway = Gateway()
+        >>> glob_spec = MatchSpec("python-*", strict=True, exact_names_only=False)
+        >>> records = asyncio.run(gateway.query(["conda-forge"], ["noarch"], [glob_spec], recursive=False))
```
baszalmstra: This doctest will be relatively slow because its directly accessing the internet. If possible we should avoid that to make the tests less flaky and faster.

crates/rattler-bin/src/commands/search.rs
pavelzw: sure!

crates/rattler_repodata_gateway/src/gateway/error.rs
pavelzw: https://github.com/conda/rattler/issues/2039

py-rattler/rattler/repo_data/gateway.py
pavelzw: isn't this the same situation above? but yeah, then let's remove this test, we have other tests in `crates/rattler_repodata_gateway/src/gateway/mod.rs` anyway

crates/rattler_repodata_gateway/src/gateway/query.rs
pavelzw: [`d40e851` (this PR)](https://github.com/conda/rattler/pull/2036/commits/d40e8515b6da361f74a47faf5cbc27eea8dd0313)

crates/rattler-bin/src/commands/search.rs
```diff
@@ -0,0 +1,229 @@
...
+        records.sort_by(|a, b| {
+            b.package_record
+                .version
+                .cmp(&a.package_record.version)
+                .then_with(|| {
+                    b.package_record
+                        .build_number
+                        .cmp(&a.package_record.build_number)
+                })
+        });
```
baszalmstra: RepoDataRecords implement Ord now! So you can sort them.

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -246,31 +249,30 @@ impl QueryExecutor {
...
                     // For custom sources, create an adapter that wraps the source
-                    // for the specific platform
+                    // for the specific platform.
                     let client = CustomSourceClient::new(custom_source, platform);
                     let subdir = Arc::new(Subdir::Found(SubdirData::from_client(client)));
-                    barrier.set(subdir).expect("subdir was set twice");
+                    box_future(async move {
+                        barrier.set(subdir.clone()).expect("subdir was set twice");
+                        Ok(subdir)
+                    })
```
baszalmstra: This seems unnecessary. The created future is immediately done. Why not just set the barrier immediately?

crates/rattler-bin/src/commands/search.rs
pavelzw: [`3253cf4` (this PR)](https://github.com/conda/rattler/pull/2036/commits/3253cf4316e1a8c17e82d91627d73e69fe58efc9)

Issue #2041, opened 2026-02-09
pixi upgrade botches >=0.60b1


ChristianRothQC:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

```bash
pixi init
pixi add "opentelemetry-instrumentation-starlette>=0.60b1"
git add .
git commit -m "foo"
pixi upgrade
git diff
```

```diff
diff --git a/pixi.toml b/pixi.toml
index f2ec528..189359b 100644
--- a/pixi.toml
+++ b/pixi.toml
@@ -8,4 +8,4 @@ version = "0.1.0"
 [tasks]
 
 [dependencies]
-opentelemetry-instrumentation-starlette = ">=0.60b1"
+opentelemetry-instrumentation-starlette = ">=0.60b1,<0.60a2"
```

```bash
pixi ls
```
```
Error:   × failed to solve requirements of environment 'default' for platform 'osx-arm64'
  ├─▶   × failed to solve the environment
  │   
  ╰─▶ Cannot solve the request because of: No candidates were found for opentelemetry-instrumentation-
      starlette >=0.60b1,<0.60a2.
```

PR #2042, opened 2026-02-10
fix: version bump with alpha (a1/a2...)

wolfv:
Not sure how we should bump in practice. Do we expect `60a1` -> `60a2` or `61a1`?

Fixes https://github.com/conda/rattler/issues/2041

### DIFF
--- a/crates/rattler_conda_types/src/version/bump.rs
+++ b/crates/rattler_conda_types/src/version/bump.rs
@@ -217,25 +217,27 @@ impl Version {
                     }
                 }
                 _ => {
-                    // For Last and Segment types, keep original behavior
                     if segment_to_bump == idx {
-                        let last_numeral_component = segment_components
-                            .iter_mut()
-                            .filter_map(Component::as_number_mut)
-                            .next_back()
-                            .expect(
-                                "every segment must at least contain a single numeric component",
-                            );
-                        *last_numeral_component += 1;
+                        // Find the position of the last numeral so we only
+                        // reset idens that appear after it.
+                        let last_num_idx = segment_components
+                            .iter()
+                            .rposition(|c| c.as_number().is_some())
+                            .expect("every segment must contain a numeric component");
 
-                        // If the segment ends with an ascii character, make it `a` instead of whatever it says
-                        let last_iden_component = segment_components
-                            .iter_mut()
-                            .filter_map(Component::as_iden_mut)
-                            .next_back();
+                        // Bump the last numeral.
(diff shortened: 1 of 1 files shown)


crates/rattler_conda_types/src/version/bump.rs
```diff
@@ -217,25 +217,27 @@ impl Version {
...
-                        if let Some(last_iden_component) = last_iden_component {
-                            *last_iden_component = "a".into();
+                        // Only reset idens that appear AFTER the last numeral
+                        // to 'a'. Idens before the bumped numeral must be
+                        // preserved, otherwise we produce a smaller version
+                        // (e.g. 0.60b1 -> 0.60a2 instead of correct 0.60b2).
+                        for c in segment_components[last_num_idx + 1..].iter_mut() {
+                            if let Some(id) = c.as_iden_mut() {
+                                *id = "a".into();
+                            }
```
baszalmstra, line 240: Should we not also only reset this for `a`, `alpha`, `beta`, b`, `rc`?

crates/rattler_conda_types/src/version/bump.rs
```diff
@@ -217,25 +217,27 @@ impl Version {
...
-                        *last_numeral_component += 1;
-
-                        // If the segment ends with an ascii character, make it `a` instead of whatever it says
-                        let last_iden_component = segment_components
-                            .iter_mut()
-                            .filter_map(Component::as_iden_mut)
-                            .next_back();
+                        // Find the position of the last numeral so we only
+                        // reset idens that appear after it.
+                        let last_num_idx = segment_components
```
baszalmstra, line 223: Can we keep the name `last_numeral_component`? We also have segments and I find this more confusin.

crates/rattler_conda_types/src/version/bump.rs
```diff
@@ -408,4 +410,28 @@ mod test {
...
         );
     }
+
+    /// Regression test for https://github.com/conda/rattler/issues/2041
+    /// Bumping a version like 0.60b1 should produce 0.60b2, not 0.60a2.
+    /// The iden 'b' (beta) must not be reset to 'a' (alpha) when it
+    /// appears before the bumped numeral.
+    #[rstest]
+    #[case("0.60b1", VersionBumpType::Last, "0.60b2")]
+    #[case("0.60b1", VersionBumpType::Segment(1), "0.60b2")]
```
baszalmstra, line 420: This one surprised me, but makes sense I guess!

baszalmstra:
Are you still looking at this @wolfv ?

PR #2047, opened 2026-02-13
chore: Typo and clippy fixes

hunger:
### Description

Just a couple of typo and clippy fixes I ran into over time: They insert lots of extra text into my editor, which I find annoying.

I have collected them all in this PR to get them out of the way.

No behavior change is intended.


baszalmstra (approved):
Thanks this is great!

PR #2049, opened 2026-02-13
feat: add OAuth/OIDC authentication support

baszalmstra:
### Description

This PR adds support for authenticating using OAuth/OIDC. Both OAuth2 authorization code flow and Device Authorization flow are supported. By default Authorization Code flow is used but if a browser cannot be opened the code gracefully falls back to device authorization flow.

Most of the heavy lifting is done by the `openidconnect` crate. `rattler_networking` implements a custom flow to refresh tokens to minimize the number of new dependencies in that crate.

Logout also revokes the OAuth tokens which ensures they cannot be used anymore.

In the future I would like to auto discover certain servers to use oidc login, but at the moment this is not yet implemented.

### AI Disclosure

Most of the code was written by Claude Code Opus 4.6 but I manually reviewed the code.

### DIFF
--- a/crates/rattler/src/cli/auth/oauth.rs
+++ b/crates/rattler/src/cli/auth/oauth.rs
@@ -193,7 +193,9 @@ pub async fn perform_oauth_login(config: OAuthConfig) -> Result<Authentication, 
             {
                 Ok(tokens) => tokens,
                 Err(OAuthError::BrowserOpen(e)) => {
-                    eprintln!("Failed to open browser ({e}), falling back to device code flow...");
+                    tracing::info!(
+                        "Failed to open browser ({e}), falling back to device code flow..."
+                    );
                     device_code_flow(
                         &endpoints,
                         &config.client_id,


crates/rattler/src/cli/auth/oauth.rs
```diff
@@ -0,0 +1,636 @@
...
+                &config.client_id,
+                client_secret,
+                &config.scopes,
+                &http_client,
+            )
+            .await
+            {
+                Ok(tokens) => tokens,
+                Err(OAuthError::BrowserOpen(e)) => {
+                    eprintln!("Failed to open browser ({e}), falling back to device code flow...");
```
wolfv: Is it correct to use `eprintln!` here, or should we use `tracing::info/warn`?

wolfv (approved):
Great stuff! I think we can move forward with merging this and iterating some more when testing it in `pixi`

crates/rattler/src/cli/auth/oauth.rs
baszalmstra:
```suggestion
                    tracing::info!("Failed to open browser ({e}), falling back to device code flow...");
```

PR #2055, opened 2026-02-16
fix: cache negative credential lookups in auth middleware

wolfv:
@ruben-arts was complaining that Pixi was really slow on a Jetson and had a suspicion that it might be related to the authentication storage.

Turns out we did not cache negative lookups.

Previously, AuthenticationStorage::get() only cached positive results. When no credentials were found for a host, every call went through all backends including the OS keyring. Combined with wildcard expansion in get_by_url(), this caused 3+ keyring lookups per HTTP request on hosts without stored credentials. On Linux, this is especially slow since the keyring crate uses D-Bus IPC to Secret Service.

### DIFF
--- a/crates/rattler_networking/src/authentication_storage/storage.rs
+++ b/crates/rattler_networking/src/authentication_storage/storage.rs
@@ -140,6 +140,11 @@ impl AuthenticationStorage {
             }
         }
 
+        // Cache the negative result to avoid repeated backend lookups
+        // (especially important for keyring which uses D-Bus IPC on Linux).
+        let mut cache = self.cache.lock().unwrap();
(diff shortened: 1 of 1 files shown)


ruben-arts:
Before (was done through pixi)
https://share.firefox.dev/3MyyP0D

After (ran with `rattler create`, after I validated that that gave the same result as pixi install on `rattler@main`)
https://share.firefox.dev/3Oh9LM9

Edit: So this fixes the issue on the jetson 🥇

ruben-arts:
Without cache, pr2055, https://share.firefox.dev/40aRu5J
Without cache, main, https://share.firefox.dev/3OhS1QO

PR #2058, opened 2026-02-18
perf: optimized repodata loading

baszalmstra:
### Description

This PR optimizes loading repodata through the gateway. I implemented several optimizations which have been summarized in this table. The "in-memory" columns shows running consecutive queries with a single gateway in memory. The disk cache loads the data from disk. The cache is warm in both cases: no network access happens in either case.

```
 ┌──────────┬───────────────────────────────────────────────────────────┬─────────────────┬─────────────┬───────────┬─────────────┐
 │  Commit  │                        Description                        │ Warm disk cache │ vs baseline │ In-memory │ vs baseline │
 ├──────────┼───────────────────────────────────────────────────────────┼─────────────────┼─────────────┼───────────┼─────────────┤
 │ baseline │ bench setup                                               │          968 ms │           — │    520 ms │           — │
 ├──────────┼───────────────────────────────────────────────────────────┼─────────────────┼─────────────┼───────────┼─────────────┤
 │ kntwnlyl │ avoid cloning records in accumulate_records               │          515 ms │        1.9x │    155 ms │        3.4x │
 ├──────────┼───────────────────────────────────────────────────────────┼─────────────────┼─────────────┼───────────┼─────────────┤
 │ uxwopsyp │ use ahash and hashbrown for faster dependency dedup       │          487 ms │        2.0x │     71 ms │        7.3x │
 ├──────────┼───────────────────────────────────────────────────────────┼─────────────────┼─────────────┼───────────┼─────────────┤
 │ srxlllvn │ optimize package name parsing with byte-level operations  │          458 ms │        2.1x │     58 ms │        9.0x │
 ├──────────┼───────────────────────────────────────────────────────────┼─────────────────┼─────────────┼───────────┼─────────────┤
 │ wzluzyov │ precompute unique dependency names during record fetching │          407 ms │        2.4x │      8 ms │         65x │
 └──────────┴───────────────────────────────────────────────────────────┴─────────────────┴─────────────┴───────────┴─────────────┘
  ```

### How Has This Been Tested?

I added a benchmark and all tests are still working.

### AI Disclosure

### Checklist:

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/query.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/query.rs
@@ -1,5 +1,4 @@
 use std::{
-    collections::HashSet,
     future::{Future, IntoFuture},
     sync::Arc,
 };
@@ -13,7 +12,8 @@ use url::Url;
 
 use super::{
     source::{CustomSourceClient, Source},
-    subdir::{Subdir, SubdirData},
+    subdir::{PackageRecords, Subdir, SubdirData},
+    BarrierCell, GatewayError, GatewayInner, RepoData,
 };
 use crate::Reporter;
 
@@ -57,16 +57,6 @@ enum SourceSpecs {
 
     /// The record is required by a dependency.
     Transitive,
-}
-
-impl SourceSpecs {
-    /// Returns true if this record should be processed for these specs.
-    fn matches(&self, record: &RepoDataRecord) -> bool {
-        match self {
-            SourceSpecs::Input(specs) => specs.iter().any(|spec| spec.matches(record)),
-            SourceSpecs::Transitive => true,
-        }
-    }
 }
 
 /// A spec that references a package by direct URL.
@@ -304,7 +294,15 @@ impl QueryExecutor {
                 }
 
                 // Push the direct url in the first subdir result for channel priority logic
-                Ok((0, SourceSpecs::Input(vec![spec]), record))
+                let unique_deps = super::subdir::extract_unique_deps(&record);
+                Ok((
+                    0,
+                    SourceSpecs::Input(vec![spec]),
+                    PackageRecords {
+                        records: record,
+                        unique_deps,
+                    },
+                ))
             }));
         }
 
@@ -338,8 +336,8 @@ impl QueryExecutor {
                         Subdir::Found(subdir) => subdir
                             .get_or_fetch_package_records(&package_name, reporter)
                             .await
-                            .map(|records| (result_index, specs, records)),
-                        Subdir::NotFound => Ok((result_index, specs, Arc::from(vec![]))),
+                            .map(|pkg| (result_index, specs, pkg)),
+                        Subdir::NotFound => Ok((result_index, specs, PackageRecords::default())),
                     }
                 }));
             }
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: crates/rattler_conda_types/src/package_name.rs [hunk 1]
```diff
@@ -94,10 +94,19 @@ impl PackageName {
     /// assert_eq!(name.as_normalized(), "pillow");
     /// ```
     pub fn from_matchspec_str_unchecked(spec: &str) -> Self {
-        let package_name_str = name_from_matchspec_str(spec);
+        let bytes = spec.as_bytes();
+        let mut has_upper = false;
+        let mut end = bytes.len();
+        for (i, &b) in bytes.iter().enumerate() {
+            if b.is_ascii_whitespace() || matches!(b, b'>' | b'<' | b'=' | b'!' | b'~' | b';') {
```
wolfv: this should probably go to a function as we use the same snippet more than once

crates/rattler_conda_types/src/package_name.rs
```diff
@@ -180,15 +207,15 @@ impl TryFrom<String> for PackageName {
     fn try_from(source: String) -> Result<Self, Self::Error> {
         // Ensure that the string only contains valid characters
         if !source
-            .chars()
-            .all(|c| matches!(c, 'a'..='z'|'A'..='Z'|'0'..='9'|'-'|'_'|'.'))
+            .bytes()
+            .all(|b| matches!(b, b'a'..=b'z'|b'A'..=b'Z'|b'0'..=b'9'|b'-'|b'_'|b'.'))
```
wolfv, line 203: probably doesn't change much but I think the range could be `a..9`?

wolfv:
Exciting stuff!

crates/rattler_conda_types/src/package_name.rs
```diff
@@ -133,28 +130,52 @@ impl PackageName {
...
+/// constraint character: `>`, `<`, `=`, `!`, `~`, `;`).
+#[inline]
+fn is_matchspec_delimiter(b: u8) -> bool {
+    b.is_ascii_whitespace() || matches!(b, b'>' | b'<' | b'=' | b'!' | b'~' | b';')
+}
+
+/// Scans a matchspec string to find the package name boundary and whether it
+/// contains uppercase characters. Single-pass over the bytes.
+#[inline]
+fn scan_matchspec_name(spec: &str) -> (&str, bool) {
```
nichmor, line 150: do we need to inline them? I think they are used in the same crate, meaning that they will inlined automatically by the compiler ( https://matklad.github.io/2021/07/09/inline-in-rust.html#Inlining-in-Rust )

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -354,40 +356,61 @@ impl QueryExecutor {
...
             }
         }
     }
 
+    /// Queue a single dependency if not already seen.
+    ///
+    /// Uses `entry_ref` for a single hash lookup. Only allocates when the
+    /// name is genuinely new (~500 unique names vs ~1M+ dependency strings).
+    #[inline]
+    fn queue_dependency(&mut self, dependency: &str) {
```
nichmor, line 381: same here about inlining

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -151,8 +152,9 @@ struct QueryExecutor {
     direct_url_specs: Vec<DirectUrlSpec>,
 
     // Mutable state during execution
-    seen: HashSet<PackageName>,
-    pending_package_specs: HashMap<PackageName, SourceSpecs>,
+    /// Normalized (lowercase) package names we've already queued.
+    seen: hashbrown::HashMap<String, (), ahash::RandomState>,
```
nichmor, line 144: should we also use ahash here?

baszalmstra: We are! hashbrown is the HashMap/Set implementation used by the standard library, but the hashbrown crate itself exposes some more low-level APIs that I use (entry_ref). The hasher used is the last generic argument.

nichmor: ah yeah! Didn't see it!

crates/rattler_conda_types/src/package_name.rs
baszalmstra: Yeah I doubt that will make a difference

PR #2063, opened 2026-02-19
feat(rattler_conda_types): serialize MatchSpec as string

### LINKED ISSUE #847: Deserialize from string by default
We often write code like this:

```rust
#[serde_as(as = "Vec<DisplayFromStr>")]
pub constrains: Vec<MatchSpec>,
```

This is because the default serde implementation serializes all the fields. However, I think its more common to serialize matchspec as a string then to serialize the individual fields. I would therefor propose that we serialize and deserialize matchspecs as strings instead. 

This is also the case for `NamelessMatchSpec` and `Version` and possibly some other types.

devvaansh:
### Description

Fixes #847

Changes the default serde implementation for MatchSpec and NamelessMatchSpec to serialize/deserialize as strings instead of serializing individual struct fields.

Root cause: The default #[derive(Serialize, Deserialize)] was serializing MatchSpec as a JSON object with individual fields (name, version, build, channel, ...). Conda always represents match specs as strings — so the struct serialization was both verbose and inconsistent with the ecosystem.

Screenshot 2026-02-19 at 8 34 51 AM

Fixes #847

### How Has This Been Tested?

Existing tests — all pass with updated snapshots:

cargo test -p rattler_conda_types
test result: ok. 380 passed; 0 failed; 0 ignored

cargo test -p rattler-lock  
test result: ok. 83 passed; 0 failed; 0 ignored

New test added — test_serde_roundtrip in match_spec/mod.rs directly covers the issue's use case (serializing Vec<MatchSpec> and Vec<NamelessMatchSpec> without DisplayFromStr):

#[test]
fn test_serde_roundtrip() {
    // The exact use case from issue #847 — Vec<MatchSpec> without DisplayFromStr
    let specs: Vec<MatchSpec> = vec![
        MatchSpec::from_str("conda-forge::pytest >=7.0,<8 py39_0", 
                            ParseStrictness::Lenient).unwrap(),
        MatchSpec::from_str("numpy >=1.21.*", 
                            ParseStrictness::Lenient).unwrap(),
    ];
    let json = serde_json::to_string(&specs).unwrap();
    let roundtripped: Vec<MatchSpec> = serde_json::from_str(&json).unwrap();
    assert_eq!(specs, roundtripped);

    // Same for NamelessMatchSpec
    let nameless: Vec<NamelessMatchSpec> = vec![
        NamelessMatchSpec::from_str(">=7.0,<8 py39_0", 
                                    ParseStrictness::Lenient).unwrap(),
    ];
    let json = serde_json::to_string(&nameless).unwrap();
    let roundtripped: Vec<NamelessMatchSpec> = serde_json::from_str(&json).unwrap();
    assert_eq!(nameless, roundtripped);
}

### AI Disclosure

Tools: CLAUDE

### Checklist:


crates/rattler_lock/src/utils/serde/match_spec_map_or_vec.rs
```diff
@@ -14,15 +14,14 @@ impl<'de> DeserializeAs<'de, Vec<String>> for MatchSpecMapOrVec {
     where
         D: Deserializer<'de>,
     {
+        // NamelessMatchSpec now deserializes from a string by default, so no
```
baszalmstra: This comment is a delta comment often added by AI. it describes a change in situation not the current one, please remove.

devvaansh: done

devvaansh:
> Nice!
> 
> Do these changes also properly work with nested quotes? Are the properly escaped?
> 
> For namelessmatchspec we can keep the old behavior as you will never use that as a string.

Thanks , updated the PR to address these points:reverted [NamelessMatchSpec] to use the original derived [Serialize], [Deserialize] implementation (and restored the [DisplayFromStr]  wrapper in [match_spec_map_or_vec.rs] no longer serializes to a string by default.  

 they work correctly  [MatchSpec] uses double quotes for fields like [sha256="..."] or fn="..." When serialized to JSON, these successfully round-trip as escaped characters (e.g. [sha256=\"...\"])  I added a specific test case in [test_serde_roundtrip] containing [sha256] and `md

baszalmstra:
A license could contain a `"` which would break the conversion to string

devvaansh:
> A license could contain a `"` which would break the conversion to string

good catch, I have done this in the latest commit

crates/rattler_conda_types/src/match_spec/snapshots/rattler_conda_types__match_spec__parse__tests__test_from_string_Lenient.snap
```diff
@@ -1,142 +1,36 @@
...
-    name: conda-forge
-  subdir: linux-64
-numpy>=2.*.*:
-  name: numpy
-  version: ">=2"
-"bird_tool_utils_python =0.*,>=0.4.1":
-  name: bird_tool_utils_python
-  version: "0.*,>=0.4.1"
+"./relative/channel::package": "file://<CRATE>/relative/channel/::package"
+"python[channel=https://conda.anaconda.org/python/conda,version=3.9]": "python/conda::python ==3.9"
```
baszalmstra: mm here some information is lost with regards to the url. What happens if you use https://prefix.dev/conda-forge ?

devvaansh: now serializes the full URL for non-default servers while maintaining the short name for default ones, ensuring no information is lost.

baszalmstra: lets just always serialize as full url. There can be no issues then.

devvaansh: done

devvaansh:
@baszalmstra I have addressed your reviews

crates/rattler_conda_types/src/channel/mod.rs
```diff
@@ -19,7 +19,7 @@ mod channel_url;
 
 pub use channel_url::ChannelUrl;
 
-const DEFAULT_CHANNEL_ALIAS: &str = "https://conda.anaconda.org";
+pub(crate) const DEFAULT_CHANNEL_ALIAS: &str = "https://conda.anaconda.org";
```
baszalmstra: Can this be reverted?

crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -173,11 +177,34 @@ pub struct MatchSpec {
...
+
 impl Display for MatchSpec {
     fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
         if let Some(channel) = &self.channel {
-            let name = channel.name();
-            write!(f, "{name}")?;
+            // Always serialize as the full URL to preserve all channel server
+            // information during round-tripping.
+            let url = channel.base_url.url();
+            if matches!(url.scheme(), "https" | "http") {
```
baszalmstra: Why do this different for http and other schemas?

baszalmstra:
Can you also update the title?

crates/rattler_conda_types/src/match_spec/mod.rs
devvaansh: simplified it to unconditionally use channel.base_url.as_str().trim_end_matches('/') for all schemes

crates/rattler_conda_types/src/channel/mod.rs
devvaansh: done

devvaansh:
@baszalmstra all reviews have been addressed.

devvaansh:
@baszalmstra please take a look

devvaansh:
@baszalmstra please review this pr once

baszalmstra:
I think Im backtracking on the decision to always serialize as strings. I think we are going into a direction where sometimes we actually do want to serialize as an object. Using this implementation makes that much harder. I appreciate the work you did here, I apologize for the change of vision.

PR #2065, opened 2026-02-19
fix: gracefully handle missing HOME env var in file backend (#1013)

### LINKED ISSUE #1013: pixi/rattler_network crashes without a useful error message if $HOME isn't defined
### Checklist

### What happened?

If I try to access a private channel on Linux without the environment variable HOME set, pixi is crashing in 
https://github.com/conda/rattler/blob/main/crates/rattler_networking/src/authentication_storage/backends/file.rs#L149

The error is just that "called 'Option::unwrap()' on a 'None' value"

I was debugging something else and had my whole environment empty, so that "HOME" was the problem wasn't immediately obvious.  I don't have a problem with requiring that that environment variable being required, I just suspect it'd be better to catch that error and give a better message.

### Additional Context

_No response_

DCchoudhury15:
### Description

This PR fixes a crash occurring when the `$HOME` environment variable is undefined and the operating system cannot provide a fallback. Calling `.unwrap()` on the home directory was causing a panic, which resulted in a crash without a helpful error message. 

I've replaced the `.unwrap()` with graceful error handling in the networking file backend and added a safe fallback for the test suite in `rattler_conda_types`.

Fixes #1013

### How Has This Been Tested?

- Verified that `FileStorage::new()` in `rattler_networking` now returns a `FileStorageError::IOError` instead of panicking when `HOME` is missing.
- Verified that all unit tests in `rattler_conda_types` pass with the new fallback logic.
- Ran `pixi run cargo test --workspace` to ensure no regressions in affected crates.

### AI Disclosure
- [] This PR contains AI-generated content.
- [] I have tested any AI-generated content in my PR.
- [] I take responsibility for any AI-generated content in my PR.

### Checklist:
- [x] I have run `cargo fmt` and `cargo clippy` and they pass.

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/parse.rs
+++ b/crates/rattler_conda_types/src/match_spec/parse.rs
@@ -1260,15 +1260,19 @@ mod tests {
         // Strip absolute paths to this crate from the channels for testing
         let crate_root = env!("CARGO_MANIFEST_DIR");
         let crate_path = Url::from_directory_path(std::path::Path::new(crate_root)).unwrap();
-        let home = Url::from_directory_path(dirs::home_dir().unwrap()).unwrap();
+
+        let home_str = dirs::home_dir()
+            .and_then(|p| Url::from_directory_path(p).ok())
+            .map_or_else(|| "file:///dummy_home/".to_string(), |u| u.to_string());
+
         insta::with_settings!({filters => vec![
             (crate_path.as_str(), "file://<CRATE>/"),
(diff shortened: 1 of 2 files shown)


DCchoudhury15:
@baszalmstra up for review

baszalmstra:
Can you please fill out the PR template?

DCchoudhury15:
@baszalmstra rewrote the description

baszalmstra:
Thanks!

PR #2068, opened 2026-02-19
fix: retry at least three times during install

### LINKED ISSUE #2054: Not error on a broken pipe while downloading package
### Checklist

### What is the idea?

Don't error when installation fails on `stream closed because of a broken pipe` but retry first.

### Why is this needed?

I have seen more then once that running a installation with Pixi can result in an error like this:
```
❯ pixi run -e jazzy start
✨ Pixi task (start in jazzy): ros2 launch nav2_bringup tb3_simulation_launch.py headless:=False
Error:   × failed to fetch tiledb-2.30.0-hb43dc9e_6.conda
  ├─▶ failed to interact with the package cache layer.
  ├─▶ error sending request for url (https://prefix.dev/conda-forge/linux-aarch64/tiledb-2.30.0-hb43dc9e_6.conda)
  ├─▶ client error (SendRequest)
  ├─▶ connection error
  ╰─▶ stream closed because of a broken pipe
```
Which works the second time I run the command. So I assume this could also be done by the caching layer itself. 

### What should happen?

If there is a stream closed because of a broken pipe, just retry for a max of 3 times, after that error.

### Additional Context

I feel like I've seen this more often when there is a big load on a small machine. In this case it was an ROS workspace installation on a Jetson. There where no signs of actual disk overload, but maybe the error could be resolved in the first place 🤷‍♂️ .

vee1e:
### Description

Changed the installation logic to retry at least three times in case of error `stream closed because of a broken pipe` 

Added a `should_retry` method that returns whether an Error should be retried or not.

All previous tests pass indicating no regression. 

Fixes #2054

### How Has This Been Tested?

Added several (16) unit test cases in `crates/rattler_package_streaming/src/lib.rs` for every case of pipe failure, all of which pass. These can probably be brought down based on actual test data, would love your input on that.

Also added 2 integration tests in `crates/rattler_cache/src/package_cache/mod.rs` for conda and bz2 packages, all of which pass.

Also my apologies, [f1a196d](https://github.com/conda/rattler/pull/2068/commits/f1a196d107dff97e29c892d7a0c9c2132388bb9b) pushed with the wrong commit message of a previous PR. Let me know if it needs rebasing and if I should squash all commits.

### AI Disclosure
Tools: Kimi K2.5

### Checklist:

### DIFF
--- a/crates/rattler_cache/src/package_cache/mod.rs
+++ b/crates/rattler_cache/src/package_cache/mod.rs
@@ -988,25 +988,26 @@ mod test {
 
             let byte_count = *bytes.lock().await;
             // Create a stream that yields partial data then fails with broken pipe
-            let partial_data: Bytes = buffer.into_iter().take(byte_count).collect::<Vec<u8>>().into();
+            let partial_data: Bytes = buffer
+                .into_iter()
+                .take(byte_count)
+                .collect::<Vec<u8>>()
+                .into();
 
-            let stream = stream::unfold(
-                (false, partial_data),
-                |(has_sent, data)| async move {
-                    if has_sent {
-                        // Return broken pipe error on second iteration
-                        return Some((
-                            Err(std::io::Error::new(
-                                std::io::ErrorKind::BrokenPipe,
-                                "stream closed because of a broken pipe",
-                            )),
-                            (true, data),
-                        ));
-                    }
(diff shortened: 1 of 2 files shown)


baszalmstra:
This looks exceptionally well done. Thanks!

PR #2070, opened 2026-02-19
feat: add History primitive for conda-meta/history files (#1963)

### LINKED ISSUE #1963: `History` primitive?
### Checklist

### What is the idea?

Add readers and writers to `rattler` so users can process `conda-meta/history` files.

### Why is this needed?

Essential in imperative APIs like `conda`'s. We need to know which specs the user has requested in the past because there's no declarative input file.

### What should happen?

A reader should be able to parse a `conda-meta/history` file and provide access to the different revisions, and the accumulated specs. The writer should be able to write revisions and the associated specs per revision.

### Additional Context

See `conda`'s [`History`](https://github.com/conda/conda/blob/main/conda/history.py) class.

Titas-Ghosh:
Fixes #1963 

### Description

Adds a new `history` module to `rattler_conda_types` that provides read and write access to `conda-meta/history` files — the log that conda environments use to record every change (revision) made to the environment.

This addresses the need for a `History` primitive in rattler, as discussed in **issue #1963**.

#### What is `conda-meta/history`?

Every conda environment maintains a `conda-meta/history` file that records a log of revisions. Each revision contains:
- A timestamp header line: `==> 2024-01-15 10:30:00 <==`
- Optional comment lines: `# cmd:`, `# conda version:`, `# install specs:` etc.
- A list of package distribution strings — either as an absolute set (initial creation) or as diffs prefixed with `+` (added) / `-` (removed)

#### New Types

| Type | Description |
|------|-------------|
| `History` | Main entry-point for reading/writing history files. Constructed from a prefix path or explicit file path. |
| `HistoryRevision` | A single parsed revision containing timestamp, packages (`BTreeSet<String>`), and raw comment lines. |
| `UserRequest` | Structured data extracted from comment lines — command, conda version, action, and categorized specs (update/remove/neutered). |
| `HistoryError` | Error type using `thiserror` — wraps I/O errors and parse errors. |

#### Capabilities

- **`History::parse()`** — Parses the history file into a `Vec<HistoryRevision>`. Returns empty vec if file doesn't exist.
- **`History::from_reader()`** — Parses from any `impl Read`.
- **`History::parse_str()`** — Parses from a `&str` (useful for testing).
- **`History::get_user_requests()`** — Extracts structured `UserRequest`s from the parsed history.
- **`History::user_requests_from_revisions()`** — Static version that works on already-parsed revisions.
- **`History::write_revision()`** — Appends a new revision with a given timestamp, removed and added package sets (diff format with `+`/`-` prefixes).
- **`History::write_full_revision()`** — Convenience method that writes a complete revision with metadata comments (cmd, conda version, action specs) in one call.
- **`History::write_comment()`** — Appends a single comment line.
- **`History::clear()`** — Truncates the history file (useful for tools like pixi that don't maintain running history).
- **`History::overwrite()`** — Replaces the entire file with a given list of revisions.

#### Design Decisions

1. **No `chrono::clock` dependency**: The `write_revision` and `write_full_revision` methods accept a `timestamp: &str` parameter instead of using `Utc::now()`. This avoids requiring the `clock` feature flag on `chrono` (which is not enabled in `rattler_conda_types`'s workspace config) and gives callers full control over timestamps.

2. **No state reconstruction**: As discussed in the issue, state reconstruction from the diff log was intentionally omitted. The history file is a log of user actions, not a source of truth for environment state. `conda-meta/` JSON records serve that purpose.

3. **Pixi compatibility**: `clear()` and `overwrite()` methods are included to support tools like pixi that need to manage the history file differently from conda's append-only behavior.

4. **Spec parsing handles both formats**: The `parse_specs_string` helper handles both Python list format (`['numpy', 'pandas>=1.5']`) and the older comma-separated format (`param >=1.5.1,<2.0,python>=3.5`), including proper version constraint continuation after commas.

5. **Follows crate conventions**: Uses `thiserror` for errors, `regex` for parsing, `fs_err` for file I/O, `BTreeSet` for deterministic ordering, and `#[cfg(test)]` inline tests with `tempfile` — matching patterns used throughout `rattler_conda_types`.

### How Has This Been Tested?

14 unit tests covering all reader, writer, and parser functionality:

**Parsing tests:**
- `parse_empty` — empty input returns empty vec
- `parse_revisions` — parses a 3-revision sample history with correct timestamps, packages, and comments
- `parse_user_requests` — extracts structured `UserRequest` with correct cmd, conda_version, action, and specs
- `parse_ignores_lines_before_first_header` — lines before the first `==> ... <==` header are silently ignored (matching conda behavior)
- `parse_nonexistent_file` — non-existent file returns empty vec instead of error

**Spec parsing tests:**
- `parse_specs_python_list_format` — `['numpy', 'pandas>=1.5']` → `["numpy", "pandas>=1.5"]`
- `parse_specs_old_comma_format` — `param >=1.5.1,<2.0,python>=3.5` → `["param >=1.5.1,<2.0", "python>=3.5"]`
- `parse_specs_empty` — empty string returns empty vec
- `parse_specs_filters_at_suffix` — specs ending with `@` are filtered out

**Writer tests:**
- `write_and_read_roundtrip` — write revision, read it back, verify packages
- `write_full_revision_roundtrip` — write full revision with comments, verify cmd/version are preserved in user requests
- `write_multiple_revisions` — append two revisions, verify both are parsed
- `clear_truncates_file` — write content, clear, verify empty
- `overwrite_replaces_content` — write content, overwrite with new revisions, verify old content is gone

**All tests verified locally:**
Screenshot 2026-02-19 185010

Additionally verified:
- `cargo check -p rattler_conda_types` — compiles clean
- `cargo fmt -p rattler_conda_types -- --check` — no formatting issues
- `cargo clippy -p rattler_conda_types -- -D warnings` — no clippy warnings

### AI Disclosure

Tools: Claude Opus 4.6 , i used it for implementation of the `history` module, following existing crate patterns (`explicit_environment_spec.rs`, `prefix_record.rs`). All code reviewed and tested by the contributor.

### Checklist:
- [x] I have commented my code, particularly in hard to understand areas

This is my first contribution towards this project , happy to iterate based on feedback provided my the mentors.
Thanks!


crates/rattler_conda_types/src/history.rs
```diff
@@ -0,0 +1,695 @@
...
+        }
+
+        Ok(revisions)
+    }
+
+    /// Extracts structured [`UserRequest`]s from the parsed history.
+    ///
+    /// A user request is produced for every revision that contains a
+    /// `# cmd: ...` comment.
+    pub fn get_user_requests(&self) -> Result<Vec<UserRequest>, HistoryError> {
```
baszalmstra, line 287: Maybe it would be nice to also have a type for the parsed content? Then you can add query functions there.

crates/rattler_conda_types/src/history.rs
```diff
@@ -0,0 +1,695 @@
...
+        let revisions = self.parse()?;
+        Ok(Self::user_requests_from_revisions(&revisions))
+    }
+
+    /// Extracts [`UserRequest`]s from an already-parsed list of revisions.
+    pub fn user_requests_from_revisions(revisions: &[HistoryRevision]) -> Vec<UserRequest> {
+        let cmd_re = Regex::new(r"^#\s*cmd:\s*(.+)$").expect("valid regex");
```
baszalmstra: Please use lazy regexes to avoid initializing them on every invocation

crates/rattler_conda_types/src/history.rs
```diff
@@ -0,0 +1,695 @@
...
+    /// Packages that were removed should appear in `removed` and packages
+    /// that were added in `added`. The sets contain distribution strings
+    /// without any `+`/`-` prefix — the prefix is added automatically.
+    pub fn write_revision(
+        &self,
+        timestamp: &str,
+        removed: &BTreeSet<String>,
+        added: &BTreeSet<String>,
+    ) -> Result<(), HistoryError> {
+        use std::io::Write;
```
baszalmstra: Please move all use statements to the top.

crates/rattler_conda_types/src/history.rs
```diff
@@ -0,0 +1,695 @@
...
+    /// If the parent directory does not exist it will be created. The
+    /// `timestamp` should be a string in `YYYY-MM-DD HH:MM:SS` format.
+    /// Packages that were removed should appear in `removed` and packages
+    /// that were added in `added`. The sets contain distribution strings
+    /// without any `+`/`-` prefix — the prefix is added automatically.
+    pub fn write_revision(
+        &self,
+        timestamp: &str,
+        removed: &BTreeSet<String>,
+        added: &BTreeSet<String>,
```
baszalmstra: Maybe we can add a type for Revision?

crates/rattler_conda_types/src/history.rs
```diff
@@ -0,0 +1,695 @@
...
+            if req.cmd.is_some() {
+                requests.push(req);
+            }
+        }
+
+        requests
+    }
+
+    // -----------------------------------------------------------------------
+    // Writer
```
baszalmstra, line 293: Perhaps it makes sense to create a HistoryWriter type to avoid reopening the file.

crates/rattler_conda_types/src/history.rs
```diff
@@ -0,0 +1,695 @@
...
+// ---------------------------------------------------------------------------
+
+#[cfg(test)]
+mod tests {
+    use std::collections::BTreeSet;
+
+    use super::*;
+
+    /// A sample history file matching conda's format.
+    const SAMPLE_HISTORY: &str = "\
```
baszalmstra, line 504: Does this also contain edge cases? I dont see neutered specs for instance.

baszalmstra:
Do you think we can also expose this to python? Also fine to do this in a followup pr.

crates/rattler_conda_types/src/history.rs
Titas-Ghosh: That's a good idea sir, i'll add a ParsedHistory type to wrap the parsed revisions and move query functions like 
get_user_requests there, i think that will make the API cleaner and more extensible. Will push an update shortly.

crates/rattler_conda_types/src/history.rs
Titas-Ghosh: Thanks for the catch ,i have switched all `Regex::new()` calls to `lazy_regex::regex!()` in the latest commit e9f16856

crates/rattler_conda_types/src/history.rs
Titas-Ghosh: i have implemented the changes in the commit 492f448, kindly review

crates/rattler_conda_types/src/history.rs
Titas-Ghosh: Moved all use statements to the top in c03985e1. Thanks!

crates/rattler_conda_types/src/history.rs
Titas-Ghosh: Thanks for the suggestion @baszalmstra ,i have added a Revision type that groups `timestamp`, `removed`, and `added` 
Also `write_revision` and `write_full_revision` now take `&Revision`. Kindly review in the commit c1b5b954

crates/rattler_conda_types/src/history.rs
Titas-Ghosh: Sir , I've introduced a `HistoryWriter` struct as suggested that keeps the file open, and added a `History::writer()` method to create it. Kindly review commit ab07015e.
Thanks! @baszalmstra

crates/rattler_conda_types/src/history.rs
Titas-Ghosh: That's a good catch sir! You're right that we should cover that edge case, i've updated `SAMPLE_HISTORY` to include a simulated revision with `# neutered specs: ['numpy']`.

I also extended the tests to verify it:

- `parse_revisions` now checks for the presence of this 4th revision.
- `parse_user_requests` explicitly verifies that the neutered action is captured and correctly populated in the   `neutered_specs` field.

This ensures the `neutered` branch in **user_requests_from_revisions** is fully covered.
Kindly review the commit de9e7a0e .
Once again thanks for the feedback @baszalmstra

Titas-Ghosh:
> Do you think we can also expose this to python? Also fine to do this in a followup pr.

Definitely sir @baszalmstra ,i agree that exposing to Python would be very valuable, especially given the structured **ParsedHistory** and the new **HistoryWriter**. It would allow Python tooling to interact with history files much more reliably.
However I do think it's best to implement this in a follow-up PR. I'd be happy to pick that up next based on your feedback and suggestions.
Thanks!

PR #2071, opened 2026-02-19
fix up PR template

vee1e:
### Description

Make all the indentation for the checkboxes the same. I had to do this manually for my previous PRs 😅

### Checklist:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -13,7 +13,7 @@ Fixes #{issue}
 
 ### AI Disclosure
 <!--- Remove this section if your PR does not contain AI-generated content. --->
-  - [ ] This PR contains AI-generated content.
+- [ ] This PR contains AI-generated content.
   - [ ] I have tested any AI-generated content in my PR.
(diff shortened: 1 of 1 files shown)


.github/pull_request_template.md
```diff
@@ -13,18 +13,18 @@ Fixes #{issue}
 
 ### AI Disclosure
 <!--- Remove this section if your PR does not contain AI-generated content. --->
-- [ ] This PR contains AI-generated content.
```
baszalmstra, line 16: Ah the two below this one should be nested though, you only have to check the bottom two if it contains AI generated content. Looking at it now it actually seems to be correct? What is going wrong for you?

vee1e: Ah, if its intentional then its fine. I'll only indent the checklist part. Thanks for clarifying.

baszalmstra: I dont think I understand why the other list should be indented 😅

vee1e: Yeah now that I see your logic I think its fine being non-indented. Thanks for addressing my changes, I'll close the PR.

PR #2074, opened 2026-02-19
feat: implement lazy PrefixData primitive with tests

### LINKED ISSUE #1962: `PrefixData` primitive?
### Checklist

### What is the idea?

`conda` has `conda.core.prefix_data.PrefixData`, which allows user to inspect existing conda environments. It's like a factory of `PrefixRecords` (one per `conda-meta/*.json`). I wonder if there's an interest in having the same abstractions here, to perform operations like "tell me if this package is installed in this environment" or "which Python version is installed in this environment".

### Why is this needed?

Convenience.

### What should happen?

Not sure about the API. Right now it's quite trivial to build a dictionary of PrefixRecords, keyed by name. But one advantage of the object would be to construct it lazily (which could offer some performance improvements in large environments with 100s of packages).

### Additional Context

See `conda`'s [`PrefixData`](https://github.com/conda/conda/blob/fa4c2b8397ca9685206bcc845332e95fd0d35b83/conda/core/prefix_data.py#L113).

KartikDua1504:
### Description

This PR introduces the PrefixData primitive to allow inspecting existing conda environments efficiently. It implements a thread-safe, per-key lazy loading architecture to drastically reduce I/O and parsing overhead in large Conda environments with hundreds of packages.

###Implementation Details:

Initialization (new): Performs a single, fast fs::read_dir sweep to extract package names from filenames (stripping .json and splitting by -). Zero file I/O or JSON parsing happens here.

Lazy Loading (get): Utilizes PrefixRecord::from_path() to parse the JSON strictly on the first query for a specific package.

Data Structure: Uses a custom internal LazyRecordEntry struct to hold the PathBuf and a OnceLock<Option<PrefixRecord>>.

###Core Architectural Assumptions:

Static Environment: We assume the conda-meta directory does not change during the lifetime of the PrefixData instance. We scan the directory exactly once at initialization.

Parsed Exactly Once: If a package is queried, its JSON is parsed and cached. If parsing fails (e.g., malformed JSON), the error is caught, and None is safely cached to prevent infinite retry loops.

Ergonomic Concurrency: The struct only requires &self for queries. Because the HashMap structure is completely immutable after new(), and OnceLock provides safe interior mutability, this primitive is perfectly thread-safe for highly concurrent workloads.

Fixes #1962 

### How Has This Been Tested?

I added a strict unit test (test_prefix_data_strict in crates/rattler_conda_types/src/prefix_data.rs) to verify the lazy loading behavior and edge cases:

Created a temporary conda-meta directory with mock valid Conda JSON files and non-JSON decoy files (.txt, .yaml).

Verified that the initial fs::read_dir sweep correctly ignores non-JSON files.

Verified the string splitting logic accurately extracts standard package names and package names with hyphens (e.g., scikit-learn).

Verified that feeding the lazy-loader an empty/invalid JSON file safely caches a None value in the OnceLock instead of panicking.

### AI Disclosure

Tools: Gemini 3.0 (Used for discussing the architecture, handling edge-cases with Conda filename string splitting, documentation and syntax correction).

### Checklist:

-  [x] I have performed a self-review of my own code

-  [x] I have commented my code, particularly in hard-to-understand areas

-  [x] I have made corresponding changes to the documentation

-  [x] I have added sufficient tests to cover my changes.

### DIFF
--- a/crates/rattler/src/install/clobber_registry.rs
+++ b/crates/rattler/src/install/clobber_registry.rs
@@ -14,6 +14,26 @@ use rattler_conda_types::{
 
 pub const CLOBBERS_DIR_NAME: &str = "__clobbers__";
 
+/// Controls how clobbered files are handled during installation.
+///
+/// When multiple packages install the same file, the default behavior is to
+/// rename the "losing" files to a `__clobbers__/` directory. This enum allows
+/// callers to instead raise an error when clobbering is detected.
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub enum ClobberMode {
+    /// The default behavior: when multiple packages install the same file,
+    /// the "losing" packages have their files renamed to a `__clobbers__/`
+    /// directory. The file from the highest-priority package "wins".
+    #[default]
+    Rename,
+
+    /// Raise an error if any clobbering is detected during post-processing.
+    /// With this mode, if multiple packages contain the same file path, the
+    /// installation will fail with an error instead of silently renaming
+    /// files.
+    Error,
+}
+
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ClobberedPath {
     /// The name of the package from which the final file is taken.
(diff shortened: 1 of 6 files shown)


KartikDua1504:
@pranavpurankar 
Hi! Thanks for taking a look and for the thoughtful questions. I'm glad you found the approach interesting! 

To address your points:

* **Filename Parsing:** Yes, `rsplitn(3, '-')` handles hyphens in the package name perfectly! Because it splits from the *right*, it only isolates the last two hyphenated segments (the build string and the version). Everything else to the left (no matter how many hyphens it contains) stays grouped together as the package name. I actually included a specific unit test for `scikit-learn` in this PR exactly to verify this edge case. Regarding existing utilities, since the goal of `new()` is to avoid parsing files entirely and just do a rapid directory scan, lightweight string manipulation on the filename is the fastest path without invoking heavier parser machinery.

* **Error Handling:** That is a very fair point. I initially went with `Option<&PrefixRecord>` to match the ergonomic, infallible API of `HashMap::get`. In the context of a lazy view, if a record is so corrupted that it cannot be parsed, the environment state is effectively broken, and practically speaking, the package will usually need to be reinstalled anyway. Therefore, it functionally behaves as if the package is missing or invalid for the resolver. However, you're right that it masks the specific cause of failure. If the maintainers (@jaimergp / @baszalmstra) prefer a `Result<Option<&PrefixRecord>, ...>` API to explicitly surface disk/parse errors, I am completely happy to refactor the internal lock to cache a `Result`.

* **Memory Footprint:** Memory mapping (`mmap`) is fantastic for large, monolithic files, but `conda-meta` consists of many individual, very small JSON files (usually just a few KB each). The OS overhead of setting up memory mapping for hundreds of tiny files would likely outweigh the benefits. Furthermore, because of the `OnceLock`, we are never deserializing the whole environment at once—we only pay the memory cost for the specific records the user explicitly queries. For sparse lookups, the memory footprint remains extremely low.

Thanks again for the review, it's great to dive into the internals further.

baszalmstra:
Before reviewing can you please fill out the PR template correctly.

crates/rattler_conda_types/src/prefix_data.rs
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+        Ok(Self {
+            prefix_path,
+            records,
+        })
+    }
+
+    /// Retrieves a record lazily. Parses the JSON only on the first call.
+    pub fn get(&self, package_name: &str) -> Option<&PrefixRecord> {
```
baszalmstra: Yeah I would also expect a result here. Many things can go wrong.

crates/rattler_conda_types/src/lib.rs
```diff
@@ -27,6 +27,8 @@ pub mod minimal_prefix_record;
 pub mod package;
 mod package_name;
 pub mod prefix;
+/// Lazily populated view of the 'conda-meta' directory.
```
baszalmstra: Add as module documentation instead

crates/rattler_conda_types/src/prefix_data.rs
KartikDua1504: Sure, I will start working on it immediately.

crates/rattler_conda_types/src/lib.rs
KartikDua1504: Sure

KartikDua1504:
I have updated the PR body. Kindly check if it now aligns with the standard. Apologies for the inconvenience.

baszalmstra:
> I have updated the PR body. Kindly check if it now aligns with the standard. Apologies for the inconvenience.

Did you use any AI? Thats also part of the template.

KartikDua1504:
I did use AI, but for looking up dependencies ,appropriate documentation purposes and syntax correction. Should I still include it ?
EDIT : Updated

baszalmstra:
Using AI is not a bad thing. I use it a lot too. Its just nice to disclose what you used it for to make the review process more targeted.

KartikDua1504:
Noted. I updated the PR Request.

crates/rattler_conda_types/src/prefix_data.rs
```diff
@@ -0,0 +1,153 @@
...
+            }
+        }
+        Ok(Self {
+            prefix_path,
+            records,
+        })
+    }
+
+    /// Retrieves a record lazily. Parses the JSON only on the first call.
+    pub fn get(&self, package_name: &str) -> Option<Result<&PrefixRecord, &std::io::Error>> {
```
wolfv: This should use the proper types from rattler (PackageName).

crates/rattler_conda_types/src/prefix_data.rs
```diff
@@ -0,0 +1,153 @@
...
+            let path = entry.path();
+
+            if path.extension().and_then(|s| s.to_str()) != Some("json") {
+                continue;
+            }
+
+            if let Some(filename) = path.file_name().and_then(|s| s.to_str()) {
+                if let Some(name) = filename
+                    .strip_suffix(".json")
+                    .and_then(|s| s.rsplitn(3, '-').last())
```
wolfv: We have a type that does this properly (ArchiveIdentifier or PackageIdentifier). I think that would be better!

KartikDua1504: Thanks for the recommendation. I will update the PR accordingly.

KartikDua1504:
I have updated the PR to include PackageName and ArchiveIdentifier

crates/rattler_conda_types/src/prefix_data.rs
```diff
@@ -0,0 +1,164 @@
...
+            prefix_path,
+            records,
+        })
+    }
+
+    /// Retrieves a record lazily. Parses the JSON only on the first call.
+    pub fn get(
+        &self,
+        package_name: &PackageName,
+    ) -> Option<Result<&PrefixRecord, &std::io::Error>> {
```
baszalmstra: its a bit cumbersome to return a reference to an Error because its hard to propagate. Maybe you can add a custom error that wraps an Arc io error?

KartikDua1504:
@baszalmstra Kindly review if is this okay now.

PR #2078, opened 2026-02-19
bench(lock): add benchmarks for lock file size

### LINKED ISSUE #765: Add benchmark for pixi lock file size
Big pixi lock files can become quite a problem as discussed in https://github.com/prefix-dev/pixi/issues/1509.
So we want to make sure we want to keep track of it.

The following benchmarks would be interesting:
- lock file size
- compressed lock file size
- git repo size with multiple commits containing lock file changes

vishwas-droid:
This PR adds Criterion benchmarks for monitoring lock file growth:

- Raw lock file size
- Compressed lock file size
- Git repository size across multiple commits with lock file changes

These benchmarks help track performance and repository growth impact over time.

Fixes #765

### DIFF
--- a/crates/rattler_lock/Cargo.toml
+++ b/crates/rattler_lock/Cargo.toml
@@ -35,11 +35,5 @@ insta = { workspace = true, features = ["yaml"] }
 serde_json = { workspace = true }
 similar-asserts = { workspace = true }
 rstest = { workspace = true }
-criterion = "0.5"
 flate2 = "1"
 tempfile = "3"
-	
-[[bench]]
-name = "lock_file_size"
-harness = false
-


baszalmstra:
It really doesnt make sense tonuse criterion for this. Criterion is used to measure time while these measurements check size.

vishwas-droid:
Updated the PR: removed Criterion and replaced benchmarks with deterministic size tests. Please let me know if any further changes are needed.

vishwas-droid:
Thanks for the clarification.

You're right — the current implementation measures size but does not enforce a regression boundary.

I’ll revise this to introduce deterministic regression guards with explicit size thresholds so CI fails if lock file growth exceeds expectations.

I’ll simplify the implementation accordingly.

baszalmstra:
Im clearly talking to an AI. Closing this PR. Please refer to our [AI_POLICY](https://github.com/conda/rattler/blob/main/AI_POLICY.md)

I find this behavior utterly disrespectful.

vishwas-droid:
That wasn’t my intention at all. I apologize if my replies came across that way.

PR #2081, opened 2026-02-19
fix(rattler_shell): force UTF-8 encoding for subprocess output on Windows (#1822)

### LINKED ISSUE #1822: Rattler Windows Tests are failing with charmap error
### Checklist

### What happened?

Recently my rattler tests for windows platform started failing with this error.  My test simply calls `app --help` which prints some utf-8 colored output.

```
'charmap' codec can't encode characters in position 405-406: character maps to <undefined>
```

I've checked my code page which currently is utf-8 and cannot reproduce the error outside of rattler.  What else can I try to debug this error?

```
❯ chcp
Active code page: 65001
```

### Additional Context

_No response_

DCchoudhury15:
### Description
Fixes #1822

On Windows, subprocess output defaults to the system locale encoding
(`charmap`/CP1252) even when `chcp` shows code page 65001 (UTF-8).
This causes a crash when the output contains non-ASCII or colored
(ANSI) characters.

Two fixes in `rattler_shell`:
- `shell/mod.rs`: Replace `String::from_utf8()` with `String::from_utf8_lossy()` 
  for cygpath output — handles non-UTF8 bytes gracefully instead of erroring
- `activation.rs`: Inject `PYTHONUTF8=1` and `PYTHONIOENCODING=utf-8` into 
  the activation subprocess environment on Windows only 
  (`#[cfg(target_os = "windows")]`), so Python child processes always 
  output UTF-8 bytes

### How Has This Been Tested?
- `cargo clippy -p rattler_shell -- -D warnings` — no warnings
- `cargo test -p rattler_shell` — 24 passed, 2 ignored
- `cargo fmt --check` — clean

### AI Disclosure
- [] This PR contains AI-generated content.
  - [] I have tested any AI-generated content in my PR.
  - [] I take responsibility for any AI-generated content in my PR.

### Checklist:

### DIFF
--- a/crates/rattler_shell/src/activation.rs
+++ b/crates/rattler_shell/src/activation.rs
@@ -655,6 +655,16 @@ impl<T: Shell + Clone> Activator<T> {
             activation_command.env_clear().envs(environment);
         }
 
+        // On Windows, force the subprocess to use UTF-8 encoding for its output.
+        // Without this, Python subprocesses may use the system 'charmap' codec
+        // (e.g. CP1252) even when the active code page is 65001, causing
+        // UnicodeEncodeError when the output contains non-ASCII characters.
+        #[cfg(target_os = "windows")]
+        {
+            activation_command.env("PYTHONUTF8", "1");
+            activation_command.env("PYTHONIOENCODING", "utf-8");
+        }
+
         let activation_result = activation_command.output()?;
 
         if !activation_result.status.success() {
--- a/crates/rattler_shell/src/shell/mod.rs
+++ b/crates/rattler_shell/src/shell/mod.rs
@@ -211,10 +211,12 @@ pub(crate) fn native_path_to_unix(path: &str) -> Result<String, std::io::Error> 
         .output();
 
     match output {
-        Ok(output) if output.status.success() => Ok(String::from_utf8(output.stdout)
-            .map_err(|_err| std::io::Error::other("failed to convert path to Unix style"))?
-            .trim()
-            .to_string()),
+        Ok(output) if output.status.success() => {
+            // Use from_utf8_lossy to handle non-UTF8 output on Windows
+            // where system encoding may not be UTF-8 even if code page is 65001
+            let path = String::from_utf8_lossy(&output.stdout).trim().to_string();
+            Ok(path)
+        }
         Err(e) if e.kind() == std::io::ErrorKind::NotFound => Err(e),
         Err(e) => Err(std::io::Error::other(format!(
             "failed to convert path to Unix style: {e}"


DCchoudhury15:
@baszalmstra , This PR is my attempt to address the Windows charmap encoding failure reported in #1822. It seems that on Windows, Python subprocesses can still fall back to a legacy encoding like CP1252, even when the active code page is 65001, which leads to UnicodeEncodeError when non-ASCII output (such as colored --help text) is printed. To reduce the chances of this happening, I’ve set PYTHONUTF8=1 and PYTHONIOENCODING=utf-8 for Python subprocesses on Windows. I also switched stdout decoding to String::from_utf8_lossy so we don’t fail hard if non-UTF8 bytes show up. The change is limited to Windows and shouldn’t affect other platforms. That said, I want to be clear that this is more of a practical workaround than a full fix. It helps in the common Python subprocess case but doesn’t solve deeper Windows console or code page inconsistencies. It also won’t change behavior for non-Python executables. If you’d prefer a more robust or lower-level encoding approach, I’m happy to continue iterating on this.

wolfv:
Thank you @DCchoudhury15 - would you be able to double check with `conda` activation logic whether they do something comparable?

Hypothetically, this is also something that could be done at the package level (ie. Python coming from conda-forge could opt to do this by default). I am not sure if this is the right "layer" to address this, although I totally sympathize with the idea.

DCchoudhury15:
@wolfv sure i'll get back to you on this asap.

wolfv:
I started a discussion here: [#general > Python force UTF8 in activation](https://conda-forge.zulipchat.com/#narrow/channel/457337-general/topic/Python.20force.20UTF8.20in.20activation/with/574896192)

DCchoudhury15:
@wolfv I looked at  PEP 686 and it actually reinforces this approach. 
The PEP explicitly states that this change "mostly affects Windows users" 
and that `PYTHONUTF8=1` is the intended mechanism — it even rejected using 
`PYTHONIOENCODING` for pipes as a solution. Since Python 3.15 will make 
UTF-8 mode the default anyway, setting `PYTHONUTF8=1` on Windows in the 
activation environment feels less like a workaround and more like getting 
ahead of where Python is already heading. Ruby did the same in 3.0 and 
Java in JDK 18, so there's solid precedent here. The scope is also pretty 
contained — it's Windows-only and only affects the activation subprocess, 
so it shouldn't surprise anyone. That said, if you'd rather this live inside 
the generated activation script instead of at the Rust level, I'm happy to 
move it there — just let me know what direction you prefer and I'll update 
the PR.

baszalmstra:
I don't appreciate the clear AI output in your last comment. We are humans, speak to us as a human. Did you read the zulip thread linked by wolfv?

I think having a special package makes sense?

DCchoudhury15:
I checked conda's activation code and it doesn't set PYTHONUTF8 or PYTHONIOENCODING anywhere
The Zulip thread ended with Nicholas suggesting a python-utf8-activation package but no final decision was made
That package doesn't exist on conda-forge yet
The PYTHONIOENCODING rejected alternative in PEP 686 is about Python's internal subprocess behavior, not about setting it manually as an env var
Based on all this, I think the cleaner move is to drop the env var changes from this PR and keep just the from_utf8_lossy fix, since that's small, safe, and consistent with how rattler already handles errors elsewhere in the same file.
If the env var approach is worth pursuing it probably makes more sense as a separate package like Nicholas suggested. Happy to close this PR if you'd prefer that direction instead.
Also i used ai to enhance my previous comment and i did not double check it which was a mishap and negligence from my side , i value the maintainers time and wanted to state that it wasnt my intention and wouldnt be repeated again.

Issue #742, opened 2024-06-10
Get `sha256`, `size` and `md5` from direct url package.


ruben-arts:
The current implementation for direct urls (#725) downloads the url to cache and then gets the package record from the `index.json`. This puts the `sha256`, `md5` and `size` to `None` as we can't currently get this info from the unpacked file in cache. 

This should be fixed as these values are part of the matchspec and do exist.

pro-utkarshM:
Direct URL packages (introduced in #725) download archives and read metadata from `index.json`, but `sha256`, `md5`, and `size` are set to `None` in the resulting `PackageRecord`.

This causes lockfile integrity checking to fail, since the lockfile has no hashes or size to verify against.

The extraction process already computes these values via `ExtractResult`, but they are discarded inside the cache layer and never propagated to `PackageRecord`.

Expected Behavior:

`PackageRecord` for direct URL packages should have `sha256`, `md5`, and `size` populated:

* For HTTP URLs: sourced from the download/extraction (`ExtractResult`)
* For local file URLs: computed directly from the file on disk

This would allow lockfiles to include proper integrity metadata and restore verification guarantees.

PR #2085, opened 2026-02-20
feat: populate sha256, md5, and size for direct URL packages

pro-utkarshM:
## Description

Direct URL–based packages were missing integrity metadata (`sha256`, `md5`, and `size`) in `PackageRecord`, which caused lockfile verification to fail.

This PR fixes that by:

* Capturing the `ExtractResult` during HTTP download/extraction and propagating the computed hash and size into `CacheMetadata`.
* Computing `sha256` and `md5` for local file URLs in a single pass using nested `HashingReader`s and reading file size from filesystem metadata.
* Wiring these computed values into `PackageRecord::from_index_json()`, preferring computed values and falling back to `MatchSpec` where necessary.

This restores proper lockfile integrity checking for both HTTP and local-file direct URLs.

Fixes [https://github.com/conda/rattler/issues/742](https://github.com/conda/rattler/issues/742)

---

## How Has This Been Tested?

* ✅ All 21 `rattler_cache` tests pass
* ✅ All 5 `rattler_repodata_gateway` direct_url tests pass (including new `sha256` / `md5` / `size` assertions)
* ✅ `cargo clippy` clean on both crates
* ✅ `cargo fmt --check` clean

---

## AI Disclosure

* [x] This PR contains AI-generated content (PR description only).

  * [x] I have tested any AI-generated content in my PR.
  * [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude Code

---

## Checklist

* [x] I have performed a self-review of my own code
* [x] I have commented my code, particularly in hard-to-understand areas
* [ ] I have made corresponding changes to the documentation (not required for this change)
* [x] I have added sufficient tests to cover my changes

---

### DIFF
--- a/crates/rattler_cache/src/package_cache/mod.rs
+++ b/crates/rattler_cache/src/package_cache/mod.rs
@@ -177,6 +177,7 @@ impl PackageCacheLayer {
             fn(PathBuf) -> _,
             Pin<Box<dyn Future<Output = Result<(), _>> + Send>>,
             std::io::Error,
+            (),
         >(
             cache_path,
             cache_entry.last_revision,
@@ -187,7 +188,7 @@ impl PackageCacheLayer {
         )
         .await
         {
-            Ok(cache_metadata) => {
+            Ok((cache_metadata, _)) => {
                 cache_entry.last_revision = Some(cache_metadata.revision);
                 cache_entry.last_sha256 = cache_metadata.sha256;
                 Ok(cache_metadata)
@@ -197,15 +198,15 @@ impl PackageCacheLayer {
     }
 
     /// Validate the package, and fetch it if invalid.
-    pub async fn validate_or_fetch<F, Fut, E>(
+    pub async fn validate_or_fetch<F, Fut, E, T>(
         &self,
         fetch: F,
         cache_key: &CacheKey,
         reporter: Option<Arc<dyn CacheReporter>>,
-    ) -> Result<CacheMetadata, PackageCacheLayerError>
+    ) -> Result<(CacheMetadata, Option<T>), PackageCacheLayerError>
     where
         F: (Fn(PathBuf) -> Fut) + Send + 'static,
-        Fut: Future<Output = Result<(), E>> + Send + 'static,
+        Fut: Future<Output = Result<T, E>> + Send + 'static,
         E: std::error::Error + Send + Sync + 'static,
     {
         let entry = self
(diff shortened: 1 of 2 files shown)


crates/rattler_repodata_gateway/src/gateway/direct_url_query.rs
```diff
@@ -57,30 +57,54 @@ impl DirectUrlQuery {
...
+                };
+
+            // Compute sha256, md5, and file size in a single pass
+            let file = std::fs::File::open(&file_path)?;
+            let file_size = file.metadata()?.len();
+            let sha256_reader = HashingReader::<_, Sha256>::new(file);
+            let mut md5_reader = HashingReader::<_, Md5>::new(sha256_reader);
+            std::io::copy(&mut md5_reader, &mut std::io::sink())?;
+            let (sha256_reader, md5_hash) = md5_reader.finalize();
+            let (_, sha256_hash) = sha256_reader.finalize();
```
baszalmstra: This is a huge waste of time for very little gain. Lets just not do it.

crates/rattler_cache/src/package_cache/mod.rs
```diff
@@ -475,12 +475,19 @@ impl PackageCache {
         let sha256 = cache_key.sha256();
         let md5 = cache_key.md5();
         let download_reporter = reporter.clone();
+
+        // Capture the extract result so we can populate md5/size on CacheMetadata
+        let extract_result_capture: Arc<Mutex<Option<rattler_package_streaming::ExtractResult>>> =
```
baszalmstra: This seems overly complex, cant we just return it?

crates/rattler_repodata_gateway/src/gateway/direct_url_query.rs
pro-utkarshM: Sure, removed the local file re-hashing and restored the original match logic. Local direct URLs now return `None` for sha256/md5/size again, and unused hashing imports are gone. Tests updated accordingly. Thanks, resolved.

crates/rattler_cache/src/package_cache/mod.rs
pro-utkarshM: Refactored this to avoid the `Arc<Mutex<Option<ExtractResult>>>` entirely. The fetch path now returns `ExtractResult` directly via a generic helper, while keeping the public API unchanged. Much simpler flow now, resolved.

pro-utkarshM:
I’ve pushed changes for this, might need a refresh. Let me know what you think.

PR #2087, opened 2026-02-20
Fix/hash mismatch cleanup panic

suhr25:
## Summary

This PR addresses two independent issues in the installer and networking layers that impact stability and performance under real-world usage.

````md id="pr_final_clean_51x2"
# Fix installer panic on cleanup + reuse HTTP client in OCI middleware

### Installer cleanup panic  
**File:** `crates/rattler/src/install/installer/mod.rs`

The `execute_transaction` function is the central entry point for install, update, and removal operations. After completing unlink operations, it performs a cleanup step:

```rust
driver.remove_empty_directories(...).unwrap();
````

This function walks the filesystem and removes directories that are left empty after package removal. Since it performs real OS-level operations, failures such as permission issues or locked directories are entirely possible.

Using `.unwrap()` here causes the process to **panic on failure**, abruptly terminating the installer instead of returning a structured error. This makes the installer fragile in scenarios where filesystem conditions are not ideal.

---

### Inefficient HTTP client usage in OCI middleware

**File:** `crates/rattler_networking/src/oci_middleware.rs`

`OciMiddleware` handles `oci://` requests and performs two auxiliary network calls per request:

* fetching a bearer token from the registry
* fetching the image manifest (when no digest is provided)

Previously, both operations created a new `reqwest::Client`:

```rust
reqwest::get(token_url).await?;
reqwest::Client::new().get(...).send().await?;
```

Each client maintains its own connection pool. Creating a new client per request prevents connection reuse and leads to unnecessary resource usage. Under concurrent OCI downloads, this results in multiple connection pools being created and discarded, degrading performance.

---

### Fix 1 — Replace panic with proper error propagation

`remove_empty_directories` already returns `Result<(), UnlinkError>`, and `InstallerError` already supports wrapping this error. The fix replaces `.unwrap()` with proper propagation:

```rust
driver
    .remove_empty_directories(&transaction.operations, transaction.unchanged_packages(), &prefix)
    .map_err(|e| InstallerError::UnlinkError("remove_empty_directories".to_string(), e))?;
```

This ensures:

* filesystem errors are returned to the caller instead of causing a panic
* behavior is consistent with the rest of `execute_transaction`

---

### Fix 2 — Reuse a single reqwest client

`OciMiddleware` is updated to hold a reusable `reqwest::Client` instance:

```rust
pub struct OciMiddleware {
    client: reqwest::Client,
}
```

Initialized once:

```rust
impl Default for OciMiddleware {
    fn default() -> Self {
        Self { client: reqwest::Client::new() }
    }
}
```

Both token and manifest requests now reuse this client:

```rust
let response = client.get(token_url).send().await?;

let manifest = client
    .get(manifest_url)
    .bearer_auth(&token)
    .send()
    .await?;
```

This allows connection pooling to work as intended and avoids repeated client construction.

---

### Installer fix

* Simulated filesystem failures (e.g., locked directories, permission errors)

* Verified:

  * no panic occurs
  * errors are returned as `InstallerError::UnlinkError`
  * successful execution remains unchanged

* Confirmed no new error types were introduced; existing error handling is reused

* Behavior is now consistent with other error paths in `execute_transaction`

---

### OCI middleware fix

* Confirmed a single `reqwest::Client` is created per middleware instance

* Verified connection reuse across multiple requests

* Observed no changes in request/response behavior

* Confirmed `Clone` remains correct since `reqwest::Client` is internally `Arc`-backed

* Updated tests to construct middleware via `OciMiddleware::default()`

---

## Impact

* Prevents unexpected crashes caused by normal filesystem failures
* Improves network efficiency by enabling connection reuse
* Keeps behavior consistent and predictable across both layers

Overall, this improves robustness in the installer and performance in OCI networking without changing external functionality.

### DIFF
--- a/crates/rattler-bin/src/commands/create.rs
+++ b/crates/rattler-bin/src/commands/create.rs
@@ -176,7 +176,7 @@ pub async fn create(opt: Opt) -> miette::Result<()> {
         .with_arc(Arc::new(
             AuthenticationMiddleware::from_env_and_defaults().into_diagnostic()?,
         ))
-        .with(rattler_networking::OciMiddleware);
+        .with(rattler_networking::OciMiddleware::default());
     #[cfg(feature = "s3")]
     let download_client = download_client.with(rattler_networking::S3Middleware::new(
         HashMap::new(),
--- a/crates/rattler-bin/src/commands/extract.rs
+++ b/crates/rattler-bin/src/commands/extract.rs
@@ -43,7 +43,7 @@ fn create_authenticated_client() -> miette::Result<reqwest_middleware::ClientWit
         .with_arc(Arc::new(AuthenticationMiddleware::from_auth_storage(
             authentication_storage.clone(),
         )))
-        .with(rattler_networking::OciMiddleware);
+        .with(rattler_networking::OciMiddleware::default());
     #[cfg(feature = "s3")]
     let client = client.with(rattler_networking::S3Middleware::new(
         std::collections::HashMap::new(),
--- a/crates/rattler-bin/src/commands/search.rs
+++ b/crates/rattler-bin/src/commands/search.rs
@@ -88,7 +88,7 @@ pub async fn search(opt: Opt) -> miette::Result<()> {
         .with_arc(Arc::new(
             AuthenticationMiddleware::from_env_and_defaults().into_diagnostic()?,
         ))
-        .with(rattler_networking::OciMiddleware);
+        .with(rattler_networking::OciMiddleware::default());
     #[cfg(feature = "s3")]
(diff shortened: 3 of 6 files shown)


baszalmstra:
Please create two seperate PRs.

suhr25:
Thanks! I’ll split this into two separate PRs (installer cleanup error propagation + OCI middleware client reuse) and update this PR accordingly

suhr25:
Thanks for the feedback! I’ve split the changes into two separate PRs to keep concerns isolated:

1) installer cleanup error handling

2) OCI middleware client reuse

Closing this PR and continuing the discussion on those.

PR #2089, opened 2026-02-20
fix(networking): reuse reqwest client in OCI middleware

suhr25:
## Summary

This PR improves the efficiency of OCI registry requests by reusing a single `reqwest::Client` instance within `OciMiddleware`.

Previously, each OCI request created a new HTTP client:
- token fetch used `reqwest::get()` (internally creates a client)
- manifest fetch used `reqwest::Client::new()`

Since each `reqwest::Client` maintains its own connection pool, this resulted in a new pool being allocated for every request. Under concurrent downloads, this led to unnecessary resource usage, lack of connection reuse, and degraded performance.

---

## Description

`OciMiddleware` is responsible for handling `oci://` scheme requests, including:
- fetching bearer tokens from registries  
- retrieving image manifests  

This PR introduces a persistent `reqwest::Client` stored within `OciMiddleware`, which is:
- initialized once via `Default`
- reused across all token and manifest requests  

### Before

```rust
reqwest::get(token_url).await?;

reqwest::Client::new()
    .get(manifest_url)
    .send()
    .await?;
````

### After

```rust
pub struct OciMiddleware {
    client: reqwest::Client,
}

impl Default for OciMiddleware {
    fn default() -> Self {
        Self { client: reqwest::Client::new() }
    }
}
```

```rust
client.get(token_url).send().await?;

client
    .get(manifest_url)
    .bearer_auth(&token)
    .send()
    .await?;
```

This ensures connection pooling is effective and avoids repeated client construction.

Fixes #{issue}

---

## How Has This Been Tested?

* Verified existing OCI-related functionality (token fetch + manifest retrieval) continues to work as expected
* Confirmed no change in request/response behavior
* Ensured compatibility with existing middleware usage by updating construction to `OciMiddleware::default()` where required
* Ran local builds and tests using the project’s development setup:

```sh
pixi install
pixi run build
```

Additionally:

* Observed that repeated requests reuse connections (no per-request client creation)
* Confirmed that `reqwest::Client` cloning remains safe due to its internal `Arc` usage

---

## Checklist

* [x] I have performed a self-review of my own code
* [x] I have commented my code where necessary

```
```

### DIFF
--- a/crates/rattler_networking/src/oci_middleware.rs
+++ b/crates/rattler_networking/src/oci_middleware.rs
@@ -13,12 +13,15 @@ use reqwest_middleware::{Middleware, Next};
 use serde::Deserialize;
 use url::{ParseError, Url};
 
-use crate::mirror_middleware::create_404_response;
+use crate::{mirror_middleware::create_404_response, LazyClient};
 
 #[derive(thiserror::Error, Debug)]
 enum OciMiddlewareError {
     #[error("Reqwest error: {0}")]
     Reqwest(#[from] reqwest::Error),
+
+    #[error("Reqwest middleware error: {0}")]
+    ReqwestMiddleware(#[from] reqwest_middleware::Error),
 
     #[error("URL parse error: {0}")]
     ParseError(#[from] ParseError),
@@ -35,13 +38,15 @@ enum OciMiddlewareError {
 pub struct OciMiddleware {
     /// Shared HTTP client reused across all OCI requests to avoid creating a
     /// new connection pool on every token fetch or manifest pull.
-    client: reqwest::Client,
+    client: LazyClient,
 }
 
 impl OciMiddleware {
     /// Create a new [`OciMiddleware`] reusing the provided HTTP client.
-    pub fn new(client: reqwest::Client) -> Self {
-        Self { client }
+    pub fn new(client: impl Into<LazyClient>) -> Self {
+        Self {
+            client: client.into(),
+        }
     }
 }
 
@@ -72,13 +77,13 @@ impl Display for OciAction {
 
 // [oci://ghcr.io/channel-mirrors/conda-forge]/[osx-arm64/xtensor]
 async fn get_token(
-    client: &reqwest::Client,
+    client: &LazyClient,
     url: &OCIUrl,
     action: OciAction,
 ) -> Result<String, OciMiddlewareError> {
     let token_url = url.token_url(action)?;
 
-    let response = client.get(token_url.clone()).send().await?;
+    let response = client.client().get(token_url.clone()).send().await?;
 
     match response.error_for_status() {
         Ok(response) => {
@@ -189,7 +194,7 @@ impl OCIUrl {
     }
 
     pub async fn get_blob_url(
-        client: &reqwest::Client,
+        client: &LazyClient,
         req: &mut Request,
     ) -> Result<(), OciMiddlewareError> {
         let oci_url = OCIUrl::new(req.url())?;
@@ -213,6 +218,7 @@ impl OCIUrl {
             let manifest_url = oci_url.manifest_url()?; // TODO: handle error
 
             let manifest = client
+                .client()
                 .get(manifest_url)
                 .bearer_auth(&token)
                 .header(ACCEPT, "application/vnd.oci.image.manifest.v1+json")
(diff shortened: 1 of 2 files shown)


wolfv (approved):

crates/rattler_networking/src/oci_middleware.rs
```diff
@@ -31,8 +31,20 @@ enum OciMiddlewareError {
 }
 
 /// Middleware to handle `oci://` URLs
-#[derive(Default, Debug, Clone)]
-pub struct OciMiddleware;
+#[derive(Debug, Clone)]
+pub struct OciMiddleware {
+    /// Shared HTTP client reused across all OCI requests to avoid creating a
+    /// new connection pool on every token fetch or manifest pull.
+    client: reqwest::Client,
```
wolfv: Maybe we can put it in a OnceLock so we only construct it when it's actually used?

baszalmstra: You can use `LazyClient` for that.

baszalmstra:
There might actually be more middlewares that use a client! From the top of my head the authentication middleware also requires one.

crates/rattler_networking/src/oci_middleware.rs
suhr25: Good point lazy initialization makes sense here. I’ll refactor this to use LazyClient so the client is only constructed when first needed.Thanks!

crates/rattler-bin/src/commands/extract.rs
```diff
@@ -43,7 +43,7 @@ fn create_authenticated_client() -> miette::Result<reqwest_middleware::ClientWit
         .with_arc(Arc::new(AuthenticationMiddleware::from_auth_storage(
             authentication_storage.clone(),
         )))
-        .with(rattler_networking::OciMiddleware);
+        .with(rattler_networking::OciMiddleware::default());
```
baszalmstra: A client is constructed a few lines above. Lets reuse that!

crates/rattler-bin/src/commands/search.rs
```diff
@@ -88,7 +88,7 @@ pub async fn search(opt: Opt) -> miette::Result<()> {
         .with_arc(Arc::new(
             AuthenticationMiddleware::from_env_and_defaults().into_diagnostic()?,
         ))
-        .with(rattler_networking::OciMiddleware);
+        .with(rattler_networking::OciMiddleware::default());
```
baszalmstra: Same

crates/rattler_networking/src/oci_middleware.rs
```diff
@@ -31,8 +31,20 @@ enum OciMiddlewareError {
...
+}
+
+impl Default for OciMiddleware {
+    fn default() -> Self {
+        Self {
+            client: reqwest::Client::new(),
```
baszalmstra: Does client also inplement default?

suhr25:
Hi @baszalmstra  @wolfv ,
 thanks for the suggestion  I have updated the implementation to reuse the existing reqwest::Client and removed the Default construction. Let me know if anything else needs adjustment.

crates/rattler_networking/src/oci_middleware.rs
suhr25: Hi, yes ,
 reqwest::Client does implement Default. However, I’ve removed the Default implementation from OciMiddleware and switched to injecting the client instead, so we consistently reuse the existing instance rather than constructing a new one.

crates/rattler-bin/src/commands/extract.rs
suhr25: Thanks @baszalmstra , I’ve updated this to reuse the existing client from above and removed the Default usage here.

crates/rattler-bin/src/commands/search.rs
suhr25: Updated this as well to reuse the existing client and removed the Default usage.

crates/rattler_networking/src/oci_middleware.rs
baszalmstra: Can you take a look at this comment still?

suhr25: Hi @baszalmstra 
Thanks for the reminder. I’ve updated this to use LazyClient so the underlying reqwest::Client is initialized lazily and reused across requests. Let me know if this aligns with what you had in mind.
Thanks!

baszalmstra (approved):

PR #2090, opened 2026-02-20
feat: add dependency override option to solver

### LINKED ISSUE #586: Add extra solve options
`uv` has some nice solver options:

- `--exclude-newer` (compare timestamps and exclude packages that were published after some date)
- `--override` to override some dependencies (basically a way of locally "repodata patching")

Premkumar-2004:
### Description

Added dependency_overrides field to SolverTask that lets you replace dependencies of specific packages during solve.
For example, to force jupyterlab 3.3.* to use python ==3.10:
`dependency_overrides: vec![("jupyterlab 3.3.*", "python ==3.10")]`
Fixes #586

### How Has This Been Tested?

Added 3 tests covering basic override, no-match case, and multiple overrides. All tests pass.

### AI Disclosure

Tools:Claude

### Checklist:

### DIFF
--- a/crates/rattler_solve/src/resolvo/mod.rs
+++ b/crates/rattler_solve/src/resolvo/mod.rs
@@ -35,6 +35,15 @@ mod conda_sorting;
 
 type MatchSpecParseCache = HashMap<String, (Vec<VersionSetId>, Option<ConditionId>)>;
 
+/// A dependency override rule.
+#[derive(Clone)]
+pub struct DependencyOverride {
+    /// Matches packages whose dependencies should be overridden.
+    pub package_matcher: MatchSpec,
+    /// Replaces matching dependencies.
+    pub override_spec: MatchSpec,
+}
+
 /// Represents the information required to load available packages into libsolv
 /// for a single channel and platform combination
 #[derive(Clone)]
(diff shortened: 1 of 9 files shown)


crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -573,6 +586,29 @@ impl<'a> CondaDependencyProvider<'a> {
                 .intern_condition(Condition::Requirement(version_set))
         })
     }
+
+    fn apply_dependency_override(&self, record: &RepoDataRecord, depends: &str) -> String {
+        let Ok(dep_spec) = MatchSpec::from_str(depends, ParseMatchSpecOptions::lenient()) else {
```
baszalmstra: This parses the matchspec twice. In this loop and at the call site. 

Maybe we can also preprocess the overrides to create a hashmap by name to skip iterating over each individual entry?

baszalmstra (changes requested):

crates/rattler_solve/src/resolvo/mod.rs
Premkumar-2004: fixed, using a hashmap now and parsing only once.

PR #2096, opened 2026-02-20
fix: Set the modification time of any file we change

hunger:
### Description

This fixes pyc files not getting updated as we patch the py-file.

I choose a modification time which is bigger than the modification time, but that is the same for all files we change to keep the file tree more reproducible.

Fixes prefix-dev/rattler-build#2147

### How Has This Been Tested?

by trying it out :-)

### AI Disclosure

- [x] AI generated contents: Claude helped to write the tests.

AI Tool: Opus 4.6

### Checklist:

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -140,6 +140,11 @@ pub struct LinkedFile {
 ///
 /// Note that usually the `target_prefix` is equal to `target_dir` but it might differ. See
 /// [`crate::install::InstallOptions::target_prefix`] for more information.
+///
+/// The `modification_time` is a timestamp we set on all files we modify. We want a value
+/// we control here to make the generated filesystem tree more reproducible. `modification_time`
+/// should be greater than any of the modification times of any of the files that were packaged
+/// up (ignoring any data conda stores).
 #[allow(clippy::too_many_arguments)] // TODO: Fix this properly
 pub fn link_file(
     path_json_entry: &PathsEntry,
@@ -152,6 +157,7 @@ pub fn link_file(
     allow_ref_links: bool,
     target_platform: Platform,
     apple_codesign_behavior: AppleCodeSignBehavior,
+    modification_time: filetime::FileTime,
 ) -> Result<LinkedFile, LinkFileError> {
     let source_path = package_dir.join(&path_json_entry.relative_path);
 
@@ -268,8 +274,7 @@ pub fn link_file(
         // Copy file permissions and timestamps
         fs::set_permissions(&destination_path, metadata.permissions())
             .map_err(LinkFileError::FailedToUpdateDestinationFilePermissions)?;
-        let file_time = filetime::FileTime::from_last_modification_time(&metadata);
-        filetime::set_file_times(&destination_path, file_time, file_time)
+        filetime::set_file_times(&destination_path, modification_time, modification_time)
             .map_err(LinkFileError::FailedToUpdateDestinationFileTimestamps)?;
 
         LinkMethod::Patched(*file_mode)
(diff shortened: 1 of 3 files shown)


crates/rattler/src/install/link.rs
```diff
@@ -268,8 +274,11 @@ pub fn link_file(
         // Copy file permissions and timestamps
         fs::set_permissions(&destination_path, metadata.permissions())
             .map_err(LinkFileError::FailedToUpdateDestinationFilePermissions)?;
-        let file_time = filetime::FileTime::from_last_modification_time(&metadata);
-        filetime::set_file_times(&destination_path, file_time, file_time)
+        eprintln!(">>> Using modification time for {destination_path:?}");
+        if destination_path.extension() == Some(std::ffi::OsString::from(".py").as_os_str()) {
+            eprintln!("PYTHON *** PYTHON *** PYTHON *** PYTHON *** PYTHON *** PYTHON *** PYTHON *** PYTHON *** ");
+        }
```
baszalmstra:
```suggestion
```

hunger:
New and improved: No more eprintln, Metadata is spelled correctly and I added tests to make sure the mtime does change on files we modify and does not change for files we leave alone.

wolfv:
yeah works for me!

hunger:
... and rustfmt the entire thing :-(

crates/rattler/src/install/mod.rs
```diff
@@ -259,6 +263,24 @@ struct LinkPath {
...
 
+/// Find a timestamp to put on all files we modify. Use `info/about.json`, as a base. This
+/// file is always written after all the packaged data is already stored.
+///
+/// We add 1s to make sure our modification time is strictly bigger than any timestamp
+/// seen in the package data, even on systems that do not have nanosecond timestamps.
+fn modification_time(package_dir: &Path) -> Result<filetime::FileTime, InstallError> {
+    let about_info_path = package_dir.join("info/about.json");
+
+    let info_metadata = fs::symlink_metadata(&about_info_path)
```
wolfv: This should use `fs_err`.

crates/rattler/src/install/mod.rs
```diff
@@ -83,6 +83,10 @@ pub enum InstallError {
     #[error("the operation was cancelled")]
     Cancelled,
 
+    /// The paths.json file could not be read.
```
wolfv: The comment seems wrong

crates/rattler/src/install/mod.rs
```diff
@@ -259,6 +263,24 @@ struct LinkPath {
     clobber_path: Option<PathBuf>,
 }
 
+/// Find a timestamp to put on all files we modify. Use `info/about.json`, as a base. This
+/// file is always written after all the packaged data is already stored.
+///
+/// We add 1s to make sure our modification time is strictly bigger than any timestamp
+/// seen in the package data, even on systems that do not have nanosecond timestamps.
+fn modification_time(package_dir: &Path) -> Result<filetime::FileTime, InstallError> {
+    let about_info_path = package_dir.join("info/about.json");
```
wolfv: I believe we have a helper for this in `rattler_conda_types` (to get this path). Might be useful to use that?

crates/rattler/src/install/mod.rs
```diff
@@ -619,6 +644,7 @@ pub fn link_package_sync(
         },
         Ok,
     )?;
+    let modification_time = modification_time(package_dir)?;
```
wolfv: I think we could fall back to `now()` if we cannot get the file time (for whatever reason) from the package dir.

hunger: Then I can also get rid of the error :-)

hunger:
@wolfv Thanks for catching all this! It's fixed now.

PR #2097, opened 2026-02-20
test(rattler_lock): track lock-file sizes with deterministic checks

### LINKED ISSUE #765: Add benchmark for pixi lock file size

devvaansh:
### Description

introduces deterministic size checks (rather than fluctuating benchmarks) for lock files to ensure stability and minimize git churn.

Key Changes:
Implements size tracking using insta snapshots and standard #[test] functions.
Monitors both raw rendered size and zstd-compressed size (at level 3).
Includes a git-history test that commits fixtures to a temporary repository. This asserts that the canonical serializer produces a zero-delta for reordered input, directly verifying that lock-file churn remains minimal across commits.

Fixes #765 

## How Has This Been Tested?

I have verified the changes using the following test suite in the rattler_lock package:
Deterministic Size Checks
Git History Table Visualization
These tests verify the metrics requested in #765: raw size, compressed size, and git repository growth stability.
### AI Disclosure

Tools: Claude
### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5081,9 +5081,11 @@ dependencies = [
  "serde_with",
  "serde_yaml",
  "similar-asserts",
+ "tempfile",
  "thiserror 2.0.18",
  "typed-path",
  "url",
+ "zstd",
 ]
 
 [[package]]
--- a/crates/rattler_lock/Cargo.toml
+++ b/crates/rattler_lock/Cargo.toml
@@ -32,6 +32,8 @@ typed-path = { workspace = true }
 
 [dev-dependencies]
 insta = { workspace = true, features = ["yaml"] }
+rstest = { workspace = true }
 serde_json = { workspace = true }
 similar-asserts = { workspace = true }
-rstest = { workspace = true }
+tempfile = { workspace = true }
+zstd = { workspace = true }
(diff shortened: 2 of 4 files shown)


hunger:
I see that #765 requested something like this, but what is the value doing it in a test? Everybody that needs to change anything in the lock file will now have to change this test. You can not get real insight into how these values develop over time. We would need some kind of dashboard for that, showing the data in relation to historical data. Unit tests are the wrong way for this IMHO.

The git test measures the wrong thing: Switch to an entirely different project configuration with different dependencies are one-off events that happen rarely. The interesting thing is the churn we have in daily work, e.g. because two devs lock files disagree on some fact for some reason and their commits then keep switching the lock file over from the "wrong" facts to their own "right" facts.

devvaansh:
Thanks for the detailed feedback, @hunger. You make a very fair point regarding the maintenance overhead—I definitely don’t want to add friction for every developer making a routine lock-file update.   I'm happy to refactor this to better align with the project's long-term needs.

PR #2099, opened 2026-02-21
fix: improve error handling in package indexing

wolfv:
### Description

This PR improves error handling in the package indexing pipeline by replacing panics and hard failures with graceful degradation:

1. **Better error messages**: Convert `.expect()` and `.unwrap()` calls to proper error handling with descriptive messages:
   - `seek::stream_conda_info()` errors now include the underlying error details
   - Unknown archive type errors now include the filename for debugging

2. **Resilient indexing**: Change the indexing behavior to skip invalid packages instead of failing the entire operation:
   - Invalid packages now log a warning and are skipped rather than aborting the entire index
   - Task panics are caught and logged as warnings instead of crashing the indexer
   - Unsupported archive types are skipped with a warning instead of panicking

This allows the indexer to be more robust when encountering malformed or unsupported packages, completing the index with valid packages while reporting issues for investigation.

### How Has This Been Tested?

The changes maintain existing test coverage. Error handling paths are exercised through:
- Existing unit tests for `package_record_from_conda_reader()` and `parse_package_buffer()`
- Integration tests for the indexing pipeline that may encounter edge cases

### Checklist:
- [x] Error handling improvements are straightforward and follow Rust best practices

https://claude.ai/code/session_01Y7k5WaeChR3djHtgze4TPg


crates/rattler_index/src/lib.rs
```diff
@@ -794,7 +789,13 @@ async fn index_subdir_inner(
...
                 conda_packages.insert(filename, package);
             }
-            _ => panic!("Unknown archive type"),
+            other => {
+                tracing::warn!(
+                    "Skipping package with unsupported archive type {:?}: {}",
+                    other,
+                    filename.to_file_name()
+                );
+            }
```
pavelzw: when would this happen? shouldn't `filename.archive_type` be always correct?

crates/rattler_index/src/lib.rs
```diff
@@ -745,24 +748,16 @@ async fn index_subdir_inner(
...
-                tasks.clear();
-                tracing::error!("Failed to process package: {}", e);
-                pb.abandon_with_message(format!(
-                    "{} {}",
-                    console::style("Failed to index").red(),
-                    console::style(subdir.as_str()).dim()
-                ));
-                return Err(RepodataError::Other(anyhow::anyhow!(e)));
+                tracing::warn!("Skipping invalid package in {}: {}", subdir, e);
+                pb.inc(1);
```
pavelzw: should we still count the errors for this subdir and print a summary for the failures later on? I think this is what we do in `conda-mirror`

crates/rattler_index/src/lib.rs
wolfv: true

crates/rattler_index/src/lib.rs
```diff
@@ -75,13 +75,24 @@ impl PreconditionChecks {
     }
 }
 
+/// A package that was skipped during indexing due to an error.
+#[derive(Debug, Clone)]
+pub struct SkippedPackage {
+    /// The filename of the skipped package
+    pub filename: String,
+    /// The error message describing why the package was skipped
+    pub error: String,
```
baszalmstra: Maybe we can use a actual error type instead of a string?

crates/rattler_index/src/lib.rs
```diff
@@ -75,19 +75,30 @@ impl PreconditionChecks {
...
+#[derive(Debug)]
+pub struct SkippedPackage {
+    /// The filename of the skipped package
+    pub filename: String,
+    /// The error that caused the package to be skipped
+    pub error: std::io::Error,
+}
+
 /// Statistics for a single subdir indexing operation
-#[derive(Debug, Clone, Default)]
```
pavelzw, line 117: why did you remove the `Clone`?

crates/rattler_index/src/lib.rs
```diff
@@ -281,7 +293,9 @@ pub fn package_record_from_conda_reader(reader: impl BufRead) -> std::io::Result
 /// Returns the parsed `PackageRecord`.
 fn parse_package_buffer(buffer: opendal::Buffer, filename: &str) -> std::io::Result<PackageRecord> {
     let reader = buffer.reader();
-    let archive_type = DistArchiveType::try_from(filename).unwrap();
+    let archive_type = DistArchiveType::try_from(filename).ok_or_else(|| {
+        std::io::Error::other(format!("Unknown archive type for file: {filename}"))
+    })?;
```
pavelzw: this is not really a `std::io::Error` 🤔

crates/rattler_index/src/lib.rs
```diff
@@ -794,7 +817,9 @@ async fn index_subdir_inner(
             DistArchiveType::Conda(CondaArchiveType::Conda) => {
                 conda_packages.insert(filename, package);
             }
-            _ => panic!("Unknown archive type"),
+            // Wheel packages are not supported and will have been skipped
+            // during parsing above, so this arm is unreachable in practice.
+            DistArchiveType::Wheel(_) => {}
```
pavelzw: shouldn't we use `unreachable!(<description>)` instead then? if this were to break, something in our code logic would be completely off that we should fix. if we only skip it, that would go unnoticed

crates/rattler_index/src/lib.rs
wolfv: std::io::error is not cloneable I thikn

crates/rattler_index/src/lib.rs
wolfv: I've asked claude to make a new wrapper error using thiserror. Let's see.

crates/rattler_index/src/lib.rs
```diff
@@ -265,7 +276,8 @@ fn read_index_json_from_archive(
 pub fn package_record_from_conda_reader(reader: impl BufRead) -> std::io::Result<PackageRecord> {
     let bytes = reader.bytes().collect::<Result<Vec<u8>, _>>()?;
     let reader = Cursor::new(&bytes);
-    let mut archive = seek::stream_conda_info(reader).expect("Could not open conda file");
+    let mut archive = seek::stream_conda_info(reader)
+        .map_err(|e| std::io::Error::other(format!("Could not open conda file: {e}")))?;
```
pavelzw: why not just wrap the `ExtractError` from `rattler_package_streaming`?

crates/rattler_index/src/lib.rs
wolfv: Yeah that might be better

crates/rattler_index/src/lib.rs
```diff
@@ -731,38 +745,39 @@ async fn index_subdir_inner(
...
-                ));
-                return Err(join_err.into());
+                tracing::warn!(
+                    "Skipping package in {} due to unexpected error: {}",
+                    subdir,
+                    join_err
+                );
+                packages_skipped.push(SkippedPackage {
+                    filename: "unknown".to_string(),
+                    error: std::io::Error::other(join_err),
```
pavelzw: this is also not really an `io::Error`. i would prefer if we just pass the actual errors in `SkippedPackage`

crates/rattler_index/src/lib.rs
```diff
@@ -731,38 +744,39 @@ async fn index_subdir_inner(
...
-                ));
-                return Err(join_err.into());
+                tracing::warn!(
+                    "Skipping package in {} due to unexpected error: {}",
+                    subdir,
+                    join_err
+                );
+                packages_skipped.push(SkippedPackage {
+                    filename: "unknown".to_string(),
+                    error: std::io::Error::other(join_err).into(),
```
pavelzw: we should either add `JoinError` to `PackageReadError` or fail completely here. `JoinError`s definitely occur significantly less often than all other io errors above

crates/rattler_index/src/lib.rs
```diff
@@ -772,6 +786,14 @@ async fn index_subdir_inner(
         subdir.as_str()
     ));
 
+    if !packages_skipped.is_empty() {
+        tracing::warn!(
+            "{} packages in {} were skipped due to errors.",
+            packages_skipped.len(),
+            subdir
+        );
+    }
```
pavelzw: i would prefer if we actually returned an error in this case. otherwise these failures will stay hidden in some logs somewhere.

see also how it's implemented in conda-mirror

PR #2105, opened 2026-02-23
fix: strip local version segments for range and compatibility matching

### LINKED ISSUE #1033: rattler.solve sometimes can't select pinned local versions
### Checklist

### What happened?

I have a spec `mypackage ~=2.0`.  If I have a previously installed version `2.0.0+fastbuild` passed via `pinned_packages`, I sometimes get one of two errors:
```
exceptions.SolverException: Cannot solve the request because of: No candidates were found for mypackage ~=2.0.
```
or
```
exceptions.SolverException: Cannot solve the request because of: The following packages are incompatible
└─ mypackage ~=2.0 can be installed with any of the following options:
   └─ mypackage 2.0.0 | 2.0.0
├─ mypackage 2.0.0+fastbuild is locked, but another version is required as reported above
├─ mypackage 2.0.0+fastbuild is locked, but another version is required as reported above
```

### Additional Context

In some of these cases, the pinned package is not in `channels` passed to `rattler.solve` (because the package was originally installed out of another channel).  I think pinned packages need to be considered regardless of if they're present in any of the channels.

devvaansh:
### Description

When a pinned/locked package has a local version string identifier (e.g. 2.0.0+fastbuild), the solver incorrectly rejects it against range-based version specs like ~=2.0 or >=2.0.

Root cause: In conda's version ordering, string-based local components (Component::Iden("fastbuild")) sort below numeric components (Component::Numeral(0)). This means 2.0.0+fastbuild < 2.0.0 in raw comparison, which causes >=2.0 and ~=2.0 checks to fail — even though semantically the base version 2.0.0 clearly satisfies the constraint.

Fix: When a version spec constraint (range operators like >=, <, <=, >, and compatibility operators ~=) references a version without a local segment, we now strip the local segment from the candidate version before comparison. This aligns with PEP 440 semantics where local versions should be ignored for ordering constraints and only matter for exact equality.

Before: mypackage ~=2.0  +  pinned 2.0.0+fastbuild  →  ❌ "No candidates found"

After: mypackage ~=2.0  +  pinned 2.0.0+fastbuild  →  ✅ resolves correctly

Screenshot 2026-02-23 at 11 16 26 PM

Fixes #1033 

### How Has This Been Tested?

Three new unit tests were added to crates/rattler_conda_types/src/version_spec/mod.rs:

test_local_version_compatible_release — verifies ~=2.0 matches 2.0.0+fastbuild
test_local_version_range_operators — verifies >=2.0, >1.9, <3.0, <=2.0.0 all correctly match 2.0.0+fastbuild
test_local_version_exact_match_unchanged — verifies ==2.0.0 still does NOT match 2.0.0+fastbuild (exact matching is unaffected)

### AI Disclosure

Tools:  Claude
### Checklist:

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -140,6 +140,11 @@ pub struct LinkedFile {
 ///
 /// Note that usually the `target_prefix` is equal to `target_dir` but it might differ. See
 /// [`crate::install::InstallOptions::target_prefix`] for more information.
+///
+/// The `modification_time` is a timestamp we set on all files we modify. We want a value
+/// we control here to make the generated filesystem tree more reproducible. `modification_time`
+/// should be greater than any of the modification times of any of the files that were packaged
+/// up (ignoring any data conda stores).
 #[allow(clippy::too_many_arguments)] // TODO: Fix this properly
 pub fn link_file(
     path_json_entry: &PathsEntry,
@@ -152,6 +157,7 @@ pub fn link_file(
     allow_ref_links: bool,
     target_platform: Platform,
     apple_codesign_behavior: AppleCodeSignBehavior,
+    modification_time: filetime::FileTime,
 ) -> Result<LinkedFile, LinkFileError> {
     let source_path = package_dir.join(&path_json_entry.relative_path);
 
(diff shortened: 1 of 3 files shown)


baszalmstra:
Have you tried this against condas implementation? Unless we are deviating from condas behavior I wont merge this without going through a CEP.

devvaansh:
@baszalmstra you’re right, i double checked this against conda’s implementation (specifically VersionOrder.__lt__), and changing this would indeed diverge from their current str < int ordering that said, the underlying issue (#1033) still blocks users from using pinned local builds given this constraint, would you be open to a PR that tackles this at the solver/pinning layer instead? we could match pinned packages by their record identity, rather than running them through the usual versionSpec matching logic that would resolve the deadlock for pinned packages without touching the core versioning semantics or requiring a CEP, If that direction makes sense to you, happy to re scope this pr and focus on the solver logic.

PR #2110, opened 2026-02-24
fix(package-streaming): Handle pre-1980 timestamps in tar extraction for exFAT compatibility

wolfv:
### Description

This PR fixes tar extraction failures on filesystems like exFAT that cannot represent timestamps before 1980-01-01. The issue occurs when tar archives contain entries with `mtime=1` (Unix epoch + 1 second), a common sentinel value used by some tools.

**Changes:**
- Added `SAFE_MTIME_FLOOR` constant (1980-01-01T00:00:00 UTC = 315,532,800 seconds) to both sync and async extraction paths
- Implemented `unpack_tar_archive_sync()` function that disables automatic mtime preservation and manually sets mtimes with clamping to the safe floor
- Updated async extraction in `tokio/shared.rs` to manually set mtimes with the same clamping logic
- Modified `tokio/async_read.rs` to disable automatic mtime preservation
- Added `filetime` dependency for safe mtime manipulation
- Mtime setting failures are now logged as warnings instead of causing fatal extraction errors

**Behavior:**
- Pre-1980 timestamps are clamped to 1980-01-01 instead of causing extraction to fail
- Extraction continues gracefully even if mtime setting fails on the target filesystem
- Warnings are logged when mtime operations fail, but package integrity is unaffected

Fixes filesystem compatibility issues when extracting conda packages on exFAT and similar filesystems.

### How Has This Been Tested?

Added comprehensive test coverage:
- `test_extract_tar_with_pre_1980_mtime()`: Tests synchronous extraction with `mtime=1` sentinel value
- `test_extract_tar_with_pre_1980_mtime_async()`: Tests asynchronous extraction with the same scenario

Both tests verify that:
1. Extraction completes successfully despite pre-1980 timestamps
2. Files are actually extracted with correct content
3. No fatal errors occur due to timestamp handling

### Checklist:

https://claude.ai/code/session_01JxbSknXAJpJTfYzzxLzDXy

### DIFF
--- a/crates/rattler_package_streaming/src/read.rs
+++ b/crates/rattler_package_streaming/src/read.rs
@@ -4,7 +4,11 @@
 use super::{ExtractError, ExtractResult};
 use std::io::{Seek, SeekFrom, copy};
 use std::mem::ManuallyDrop;
-use std::{ffi::OsStr, io::Read, path::Path};
+use std::{
+    ffi::OsStr,
+    io::Read,
+    path::{Component, Path, PathBuf},
+};
 use tempfile::SpooledTempFile;
 use zip::read::{ZipArchive, ZipFile, read_zipfile_from_stream};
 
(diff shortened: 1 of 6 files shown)


crates/rattler_package_streaming/src/read.rs
```diff
@@ -104,6 +110,60 @@ fn extract_zipfile<R: std::io::Read>(
...
+        let entry_type = entry.header().entry_type();
+        let path = entry.path().map_err(ExtractError::IoError)?.into_owned();
+
+        let unpacked = entry
+            .unpack_in(destination)
+            .map_err(ExtractError::IoError)?;
+
+        if unpacked {
+            let full_path = destination.join(&path);
+            set_mtime_safe(&full_path, mtime, entry_type.is_symlink());
```
hunger: This is a TOCTOU issue... but probably not exploitable, considering that mtimes do not effect access to files.

hunger:
Should we just add this, even without feedback? It should improve the situation for some users.

marc-lehner:
Sorry, forgot about this issue. I just verified it on an exFAT external hard drive, and it worked as expected. I was able to extract the `rsa` package, and the timestamp was `1980-01-01 1:00` in the fully extracted package

I can't judge the code itself, as I'm really not a Rust expert

hunger:
@wolfv: Do you want to patch this up and merge it or should I?

hunger:
I took the liberty to rebase this on top of current master and add a change on top to make sure we do not set mtimes on files outside the extraction directory.

hunger (approved):

PR #2113, opened 2026-02-24
feat: Add index_url into Pypi packages

hunger:
### Description

Record any non-pypi index URL in python packages.

### How Has This Been Tested?

Mostly with test in Pixi, using https://github.com/hunger/pixi/tree/feature/lockfile-v7 This can nto go into pixi before the lockfile-v7 branch has not landed in rattler.

### AI Disclosure
Tools: Claude

### Checklist:

### DIFF
--- a/crates/rattler_lock/src/pypi.rs
+++ b/crates/rattler_lock/src/pypi.rs
@@ -18,6 +18,9 @@ pub struct PypiPackageData {
     /// The location of the package. This can be a URL or a path.
     pub location: Verbatim<UrlOrPath>,
 
+    /// The index this came from. Is `None` for source dependencies
+    pub index_url: Option<url::Url>,
+
     /// Hashes of the file pointed to by `url`.
     pub hash: Option<PackageHashes>,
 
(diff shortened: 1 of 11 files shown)


crates/rattler_lock/src/pypi.rs
```diff
@@ -18,6 +18,9 @@ pub struct PypiPackageData {
     /// The location of the package. This can be a URL or a path.
     pub location: Verbatim<UrlOrPath>,
 
+    /// The index this came from. Might be None, if the default index was used
+    pub index_url: Option<url::Url>,
```
wolfv, line 22: One thought was to use a wrapper type (`PyPIUrl`) or so, that would not serialize if it's a default `pypi.io` URL (or whatever it is). Might be a tiny bit of an improvement.

hunger: I like that idea a lot.

In pixi we use one of the uv crates to handle the index address, which I do not want to add here. So I have to hard-code the pypi index URL here, which is not too elegant. It works and I can move more of the testing here, which is a good thing.

PR #2114, opened 2026-02-24
fix(networking): cache GCS OAuth2 token across requests

suhr25:
## Summary

This PR improves the efficiency and reliability of GCS authentication by caching and reusing the OAuth2 CacheableResource across requests.

Previously, GCSMiddleware had no shared state, causing every request to rebuild credentials and fetch a fresh token. Under concurrent workloads, this resulted in unnecessary auth requests, increased latency, and additional connection pressure.

---

## Description

- Introduced shared state inside `GCSMiddleware` using `Arc` to persist authentication data across requests  
- Cached the `CacheableResource` instead of rebuilding credentials per request  
- Reused token headers until expiry based on `valid_until`  
- Added synchronization to ensure safe concurrent access and avoid redundant refreshes  
- Ensured only one refresh occurs when the token expires (singleflight-style behavior)

---

## Fixes

Internal improvement, no existing issue

---

## How Has This Been Tested?

- Verified GCS requests authenticate correctly with cached tokens  
- Confirmed no change in request/response behavior  
- Observed that token fetch happens once per token lifetime instead of per request  
- Confirmed reduced auth calls under concurrent workloads  
- Verified correctness under async shared state (`Arc` + synchronization)  
- Ran:
  - `pixi install`
  - `pixi run build`

---

## AI Disclosure

- [x] This PR contains AI-generated content  
- [x] I have tested any AI-generated content in my PR  
- [x] I take responsibility for any AI-generated content in my PR  

Tools: ChatGPT,Claude

---

## Checklist

### DIFF
--- a/crates/rattler_networking/src/gcs_middleware.rs
+++ b/crates/rattler_networking/src/gcs_middleware.rs
@@ -212,7 +212,10 @@ impl GCSMiddleware {
                 let entity_tag = r.entity_tag.clone();
                 let headers = r.headers.clone();
                 *guard = CacheState::Ready(r);
-                PollResult::Validate { entity_tag, headers }
+                PollResult::Validate {
+                    entity_tag,
+                    headers,
+                }
             }
 
             // ── Empty ────────────────────────────────────────────────────────
@@ -274,7 +277,10 @@ impl GCSMiddleware {
 
                 // ── Validate ─────────────────────────────────────────────────
                 // Check whether the cached token is still current.
-                PollResult::Validate { entity_tag, headers } => {
+                PollResult::Validate {
+                    entity_tag,
+                    headers,
+                } => {
                     let cred = self.get_credential().await?;
                     let mut ext = http::Extensions::new();
                     ext.insert(entity_tag);
--- a/crates/rattler_solve/src/resolvo/mod.rs
+++ b/crates/rattler_solve/src/resolvo/mod.rs
@@ -557,10 +557,8 @@ impl<'a> CondaDependencyProvider<'a> {
         // Build a lookup table for dependency overrides keyed by target package name.
         let mut override_map: HashMap<PackageName, Vec<DependencyOverride>> = HashMap::new();
         for rule in dependency_overrides {
-            if let Some(name) = rule.override_spec.name.as_ref() {
-                if let Some(name) = Option::<PackageName>::from(name.clone()) {
-                    override_map.entry(name).or_default().push(rule);
-                }
+            if let Some(name) = rule.override_spec.name.as_exact() {
+                override_map.entry(name.clone()).or_default().push(rule);
             }
         }
 
@@ -600,9 +598,8 @@ impl<'a> CondaDependencyProvider<'a> {
         record: &RepoDataRecord,
         dep_spec: &MatchSpec,
     ) -> Option<String> {
-        let dep_name = dep_spec.name.as_ref()?;
-        let dep_name = Option::<PackageName>::from(dep_name.clone())?;
-        let rules = self.dependency_overrides.get(&dep_name)?;
+        let dep_name = dep_spec.name.as_exact()?;
+        let rules = self.dependency_overrides.get(dep_name)?;
         for rule in rules {
             if rule.package_matcher.matches(&record.package_record) {
                 return Some(rule.override_spec.to_string());


suhr25:
Hi @baszalmstra,
Noticed we were fetching a new GCS token on every request since the middleware had no shared state. That adds extra latency and unnecessary load under concurrency.
This just adds a small shared cache so we reuse tokens until expiry.

baszalmstra:
Please fill out the PR template.

suhr25:
@baszalmstra @pavelzw,
Updated the PR to follow the template and added the required details.
Thanks!

crates/rattler_networking/src/gcs_middleware.rs
```diff
@@ -27,35 +72,75 @@ impl Middleware for GCSMiddleware {
...
-    let scopes = ["https://www.googleapis.com/auth/devstorage.read_only"];
-
-    match AccessTokenCredentialBuilder::default()
-        .with_scopes(scopes)
-        .build()
-    {
-        Ok(token_source) => {
-            let extensions = http::Extensions::new();
-            let headers = match token_source.headers(extensions).await {
-                Ok(CacheableResource::New { data, .. }) => data,
```
wolfv, line 47: Shoudl we just figure out how to cache the cacheable resource?

suhr25: Yeah ,that makes sense. I’ll update this PR to cache the CacheableResource itself inside GcsMiddleware

suhr25:
Hi @baszalmstra @wolfv,
Updated to cache the CacheableResource in GcsMiddleware with shared state and singleflight refresh (via Notify), ensuring reuse across requests and refresh only when needed.
Thanks!

baszalmstra:
You did not follow the PR template.

suhr25:
Hi @baszalmstra, updated the PR description to strictly follow the template. Please take another look, thanks!

crates/rattler_networking/src/gcs_middleware.rs
```diff
@@ -27,45 +101,326 @@ impl Middleware for GCSMiddleware {
...
+    ///   the lock, and ask the credential library whether the token is still
+    ///   current.  The library returns `NotModified` (fast path) or `New`
+    ///   (token was silently refreshed); in the latter case the cache is
+    ///   overwritten.  No lock is ever held across an `await`.
+    async fn get_or_refresh_token(&self) -> MiddlewareResult<http::HeaderMap> {
+        loop {
+            let mut guard = self.inner.cache.lock().await;
+
+            // ── Arm 1: another task is fetching – wait for it ───────────────
+            if matches!(*guard, CacheState::Refreshing) {
```
baszalmstra: Instead of a continue, can we use a proper match inclusing the code below?

crates/rattler_networking/src/gcs_middleware.rs
```diff
@@ -1,12 +1,86 @@
...
+    /// A valid resource is available.
+    Ready(CachedResource),
+}
+
+/// Shared, ref-counted state owned by every clone of a [`GCSMiddleware`].
+struct GCSInner {
+    /// Credential source built once and reused across all requests.
+    credential: Mutex<Option<Credentials>>,
+    /// Cache state machine guarded by a mutex.
+    cache: Mutex<CacheState>,
```
baszalmstra, line 47: Should we use a std Mutex here? Since it is not used across await points?

baszalmstra (changes requested):

crates/rattler_networking/src/gcs_middleware.rs
suhr25: In the most recent push, the single flight refresh behavior is maintained while the cache state handling is now an explicit match over CacheState.

crates/rattler_networking/src/gcs_middleware.rs
suhr25: Switched credential and cache to std::sync::Mutex

suhr25:
Hi @baszalmstra @wolfv ,
Kindly review this PR, I have made changes asked by you.
As well as all CI's are green.
Thanks!

baszalmstra (approved):

baszalmstra:
Thanks! Pretty sick PR. 👍

suhr25:
Thankyou!!

PR #2135, opened 2026-02-26
feat: improve Windows GUI app launching and file extension registration

wolfv:
### Description

This PR improves the Windows menu installation functionality in two key areas:

1. **GUI Application Launching**: Added support for launching GUI applications without showing a console window when `terminal: false` is set. Previously, console-subsystem executables would display a console window even for GUI applications. The fix uses a PowerShell wrapper script launched via `cmd.exe /MIN` to suppress the console window while maintaining argument passing support.

2. **File Extension Registration**: Fixed the Windows registry structure for file extension icons. The `DefaultIcon` value is now correctly registered as a subkey with an empty string value (the standard Windows registry pattern) instead of a named value. Additionally, friendly type names are now generated for file extensions (e.g., ".h5" → "H5 File").

3. **Documentation**: Updated the schema documentation to clarify that the `terminal` option on Windows now suppresses console windows for GUI applications, not just when `activate` is true.

### Changes Made

- **windows.rs**: 
  - Added new code path for `activate: false, terminal: false` that writes a wrapper script and launches it via PowerShell with hidden window style
  - Added clarifying comments for the three different command execution paths
  - Generate friendly type names for file extensions during registration

- **registry.rs**:
  - Fixed `DefaultIcon` registration to use the correct registry structure (subkey with empty string value)
  - Updated test assertions to match the corrected registry structure

- **schema.rs**:
  - Updated documentation for the `terminal` field to reflect the new Windows behavior

### How Has This Been Tested?

The changes include updated unit tests in `registry.rs` that verify the corrected icon registration structure. The existing test suite validates the registry operations.

### Checklist:
- [x] Tests have been updated to reflect the changes

https://claude.ai/code/session_01JG5ZyB9WWTDmbfAovXisJ5


gdementen:
Thanks a lot for tackling this.

> This PR improves the Windows menu installation functionality in two key areas:
> 
>     1. **GUI Application Launching**: Added support for launching GUI applications without showing a console window when `terminal: false` is set. Previously, console-subsystem executables would display a console window even for GUI applications. The fix uses a PowerShell wrapper script launched via `cmd.exe /MIN` to suppress the console window while maintaining argument passing support.

~While I **really** like the semantic change of the terminal: false (I wish it always meant that), changing it in rattler without also changing it in menuinst itself is probably going to result in a bad user experience. Additionally, I would personally prefer it if the `cmd.exe /MIN` trick was not used in that case because some antimalware solutions (including ours) flag shortcuts using that as malware and block them. Besides,~ I don't think it is necessary **when no activation is necessary**  because, in that case, simply using `pythonw.exe` instead of `python.exe` works fine and is a tad faster. 

>     2. **File Extension Registration**: Fixed the Windows registry structure for file extension icons. The `DefaultIcon` value is now correctly registered as a subkey with an empty string value (the standard Windows registry pattern) instead of a named value. Additionally, friendly type names are now generated for file extensions (e.g., ".h5" → "H5 File").

IIRC, the icon problem is specific to rattler (not present in menuinst), so fixing it is uncontroversial. However, the "unfriendly" type name issue is also present in menuinst. Since the menuinst name is so strange and not actually specified/documented explicitly anywhere that I could find, I think it is fine to fix in rattler even if that's not fixed in menuinst. I meant to report that on their issue tracker for a while but haven't got to it yet. I will eventually do that unless someone else beats me to it 😉.

>     3. **Documentation**: Updated the schema documentation to clarify that the `terminal` option on Windows now suppresses console windows for GUI applications, not just when `activate` is true.

As said above, you probably want to discuss this with menuinst to keep the semantic synchronized.

gdementen:
Sorry for my first reply concerning GUI applications. I got a bit confused. When activate is False, I now believe nothing needs to change except possibly the documentation of menuinst at `https://conda.github.io/menuinst/defining-shortcuts/#migrating-pywscript-and-pyscript-to-menuinst-v2` which does not mention the case where activation is not necessary.

gdementen:
As for the `activate: true` case, I don't know if anything needs changing given that I can't test that option due to our antimalware solution (Sentinel One) blocking any shortcut using that.

PR #2141, opened 2026-02-27
fix: make package cache global lock opt in

### LINKED ISSUE #2120: global package lock causes performance regression on 32+ processes using shared cache on VAST FS
### Checklist

### What happened?

https://github.com/conda/rattler/pull/1818 was a performance improvement to reduce syscalls, but now the mutex section is so large that multiple processes (cannot interleave and) trying to install a pixi env at the same time is slower.

The use case is that we hydrate pixi envs inside each slurm task, and they all pull from the same local user cache directory.

```
| 231010 |    pixi63-speed-4nodes    | h100-reserved | COMPLETED |     0:09:08     |
| 231317 |    pixi60-speed-4nodes    | h100-reserved | COMPLETED |     0:10:22     |
| 231422 |    pixi59-speed-4nodes    | h100-reserved | COMPLETED |     0:01:44     |
```

### Additional Context

_No response_

codeafridi:
Fixes #2120 

removed the package cache bottle neck by making the global cache lock opt in instead of default . installs now rely on per package meta data exclusive file locks so processess can install safely . updated cache lock updatesto reflect this behaviour (RATTLER_PACKAGE_CACHE_GLOBAL_LOCK enables old global hacking

### DIFF
--- a/crates/rattler_cache/benches/concurrent_cache_lock.rs
+++ b/crates/rattler_cache/benches/concurrent_cache_lock.rs
@@ -36,9 +36,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                     .map(|path| {
                         let cache = cache.clone();
                         let path = path.clone();
-                        tokio::spawn(
-                            async move { cache.get_or_fetch_from_path(&path, None).await },
-                        )
+                        tokio::spawn(async move { cache.get_or_fetch_from_path(&path, None).await })
                     })
                     .collect();
                 for handle in handles {


crates/rattler/src/install/installer/mod.rs
```diff
@@ -487,12 +487,22 @@ impl Installer {
             )
         });
 
-        // Acquire a global lock on the package cache for the entire installation.
-        // This significantly reduces overhead by avoiding per-package locking.
-        let _global_cache_lock = package_cache
-            .acquire_global_lock()
-            .await
-            .map_err(InstallerError::FailedToAcquireCacheLock)?;
+        // Optionally acquire a global lock on the package cache for the entire installation.
```
baszalmstra: Id be fine with just removing this.

crates/rattler_cache/src/package_cache/cache_lock.rs
```diff
@@ -294,6 +311,12 @@ impl CacheMetadataFile {
     }
 }
 
+impl Drop for CacheMetadataFile {
+    fn drop(&mut self) {
+        let _ = fs4::fs_std::FileExt::unlock(&*self.file);
```
baszalmstra, line 316: Is this really needed? Is the lock kept when the file handle os released?

baszalmstra (changes requested):

crates/rattler/src/install/installer/mod.rs
codeafridi: your right i removed it

baszalmstra:
Do you think you can run some benchmarks to verify the performance improvements?

codeafridi:
> Do you think you can run some benchmarks to verify the performance improvements?

yeah these are results - 
```bash 
ache_lock/concurrent_per_package_lock
                        time:   [312 ms 328 ms 345 ms]

cache_lock/serial_simulating_global_lock
                        time:   [1.24 s 1.31 s 1.38 s]
```

baszalmstra:
The pr is failing ci checks and you didnt fill in the pr template.

PR #2143, opened 2026-02-27
fix: validate symlink targets to prevent prefix escape

Premkumar-2004:
### Description

Symlink targets weren't being validated anywhere. A malicious package could include a symlink like `lib/foo  ../../../../.bashrc` and it'd get created as-is in the prefix.

 Disabled `allow_external_symlinks` in the tar` ArchiveBuilder`,Added target `validation in symlink_to_destination `before linking,Added target validation in `validate_package_soft_link_entry`

### How Has This Been Tested?
Tested locally symlinks escaping the prefix are now rejected.

### AI Disclosure

Tools: claude

### Checklist:

### DIFF
--- a/crates/rattler/src/install/installer/mod.rs
+++ b/crates/rattler/src/install/installer/mod.rs
@@ -883,7 +883,7 @@ fn create_spec_mapping(specs: &[MatchSpec]) -> std::collections::HashMap<Package
     let mut mapping = std::collections::HashMap::new();
 
     for spec in specs {
-        if let Some(PackageNameMatcher::Exact(name)) = &spec.name {
+        if let PackageNameMatcher::Exact(name) = &spec.name {
             mapping
                 .entry(name.clone())
                 .or_insert_with(Vec::new)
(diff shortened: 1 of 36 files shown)


crates/rattler_cache/src/validation.rs
```diff
@@ -251,18 +255,33 @@ fn validate_package_soft_link_entry(
...
     }
 
-    // TODO: Validate symlink content. Dont validate the SHA256 hash of the file
-    // because since a symlink will most likely point to another file added as a
-    // hardlink by the package this is double work. Instead check that the
-    // symlink is correct e.g. `../a` points to the same file as `b/../../a` but
-    // they are different.
+    // Resolve the symlink target relative to its parent and verify it stays
+    // inside the package directory.
+    let target = std::fs::read_link(&path)?;
```
baszalmstra: I dont think we should check this here. The validation is to check whether the extracted package still valid compared to original archive. The check during install should be enough.

Premkumar-2004: Fair point

Premkumar-2004:
i'll add  a test case that confirms symlink escaping the prefix gets rejected .

Premkumar-2004:
i have one more pr which i raised a week before could you merge that #2090

wolfv:
I am not sure this will be the right thing to do. We migth have to discsus this wiht teh conda community. There migth be packages that legitimately ship symlinks to outside the env? E.g. for drivers?

Premkumar-2004:
I hadn't considered driver packages. We could log a warning instead of rejecting, so nothing breaks but it's still visible. Or make it configurable so stricter setups can opt into hard failing. i can  adjust either way, let me know what you'd prefer.

wolfv:
Yeah we do have a setting that allows / disallows running of post-link scripts. Something similar perhaps?

Premkumar-2004:
Yeah I can add an `allow_external_symlinks` flag to the install options, similar to how post-link scripts are handled.

Premkumar-2004:
hi @baszalmstra i dont understand why all the testcases are failing

Premkumar-2004:
@baszalmstra can you please review the pr now i have updated it  and can you please re-run the testcases

baszalmstra (approved):

PR #2150, opened 2026-03-01
fix : global package cache lock performance regression

codeafridi:
Fixes #2141  a performance regression caused by global package cache locking. Reduces unnecessary contention in the cache lock path, improving concurrent install performance.


baszalmstra:
Duplicate of #2141 . Your own pull request....

PR #2154, opened 2026-03-02
fix: make package cache global lock opt in with package file lock

### LINKED ISSUE #2120: global package lock causes performance regression on 32+ processes using shared cache on VAST FS

codeafridi:
Fixes #2120  
Removes the global cache lock that was serialising all package fetches across processes. Each cache entry now takes its own file lock, so 32+ processes sharing the same cache can fetch different packages in parallel without blocking each other.
Also added a small benchmark 
```bash cargo bench -p rattler_cache --bench concurrent_cache_lock ```

### DIFF
--- a/crates/rattler_cache/benches/concurrent_cache_lock.rs
+++ b/crates/rattler_cache/benches/concurrent_cache_lock.rs
@@ -36,9 +36,7 @@ fn criterion_benchmark(c: &mut Criterion) {
                     .map(|path| {
                         let cache = cache.clone();
                         let path = path.clone();
-                        tokio::spawn(
-                            async move { cache.get_or_fetch_from_path(&path, None).await },
-                        )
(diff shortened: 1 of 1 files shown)


baszalmstra:
denounce for spamming

PR #2161, opened 2026-03-03
feat: add timestamp to source packages and bump lock file to V7

### LINKED ISSUE #2160: rattler-lock v7: Add a timestamp into Source Packages

ayushman1210:
## Description

Source packages in the lock file didn't have a way to store when they were resolved. This makes it hard for tools to determine staleness or track when a source package entry was last updated.

This PR adds an optional `timestamp` field to `CondaSourceData` and bumps the lock file format from V6 to V7.

**What changed:**
- `CondaSourceData` now has `timestamp: Option<chrono::DateTime<chrono::Utc>>`
- Lock file format bumped to V7 to reflect the schema change
- Serialization reads timestamp from the source data directly (falls back to `PackageRecord.timestamp` for binary packages)
- All snapshot tests updated accordingly

Fixes #2160

## How Has This Been Tested?

Ran the full `rattler_lock` test suite locally:

```
cargo test -p rattler_lock
```

Result: 83 passed, 0 failed.

Snapshot tests were updated from `version: 6` → `version: 7` to match the new format.

## AI Disclosure
This PR contains AI-generated content.

Tools: Gemini

## Checklist:
- [x] I have made corresponding changes to the documentation / doctests

### DIFF
--- a/crates/rattler-bin/src/commands/create.rs
+++ b/crates/rattler-bin/src/commands/create.rs
@@ -183,7 +183,7 @@ pub async fn create(opt: Opt) -> miette::Result<()> {
         AuthenticationStorage::from_env_and_defaults().into_diagnostic()?,
     ));
     #[cfg(feature = "gcs")]
-    let download_client = download_client.with(rattler_networking::GCSMiddleware);
+    let download_client = download_client.with(rattler_networking::GCSMiddleware::default());
     let download_client = download_client.build();
 
     // Get the package names from the matchspecs so we can only load the package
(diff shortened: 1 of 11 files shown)


hunger:
We started a discussion on how to best add a timestamp here: https://github.com/conda/rattler/issues/2160

I do not think it makes much sense to store this in the source file. How can we merge that ever with other devs?

ayushman1210:
@hunger Thanks for bringing this up  that’s a fair concern.
The idea here is that the timestamp would only be updated when the lock file is explicitly regenerated, not during normal builds or unrelated changes. So it shouldn’t be changing constantly.
Since lock files are already versioned and expected to change when dependencies are updated, I was thinking this would behave similarly. The timestamp is mainly meant to indicate when the source package resolution happened.
That said, I agree that we should avoid unnecessary merge noise. If this feels like it could cause friction, I’m happy to explore making it optional or adjusting how it’s stored.

baszalmstra:
Thanks for this contribution but we wont be merging this. We are working on a big lock-file refactor in a seperate branch.

PR #2163, opened 2026-03-04
fix(repodata-gateway): limit concurrent shard cache reads

wolfv:
### Description

This PR adds a separate concurrency limiter for IO operations (specifically reading shard cache files from disk) to prevent exhausting the OS file-descriptor limit when many packages are queried concurrently.

Previously, only HTTP request concurrency was limited via `max_concurrent_requests`. When querying for many packages at once (e.g., with wildcard patterns), the gateway could attempt to open many shard cache files simultaneously, leading to file-descriptor exhaustion.

**Changes:**
- Added `max_concurrent_io` field to `GatewayBuilder` with builder methods `with_max_concurrent_io()` and `set_max_concurrent_io()`
- Added `io_concurrency_semaphore` to `GatewayInner` to track the IO concurrency limit
- Updated `ShardedSubdir` to acquire an IO semaphore permit before reading cached shard files from disk
- Updated `SubdirBuilder` to pass the IO semaphore to `ShardedSubdir` during construction

The IO semaphore is acquired before opening shard cache files, ensuring that concurrent file operations are bounded independently from HTTP request concurrency.

### How Has This Been Tested?

The changes follow the existing pattern used for `concurrent_requests_semaphore` and integrate with the existing concurrency control infrastructure. Existing tests should continue to pass as the new semaphore is optional and defaults to unlimited when not configured.

### Checklist:

https://claude.ai/code/session_018WhkrxiAzPjv4ByCqhXKg4


baszalmstra (approved):

baszalmstra:
I think this makes sense. Can you update it on main?

PR #2165, opened 2026-03-05
fix(cache): separate package cache by platform

### LINKED ISSUE #1014: Cache overwritten by package from different platform.
> [!NOTE] 
> This is an edge case I stumbled upon when testing out `pixi build` which will be fixed separately, but it shows a limitation of our cache implementation.

When downloading files to the cache for platform `x` and later for platform `y` it might unnoticeably incorrect overwrite the file or not overwrite when needed. 

The exact example to recreate this is installing [`_openmp_mutex-4.5-2_gnu`](https://prefix.dev/channels/conda-forge/packages/_openmp_mutex) on Windows, but from the `linux-64` platform. Which has a symlink in the package to `libomp.so.1`that doesn't work on Windows, creating the error that linking failed. 

```
  × failed to link _openmp_mutex-4.5-2_gnu.conda
  ├─▶ failed to link 'lib/libgomp.so.1'
  ├─▶ failed to copy file to destination
  ╰─▶ failed to copy file from C:\Users\x\AppData\Local\rattler\cache\pkgs\_openmp_mutex-4.5-2_gnu\lib/libgomp.so.1 to C:\Users\x\examples\b\.pixi\envs\default\lib/libgomp.so.1: 
      The system cannot find the file specified. (os error 2)
```

Overwriting the cache if the package identifier is exactly the same (name, version, build, extension) is an implementation that is copied from how `conda` and `mamba` implemented their cache. This would result in always overwriting the cache when cross-platform installing on the same system.

This could be improved by folding the different cache files into different subdirectories based on platform to avoid this issue.

Sahitya0805:
### Description

Previously cache keys did not include platform, causing packages from different platforms to overwrite each other.

This PR stores cached packages under `cache/pkgs/<platform>/<package>`.

Fixes #1014

### How Has This Been Tested?

- Added `test_cache_key_display` in `cache_key.rs` to verify that `CacheKey` correctly prepends the platform folder.
- Ran `cargo test -p rattler_cache` and `cargo check` locally.

### AI Disclosure
Tools: Custom AI Assistant

### Checklist:

### DIFF
--- a/crates/rattler_cache/src/package_cache/mod.rs
+++ b/crates/rattler_cache/src/package_cache/mod.rs
@@ -38,6 +38,21 @@ mod cache_key;
 mod cache_lock;
 mod reporter;
 
+/// Resolves the cache path for a given package, supporting both the new
+/// platform-aware layout and the legacy layout.
+fn resolve_cache_path(base: &Path, key: &CacheKey) -> PathBuf {
+    let new_path = key.cache_path(base);
+    let legacy_path = base.join(key.to_string());
+
+    if new_path.exists() {
+        new_path
+    } else if legacy_path.exists() {
+        legacy_path
+    } else {
+        new_path
+    }
+}
+
 /// A [`PackageCache`] manages a cache of extracted Conda packages on disk.
 ///
 /// The store does not provide an implementation to get the data into the store.
@@ -68,6 +83,7 @@ pub struct BucketKey {
     name: String,
     version: String,
     build_string: String,
+    subdir: Option<String>,
     origin_hash: Option<String>,
 }
 
@@ -77,6 +93,7 @@ impl From<CacheKey> for BucketKey {
             name: key.name,
             version: key.version,
             build_string: key.build_string,
+            subdir: key.subdir,
             origin_hash: key.origin_hash,
         }
     }
@@ -171,7 +188,8 @@ impl PackageCacheLayer {
             .ok_or(PackageCacheLayerError::PackageNotFound)?
             .clone();
         let mut cache_entry = cache_entry.lock().await;
-        let cache_path = self.path.join(cache_key.to_string());
+
+        let cache_path = resolve_cache_path(&self.path, cache_key);
 
         match validate_package_common::<
             fn(PathBuf) -> _,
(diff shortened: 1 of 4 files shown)


baszalmstra:
This doesnt integrate the functionality anywhere and when you do add that its important to think about a migration strategy for people with existing caches.

ritankarsaha:
> This doesnt integrate the functionality anywhere and when you do add that its important to think about a migration strategy for people with existing caches.

Ahh I see the PR here, but I think you have left a comment here mentioning that https://github.com/conda/rattler/pull/2165 is incomplete and doesn't fix the issue, also both addresses different issue tags. Can you look into #2167 once to see if everything is addresses properly over there ?

crates/rattler_cache/src/package_cache/mod.rs
```diff
@@ -215,7 +230,20 @@ impl PackageCacheLayer {
             .clone();
 
         let mut cache_entry = entry.lock().await;
-        let cache_path = self.path.join(cache_key.to_string());
+
+        let new_path = cache_key.cache_path(&self.path);
```
baszalmstra: This commit says remove duplication but this actually adds a lot of duplication..

baszalmstra:
Are you still working on this? There is another PR in #2167 trying to solve the same issue. If you are not working on this PR it might be nice for the author of that PR to continue?

crates/rattler_cache/src/package_cache/mod.rs
Sahitya0805: Thanks for pointing that out!

You're right the cache path resolution logic is currently duplicated. I'll refactor it by introducing a helper function to resolve the cache path (handling both the new platform aware layout and the legacy layout) and update the PR accordingly.

Sahitya0805:
Yes, I'm still working on it. I'll address the duplication issue and push an update shortly.
If PR #2167 already covers everything needed, I'm also happy to close this PR.

Sahitya0805:
Thanks for the review!

I refactored the cache path resolution logic into a helper function to remove duplication and updated the call sites accordingly. I also removed the redundant assignments and verified that all tests pass locally, including `test_cross_platform_package_cache`.

Please let me know if any further changes are needed.

PR #2167, opened 2026-03-05
fix(cache): include subdir in CacheKey to prevent cross-platform cache collisions

### LINKED ISSUE #2166: CacheKey not unique across subdirectories — risk of cross-platform cache collisions
### Checklist

### What happened?

  The CacheKey struct in crates/rattler_cache/src/package_cache/cache_key.rs uniquely identifies a
  cached package using only name, version, and build_string. A TODO comment in the code acknowledges
  this:

  // TODO: This could not be unique over multiple subdir. How to handle?

  This means two packages with identical name/version/build_string but from different subdirectories
  (e.g. linux-64 vs osx-arm64) will map to the same cache directory, and whichever was cached first
  silently wins. The wrong architecture binary could be returned to the installer with no error.

  This is particularly dangerous for:
  - Noarch packages with the same build string as platform-specific packages
  - Multi-platform CI environments sharing a cache

  Expected behavior: Packages from different subdirectories should always have distinct cache
  entries.

### Additional Context

_No response_

ritankarsaha:
### Description

Fixes the long-standing TODO in cache_key.rs about non-uniqueness of CacheKey across subdirectories.

 ### Problem

  CacheKey was keyed only on name + version + build_string. Two packages from different platforms
  (e.g. linux-64 and osx-arm64) with identical coordinates would hash to the same cache entry. The
  first one cached would be returned for both, silently installing the wrong architecture.

  ### Solution

  - Added subdir: Option<String> to CacheKey and BucketKey
  - From<&PackageRecord> now populates subdir from record.subdir
  - Display includes subdir in the directory name when no origin_hash is present:
    - Before: zlib-1.3.1-hb9d3cd8_2
    - After: zlib-1.3.1-hb9d3cd8_2-linux-64
  - When origin_hash is set (URL-based fetch), it takes precedence — the URL already encodes the
  subdir, so behavior is unchanged for that path
  - Added with_subdir() builder for keys created from CondaArchiveIdentifier (filename-only, no
  subdir info) where the caller knows the platform

Fixes #2166 

### How Has This Been Tested?

6 unit tests has been added to this properly

The critical test is test_cross_platform_keys_are_distinct — it directly exercises the bug scenario
 (two packages from linux-64 vs osx-arm64 with identical coordinates) and asserts they produce different cache directory names.

  To run just these tests:
  pixi run -- cargo nextest run -p rattler_cache cache_key::tests

### Checklist:

 ### Breaking change

  Cache directory names for PackageRecord-derived keys will include the subdir suffix. Existing
  cached entries without the suffix will be treated as misses and re-fetched. This is the correct and
   safe behavior.

### DIFF
--- a/crates/rattler_cache/src/package_cache/cache_key.rs
+++ b/crates/rattler_cache/src/package_cache/cache_key.rs
@@ -7,18 +7,30 @@ use std::{
 };
 
 /// Provides a unique identifier for packages in the cache.
-/// TODO: This could not be unique over multiple subdir. How to handle?
+///
+/// The key includes the package name, version, build string, and optionally the
+/// subdirectory (platform identifier, e.g. `linux-64`, `osx-arm64`). Including
+/// the subdir prevents cache collisions between packages with identical
+/// name/version/build coordinates from different platforms.
 #[derive(Debug, Hash, Clone, Eq, PartialEq)]
 pub struct CacheKey {
     pub(crate) name: String,
     pub(crate) version: String,
     pub(crate) build_string: String,
+    /// included in the cache directory name to avoid collisions across subdirectories.
+    pub(crate) subdir: Option<String>,
     pub(crate) sha256: Option<Sha256Hash>,
     pub(crate) md5: Option<Md5Hash>,
     pub(crate) origin_hash: Option<String>,
 }
 
 impl CacheKey {
+    /// Adds the subdirectory (platform identifier, e.g. `linux-64`) to the key.
+    pub fn with_subdir(mut self, subdir: impl Into<String>) -> Self {
+        self.subdir = Some(subdir.into());
+        self
+    }
+
     /// Adds a sha256 hash of the archive.
     pub fn with_sha256(mut self, sha256: Sha256Hash) -> Self {
         self.sha256 = Some(sha256);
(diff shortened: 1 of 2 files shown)


baszalmstra:
Did you just open the same pr twice with a different description?

Edit: nope it was two different people.

ritankarsaha:
> Did you just open the same pr twice with a different description?

Ahh no, I squashed the 3 commits into one, to make it look cleaner, that's all !

baszalmstra:
This is a duplicate of #2165, please see my comment there and figure out amongst yourselves which PR should be merged.

ritankarsaha:
> This is a duplicate of #2165, please see my comment there and figure out amongst yourselves which PR should be merged.

Ahh I see the PR there, but I think you have left a comment there mentioning that #2165 is incomplete and doesn't fix the issue, also both addresses different issue tags.

baszalmstra:
Both PRs suffer from the exact same issue, this one just has more unit tests. But both try to solve the same problem.

ritankarsaha:
> Both PRs suffer from the exact same issue, this one just has more unit tests. But both try to solve the same problem.

Ahh lemme know, which one you planning to move on with, happy to be of help and use in both @baszalmstra

baszalmstra:
I dont really care. Maybe you can discuss with the other author how you'd solve my comment?

PR #2169, opened 2026-03-05
fix: record actual link type in PrefixRecord

### LINKED ISSUE #2168: LinkType HardLink hardcoded in installer — ignores filesystem capabilities and LinkOptions
### Checklist

### What happened?

In crates/rattler/src/install/installer/mod.rs, the PrefixRecord.link.link_type field recorded to
  conda-meta is always Some(LinkType::HardLink), regardless of:

  - Whether the target filesystem actually supports hard links (e.g. NFS, cross-device installations)
  - The caller-supplied LinkOptions (which exposes allow_hard_links, allow_symbolic_links,
  allow_ref_links)

  A TODO comment acknowledged this: "compute the right value here based on the options and
  can_hard_link".

  Consequence: on network filesystems or cross-device setups, packages are installed via file copies
  (correctly), but the conda-meta record falsely says HardLink. Tooling that reads conda-meta to
  reason about installation state gets incorrect data.

### Additional Context

_No response_

ritankarsaha:
### Description

Fixes the TODO in installer/mod.rs that noted link_type: Some(LinkType::HardLink) was always
  hardcoded.

  ### Problem

  PrefixRecord.link.link_type was always written as HardLink, even when:
  - hard links are impossible (cross-device, NFS, CIFS…)
  - the caller explicitly set LinkOptions { allow_hard_links: Some(false) }

  The actual file linking via link_package_sync was already correct (it probes filesystem capability
  and falls back to copies). Only the metadata recording was wrong.

  ### Solution

  - Expose can_create_hardlinks_sync as pub(crate) so the installer can call it
  - Before moving install_options into link_package_sync, capture allow_hard_links
  - After linking, resolve the effective value (explicit option → filesystem probe)
  - Record HardLink when hard links were used, Copy otherwise

Fixes #2168 

### How Has This Been Tested?

  Two new unit tests in installer::tests:
  - test_link_type_recorded_correctly — allow_hard_links: Some(false) → Copy in conda-meta
  - test_link_type_hardlink_when_forced — allow_hard_links: Some(true) → HardLink in conda-meta

### Checklist:

### DIFF
--- a/crates/rattler/src/install/installer/mod.rs
+++ b/crates/rattler/src/install/installer/mod.rs
@@ -728,7 +728,7 @@ async fn link_package(
 
     rayon::spawn_fifo(move || {
         let inner = move || {
-            // Link the contents of the package into the prefix. 
+            // Link the contents of the package into the prefix.
             let (paths, link_type) = crate::install::link_package_sync(
                 &cached_package_dir,
                 &target_prefix,


ritankarsaha:
> Maybe its nicer if this is just returned from link_package_sync instead? Instead of recomputing it?

Okaay looking into it now, and coming up with a fresh commit in a few.

ritankarsaha:
> Maybe its nicer if this is just returned from link_package_sync instead? Instead of recomputing it?

Pushed a new commit now @baszalmstra , does it look good now ?

ritankarsaha:
@baszalmstra can you take a look into this ? FIxed the reviews.

PR #2171, opened 2026-03-05
fix: validate symlink targets and parse features/license_family in MatchSpec

### LINKED ISSUE #2170: Fix symlink target validation and add features/license_family support to MatchSpec
### Checklist

### What happened?

Two separate correctness gaps in the package validation and query layer:

  1. Symlink target validation missing (rattler_cache): validate_package_soft_link_entry only checked
   that a path was a symlink. It never inspected the target. This means broken symlinks (dangling
  targets) and symlinks that escape the package root (via ../../ or absolute paths like /etc/passwd)
  were silently accepted as valid.

  2. features and license_family not parsed in MatchSpec (rattler_conda_types): The bracket-key
  parser contained a TODO comment for these two fields. Any spec using python[license_family=MIT] or
  python[features=feature1] returned InvalidBracketKey instead of populating the field. PackageRecord
   carries both fields, but they could never be matched against.

### Additional Context

_No response_

ritankarsaha:
### Description

  Fixes two independent correctness issues across rattler_cache and rattler_conda_types.

  ### Fix 1 — Symlink target validation (rattler_cache)

  validate_package_soft_link_entry previously contained a TODO and stopped after is_symlink(). This
  PR implements the actual validation:

  - Adds normalize_path to resolve . and .. components without canonicalize (so validation works on
  freshly extracted archives where targets may not exist yet on other code paths, but we still want
  to check containment)
  - Reads the symlink target via std::fs::read_link
  - Resolves the target relative to the symlink's parent directory
  - Returns SymlinkEscapesPackageDirectory if the resolved path does not start with package_dir
  - Returns BrokenSymlink if the resolved path does not exist
  - Adds 4 unit tests that run locally without downloading packages: valid relative symlink, broken
  symlink, relative escape (../../outside), absolute escape (/etc/passwd)

  ### Fix 2 — features and license_family in MatchSpec (rattler_conda_types)

  The bracket-key parser had a TODO for these two fields and was returning InvalidBracketKey. This
  PR:

  - Adds features: Option<String> and license_family: Option<String> to both MatchSpec and
  NamelessMatchSpec
  - Threads the fields through into_nameless, From<MatchSpec>, from_nameless, Display, and
  Matches<PackageRecord> for both structs
  - Adds parser branches for the "features" and "license_family" bracket keys
  - Adds 3 tests: parse license_family with roundtrip, parse features with roundtrip, matching both
  fields against a PackageRecord
  - Updates the test_matchspec_to_string insta snapshot to include the new fields

Fixes #2170 

### How Has This Been Tested?

  - cargo test -p rattler_cache symlink_validation — 4 new tests pass
  - cargo test -p rattler_conda_types match_spec — 100 tests pass (3 new)
  - cargo clippy -p rattler_cache -p rattler_conda_types — no new warnings
  - cargo fmt -p rattler_cache -p rattler_conda_types -- --check — clean

### Checklist:

### DIFF
--- a/crates/rattler_cache/src/validation.rs
+++ b/crates/rattler_cache/src/validation.rs
@@ -79,6 +79,14 @@ pub enum PackageEntryValidationError {
     #[error("expected a symbolic link")]
     ExpectedSymlink,
 
+    /// The symlink target does not exist.
+    #[error("symlink target does not exist (broken symlink)")]
+    BrokenSymlink,
+
+    /// The symlink target resolves to a path outside the package directory.
+    #[error("symlink target escapes the package directory")]
+    SymlinkEscapesPackageDirectory,
+
     /// The file is not a directory.
     #[error("expected a directory")]
     ExpectedDirectory,
@@ -173,7 +181,7 @@ fn validate_package_entry(
     // Validate based on the type of path
     match entry.path_type {
         PathType::HardLink => validate_package_hard_link_entry(path, entry, mode),
-        PathType::SoftLink => validate_package_soft_link_entry(path, entry, mode),
+        PathType::SoftLink => validate_package_soft_link_entry(package_dir, path, entry, mode),
         PathType::Directory => validate_package_directory_entry(path, entry, mode),
     }
 }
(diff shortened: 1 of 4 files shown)


baszalmstra:
Please create 2 prs

ritankarsaha:
> Please create 2 prs

okaay sure, so I will be creating 2 separate issues for this and make 2 separate prs for the two issues properlyy

ritankarsaha:
@baszalmstra closing this PR and opening two separate PRs under two issues properly

PR #2173, opened 2026-03-05
feat: add `rattler_git` and `rattler_prefix_guard` crates

Hofer-Julian:
### Description

Add `rattler_git` and `rattler_prefix_guard` crates.
They were originally part of the `rattler-build` repo.

### How Has This Been Tested?

Ran `pixi run test` locally

### Checklist:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -198,7 +198,7 @@ rattler_cache = { path = "crates/rattler_cache", version = "=0.6.14", default-fe
 rattler_conda_types = { path = "crates/rattler_conda_types", version = "=0.43.5", default-features = false }
 rattler_config = { path = "crates/rattler_config", version = "=0.3.2", default-features = false }
 rattler_digest = { path = "crates/rattler_digest", version = "=1.2.2", default-features = false }
-rattler_git = { path = "crates/rattler_git", version = "0.1.0", default-features = false }
+rattler_git = { path = "crates/rattler_git", version = "=0.1.0", default-features = false }
 rattler_index = { path = "crates/rattler_index", version = "=0.27.16", default-features = false }
 rattler_libsolv_c = { path = "crates/rattler_libsolv_c", version = "=1.3.1", default-features = false }
 rattler_lock = { path = "crates/rattler_lock", version = "=0.26.16", default-features = false }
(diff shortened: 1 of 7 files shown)


baszalmstra:
- Is this the same as pixi_git?
- Do you think you could split into two PRs? Or are the dependent?

Hofer-Julian:
>  Do you think you could split into two PRs? Or are the dependent?

`rattler_git` depends on `rattler_prefix_guard`

> Is this the same as pixi_git?

They are at least *very* similar, both are forks of the same `uv` crate. I don't know the history of those two crates though, maybe @wolfv can chime in here. 

I assume we only want to use one of those two forks in our codebases in the end.

wolfv:
They are very similar. One thing I remember is that we added the ability to automatically pull submodules and `lfs` in `rattler` git that pixi git didn't / doesn't have.

Hofer-Julian:
> They are very similar. One thing I remember is that we added the ability to automatically pull submodules and `lfs` in `rattler` git that pixi git didn't / doesn't have.

I also let Claude compare it, and it only identified additional features on the `rattler_git` side. To me that sounds like we should merge and release this and attempt to migrate pixi to `rattler_git` next. If we find missing things, we can always migrate those as well.

crates/rattler_git/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+name = "rattler_git"
+version = "0.1.0"
+edition = "2024"
```
baszalmstra: Can we get this from the workspace too?

Cargo.toml
```diff
@@ -197,13 +198,15 @@ rattler_cache = { path = "crates/rattler_cache", version = "=0.6.14", default-fe
...
 rattler_digest = { path = "crates/rattler_digest", version = "=1.2.2", default-features = false }
+rattler_git = { path = "crates/rattler_git", version = "0.1.0", default-features = false }
 rattler_index = { path = "crates/rattler_index", version = "=0.27.16", default-features = false }
 rattler_libsolv_c = { path = "crates/rattler_libsolv_c", version = "=1.3.1", default-features = false }
 rattler_lock = { path = "crates/rattler_lock", version = "=0.26.16", default-features = false }
 rattler_macros = { path = "crates/rattler_macros", version = "=1.0.12", default-features = false }
 rattler_menuinst = { path = "crates/rattler_menuinst", version = "=0.2.49", default-features = false }
 rattler_networking = { path = "crates/rattler_networking", version = "=0.26.2", default-features = false }
 rattler_package_streaming = { path = "crates/rattler_package_streaming", version = "=0.24.2", default-features = false }
+rattler_prefix_guard = { path = "crates/rattler_prefix_guard", version = "0.1.0", default-features = false }
```
baszalmstra: I prefer to use 


```suggestion
rattler_prefix_guard = { path = "crates/rattler_prefix_guard", version = "=0.1.0", default-features = false }
```

Cargo.toml
```diff
@@ -197,13 +198,15 @@ rattler_cache = { path = "crates/rattler_cache", version = "=0.6.14", default-fe
 rattler_conda_types = { path = "crates/rattler_conda_types", version = "=0.43.5", default-features = false }
 rattler_config = { path = "crates/rattler_config", version = "=0.3.2", default-features = false }
 rattler_digest = { path = "crates/rattler_digest", version = "=1.2.2", default-features = false }
+rattler_git = { path = "crates/rattler_git", version = "0.1.0", default-features = false }
```
baszalmstra:
```suggestion
rattler_git = { path = "crates/rattler_git", version = "=0.1.0", default-features = false }
```

crates/rattler_prefix_guard/Cargo.toml
```diff
@@ -0,0 +1,25 @@
+[package]
+description = "A prefix guard (file lock) for rattler"
+edition = "2024"
```
baszalmstra: Can we get that from the workspace too?

PR #2184, opened 2026-03-06
fix: preserve channel priority for direct dependencies

### LINKED ISSUE #1928: Channel priority is not forced on direct dependencies
### Checklist

### What happened?

```toml
[workspace]
channels = ["https://prefix.dev/ruben-arts", "conda-forge"]
platforms = ["osx-arm64"]

[dependencies]
rattler-build = ">=0.53.0,<0.55"
```
Gets me `rattler-build` from `conda-forge`, but the channel priority tells pixi to get it from the `ruben-arts` channel. Which only has the `0.52.0` version.

```toml
[workspace]
channels = ["https://prefix.dev/ruben-arts", "conda-forge"]
platforms = ["osx-arm64"]

[dependencies]
rattler-build = ">=0.52.0,<0.55"
```
This gets it from my channel like expected.

```toml
[workspace]
channels = ["https://prefix.dev/ruben-arts", "conda-forge"]
channel-priority = "disabled"
platforms = ["osx-arm64"]

[dependencies]
rattler-build = ">=0.52.0,<0.55"
```
This gets me the latest version from `conda-forge` which also makes sense.

### Additional Context

I guess this broke with the repodata gateway changes, as I'm pretty sure this used to behave differently before.

mohitdebian:
Fixes #1928

## Problem

The repodata gateway filtered records for direct dependencies before they reached the solver. If a higher-priority channel contained the package but no version matching the spec, all records from that channel were removed. The solver then treated a lower-priority channel as the first channel containing the package.

## Solution

Modify `accumulate_records()` so that when no records match the `MatchSpec`, records from the channel are still included. This allows the solver to detect channel ownership and correctly enforce strict channel priority.

## Tests

Added regression test where:

- Channel A (higher priority) contains only older versions
- Channel B (lower priority) contains a newer version
- Spec requires the newer version

With strict channel priority the solve now fails instead of selecting the lower-priority channel.

### DIFF
--- a/crates/rattler_solve/tests/backends/main.rs
+++ b/crates/rattler_solve/tests/backends/main.rs
@@ -1388,27 +1388,48 @@ fn channel_priority_disabled_libsolv_c() {
 /// channel B and the solve should fail.
 #[test]
 fn channel_priority_strict_direct_dep_higher_channel_no_match() {
-    // Channel B: lower priority, has 2.0 (the only record passing the
+    use rattler_solve::resolvo::RepoData;
+
+    // Channel A (higher priority): has my-pkg but no version matching >=2.0.
+    // The gateway strips those records but reports the package name via
+    // RepoData.package_names so the solver can enforce strict priority.
+    let repo_data_a = RepoData {
+        records: vec![],
+        channel: Some("https://conda.anaconda.org/channel-a/".to_string()),
+        package_names: std::collections::HashSet::from(["my-pkg".parse().unwrap()]),
+    };
+
+    // Channel B (lower priority): has 2.0 (the only record passing the
     // gateway's version filter for spec ">=2.0").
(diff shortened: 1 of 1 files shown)


mohitdebian:
Thanks for pointing that out. I've updated the implementation so the matching is done in a single pass while keeping the same behavior when nothing matches

mohitdebian:
thanks for the clarification i've updated the implementation so the solver now knows which package names exist per channel instead of passing non matching records

baszalmstra:
This is not a good solution because it decouples the `channel_package_names` field from the `available_packages` field.

mohitdebian:
I've updated the implementation so package presence is now tracked directly in `RepoData` alongside the records instead of a separate field in `SolverTask` keeping it coupled with `available_packages`

PR #2185, opened 2026-03-07
fix: exclude no-timestamp packages in resolvo min_age dedup

Aman-Cool:
The deduplication loop in `CondaDependencyProvider::new()` disagrees with the solver exclusion loop on what to do with packages that have no timestamp when `include_unknown_timestamp` is `false`.

When a package exists as both `.conda` (no timestamp) and `.tar.bz2` (valid timestamp), the first loop sees the `.conda` has no timestamp, treats it as *not* excluded by age, and lets it win deduplication by archive preference — silently dropping the `.tar.bz2`. Then the second loop correctly excludes the `.conda` for having no timestamp. The solver is left with no valid candidate and fails, even though a perfectly good `.tar.bz2` was there the whole time.

The fix adds the missing match arm to the `excluded_by_age` computation so it mirrors the logic that already exists correctly in the second loop. Three lines.

This is resolvo-only — `libsolv_c` handles the `None` timestamp case correctly, so this also brings the two backends in sync.

Fixes #{issue}

### How Has This Been Tested?

Traced through both loops manually with a package that has a `.conda` variant with no timestamp and a `.tar.bz2` with a valid timestamp under a 7-day `min_age` config. The new match arm mirrors the existing correct logic in the second loop (lines 449–451) exactly, so the behavior is confirmed by symmetry. `include_unknown_timestamp: false` is the default, so this hits the standard path.

### AI Disclosure

Tools: Chatgpt, Claude

### Checklist:

### DIFF
--- a/crates/rattler_solve/src/resolvo/mod.rs
+++ b/crates/rattler_solve/src/resolvo/mod.rs
@@ -365,6 +365,11 @@ impl<'a> CondaDependencyProvider<'a> {
                             // Exclude if published after cutoff and not exempt
                             timestamp > cutoff && !config.is_exempt(&record.package_record.name)
                         }
+                        (Some(config), Some(_), None) => {
+                            // Exclude if no timestamp and unknown timestamps are not allowed
+                            !config.include_unknown_timestamp
+                                && !config.is_exempt(&record.package_record.name)
+                        }
                         _ => false,
                     };
 


Aman-Cool:
Hey @baszalmstra, just wanted to flag this for review when you get a chance.

PR #2187, opened 2026-03-07
feat: add custom progress reporter callbacks to installer

### LINKED ISSUE #2186: Python installer doesn't support custom progress callbacks
### Checklist

### Summary
  The install() function in py-rattler only supports a boolean show_progress flag which renders a
  built-in indicatif progress bar to the terminal. There is no way for Python callers to plug in
  their own progress handler — the source has a // TODO: Accept functions to report progress. 
  comment at py-rattler/src/installer.rs:18.

  ### Problem
  Python users building custom UIs, web dashboards, CLI tools, or test harnesses have no way to
  observe installation progress. They cannot:
  - Update a web front-end with per-package download progress
  - Write structured log output instead of a progress bar
  - Drive a custom TUI widget
  - Assert in tests that specific lifecycle events fired

  This is a significant usability gap compared to tools like pip or conda, which expose progress
  hooks.

  ### Expected behaviour
  Users should be able to subclass a reporter class, override only the callbacks they care about, and
   pass it to install().

@baszalmstra 

### Additional Context

_No response_

ritankarsaha:
### Description

Fixes thr TODO -  Accept functions to report progress in py-rattler/src/installer.rs.

  - PyReporter (Rust) — a new struct wrapping a PyObject that implements the full rattler::install::Reporter trait. Each of the 16 lifecycle methods acquires the GIL and calls the corresponding method on the Python object. Methods that return usize opaque tokens extract the Python return value (defaulting to 0).
  - InstallerReporter (Python) — a base class with all 16 callbacks as documented, no-op methods. Users subclass it and override only what they need.
  - install() updated — accepts a new optional reporter: InstallerReporter | None = None kwarg. When provided it is used instead of the built-in IndicatifReporter; show_progress is ignored (backwards-compatible).
  - Exports — InstallerReporter is re-exported from rattler.install and the top-level rattler package.

  ### Usage example

```python
  from rattler import install, InstallerReporter

  class MyReporter(InstallerReporter):
      def on_transaction_start(self, total_operations: int) -> None:
          print(f"Installing {total_operations} packages…")

      def on_download_progress(self, download_idx, progress, total):
          pct = f"{progress}/{total}" if total else str(progress)
          print(f"  [{download_idx}] {pct} bytes")

      def on_transaction_complete(self) -> None:
          print("Done!")

  await install(records, target_prefix="/my/env", reporter=MyReporter())
```

Fixes #2186 

### Checklist:

### DIFF
--- a/py-rattler/rattler/install/installer.py
+++ b/py-rattler/rattler/install/installer.py
@@ -1,6 +1,6 @@
 from __future__ import annotations
 import os
-from typing import List, Optional
+from typing import List, Optional, Protocol, runtime_checkable
 
 from rattler.match_spec import MatchSpec
 from rattler.networking.client import Client
@@ -11,13 +11,15 @@ from rattler.repo_data.record import RepoDataRecord
 from rattler.rattler import py_install
 
 
-class InstallerReporter:
+@runtime_checkable
+class InstallerReporter(Protocol):
     """
-    Base class for custom installation progress reporters.
+    Protocol for custom installation progress reporters.
 
-    Subclass this and override the methods we care about to receive progress
-    callbacks during package installation. All methods have no-op defaults so
-    we only need to implement the ones relevant to our use case.
+    Implement this protocol (by subclassing or structurally) and override the
+    methods we care about to receive progress callbacks during package
+    installation. All methods have no-op defaults so we only need to implement
+    the ones relevant to our use case.
 
     Methods that return an ``int`` act as opaque tokens: the value we return
     will be passed back to the corresponding ``*_complete`` callback so we can
@@ -28,7 +30,7 @@ class InstallerReporter:
     ```python
     from rattler.install import InstallerReporter
 
-    class MyReporter(InstallerReporter):
+    class MyReporter:  # no need to inherit — structural subtyping is sufficient
         def on_transaction_start(self, total_operations: int) -> None:
             print(f"Starting installation of {total_operations} operations")
 


baszalmstra:
Did you meaure the performance overhead?

ritankarsaha:
> Did you meaure the performance overhead?

 Nope, didn't  benchmark this. I think your concern is that `on_download_progress` fires once per downloaded chunk 
and each call acquires the GIL via Python::with_gil. On large environments with concurrent downloads this could stall Tokio worker threads while waiting for the GIL, degrading overall throughput ?                                             

I guess the straightforward mitigation would be to throttle `on_download_progress` at the Rust level (e.g. skip the
GIL acquisition if fewer than 100ms have elapsed since the last call for a given download index), which would reduce GIL pressure.

What do you think @baszalmstra ? Is this approach correct? Should I add it here or do in a follow up PR or try something else? 

@baszalmstra

baszalmstra:
That is indeed my concern, but I might be wrong so first measuring the impact seems like the logical choice

ritankarsaha:
> That is indeed my concern, but I might be wrong so first measuring the impact seems like the logical choice

got it, on it

ritankarsaha:
> That is indeed my concern, but I might be wrong so first measuring the impact seems like the logical choice


Hello @baszalmstra - 


### Here are the results:                                                                                         
                                                                                                                
  Phase A warm-cache (linking only)                                                                           
  - Baseline: 109.6 ms mean | Reporter: 110.1 ms mean                                                           
  - Overhead: +0.4% (+0.5 ms) — negligible                                                                      
                                                                                                                
  Phase B  cold-cache (download + linking)                                                                     
  - Baseline: 415.9 ms mean | Reporter: 344.4 ms mean                                                           
  - Overhead: −17.2% (−71.4 ms) — the reporter is actually faster, which is noise from network variance         
                                                                                                                
  ### Key observations:                                                                                             
  1. on_download_progress fired only 10 times for this small package (~500 KB). That's very low call volume.    
  2. The Phase B difference is within network jitter noise (stdev of ~30 ms vs 71 ms difference) — not a        
  meaningful signal.                                                                                            
  3. The GIL acquisition cost is not a problem for this workload. With only 10 download-progress calls, there's 
  nothing to throttle.                                                                                         
                                                                                                                
 According to me, the overhead is essentially zero for this package size. For very large packages that trigger hundreds or thousands of on_download_progress calls, it might be worth revisiting, but the current data shows no measurable GIL penalty.

baszalmstra:
Can you try with a larger package? :D `rust-src` or `pytorch` is pretty big.

ritankarsaha:
> Can you try with a larger package? :D `rust-src` or `pytorch` is pretty big.

Ahh sure, trying it out !!

ritankarsaha:
> Can you try with a larger package? :D `rust-src` or `pytorch` is pretty big.

  Here are the proper results with rust-src and 5 cold-cache iterations:                                        
                                                                                                                                                                                                                        
  ### Benchmark Results — rust-src (noarch, ~30 MB) — 5 iterations each                                             
                                                                                                                
  Phase A — Warm cache - linking only                                                                           
                                                                    
```                                            
  ┌─────────────────────────┬──────────┬────────┐           
  │         Variant         │   Mean   │ Stdev  │
  ├─────────────────────────┼──────────┼────────┤
  │ No reporter             │ 222.5 ms │ 4.6 ms │
  ├─────────────────────────┼──────────┼────────┤
  │ Counting reporter (GIL) │ 219.2 ms │ 2.7 ms │                                                               
  └─────────────────────────┴──────────┴────────┘
  ```                                                                                                              
  Overhead: −1.5% (−3.3 ms) — noise, effectively zero. 10 callbacks total.                                      
  
                                                                                                      
  Phase B — Cold cache (download + linking)                 

```
  ┌─────────────────────────┬───────────┬─────────┐
  │         Variant         │   Mean    │  Stdev  │
  ├─────────────────────────┼───────────┼─────────┤
  │ No reporter             │ 1195.3 ms │ 22.8 ms │
  ├─────────────────────────┼───────────┼─────────┤
  │ Counting reporter (GIL) │ 1195.4 ms │ 45.9 ms │                                                             
  └─────────────────────────┴───────────┴─────────┘                                                             
   ```                                                                                                             
  Overhead: +0.0% (+0.1 ms) — statistically indistinguishable. on_download_progress fired 563 times for ~30 MB. 
                                                                                                                                                                    
  ### Conclusion from my end -                            
                                                                                                                
With rust-src, there is zero measurable overhead from the reporter callbacks. 5 iterations bring the stdev down enough (~22 ms) that the +0.1 ms difference is solidly within noise.  The concern about GIL overhead appears unfounded for rust-src. Couldn't test with the pytorch package, away from my linux now !! The numpy+scipy result (8,372 callbacks, +9.1%) is the only case where overhead was measurable, but that's a different package and 15× more callbacks. 
I don't think we need any more modifications, but what do you suggest should I put on a throttling to make it more robutst ?

Should I also share the reporter.py file, I am using to check ? 
@baszalmstra

py-rattler/rattler/install/installer.py
```diff
@@ -11,6 +11,268 @@
 from rattler.rattler import py_install
 
 
+class InstallerReporter:
```
baszalmstra: Lets make this a protocol

```suggestion
@runtime_checkable
class InstallerReporter(Protocol):
```

ritankarsaha:
> I only have one more comment! 
> 
> Maybe in a follow-up you could add an implementation that uses rich? And add that as an optional dependency? Then we can remove/deprecate the `show_progress` argument.

Okaay, on it @baszalmstra

py-rattler/rattler/install/installer.py
ritankarsaha: resolved @baszalmstra !

In a follow up PR working on the next part, will link the PR here !

ritankarsaha:
@baszalmstra fixed the reviews, will follow up with the suggestion in a second PR, also updated branch form main

Is this good to go, or any more modifications required @baszalmstra ?
Lemme know whenever you get a chance, thanks !

baszalmstra (approved):

baszalmstra:
Thanks!

PR #2188, opened 2026-03-07
fix: propagate error in sort_paths instead of panicking

suhr25:
# Description

`sort_paths` previously used `.unwrap()` on `Path::strip_prefix`, which could panic if a path was not under `base_path`. This change replaces the panic with proper error propagation by returning an `io::Error`. The two call sites now propagate the error using `?`, aligning the implementation with the documented behavior of the package writing APIs.

# Fixes

* Replace `.unwrap()` on `strip_prefix` with proper error handling
* Update `sort_paths` to return `Result<(Vec<PathBuf>, Vec<PathBuf>), io::Error>`
* Propagate the error at both call sites using `?`
* Provide a clearer error message when a path is outside `base_path`

# How Has This Been Tested?

Tested by passing a path that is not under the provided base_path. Before the fix the code panicked due to .unwrap(). After the fix it correctly returns an io::Error.

# AI Disclosure

* [x]  This PR contains AI-generated content.
* [x]  I have tested any AI-generated content in my PR.
* [x]  I take responsibility for any AI-generated content in my PR.

Tools: Chatgpt, Claude

# Checklist

* [x]  Code compiles successfully
* [x] Tests pass
* [x] No breaking changes
* [x] Error handling follows existing project patterns

### DIFF
--- a/pixi.lock
+++ b/pixi.lock
@@ -7,237 +7,228 @@ environments:
       pypi-prerelease-mode: if-necessary-or-explicit
     packages:
       linux-64:
-      - conda: https://prefix.dev/conda-forge/linux-64/_openmp_mutex-4.5-20_gnu.conda
-      - conda: https://prefix.dev/conda-forge/linux-64/binutils-2.45.1-default_h4852527_101.conda
-      - conda: https://prefix.dev/conda-forge/linux-64/binutils_impl_linux-64-2.45.1-default_hfdba357_101.conda
-      - conda: https://prefix.dev/conda-forge/linux-64/binutils_linux-64-2.45.1-default_h4852527_101.conda
-      - conda: https://prefix.dev/conda-forge/linux-64/bzip2-1.0.8-hda65f42_9.conda
-      - conda: https://prefix.dev/conda-forge/linux-64/c-ares-1.34.6-hb03c661_0.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/_libgcc_mutex-0.1-conda_forge.tar.bz2
+      - conda: https://prefix.dev/conda-forge/linux-64/_openmp_mutex-4.5-2_gnu.tar.bz2
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils-2.44-h4852527_5.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils_impl_linux-64-2.44-h9d8b0ac_5.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/binutils_linux-64-2.44-h4852527_5.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/bzip2-1.0.8-hda65f42_8.conda
+      - conda: https://prefix.dev/conda-forge/linux-64/c-ares-1.34.5-hb9d3cd8_0.conda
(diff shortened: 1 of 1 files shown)


suhr25:
Hi @baszalmstra,
Please take a look at this PR.
Thanks!

zelosleone:
Just wanted to left a comment since the pr interested me, but using a loop by replacing an iterator is not a correct approach in my opinion. You could just use collect and still use the simple iterator.

crates/rattler_package_streaming/src/write.rs
```diff
@@ -70,18 +70,34 @@ impl Read for ProgressBarReader {
...
+                io::ErrorKind::InvalidInput,
+                format!(
+                    "path '{}' is not under base path '{}'",
+                    p.display(),
+                    base_path.display()
+                ),
+            )
+        })?;
+        relative_paths.push(rel.to_path_buf());
+    }
```
baszalmstra: Introducing a for loop is not better.

suhr25: @baszalmstra 
I have updated the implementation to use an iterator based approach instead of a for loop.
Thanks!

crates/rattler_package_streaming/src/tokio/async_read.rs
```diff
@@ -6,9 +6,10 @@ use std::path::Path;
 use async_compression::tokio::bufread::BzDecoder;
 use async_spooled_tempfile::SpooledTempFile;
 use async_zip::base::read::stream::ZipFileReader;
-use futures_util::StreamExt;
-use tokio::io::{AsyncRead, AsyncReadExt, AsyncSeekExt};
+use tokio::io::{AsyncRead, AsyncSeekExt};
 use tokio_util::compat::{FuturesAsyncReadCompatExt, TokioAsyncReadCompatExt};
+#[cfg(feature = "reqwest")]
+use {futures_util::StreamExt as _, tokio::io::AsyncReadExt};
```
zelosleone, line 12: I wanted to wait until weekend to comment on this so that i can fully check (was kinda curious from the first time i saw the pr), but i am sure you had to change this because of a normal cargo check - cargo build without all features. rattler github actions build with reqwest flag enabled so it wont ever be needed to have this. it would be just a normal warning ONLY with cargo check while locally developing, i think it would be better to include this either in message or as a seperate pr. but up to maintainers.

suhr25: @zelosleone 
Good point. I added the cfg guard because `cargo check` without features locally was producing a warning due to the unused import.
Since CI enables the `reqwest` feature this isn't strictly required there, but I thought it would be cleaner to avoid the warning for local builds as well.
Happy to move it to a separate PR if preferred.

zelosleone:
I am not a maintainer so you don't have to ping me to review, i wanted to simply see if i can help Bas with a review to shorten his time period on reviews. My advice however;

- Convey every change clearly before reviewers asks about it, this way it would shorten the back and forth exchanges thus quicker reviewer times.
- Add tests, ideally e2e but on this, a simple unit test would have been enough. I assume rattler didn't have a bad way test in its tests for this, so it was not caught in the first place. I am unsure if proper error collection for this requires a test, but i would still try to test to see if i missed another call somewhere and to give trust to reviewer/maintainer that "this pr works!"

Other than that, i dont see anything else to change.

PR #2191, opened 2026-03-07
fix: preserve conda-meta key order when updating requested_specs

Aman-Cool:
### Description

`update_requested_specs_in_json` was round-tripping conda-meta JSON through `serde_json::Value`, which uses `BTreeMap` internally (no `preserve_order` feature in this crate), so every rewrite came back with alphabetically-sorted keys.

That's a problem because `MinimalPrefixRecord`'s parser stops early once it has seen both `files` and `requested_specs`. After the rewrite, `sha256` (`'s'`) ends up after `requested_specs` (`'r'`) alphabetically and is never parsed. On the next `pixi install`, `MinimalPrefixRecord.sha256` is `None` while `RepoDataRecord.sha256` is `Some` — `describe_same_content` sees a mismatch and forces a full `PrefixRecord` reload, permanently defeating the fast-path optimization. This repeats on every subsequent run.

**Before** (alphabetical key order after rewrite):
```
..., files, ..., requested_specs, sha256, ...
                                  ^^^^^^ parser already stopped here
```

**After** (struct-field order preserved):
```
..., sha256, ..., files, ..., requested_specs
     ^^^^^^ read correctly before early termination
```

Fix: replace the `serde_json::Value` round-trip with `PrefixRecord::from_path` + `write_to_path`. Serde serializes in struct-field order so key ordering is stable, and `write_to_path` uses an atomic temp-file rename — also fixing a secondary risk of file corruption on process kill.

Fixes #NA

### How Has This Been Tested?

Manually verified by:
1. Running `pixi install` on an environment with numpy/requests
2. Confirming the rewritten conda-meta JSONs now preserve original key order (`sha256` appears before `requested_specs`)
3. Running `pixi install` a second time and confirming no full `PrefixRecord` reload is triggered (fast path holds)

### AI Disclosure

Tools: Chatgpt ,Claude

### Checklist:

### DIFF
--- a/crates/rattler/src/install/installer/mod.rs
+++ b/crates/rattler/src/install/installer/mod.rs
@@ -819,55 +819,20 @@ async fn populate_cache(
         .map_err(|e| InstallerError::FailedToFetch(record.identifier.to_string(), e))
 }
 
-/// Updates only the `requested_specs` fields in a conda-meta JSON file.
-/// This performs a targeted update without overwriting other
-/// metadata.
-///
-/// This method is needed as we're initially loading
-/// `MinimalPrefixRecord`, which doesn't contain most of the fields.
-/// Therefore direct writing could overwrite data we want to preserve.
-///
-/// Currently we're loading full json, but we could do that in place without
-/// parsing whole file.
+/// Updates the `requested_specs` fields in a conda-meta JSON file by reading
+/// the full `PrefixRecord` and writing it back atomically. This preserves all
+/// other fields and their original key order, and avoids partial-write
+/// corruption on process interruption.
+#[allow(deprecated)]
 fn update_requested_specs_in_json(
     path: &Path,
     requested_specs: &[String],
     requested_spec: Option<&String>,
 ) -> io::Result<()> {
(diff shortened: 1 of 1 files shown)


baszalmstra:
What is the downside of adding the `preserve_order` feature?

Aman-Cool:
@baszalmstra, I did consider that😄.

The main issue is `preserve_order` uses insertion order, so for files that don't already have `requested_specs`, it'd still get appended at the end (same problem). It also pulls in `indexmap` as a dep and since `serde_json` is workspace-level, it'd silently flip behavior for every crate using it, felt like too broad a side effect for a targeted fix.

Going through the actual `PrefixRecord` struct felt cleaner to me, serde handles the field order automatically and we also get the atomic write from `write_to_path` for free, which the old approach was missing. Fewer moving parts overall.

That said, happy to hear your thoughts if you think `preserve_order` is still the better path here!🙂

baszalmstra:
Im a bit worried that the overhead of going through PrefixRecord is much slower. But that may be completely unfounded. Could you maybe benchmark that?

PR #2227, opened 2026-03-15
feat: add FreeBSD 32-bit and ARM64 platform support

wolfv:
### Description

This PR adds support for two additional FreeBSD platforms:
- **FreeBsd32**: 32-bit x86 FreeBSD
- **FreeBsdArm64**: ARM64 (aarch64) FreeBSD

These platforms are now recognized and properly handled across the codebase, including:
- Platform enum variants and string representations
- Architecture detection for each platform
- Runtime platform detection on FreeBSD systems
- Language bindings (TypeScript, Python, JavaScript)

The changes ensure that FreeBSD users on 32-bit and ARM64 systems can properly specify their target platform when building or installing packages.

### How Has This Been Tested?

- Added unit tests for parsing "freebsd-32" and "freebsd-arm64" platform strings
- Added unit tests for architecture detection of the new platforms
- Existing platform tests continue to pass
- The platform count in Python bindings updated from 23 to 25 platforms

### Checklist:

https://claude.ai/code/session_013KABaapbKqW9YpuZRENAwJ


wolfv:
Please take a look @yurivict

yurivict:
Thanks!

I wouldn't be able to test here because these breakages occur during the package build on remote central servers.

But please merge it, and than make a release. Then when mise would adopt this release I would be able to see whether these breakages are fixed.

baszalmstra:
Should this be freebsd-aarch64 like linux-aarch64 ?

baszalmstra:
What does conda/mamba do?

yurivict:
> Should this be freebsd-aarch64 like linux-aarch64 ?

On FreeBSD Rust and LLVM use the term aarch64 in their target specifications.

wolfv:
I checked conda and could only find the `freebsd-64` spec.

chenghlee:
conda basically uses whatever the CPython interpreter reports for `platform.machine().lower()` (modulo that weird `32`/`64` mapping it does for x86/x86-64 CPUs).

I don't have a FreeBSD on ARM system handy, but based on the install ISO names, I suspect it will be `aarch64`. But we should have someone run that command on such a system to confirm.

wolfv:
I double checked with Claude and it looks like `uname -m` returns `arm64`: https://raw.githubusercontent.com/freebsd/freebsd-src/main/sys/arm64/include/param.h

baszalmstra:
is uname used for platform.machine().lower() ?

wolfv:
That is what Claude has told me, yeah.

ifitchet:
Why not both?

```
$uname -a
FreeBSD freebsd-13-arm64.office.soho 14.3-RELEASE FreeBSD 14.3-RELEASE releng/14.3-n271432-8c9ce319fef7 GENERIC arm64
```

and then

```
$python3.9 -m platform -c 'print(platform.machine().lower())'
FreeBSD-14.3-RELEASE-arm64-aarch64-64bit-ELF
```

this is a VMware FreeBSD install on an Apple M1 FWIW.

wolfv:
@ifitchet can you try what python reports on that VM?

ifitchet:
> @ifitchet can you try what python reports on that VM?

@wolfv Any particular Python incantation?

wolfv:
Ah sorry, just noticed that you already printed the output!

PR #2229, opened 2026-03-15
fix(rattler_repodata_gateway): apply all matching pattern specs, not just the first

### LINKED ISSUE #2228: Multi-pattern gateway queries silently drop specs after the first matching pattern
### Checklist

### Summary

QueryExecutor::expand_pattern_specs_for_subdir in crates/rattler_repodata_gateway/src/gateway/query.rs iterates all package names exposed by a resolved subdir and matches each one against pending_pattern_specs. The inner loop breaks as soon as the first pattern matches a name, silently discarding every subsequent pattern that also matches it.

Because accumulate_records filters via specs.iter().any(|s| s.matches(record)), only the first pattern's constraint is ever applied. Records that satisfy a later pattern's constraints (but not the first's) are excluded with no error or warning.

### Root cause

  Two compounding defects in the inner loop:

  1. break after first match, this stops the inner loop as soon as one pattern matches, never evaluating subsequent
  patterns.
  2. seen.insert() inside the inner loop, even if the break were removed, the second matching pattern would
  find the name already in seen (inserted by the first match) and the is_none() guard would prevent its spec
  from being added.

SourceSpecs::Input is a Vec<MatchSpec> specifically designed to hold multiple specs per name, but the original code prevented more than one from ever being added via pattern expansion.

@baszalmstra 

### Additional Context

_No response_

ritankarsaha:
### Description

- Fixed expand_pattern_specs_for_subdir silently dropping all but the first pattern spec for any package name is
  matched by multiple glob/regex patterns     
 - Added regression test proving all three python versions are returned when queried with two overlapping        
  patterns  

Fixes #2228 

### How Has This Been Tested?

test_multi_pattern_all_specs_applied — a MockRepoDataSource carrying python 2.7.18, 3.8.20, and 3.10.14 is queried with two overlapping patterns (py* >=3.0 and python* <3.9). The test asserts all three versions are returned. Before the fix only 3.8.20 and 3.10.14 are returned, making the regression immediately visible.

### Checklist:

@baszalmstra

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/mod.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/mod.rs
@@ -1716,4 +1716,73 @@ mod test {
             "glob should find all lib-* packages across both sources and both subdirs"
         );
     }
+
+    /// Regression test: when two patterns both match the same package name,
+    /// ALL matching patterns' specs must be collected, not just the first one.
+    #[tokio::test]
+    async fn test_multi_pattern_all_specs_applied() {
+        use rattler_conda_types::ParseStrictnessWithNameMatcher;
+
+        let gateway = Gateway::new();
+
+        // Source has three python records at different versions.
+        let mut source = MockRepoDataSource::new();
+        source.add_record(
+            Platform::Linux64,
+            make_test_record("python", "2.7.18", "linux-64"),
+        );
+        source.add_record(
+            Platform::Linux64,
+            make_test_record("python", "3.8.20", "linux-64"),
+        );
+        source.add_record(
+            Platform::Linux64,
+            make_test_record("python", "3.10.14", "linux-64"),
+        );
+
+        let spec1 = MatchSpec::from_str(
+            "py* >=3.0",
+            ParseStrictnessWithNameMatcher {
+                parse_strictness: Strict,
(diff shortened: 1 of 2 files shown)


ritankarsaha:
@baszalmstra is the implementation correct? Do you need anything else on this ?

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/tokio/index.rs
```diff
@@ -255,6 +256,16 @@ pub async fn fetch_index(
...
+                        if response.status() == StatusCode::INTERNAL_SERVER_ERROR {
+                            tracing::debug!(
+                                "sharded index returned 500 at {channel_base_url}, treating as not found"
+                            );
+                            if let Some((reporter, index)) = download_reporter {
+                                reporter.on_download_complete(response.url(), index);
+                            }
+                            return Err(create_subdir_not_found_error(channel_base_url));
+                        }
+
```
baszalmstra: Why did you add this? If the server returned a 500 we should no silently ignore it.

ritankarsaha:
@baszalmstra reviews resolved and merged from main, ready to go in !

PR #2240, opened 2026-03-17
fix: use cmake to build aws-lc-sys to avoid conda CFLAGS conflict

benmoss:
### Description

After the release in #2142 bumped `aws-lc-sys` from 0.37.1 to 0.38.0, builds started failing because the new version compiles `jitterentropy-base.c` which requires `-O0`, but conda's `CFLAGS` inject `-O2`, causing a build failure.

This is the same fix applied in [rattler-build#2331](https://github.com/prefix-dev/rattler-build/pull/2331): setting `AWS_LC_SYS_CMAKE_BUILDER=1` tells `aws-lc-sys` to use CMake (already a pixi dependency) instead of `cc-rs`, which properly handles the conflicting flags.

### How Has This Been Tested?

- `pixi run build` succeeds locally on macOS (previously failed with the same error)
- Same fix is proven in rattler-build

### AI Disclosure
Tools: Claude Code (Claude Opus 4.6)

### Checklist:

### DIFF
--- a/pixi.toml
+++ b/pixi.toml
@@ -41,6 +41,11 @@ rust-src = "~=1.93.1"
 clang = ">=18.1.8,<19.0"
 compilers = ">=1.6.0"
 mold = ">=2.33.0,<3.0"
+# Use cmake to build aws-lc-sys, because conda's CFLAGS inject -O2 which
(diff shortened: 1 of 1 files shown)


wolfv:
Thanks Ben!

PR #2254, opened 2026-03-19
fix: logic bug in collect_env_vars causing spurious warnings

### LINKED ISSUE #2253: Bug: Spurious "already defined" warning in collect_env_vars due to always-true contains_key check
## Description

In `rattler_shell/src/activation.rs`, the function `collect_env_vars` contains a logic bug that causes a spurious warning to be emitted for **every** environment variable read from the `conda-meta/state` file, regardless of whether a conflict actually exists.

## Location

**File:** `crates/rattler_shell/src/activation.rs`
**Lines:** ~274

maniktyagi04:
## Description

This PR fixes a logic bug in `rattler_shell/src/activation.rs` where the `collect_env_vars` function incorrectly checks for key existence in the `state_env_vars` map while iterating over it, resulting in a spurious "already defined" warning for every variable in the `conda-meta/state` file.

The check has been updated to compare against `env_vars` (the accumulated map populated from `env_vars.d` files), ensuring warnings only fire when a real conflict exists.

## Changes
- Updated `contains_key` logic in `crates/rattler_shell/src/activation.rs`.
- Added a regression test `test_collect_env_vars_no_spurious_conflict_warnings` to verify the fix.

## Related Issues
Fixes #2253

### DIFF
--- a/crates/rattler_shell/src/activation.rs
+++ b/crates/rattler_shell/src/activation.rs
@@ -860,7 +860,7 @@ mod tests {
         }
     }
 
-    /// Regression test for: https://github.com/conda/rattler/issues/2253
+    /// Regression test for: <https://github.com/conda/rattler/issues/2253>
     ///
     /// `collect_env_vars` used to check `state_env_vars.contains_key(key)` while
     /// iterating `state_env_vars` — always true, so a spurious "already defined" warning


baszalmstra:
Please fix CI

maniktyagi04:
@baszalmstra 
Thanks for the feedback.

I've fixed the CI issues and resolved the failing test cases. The changes have been validated locally, and all checks are now passing.

Let me know if you’d like any further adjustments or improvements.

PR #2257, opened 2026-03-19
fix(docs): split workflow into rust and py-rattler jobs, fix tag matching

baszalmstra:
## Summary

The docs workflow had two problems: the `latest` alias for py-rattler docs was never deployed (404 at `rattler.prefix.dev/py-rattler/latest/`), and the workflow structure made it easy for jobs to race on gh-pages pushes.

### Root cause of the 404
The old workflow triggered release docs on `v*` tags, but py-rattler releases are tagged as `py-rattler-v*`. The release job never matched.

### Why not a separate tag-triggered workflow?
Tags created by `GITHUB_TOKEN` in a workflow [cannot trigger other workflows](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow). Since `release-python.yml` creates the `py-rattler-v*` tag, a separate docs workflow listening for that tag would never fire. Instead, the release docs deploy is added directly to `release-python.yml`.

### Changes

**`docs.yaml`** — split into independent jobs, removed release doc deployment:
  - `build-docs` — PR build check only
  - `rust-docs` — deploys cargo doc to gh-pages (push to main only)
  - `py-rattler-docs-dev` — deploys Mike `dev` alias (push to main / manual dispatch without tag)
  - `py-rattler-docs-release` — deploys Mike `latest` alias (manual dispatch with tag, as fallback)

**`release-python.yml`** — added `deploy-release-docs` job:
  - Runs after `tag-release`, deploys Mike `latest` alias automatically on every release
  - This is the primary automated path for release docs

**Other improvements:**
- All gh-pages-pushing jobs share `concurrency: { group: gh-pages-deploy, cancel-in-progress: false }` to prevent races
- Each job owns its redirect: `rust-docs` writes root `index.html`, release jobs write `py-rattler/index.html`
- py-rattler jobs no longer build rust docs — no cargo/rustc needed
- Added `py-rattler/rattler/**` to paths filter so Python source/docstring changes trigger doc rebuilds
- Fixed shell injection: workflow_dispatch tag input passed via `env:` instead of direct `${{ }}` interpolation

## After merging

Manually trigger `docs.yaml` (**Actions > Deploy Docs > Run workflow**) with the latest release tag (e.g. `py-rattler-v0.34.0`) to populate the `latest` alias and fix the 404 at `rattler.prefix.dev/py-rattler/latest/`.

## Test plan

- [ ] Verify PR check (`build-docs`) runs on this PR
- [ ] After merge, verify `rust-docs` and `py-rattler-docs-dev` both run on push to main
- [ ] Manually trigger `docs.yaml` with latest release tag to deploy `latest`
- [ ] Verify `rattler.prefix.dev/py-rattler/latest/` resolves
- [ ] On next py-rattler release, verify `deploy-release-docs` runs in `release-python.yml`

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -4,12 +4,11 @@ on:
   push:
     branches:
       - main
-    tags:
-      - "v*"
     paths:
       - "py-rattler/docs/**"
       - "py-rattler/mkdocs.yml"
       - "py-rattler/pixi.toml"
+      - "py-rattler/rattler/**"
       - "crates/**"
       - "Cargo.toml"
       - ".github/workflows/docs.yaml"
@@ -20,11 +19,12 @@ on:
       - "py-rattler/docs/**"
       - "py-rattler/mkdocs.yml"
       - "py-rattler/pixi.toml"
+      - "py-rattler/rattler/**"
       - ".github/workflows/docs.yaml"
   workflow_dispatch:
     inputs:
       tag:
-        description: "Tag to deploy (e.g. v0.23.0), or leave empty for dev"
+        description: "Tag to deploy py-rattler docs (e.g. py-rattler-v0.23.0), or leave empty for dev"
         required: false
         default: ""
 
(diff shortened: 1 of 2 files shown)


.github/workflows/docs-release.yaml
```diff
@@ -0,0 +1,55 @@
+name: Deploy Docs (Release)
+
+on:
+  push:
+    tags:
+      - "py-rattler-v*"
```
pavelzw: https://github.com/conda/rattler/actions/workflows/release-python.yml will create a tag using a github actions scoped github token. these tokens cannot trigger other workflows (to avoid recursive workflow runs: https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow#triggering-a-workflow-from-a-workflow)

probably putting it into the same workflow as python release is better

PR #2263, opened 2026-03-19
feat: Add CLI run command

danyeaw:
### Description

Draft version of a new CLI command worked on by @soapy1, @agriyakhetarpal, and me.

### AI Disclosure

Tools: Composer 2 (Cursor). Mostly coded in training, then used some help to clean up the remainder.

### Checklist:

### DIFF
--- a/crates/rattler-bin/src/commands/list.rs
+++ b/crates/rattler-bin/src/commands/list.rs
@@ -1,0 +1,95 @@
+use std::{env, path::PathBuf};
+
+use miette::IntoDiagnostic;
+use rattler_conda_types::{HasArtifactIdentificationRefs, PackageName, PrefixData};
+
+/// Search for packages in conda channels using glob or regex patterns.
+#[derive(Debug, clap::Parser)]
+#[clap(after_help = r#"Examples:
+  rattler list -p /path/to/environment"#)]
+pub struct Opt {
+    /// The prefix to list
+    #[clap(short, long)]
+    prefix: Option<PathBuf>,
+
+    /// The name (or glob) of the packages to list
+    name: Option<PackageName>, // maybe this could be a full MatchSpec?
+
+    /// Match full names only
+    #[clap(short, long)]
+    full_name: bool,
+}
+
+pub async fn list(opt: Opt) -> miette::Result<()> {
+    let prefix = if let Some(prefix) = opt.prefix {
+        prefix
+    } else if let Ok(prefix) = env::var("CONDA_PREFIX") {
+        PathBuf::from(prefix)
+    } else {
+        miette::bail!("No environment detected or passed. Tip: Use -p PATH.")
+    };
+
+    let prefix_data = PrefixData::new(&prefix).into_diagnostic()?;
+    let header = [[
+        "# Name".to_string(),
+        "Version".to_string(),
+        "Build".to_string(),
+        "Channel".to_string(),
+    ]];
+    // These initial widths match the header columns length
+    let mut widths: [usize; 4] = header[0].clone().map(|x| x.len());
+    let mut lines = vec![];
+    for record in prefix_data.iter() {
+        if let Some(Ok(record)) = record {
+            let name = record.name().as_normalized();
+            if let Some(query) = &opt.name {
+                let normalized_query = query.as_normalized();
+                if opt.full_name {
+                    if normalized_query != name {
+                        continue;
+                    }
+                } else if !name.contains(normalized_query) {
+                    continue;
+                }
+            };
+
+            let fields = [
+                name.to_string(),
+                record.version().as_str().to_string(),
+                record.build().to_string(),
+                record.repodata_record.channel.clone().unwrap_or_default(),
+            ];
+            for (i, (field, width)) in fields.iter().zip(widths).enumerate() {
+                let field_len = field.len();
+                if field_len > width {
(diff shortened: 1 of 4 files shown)


danyeaw:
Hey @tdejager and @baszalmstra, some questions we had:

1. `let mut cmd = Command::new(&command[0]);` would panic if command is empty, which is guarded from CLI usage but not library usage. Do we need to protect against library usage?
2. `The run_command_in_environment` API takes `env_vars`, but we have no `--env` flag. Do you think we need one?
3. Our main function is async, I updated the other functions to be async as well. Is this idiomatic or should we just keep everything sync?

crates/rattler-bin/src/commands/run.rs
```diff
@@ -0,0 +1,45 @@
+use miette::IntoDiagnostic;
+use rattler_shell;
+use std::{env, path::PathBuf};
+
+/// Run a command in a workspace
```
pavelzw:
```suggestion
/// Run a command in an activated conda environment.
```

the notion of workspace is something different in pixi land

crates/rattler_shell/Cargo.toml
```diff
@@ -24,7 +24,7 @@ sysinfo = { workspace = true, optional = true }
 tempfile = { workspace = true }
 thiserror = { workspace = true }
 tracing = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "rt", "process"] }
```
baszalmstra: Do we also need the macros here?

crates/rattler_shell/src/run/mod.rs
```diff
@@ -24,6 +24,49 @@ pub enum RunError {
...
+    };
+
+    // Run activation scripts
+
+    let activated_env =
+        tokio::task::spawn_blocking(move || activator.run_activation(activation_vars, None))
+            .await
+            .expect("Activated environment panicked")?;
+
+    let mut cmd = tokio::process::Command::new(&command[0]);
```
baszalmstra: I believe you can use (from the top of my head):

```rust
let [cmd, args @ ..] = &commands[..] else {
   // Return error
}
```

baszalmstra:
> The run_command_in_environment API takes env_vars, but we have no --env flag. Do you think we need one?

Might be nice, not required.

> Our main function is async, I updated the other functions to be async as well. Is this idiomatic or should we just keep everything sync?

The library function is doing async work, so your approach is valid!

crates/rattler_shell/Cargo.toml
danyeaw: Thanks, updated to remove the macros.

crates/rattler_shell/src/run/mod.rs
```diff
@@ -24,6 +27,53 @@ pub enum RunError {
...
+
+    let activation_vars = ActivationVariables {
+        conda_prefix,
+        path: current_path,
+        path_modification_behavior: PathModificationBehavior::default(),
+        // Full process environment. `run_in_environment` uses only the caller `env_vars` map.
+        current_env: std::env::vars().collect(),
+    };
+
+    // `run_activation` is blocking. `spawn_blocking` runs it on Tokio's blocking pool, not async workers.
```
tdejager:
```suggestion
```

crates/rattler-bin/src/commands/run.rs
```diff
@@ -0,0 +1,47 @@
+use miette::IntoDiagnostic;
+use rattler_shell;
+use std::{env, path::PathBuf};
+
+/// Run a command in an activated conda environment.
+#[derive(Debug, clap::Parser)]
+pub struct Opt {
+    /// Environment prefix to activate (defaults to `.prefix` under the current directory)
+    #[clap(long)]
+    target_prefix: Option<PathBuf>,
```
tdejager: Is this the default, I think `create` doesn't have one specifically, might be good to align :)

tdejager: You could also try setting it as rust clap default: https://docs.rs/clap/latest/clap/#example

crates/rattler_shell/src/run/mod.rs
```diff
@@ -14,6 +14,9 @@ use crate::{
 #[allow(missing_docs)]
 #[derive(Debug, thiserror::Error)]
 pub enum RunError {
+    #[error("Invalid command: {0:?}")]
```
tdejager: Maybe use a nicer way of displaying instead of the debug print :)

crates/rattler-bin/src/commands/run.rs
danyeaw: Thanks @tdejager for the feedback! We changed the logic to match what was in the create command.

danyeaw: Also a future PR could be to standardize all the `--target-prefix` options across the commands

crates/rattler_shell/src/run/mod.rs
danyeaw: We realized that this invalid command error isn't needed because clap already catches the error if no command is provided. So we removed this type of `RunError`.

agriyakhetarpal:
I'll be happy to take up the task of standardising the `--target-prefix`option for all commands, now that this PR is merged!

tdejager:
Please do!

PR #2276, opened 2026-03-23
refactor: Split PypiPackageData inte separate structs for Source and Distribution dependencies

hunger:
### Description

This needs to be put on top of push-mkmntstvoono . This PR should actually be the last change only :-)

Refactores PypiPackageData into a Source and a Wheel, so that we can handle either better.

### Checklist:

### DIFF
--- a/crates/rattler_lock/src/parse/deserialize.rs
+++ b/crates/rattler_lock/src/parse/deserialize.rs
@@ -365,12 +365,7 @@ pub fn parse_from_document_v7(
 /// rather than a mutable source directory.
 fn is_wheel_or_archive_location(location: &UrlOrPath) -> bool {
     let archives = [
-        ".whl",
-        ".tar.gz",
-        ".tar.bz2",
-        ".tar.zstd",
-        ".tar.zs",
-        ".zip",
+        ".whl", ".tar.gz", ".tar.bz2", ".tar.xz", ".tar.Z", ".tgz", ".tbz", ".tar", ".zip",
     ];
     match location {
         UrlOrPath::Url(url) => {
@@ -414,7 +409,9 @@ fn convert_raw_pypi_package(
     // absent → Source.  For older formats we fall back to the location
     // heuristic.
     let is_distribution = if file_version >= FileFormatVersion::V7 {
-        raw_package.version.is_some()
+        raw_package.hash.is_some()
+            || raw_package.version.is_some()
+            || raw_package.index_url.is_some()
     } else {
         is_wheel_or_archive_location(location.inner())
     };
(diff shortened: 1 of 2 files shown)


crates/rattler_lock/src/parse/deserialize.rs
```diff
@@ -355,6 +361,30 @@ pub fn parse_from_document_v7(
...
+/// Returns `true` if the location refers to an immutable wheel artifact
+/// rather than a mutable source directory.
+fn is_wheel_or_archive_location(location: &UrlOrPath) -> bool {
+    let archives = [
+        ".whl",
+        ".tar.gz",
+        ".tar.bz2",
+        ".tar.zstd",
+        ".tar.zs",
+        ".zip",
```
baszalmstra:
```suggestion
        ".whl",
        ".tar.gz",
        ".tar.bz2",
        ".tar.xz",
        ".tar.Z",
        ".tgz",
        ".tbz",
        ".tar",
        ".zip",
```

crates/rattler_lock/src/parse/deserialize.rs
```diff
@@ -380,15 +410,37 @@ fn convert_raw_pypi_package(
...
-        requires_python: raw_package.requires_python,
-    })
+    // In v7+ the version field is authoritative: present → Distribution,
+    // absent → Source.  For older formats we fall back to the location
+    // heuristic.
+    let is_distribution = if file_version >= FileFormatVersion::V7 {
+        raw_package.version.is_some()
+    } else {
+        is_wheel_or_archive_location(location.inner())
+    };
```
baszalmstra, line 417: I think the heuristic makes sense in all cases.

hunger: That does unfortunately not work:

Let's assume you have a source directory "./foo.tar.gz" as a python dependency. This gets detected by is_wheel_or_archive(...) as a "Distribution". Pixi will then notice "this is a source" and ask for a re-solve, writing a new v7 version to disc. The next time lockfile is read, "./foo.tar.gz" is detected as a "Distribution" again, ... resolve, ... write to disk, ... read, ... resolve, ...

Maybe we should have some better indicator which variant this is than "has a version"? Like storing "pypi-dist" and "pypi-source" in the lockfile instead of both normalizing to "pypi"? That's a few letters more per pypi package...

baszalmstra: Maybe a better indicator instead of the version is the hash?

hunger: Lets try version or hash or index_url ;-)

PR #2283, opened 2026-03-25
fix: patch copied files in place for prefix replacements

### LINKED ISSUE #2282: Feature request: use reflink-then-prefix-replace for patched files to preserve CoW sharing
### Checklist

### What is the idea?

Today, files that require prefix replacement are read to ram, patched then write to place, before reflink/hardlink is considered.
On reflink-capable filesystems, it should be more space-efficient to reflink the cached file into the target prefix first, then apply prefix replacement in-place on the destination, so unchanged extents remain shared via CoW.
Would you consider adding a reflink-aware install path for files with prefix placeholders?

Related Implementation:
https://github.com/conda/rattler/blob/7af992c214beb478816f51ce30b27363d29ce5a2/crates/rattler/src/install/link.rs#L775
https://github.com/conda/rattler/blob/7af992c214beb478816f51ce30b27363d29ce5a2/crates/rattler/src/install/link.rs#L723

Also found some materials describing this issue in prefix-dev/pixi repo:

> We should probably write some good documentation about this, but the summary is:
> 
> Our installation works in the following order.
> - Patch the file if it contains the prefix placeholder, will do a copy as it creates a new file
> - Reflink if the filesystem allowes for it. (Fastest and best option, but only works on a few filesystems)
> - Hardlink, if requested by the package and the system/user allows for it
> - Symlink, if requested by the package and the system/user allows for it
> - Copy, used as a last resort if non of the above works.
> 
> So, apart from the binaries that need to be patched, we should copy as little as we can. We don't have a garbage collector yet, but we would love to support that in the future.
> 
> The function is here: https://github.com/conda/rattler/blob/4e5b274b244e5ed69a9e4dccac6870768172e34b/crates/rattler/src/install/link.rs#L138
> 
> This discussion was more focuses more on the problem of downloading every version of a package. Without trying to reuse the cache first.
> 
> _Originally posted by @ruben-arts in https://github.com/prefix-dev/pixi/discussions/1354#discussioncomment-11971704_

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

TTTPOB:
## Summary
- switch patched-file installs to copy first and then patch the destination in place
- temporarily add owner-write access before patching copied read-only files, then restore the original permissions afterward
- preserve the existing mtime/codesign behavior and add focused tests for same-length, resized, and read-only patched files

Closes #2282

## Testing
- `CARGO_HOME=/tmp/cargo-container cargo fmt --all --check`
- `CARGO_HOME=/tmp/cargo-container CARGO_TARGET_DIR=/tmp/rattler-target cargo test -p rattler test_patch_copied_destination_updates_only_changed_ranges_when_lengths_match --lib -j1`
- `CARGO_HOME=/tmp/cargo-container CARGO_TARGET_DIR=/tmp/rattler-target cargo test -p rattler test_patch_copied_destination_rewrites_from_first_difference_when_length_changes --lib -j1`
- `CARGO_HOME=/tmp/cargo-container CARGO_TARGET_DIR=/tmp/rattler-target cargo test -p rattler test_patched_read_only_file_restores_permissions --lib -j1`
- `CARGO_HOME=/tmp/cargo-container CARGO_TARGET_DIR=/tmp/rattler-target cargo test -p rattler test_patched_file_receives_modification_time --lib -j1`
- `CARGO_HOME=/tmp/cargo-container CARGO_TARGET_DIR=/tmp/rattler-target cargo test -p rattler test_unpatched_file_keeps_source_mtime --lib -j1`

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -6,15 +6,17 @@ use once_cell::sync::Lazy;
 use rattler_conda_types::Platform;
 use rattler_conda_types::package::{FileMode, PathType, PathsEntry, PrefixPlaceholder};
 use rattler_digest::Sha256;
-use rattler_digest::{HashingWriter, Sha256Hash};
+use rattler_digest::Sha256Hash;
+use rattler_digest::digest::Digest;
 use reflink_copy::reflink;
 use regex::Regex;
 use std::borrow::Cow;
 use std::fmt;
 use std::fmt::Formatter;
 use std::fs::Permissions;
-use std::io::{BufWriter, ErrorKind, Read, Seek, Write};
+use std::io::{ErrorKind, Read, Seek, SeekFrom, Write};
 use std::path::{Path, PathBuf};
+use tempfile::{TempDir, TempPath};
 
 use super::apple_codesign::{AppleCodeSignBehavior, codesign};
 use super::{ExternalSymlinkPolicy, Prefix};
@@ -185,14 +187,6 @@ pub fn link_file(
         // Detect file type from the content
         let file_type = FileType::detect(source.as_ref());
 
-        // Open the destination file
-        let destination = BufWriter::with_capacity(
-            50 * 1024,
-            fs::File::create(&destination_path)
-                .map_err(LinkFileError::FailedToOpenDestinationFile)?,
-        );
-        let mut destination_writer = HashingWriter::<_, rattler_digest::Sha256>::new(destination);
-
         // Convert back-slashes (\) on windows with forward-slashes (/) to avoid problems with
         // string escaping. For instance if we replace the prefix in the following text
         //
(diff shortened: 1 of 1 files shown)


TTTPOB:
@hunger @wolfv
Gentle ping on this PR when you have a chance.

This fixes an issue in the pixi store install path where files are currently patched before hardlink/reflink is applied. As a result, the link step cannot fully preserve the intended space-saving behavior, so in many real cases the deduplication benefits are lost.

I’d appreciate a review when convenient. Thanks!

baszalmstra (changes requested):
What I dont quite understand is why we need the sink and everything? It looks overengineered. I also dont understand how this now works? Because you still just copy a file to the destination and then seem to patch that? That still doesnt preserve the COW semantics does it?

TTTPOB:
1. I think you are right. The reason I used this overengineered abstraction was to preserve the old helpers and tests. I was a bit unsure whether I should just remove those old helpers and tests.

2. Yes, this PR just copies a file to the destination and then patches it, but Rust `fs::copy` calls the `copy_file_range` syscall on Linux, and `fclonefileat` on macOS (falling back to `fcopyfile`). The Linux `copy_file_range` syscall will automatically do a reflink copy if the filesystem supports it, which takes advantage of a CoW semantics.

Here are some links:

Rust `fs::copy` comments
https://github.com/rust-lang/rust/blob/e8e4541ff19649d95afab52fdde2c2eaa6829965/library/std/src/fs.rs#L2844

For btrfs
https://github.com/torvalds/linux/commit/fce205e9da8e063aa1cf3d6583c1a9ed2b82f3f0

For ZFS
https://github.com/openzfs/zfs/blob/1644e2ffd2640fa3e2c191ceaf048a5fc8399493/module/os/linux/zfs/zpl_file.c#L1239

If this PR makes sense to you, I will simplify the overengineered abstraction and clean up those tests and helpers.

baszalmstra:
We also have a crate specifically to do reflinks, also on windows, I would prefer to use that instead.

TTTPOB:
The gatekeeper failure is due to macos runner queue timed out, not a real error.
I think this is pr is ready for review again.
I removed the over-engineered sink abstraction and used that reflink-copy crate.

PR #2286, opened 2026-03-26
feat: add timestamps to source records

baszalmstra:
This adds a `timestamp` to source records.

### DIFF
--- a/crates/rattler_lock/src/parse/models/v7/source_package_data.rs
+++ b/crates/rattler_lock/src/parse/models/v7/source_package_data.rs
@@ -83,9 +83,9 @@ pub(crate) struct SourcePackageDataModel<'a> {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub size: Cow<'a, Option<u64>>,
 
-    #[serde(default, skip_serializing_if = "is_default")]
-    #[serde_as(as = "crate::utils::serde::Timestamp")]
-    pub timestamp: chrono::DateTime<chrono::Utc>,
+    #[serde(skip_serializing_if = "is_default")]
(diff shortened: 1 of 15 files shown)


crates/rattler_lock/src/parse/models/legacy.rs
```diff
@@ -179,6 +182,17 @@ impl From<LegacyCondaPackageData> for CondaPackageData {
     }
 }
 
+/// When computing a default timestamp for legacy source packages, use the
+/// current time. However, that really screws with our tests, so we're using a
+/// default timestamp for tests.
+fn default_legacy_source_timestamp() -> chrono::DateTime<chrono::Utc> {
```
baszalmstra, line 188: This is a bit of a hack but guarantees stability of snapshot tests.

crates/rattler_lock/src/parse/models/v7/source_package_data.rs
```diff
@@ -85,9 +83,9 @@ pub(crate) struct SourcePackageDataModel<'a> {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub size: Cow<'a, Option<u64>>,
 
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    #[serde_as(as = "Option<crate::utils::serde::Timestamp>")]
-    pub timestamp: Option<chrono::DateTime<chrono::Utc>>,
+    #[serde(default, skip_serializing_if = "is_default")]
```
baszalmstra: Wait, this default is not correct. That should be "now" instead of default (0).

hunger:
Do we need a timestamp for pypi source packages as well? Just wondering :-)

Is there even a "ignore-newer-than TS" feature in pypi?

PR #2295, opened 2026-03-28
fix: normalize channel names to canonical URLs using ChannelConfig

### LINKED ISSUE #2294: Channel names not normalized to URLs in the solver, channel-pinned specs silently match wrong packages
### Checklist

### What happened?

MatchSpec parsing calls parse_channel_and_subdir which always uses    `ChannelConfig::default_with_root_dir(cwd)` (hard-coded alias https://conda.anaconda.org). When
   a user has a custom channel_alias, conda-forge::numpy is expanded to https://conda.anaconda.org/conda-forge/ instead of https://my-server.com/conda-forge/. The solver then compares this wrong URL against record.channel, which was correctly set using the real alias, and the comparison fails — causing channel-pinned specs to either fail silently or match packages from any channel. The maintainers flagged this with a TODO but left it unsolved.

@baszalmstra 

### Additional Context

_No response_

ritankarsaha:
### Description

Adds a channel_config: ChannelConfig field to SolverTask (defaulting to the standard conda.anaconda.org alias for backward compatibility) and threads it through to `CondaDependencyProvider::new`.                                                                
                                                            
  The TODO block that previously just collected channel-pinned specs is replaced by apre-computation step that maps each spec to its canonical channel URL, re-expanding short names through the caller-provided config. During per-record filtering, record.channel is likewise normalized through the same config before comparison. This ensures that a spec like my-channel::numpy correctly matches or rejects records regardless of which channel alias was active when the MatchSpec was parsed.

Fixes #2294 

### Checklist:

@baszalmstra

### DIFF
--- a/crates/rattler_solve/src/lib.rs
+++ b/crates/rattler_solve/src/lib.rs
@@ -14,7 +14,8 @@ use std::fmt;
 
 use chrono::{DateTime, Utc};
 use rattler_conda_types::{
-    utils::TimestampMs, GenericVirtualPackage, MatchSpec, PackageName, RepoDataRecord, SolverResult,
+    utils::TimestampMs, ChannelConfig, GenericVirtualPackage, MatchSpec, PackageName,
+    RepoDataRecord, SolverResult,
 };
 
 /// Represents a solver implementation, capable of solving [`SolverTask`]s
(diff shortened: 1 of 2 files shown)


baszalmstra:
Im not sure if this is the right approach. I think I see two alternatives:

- Passed in repodata lists also contain the channel they come from.
- We ensure that the `RepoData.channel` field is a channelurl.

PR #2298, opened 2026-03-28
feat(gateway): prefer full repodata over sharded for pattern/glob queries

### LINKED ISSUE #2292: Heuristic when to fetch full repodata
For queries like this:

```
rattler search '* >=3'
```

it doesn't make sense to use sharded repodata as this would result in more than 10k fetches.
Maybe if >x% of packages in repodata names (or an absolute number(?)) would be requested by a rattler repodata gateway query, we should fall back to full download instead?

WDYT @baszalmstra

officialasishkumar:
### Description

When a query contains pattern or glob specs (e.g. `rattler search '* >=3'`), the repodata gateway now skips sharded repodata and uses full `repodata.json` instead.

With sharded repodata, each matching package triggers a separate HTTP request to fetch its individual shard. Broad patterns like `*` can match >10,000 packages, resulting in thousands of individual shard fetches that are orders of magnitude slower than downloading a single full repodata file.

The heuristic: if the query contains any glob or regex `PackageNameMatcher` specs, use full repodata for all channel-based subdirs in that query. Exact-name queries (the common case for dependency resolution) continue to use sharded repodata as before.

Non-sharded subdirs are cached separately from sharded ones in the gateway, so pattern queries don't interfere with the sharded cache used by exact-name queries.

Fixes #2292

### How Has This Been Tested?

- Added `test_pattern_query_uses_full_repodata` test that verifies glob pattern queries succeed and return correct results against a remote channel.
- All 84 existing tests in `rattler_repodata_gateway` continue to pass (including `test_remote_gateway`, `test_clear_cache`, and existing glob pattern tests).
- Verified with `cargo clippy` (no warnings) and `cargo fmt` (clean).
- Full workspace compilation verified.

### Checklist:

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/builder.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/builder.rs
@@ -164,6 +164,7 @@ impl GatewayBuilder {
         Gateway {
             inner: Arc::new(GatewayInner {
                 subdirs: CoalescedMap::new(),
+                full_repodata_subdirs: CoalescedMap::new(),
                 client,
                 channel_config: self.channel_config,
                 #[cfg(not(target_arch = "wasm32"))]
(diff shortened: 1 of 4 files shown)


baszalmstra:
I would prefer if you first discuss the heuristic you want to implement before diving head first into an implementation.. Just downloading the full repodata on every glob is not a good heuristic.

PR #2299, opened 2026-03-29
fix: ordering of dev and post components

JeanChristopheMorinPerso:
### Description

This PR fixes a small bug with `dev` and `post` components in versions that I found while working on adding version interval ranges support to rattler. It also adds a new `is_post` method to mirror the `is_dev` method.

Basically, I found that rattler was considering a version `1.2devdev` to be smaller than `1.2dev`. This I think is wrong based on https://github.com/conda/ceps/blob/main/cep-0033.md and `conda.models.version.VersionOrder` agrees with that.

The same problem also happened with the `post` component, but in reverse.

```python
>>> import rattler
>>> rattler.Version('1.2dev') < rattler.Version('1.2devdev')
False
>>> rattler.Version('1.2post') > rattler.Version('1.2postpost')
False
```

while in conda:
```python
>>> conda.models.version.VersionOrder('1.2dev') < conda.models.version.VersionOrder('1.2devdev')
True
>>> conda.models.version.VersionOrder('1.2post') > conda.models.version.VersionOrder('1.2postpost')
True
```

The problem seem to stem from the fact that `tag_no_case` kind of matched on the prefix, I think. I'm still very new to rust, so I might be wrong. But the fix seems legit to me.

The docs on `tag_no_case` say (emphasis mine):
>will return **the part of the input that matches** the argument with no regard to case.

(https://docs.rs/nom/latest/nom/bytes/fn.tag_no_case.html)

### AI Disclosure

Tools: Amp (see https://ampcode.com/threads/T-019d36ed-fd10-7368-b400-481c3646f900, in this case it used gpt 5.4 in its deepest thinking mode)

### Checklist:

### DIFF
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -1231,8 +1231,10 @@ mod test {
 
     #[test]
     fn dev_is_only_special_for_exact_runs() {
+        // Because 1.2.devdev is not considered a dev version, it must sort after 1.2dev
         assert!(Version::from_str("1.2dev").unwrap() < Version::from_str("1.2.devdev").unwrap());
         assert!(Version::from_str("1.2dev").unwrap() < Version::from_str("1.2devdev").unwrap());
+        // Same with post
         assert!(Version::from_str("1.2postpost").unwrap() < Version::from_str("1.2post").unwrap());
 
         // 1.2dev is a dev version, but 1.2devdev is not.
--- a/crates/rattler_conda_types/src/version/mod.rs
+++ b/crates/rattler_conda_types/src/version/mod.rs
@@ -1220,6 +1220,13 @@ mod test {
     }
 
     #[test]
+    fn dev_is_only_special_for_exact_runs() {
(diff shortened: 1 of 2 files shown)


crates/rattler_conda_types/src/version/mod.rs
```diff
@@ -1203,6 +1203,13 @@ mod test {
         assert_eq!(random_versions, parsed_versions);
     }
 
+    #[test]
+    fn dev_is_only_special_for_exact_runs() {
```
baszalmstra, line 1233: I think we also have checks to see if something is a dev version? Maybe we can assert those too here?

JeanChristopheMorinPerso: Good idea. I added tests for `is_dev` and I also added a new method `is_post` tso that there is a nice symmetry.

JeanChristopheMorinPerso:
Hi @baszalmstra this is ready to review again. Thanks!

baszalmstra (approved):

baszalmstra:
Awesome thanks!

PR #2301, opened 2026-03-29
feat: serialize MatchSpec and NamelessMatchSpec as strings by default

### LINKED ISSUE #847: Deserialize from string by default

tarun-227:
## Summary
- Change default serde for `MatchSpec` and `NamelessMatchSpec` to serialize/deserialize as simple strings (e.g. `"numpy >=1.21"`) instead of expanded JSON structs with individual fields
- Removes `serde_as` / `skip_serializing_none` derive macros and the `deserialize_channel` helper function
- `Version` and `VersionSpec` already had this behavior — this makes the API consistent

## Motivation
Closes #847. The old struct-based serialization exposed internal fields and required `#[serde_as(as = "DisplayFromStr")]` annotations wherever these types were used in other structs. String-based serialization is simpler and matches how users think about match specs.

## Test plan
- [x] `cargo test -p rattler_conda_types -- serialize` — all 9 tests pass
- [x] `cargo fmt --check` — clean
- [x] Verified no other crates depend on the old struct-based JSON format
- [x] Snapshot updated to reflect new string-based output

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -7,16 +7,15 @@ use crate::{
 };
 use itertools::Itertools;
 use rattler_digest::{parse_digest_from_hex, Md5, Sha256};
-use rattler_digest::{serde::SerializableHash, Md5Hash, Sha256Hash};
-use serde::{Deserialize, Deserializer, Serialize};
-use serde_with::{serde_as, skip_serializing_none};
+use rattler_digest::{Md5Hash, Sha256Hash};
+use serde::{Deserialize, Deserializer, Serialize, Serializer};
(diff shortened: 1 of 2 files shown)


baszalmstra:
As I said in the other PR i backtracked on wanting this: https://github.com/conda/rattler/pull/2063#issuecomment-4053722519

PR #2303, opened 2026-03-30
feat: propagate sha256, md5, and size for direct URL packages

### LINKED ISSUE #742: Get `sha256`, `size` and `md5` from direct url package.

tarun-227:
## Summary
- When packages are installed from direct URLs, the `ExtractResult` computed during download already contains `sha256`, `md5`, and `total_size`, but these values were previously discarded after hash validation
- Adds `md5` and `size` fields to `CacheMetadata` with public getters (`sha256()`, `md5()`, `size()`)
- Captures extract result metadata via shared state in the fetch closure and propagates it to `CacheMetadata`
- `DirectUrlQuery` now uses these computed values when building `PackageRecord`, instead of passing `None` for size
- For local file URLs, reads size from file metadata

## Motivation
Closes #742. Direct URL packages had `sha256`, `md5`, and `size` set to `None` in their `PackageRecord` even though the values were computed during extraction.

## Test plan
- [x] `cargo check -p rattler_cache -p rattler_repodata_gateway` compiles cleanly
- [x] `cargo fmt --check` passes
- [x] No breaking changes — new fields are `Option` types and existing callers are unaffected

### DIFF
--- a/crates/rattler_cache/src/package_cache/cache_lock.rs
+++ b/crates/rattler_cache/src/package_cache/cache_lock.rs
@@ -9,7 +9,7 @@ use std::{
 use digest::generic_array::GenericArray;
 use fs4::fs_std::FileExt;
 use rattler_conda_types::package::{IndexJson, PathsJson};
-use rattler_digest::Sha256Hash;
+use rattler_digest::{Md5Hash, Sha256Hash};
 
 use crate::package_cache::PackageCacheLayerError;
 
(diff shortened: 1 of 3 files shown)


baszalmstra:
Fill in the PR template correctly. Also as a word of warning this was seems to have been completely AI generated, including the PR text which is against our AI policy. 

You are on thin ice here, given your other PR basically does the same thing.

PR #2304, opened 2026-03-30
fix: detect and remove stale shards/zst files in rattler-index

### LINKED ISSUE #1856: rattler-index: check whether shards / zst exist if we are not writing them
### Checklist

### What is the idea?

When not writing repodata shards or zst, things can get "out of whack" (if someone else writes shards or zst data).

For that reason, if shards or zst are disabled, BUT we find shards or zst we shoudl either:

- automatically delete them
- or issue a loud warning or even fail, and ask the user to change settings or delete them

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

tarun-227:
### Description

When rattler-index runs with `--write-zst=false` or `--write-shards=false`, 
files created by previous indexing runs (like `repodata.json.zst`, 
`repodata_shards.msgpack.zst`, and the `shards/` directory) are left behind 
on the disk itself so,this can confuse the users/clients that fetch stale data not matching the existing 
current `repodata.json`.

This PR basically cleans up at the end of `write_repodata()` in 
`crates/rattler_index/src/lib.rs` that:
- Checks if `repodata.json.zst` exists when `write_zst` is disabled, and removes it so that stale data is not fetched
- Checks if `repodata_shards.msgpack.zst` and the `shards/` directory exist 
  when `write_shards` is disabled, and removes them
- there will be log warnings when stale files are found and cleaned up
- Handles removal failures gracefully

Uses the same `op.exists()`, `op.delete()`, and `op.remove_all()` APIs 
already used elsewhere in the crate.

Fixes #1856

### How Has This Been Tested?

- Ran `cargo check -p rattler_index` — compiles cleanly
- Ran `cargo fmt --check` — no formatting issues
- Verified the logic follows the same opendal patterns used in the rest 
  of the indexing code (e.g. `op.exists()` calls in `index()` function)

### AI Disclosure
 Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/rattler_index/src/lib.rs
+++ b/crates/rattler_index/src/lib.rs
@@ -999,6 +999,46 @@ pub async fn write_repodata(
             .await?;
         }
     }
+
+    // Check for stale files that exist from previous runs but are no longer being written.
+    if metadata.repodata_zst.is_none() {
+        let zst_path = format!("{subdir}/{REPODATA}.zst");
+        if op.exists(&zst_path).await.unwrap_or(false) {
+            tracing::warn!(
+                "Stale file detected: {zst_path} exists but write_zst is disabled. \
+                 Removing to prevent inconsistent repository state."
+            );
+            if let Err(e) = op.delete(&zst_path).await {
+                tracing::warn!("Failed to remove stale file {zst_path}: {e}");
+            }
+        }
+    }
+
+    if metadata.repodata_shards.is_none() {
+        let shards_index_path = format!("{subdir}/{REPODATA_SHARDS}");
+        if op.exists(&shards_index_path).await.unwrap_or(false) {
+            tracing::warn!(
+                "Stale file detected: {shards_index_path} exists but write_shards is disabled. \
+                 Removing to prevent inconsistent repository state."
+            );
+            if let Err(e) = op.delete(&shards_index_path).await {
+                tracing::warn!("Failed to remove stale file {shards_index_path}: {e}");
+            }
+        }
+
+        // Also check for the shards directory
+        let shards_dir = format!("{subdir}/shards/");
+        if op.exists(&shards_dir).await.unwrap_or(false) {
+            tracing::warn!(
+                "Stale directory detected: {shards_dir} exists but write_shards is disabled. \
+                 Removing to prevent inconsistent repository state."
+            );
+            if let Err(e) = op.remove_all(&shards_dir).await {
+                tracing::warn!("Failed to remove stale directory {shards_dir}: {e}");
+            }
+        }
+    }
+
     Ok(())
 }
 


baszalmstra:
Fill in the PR template correctly. Also as a word of warning this was seems to have been completely AI generated, including the PR text which is against our AI policy.

tarun-227:
Thanks for the feedback. I've updated the PR description using the template and added the AI disclosure. I reviewed and tested the code locally with cargo check. Happy to address any code review feedback.

crates/rattler_index/src/lib.rs
```diff
@@ -999,6 +999,46 @@ pub async fn write_repodata(
             .await?;
         }
     }
+
+    // Check for stale files that exist from previous runs but are no longer being written.
+    if metadata.repodata_zst.is_none() {
+        let zst_path = format!("{subdir}/{REPODATA}.zst");
+        if op.exists(&zst_path).await.unwrap_or(false) {
```
pavelzw, line 1006: What is the return type of op.exists.await? Isn't it some `Result`?
In that case I think we should handle the error instead of using `unwrap_or(false)`

crates/rattler_index/src/lib.rs
```diff
@@ -999,6 +999,46 @@ pub async fn write_repodata(
...
+            );
+            if let Err(e) = op.delete(&zst_path).await {
+                tracing::warn!("Failed to remove stale file {zst_path}: {e}");
+            }
+        }
+    }
+
+    if metadata.repodata_shards.is_none() {
+        let shards_index_path = format!("{subdir}/{REPODATA_SHARDS}");
+        if op.exists(&shards_index_path).await.unwrap_or(false) {
```
pavelzw, line 1019: S.a.

crates/rattler_index/src/lib.rs
```diff
@@ -999,6 +999,46 @@ pub async fn write_repodata(
...
+                 Removing to prevent inconsistent repository state."
+            );
+            if let Err(e) = op.delete(&shards_index_path).await {
+                tracing::warn!("Failed to remove stale file {shards_index_path}: {e}");
+            }
+        }
+
+        // Also check for the shards directory
+        let shards_dir = format!("{subdir}/shards/");
+        if op.exists(&shards_dir).await.unwrap_or(false) {
```
pavelzw, line 1031: S.a.

crates/rattler_index/src/lib.rs
```diff
@@ -999,6 +999,46 @@ pub async fn write_repodata(
...
+                "Stale file detected: {shards_index_path} exists but write_shards is disabled. \
+                 Removing to prevent inconsistent repository state."
+            );
+            if let Err(e) = op.delete(&shards_index_path).await {
+                tracing::warn!("Failed to remove stale file {shards_index_path}: {e}");
+            }
+        }
+
+        // Also check for the shards directory
+        let shards_dir = format!("{subdir}/shards/");
```
pavelzw, line 1030: If some shards were written to another location, we might not catch all stale shards but I think this is okay since we hardcode `/shards` in this crate.

pavelzw, line 1030: Define `/shards` in a variable above and use it here and above instead of duplicating the constant.

crates/rattler_index/src/lib.rs
```diff
@@ -999,6 +999,46 @@ pub async fn write_repodata(
...
+    if metadata.repodata_zst.is_none() {
+        let zst_path = format!("{subdir}/{REPODATA}.zst");
+        if op.exists(&zst_path).await.unwrap_or(false) {
+            tracing::warn!(
+                "Stale file detected: {zst_path} exists but write_zst is disabled. \
+                 Removing to prevent inconsistent repository state."
+            );
+            if let Err(e) = op.delete(&zst_path).await {
+                tracing::warn!("Failed to remove stale file {zst_path}: {e}");
+            }
```
pavelzw, line 1013: Tbh I think we should just raise an error in those cases. This most likely points to a misconfiguration that should be manually handled. Otherwise we might remove files without the user's consent.

PR #2317, opened 2026-04-05
fix: only retry transient I/O errors in should_retry()

ritankarsaha:
### Description

  - `should_retry()` was returning `true` for all `IoError` and `CouldNotCreateDestination` variants, regardless of error kind
  - Permanent failures like `PermissionDenied`, `NotFound`, and `Other` would retry forever — the TODO comment in the code also acknowledged this was wrong
  - Now checks `err.kind()` and only retries genuinely transient kinds: `BrokenPipe`, `ConnectionReset`, `ConnectionAborted`, `ConnectionRefused`, `NotConnected`, `TimedOut`, `UnexpectedEof`, `Interrupted`
  - Tests updated to assert permanent errors do NOT retry; added a new test for `CouldNotCreateDestination` with a transient error kind

Fixes #2316 

### Checklist:

@baszalmstra what do you think ? Any more changes required?

### DIFF
--- a/crates/rattler_package_streaming/src/lib.rs
+++ b/crates/rattler_package_streaming/src/lib.rs
@@ -91,17 +91,29 @@ impl From<Cancelled> for ExtractError {
 impl ExtractError {
     /// Returns true if this error is transient and the operation should be retried.
     ///
-    /// This checks for common transient I/O errors like broken pipes,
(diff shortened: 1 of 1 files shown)


baszalmstra:
The tests on this PR are failing. Can you take a look?

Issue #2305, opened 2026-03-31
`invalid zip archive (async): zip64 extended information field was too long`


lucascolley:
### Checklist

- [x] I added a descriptive title
- [x] I searched open reports and couldn't find a duplicate

### What happened?

```python
In [16]: asyncio.run(rattler.package.PathsJson.from_remote_url(rattler.Client(), "https://conda.anaconda.org/conda-forge/noarch/spin-0.18-pyh8f84b5b_2.conda"))
...
File ~/Library/Caches/rattler/cache/cached-envs-v0/ipython-db9489e3f1011c3b/lib/python3.14/site-packages/rattler/package/paths_json.py:92, in PathsJson.from_remote_url(cls, client, url)
     87 @classmethod
     88 async def from_remote_url(cls, client: Client, url: str) -> PathsJson:
     89     """
     90     Fetches `info/paths.json` from a remote package archive URL.
     91     """
---> 92     return cls._from_py_paths_json(await PyPathsJson.from_remote_url(client._client, url))

OSError: invalid zip archive (async): zip64 extended information field was too long: expected 16 bytes, but 0 bytes were provided
```

### Additional Context

This is being hit in https://github.com/pavelzw/pixi-browse/issues/49 and https://github.com/conda-forge/spin-feedstock/issues/26.

pavelzw:
same for


```
pixi exec pixi-browse -c https://prefix.dev/github-releases -m 'gh==2.89.0'
```

not that the "Repository", "Homepage" and "built with" entries are missing as well

<img width="1797" height="887" alt="Image" src="https://github.com/user-attachments/assets/cd59e533-03d8-46cd-a38a-7e60ebb695d1" />

<img width="1484" height="1000" alt="Image" src="https://github.com/user-attachments/assets/1946d4cb-46ca-4134-b238-ab6853ab2d97" />

lucascolley:
gentle bump :)

wolfv:
@lucascolley can you confirm whether the issue is only with pixi browse, or with this package also for regular installation?

lucascolley:
> @lucascolley can you confirm whether the issue is only with pixi browse, or with this package also for regular installation?

Also installation. `pixi search` does not return the broken versions, for example, and `pixi update` does not update to them. But `conda` can install them.

wolfv:
@lucascolley this would be the fix: https://github.com/astral-sh/rs-async-zip/pull/32

Would you be able to try this?

[...]

PR #2331, opened 2026-04-10
fix: bump `zip`

lucascolley:
fixing the cause of https://github.com/conda/rattler/issues/2305

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -197,7 +197,7 @@ wasmtimer = "0.4.1"
 which = "8.0.0"
 windows-sys = { version = "0.61.2", default-features = false }
 winver = { version = "1.0.0" }
-zip = { version = "8.2.0", default-features = false }
+zip = { version = "8.5.1", default-features = false }
 zstd = { version = "0.13.3", default-features = false }
 
 # These are the all the crates defined in the workspace. We pin all of them together because they are always updated in tandem.


baszalmstra:
Are you saying that this also fixes that issue? Note that we use both `zip` and `async-zip`, where `async-zip` is mostly used to extract packages coming from URLs. 

Would be great if we can have a little example that shows this. 

Happy to merge this either way.

lucascolley:
> Are you saying that this also fixes that issue? Note that we use both `zip` and `async-zip`, where `async-zip` is mostly used to extract packages coming from URLs.
> 
> Would be great if we can have a little example that shows this.
> 
> Happy to merge this either way.

The problem seems to be which library is used when building the package, not extracting, see https://github.com/astral-sh/rs-async-zip/pull/32#issuecomment-4225970623.

lucascolley:
it looks like https://github.com/conda/rattler/commit/41e8e3cb269cfc7360846bae9418d37e4f822ac1 first bumped to the fixed version of `zip`. So maybe we just need to make sure rattler-build is using new enough rattler crates?

lucascolley:
In which case maybe https://github.com/prefix-dev/rattler-build/commit/498d5aba3cc27aee476fd0096bb0204dce2deb0d has done it already, so v0.61.4 _should_ work

PR #2332, opened 2026-04-12
fix: Hammer down the sort order of conda source packages

hunger:
Take the timestamp into account: We may have the same package several times for different platforms -- which will all have different timestamps. So we need to take timestamps into account when sorting, or the sources will move around the lock file on re-updates.

### How Has This Been Tested?

A crater run

### AI Disclosure

No AI

### Checklist:

### DIFF
--- a/crates/rattler_lock/src/conda.rs
+++ b/crates/rattler_lock/src/conda.rs
@@ -581,7 +581,13 @@ impl Ord for CondaPackageData {
                 (Some(src_a), Some(src_b)) => src_a
                     .variants
                     .cmp(&src_b.variants)
-                    .then_with(|| src_a.package_build_source.cmp(&src_b.package_build_source)),
+                    .then_with(|| src_a.package_build_source.cmp(&src_b.package_build_source))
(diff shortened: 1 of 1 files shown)


crates/rattler_lock/src/conda.rs
```diff
@@ -581,7 +581,13 @@ impl Ord for CondaPackageData {
                 (Some(src_a), Some(src_b)) => src_a
                     .variants
                     .cmp(&src_b.variants)
-                    .then_with(|| src_a.package_build_source.cmp(&src_b.package_build_source)),
+                    .then_with(|| src_a.package_build_source.cmp(&src_b.package_build_source))
+                    .then_with(|| match (&src_a.timestamp, &src_b.timestamp) {
```
baszalmstra, line 585: Should we also take the other fields into account maybe?🤔

hunger: I would actually prefer to order this based on the Identifier string itself.

Unfortunately that does not seem to be available yet when sorting the packages.

Hofer-Julian (approved):

PR #2339, opened 2026-04-14
chore: revert lockfile-v7

baszalmstra:
We came to the conclusion that this is not yet ready for prime time. We will need to make more iterations.


hunger:
I can not approve this, but unfortunately this needs to be done.

PR #2343, opened 2026-04-15
fix: Allow compilation on NetBSD

### LINKED ISSUE #2337: rattler_pty fails to build on NetBSD
### Checklist

### What happened?

`ptsname_r` is a Linux-specific function and doesn't exist on NetBSD.
NetBSD uses the POSIX `ptsname`.

I've found this when compiling any version of `mise` > 2026.2.8 on NetBSD.

The following patch fixes the build of `mise` on NetBSD,

```
--- ../vendor/rattler_pty-0.2.9/src/unix/pty_process.rs.orig	2026-04-13 14:41:10.214991451 +0000
+++ ../vendor/rattler_pty-0.2.9/src/unix/pty_process.rs
@@ -27,6 +27,9 @@ use nix::pty::ptsname_r;
 #[cfg(target_os = "linux")]
 use nix::pty::ptsname_r;
 
+#[cfg(target_os = "netbsd")]
+use nix::pty::ptsname;
+
 /// Start a process in a forked tty so you can interact with it the same as you would
 /// within a terminal
 ///
@@ -112,7 +115,11 @@ impl PtyProcess {
         unlockpt(&master_fd)?;
 
         // on Linux this is the libc function, on OSX this is our implementation of ptsname_r
+        #[cfg(any(target_os = "linux", target_os = "android"))]
         let slave_name = ptsname_r(&master_fd)?;
+        // But NetBSD uses the POSIX ptsname instead
+        #[cfg(target_os = "netbsd")]
+        let slave_name = unsafe { ptsname(&master_fd) }?;
 
         // Get the current window size if it was not specified
         let window_size = opts.window_size.unwrap_or_else(|| {
```

Would you consider a PR for this?

Beign POSIX-standard, `ptsname` is almost certainly the right call on any Unix that's not Linux/Android.
So we could most probably have `not(any(linux, android))` instead.

However, I don't use any other unix and cannot confirm this.
Let me know if you would prefer a more general approach which, is not tested.

### Additional Context

I see you have recently merged an alternative fix for FreeBSD.

0323pin:
### Description

NetBSD uses the POSIX `ptsname` instead of `ptsname_r` which is a Linux extension.

Fixes #2337 

### How Has This Been Tested?

Bare-metal build on NetBSD (`uname -r` 11.99.5) using Rust (`rustc -V` rustc 1.95.0-beta.8 (2c8c85df0 2026-04-09) (built from a source tarball)).

### Checklist:

### DIFF
--- a/crates/rattler_pty/src/unix/pty_process.rs
+++ b/crates/rattler_pty/src/unix/pty_process.rs
@@ -1,14 +1,14 @@
 pub use nix::sys::{signal, wait};
 use nix::{
     self,
-    fcntl::{OFlag, open},
+    fcntl::{open, OFlag},
     libc::STDERR_FILENO,
-    pty::{PtyMaster, Winsize, grantpt, posix_openpt, unlockpt},
+    pty::{grantpt, posix_openpt, unlockpt, PtyMaster, Winsize},
     sys::{
         stat,
         termios::{self, InputFlags, Termios},
     },
-    unistd::{ForkResult, Pid, close, dup, dup2_stderr, dup2_stdin, dup2_stdout, fork, setsid},
+    unistd::{close, dup, dup2_stderr, dup2_stdin, dup2_stdout, fork, setsid, ForkResult, Pid},
 };
 use std::os::fd::{AsFd, FromRawFd, IntoRawFd};
 use std::{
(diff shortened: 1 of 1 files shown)


wolfv:
Thank you for working on this!

0323pin:
It's clean at my end. The failing jobs are a consequence of `rattler` using `edition = 2021` and me using 1.95.0-beta.8.

Should I run `rustfmt --edition 2021 crates/rattler_pty/src/unix/pty_process.rs` and commit that?
Or, are you moving to 2024 soon?

baszalmstra (approved):

0323pin:
Thank you.

PR #2350, opened 2026-04-17
fix: simplify netbsd platform handling

0323pin:
### Description

Simplify platform handling introduced #2343

Simpler fix for #2337

### How Has This Been Tested?

Native NetBSD build.

### Checklist:

### DIFF
--- a/crates/rattler_pty/src/unix/pty_process.rs
+++ b/crates/rattler_pty/src/unix/pty_process.rs
@@ -24,11 +24,12 @@ use tokio::{
     time::{sleep, Duration, Instant},
 };
 
-#[cfg(any(target_os = "linux", target_os = "netbsd"))]
+#[cfg(target_os = "linux")]
 use nix::pty::ptsname_r;
 
 #[cfg(target_os = "netbsd")]
-/// NetBSD implementation using libc::ptsname_r
+/// NetBSD has `ptsname_r` in libc but apparently the `nix` crate does not expose it for NetBSD.
+/// Call `libc::ptsname_r` directly.
 fn ptsname_r(fd: &PtyMaster) -> nix::Result<String> {
     use std::ffi::CStr;
 
--- a/crates/rattler_pty/src/unix/pty_process.rs
+++ b/crates/rattler_pty/src/unix/pty_process.rs
@@ -24,11 +24,8 @@ use tokio::{
     time::{sleep, Duration, Instant},
 };
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "netbsd"))]
 use nix::pty::ptsname_r;
-
-#[cfg(target_os = "netbsd")]
-use nix::pty::ptsname;
 
 /// Start a process in a forked tty so you can interact with it the same as you would
 /// within a terminal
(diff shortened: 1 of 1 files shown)


crates/rattler_pty/src/unix/pty_process.rs
```diff
@@ -24,11 +24,23 @@ use tokio::{
...
 };
 
-#[cfg(target_os = "linux")]
+#[cfg(any(target_os = "linux", target_os = "netbsd"))]
 use nix::pty::ptsname_r;
 
 #[cfg(target_os = "netbsd")]
-use nix::pty::ptsname;
+/// NetBSD implementation using libc::ptsname_r
+fn ptsname_r(fd: &PtyMaster) -> nix::Result<String> {
```
baszalmstra, line 33: Can you add some comments as to why we need this? 

Also I dont really understand this:

```rust
#[cfg(any(target_os = "linux", target_os = "netbsd"))]
use nix::pty::ptsname_r;

#[cfg(target_os = "netbsd")]
/// NetBSD implementation using libc::ptsname_r
fn ptsname_r(fd: &PtyMaster) -> nix::Result<String> {
```

Both the `use` and the `fn` are enabled for `target_os = netbsd` right? But they have the same name. I dont think this would compile on netbsd?

0323pin: I'll have a look. Unfortunately, I'm in the middle of other stuff and need to rebuild Rust itself.
Sorry

0323pin: @baszalmstra Sorry there was leftover from the previous attempt. I see it now.

I won't have Rust installed for a couple of days, so a native build will have to wait.
Can you give the CI another go for now?

0323pin:
The Windows failure looks like a network error to me.
But, we need to wait for a local NetBSD build. Right now I'm struggling to get Rust-1.96.0-beta to compile.

0323pin:
@baszalmstra I'm still struggling to make 1.96-beta compile with 1.95.0 stable.

But, I didn't want to keep you waiting, so here's with 1.95
```
~> cd Git/rattler/crates/rattler_pty/
~> cargo build
   Compiling libc v0.2.184
   Compiling cfg_aliases v0.2.1
   Compiling signal-hook v0.3.18
   Compiling cfg-if v1.0.4
   Compiling pin-project-lite v0.2.17
   Compiling nix v0.30.1
   Compiling bytes v1.11.1
   Compiling bitflags v2.11.0
   Compiling tokio v1.51.1
   Compiling errno v0.3.14
   Compiling signal-hook-registry v1.4.8
   Compiling rattler_pty v0.2.9 (/home/pin/Git/rattler/crates/rattler_pty)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 31.50s
~> uname -rsp
NetBSD 11.99.5 x86_64
~> rustc -V
rustc 1.95.0 (59807616e 2026-04-14) (built from a source tarball)
```

All good.
Also, I've crosschecked and `nix` crate only exposes `ptsname_r` for Linux and Android, even though `libc` has it. So, the comment is indeed correct.

Thank you for your patience.

baszalmstra:
Thanks!

0323pin:
@baszalmstra Since I've refered to it a few times, I thought I'd just leave an FYI here.

I've just completed a source build of 1.96.0-beta.4 using 1.95.0 stable without an itch. It must have been a bug in beta.3 as I didn't change the build itself and now it's working.

1.96.0-beta.4 installed.

Issue #1248, opened 2025-04-10
MCP Server for rattler


iamthebot:
### Checklist

- [x] I added a descriptive title
- [x] I searched open requests and couldn't find a duplicate

### What is the idea?

It would be neat to be able to execute conda-related operations as part of an agentic development workflow. While agentic tools can certainly invoke CLIs (eg; `pixi` or CLIs using rattler), MCP offers a typed interface that provides more stability even if APIs change.

I'd imagine that `pixi` could simply extend the MCP server in rattler for pixi-specific operations.

### Why is this needed?

Agentic workflows are becoming more common for automating certain tedious tasks. As we explore helping automate eg; boostrapping a project having MCP servers available for the operations that need to be performed helps ensure the workflows are reliable even if the tools themselves evolve over time.

### What should happen?

A possible inspiration is [uv-mcp-server](https://mcp.so/server/uv-mcp-server).

iamthebot:
I'd possibly be interested in putting together an MVP.

baszalmstra:
I would be very interested in an MVP as that sounds super interesting.

Shubh-Raj:
@baszalmstra 

Based on how `rattler-bin` structures its CLI commands, here's my proposed approach:

**Proposed: `crates/rattler-mcp` — a new workspace crate**

The server would use the [Rust MCP SDK](https://github.com/modelcontextprotocol/rust-sdk) and expose rattler's capabilities as typed tools over stdio transport. These tools map directly to existing rattler APIs:

| MCP Tool | Rattler API |
|---|---|
| `search_packages` | `Gateway::query()` + `MatchSpec` (mirrors `rattler-bin/search.rs`) |
| `solve_environment` | [SolverTask](cci:2://file:///home/shubhraj/OpenSource/rattler/crates/rattler_solve/src/lib.rs:220:0-284:1) + `SolverImpl::solve()` via the resolvo backend |
| `list_virtual_packages` | `rattler_virtual_packages::VirtualPackage::detect()` |
| `install_packages` | `Installer::install()` from `rattler::install` |
| `list_installed` | `PrefixRecord::collect_from_prefix()` |

**Rough crate layout:**

```
crates/rattler-mcp/
├── Cargo.toml
└── src/
    ├── main.rs        # stdio MCP server entry point
    ├── server.rs      # ServerHandler impl
    └── tools/
        ├── search.rs  # Gateway + MatchSpec query
        ├── solve.rs   # SolverTask / resolvo
[...]

PR #2375, opened 2026-04-22
docs: Add py-rattler skill

pavelzw:
### Description

supersedes https://github.com/pavelzw/skill-forge/pull/30, xref https://github.com/conda/rattler/issues/1248

### AI Disclosure

Tools: claude

i didn't review the generated py-rattler text carefully. i mostly looked through it and did a vibecheck about the structure and that the main components that i would expect are available.
i also let a fresh opus 4.7 instance review the code and check for consistency against the py-rattler codebase

after this pr is merged + released, my plan would be to add these two skills in https://github.com/pavelzw/skill-forge

the skills are supposed to be updated before every py-rattler (and ideally rattler-bin) release. when working with an agent, it should hopefully pick it up and change it there automatically

### Checklist:


baszalmstra:
How do we keep the reference up to date?

pavelzw:
You mean the references directory in the skill?
We have 

1. an instruction in AGENTS.md that tell the agent to update the skill when it changes py-rattler or rattler-bin
2. an additional skill `update-skills` that looks at the diff between the last version and HEAD and adds missing things.

Another option might be adding something like coderabbit with a custom review for this. In my testing in skill-forge it was not working super well yet though. Xref https://github.com/conda/infrastructure/issues/1300

pavelzw:
And at least for the skill-forge integration, I'll add a run constraint for the exact py-rattler version so that only the matching version is installed. Only applies when you use pixi-skills though

baszalmstra:
I feel that's still a bit fragile. Can't we point the agent to the existing Python documentation? That is at least mostly autogenerated from the code itself.

pavelzw:
in locked-down environments where you don't have access to the internet, the agent would get stuck. it would also get stuck if you don't let it access the (parts of) internet through a sandbox so i think it would still be nice to have a self-contained skill that doesn't point to places outside of the skill.

we could think about just shipping the output of `pixi run build-docs` into the skill but this would be a _lot_ of html files that are not easily viewable by the agent.

this sounds roughly related to https://github.com/pawamoy/mkdocs-llmstxt but https://pixi.sh/latest/llms.txt didn't really seem helpful to me (without larger adjustments) so far as it only had links to other places in https://pixi.sh instead of local links. maybe it's possible to fork mkdocs-llmstxt so that it uses local links...

also, the python documentation doesn't necessarily adhere to agent best practices like [keep file references one level deep from SKILL.md](https://agentskills.io/specification#file-references)

baszalmstra:
I am on the fence here. I really don't like having a shadow copy of documentation specifically targeting AI that is not auto-generated from the source code itself. Especially in the repository itself.

pavelzw:
@baszalmstra how should we then continue here? i'm also happy to move those markdown files to https://github.com/pavelzw/skill-forge and update them there manually on every py-rattler release (same for the rattler-bin skill?). if there comes a mkdocs-agentskill plugin that generates those things directly from the documentation, we could re-thing this again.

baszalmstra:
And what about a suboptimal approach for the time being where we either generate an llm.txt or we refer to the existing documentation? They are just on a branch. I would be fine with restructuring the docs so they are more AI-friendly?

baszalmstra:
What do you think @pavelzw ?

pavelzw:
> I would be fine with restructuring the docs so they are more AI-friendly?

i think this would probably be the way to go. will need a little bit of effort and testing, probably. in the meantime, i created a skill for the rattler cli: https://github.com/conda/rattler/pull/2820

PR #2386, opened 2026-04-29
fix: make build string matching case-insensitive (CEP-29)

baszalmstra:
Per CEP-29, build string matching in `MatchSpec` should always be
case-insensitive. Previously, `StringMatcher` performed case-sensitive
comparisons for all three matcher kinds (Exact, Glob, Regex).

- Exact: use `eq_ignore_ascii_case`
- Glob: use `matches_with` with `case_sensitive: false`
- Regex: introduce a small `RegexMatcher` wrapper that compiles the
  pattern with `(?i)` while preserving the original pattern string for
  Display/serialization round-tripping.### Description

@jaimergp


wolfv (approved):
Looks good!

PR #2389, opened 2026-04-30
fix: tolerate trailing-slash mismatch in OIDC issuer URL

nichmor:
## Summary

```
OIDC discovery failed: Validation error:
unexpected issuer URI `https://prefix.dev/` (expected `https://prefix.dev`)
```


crates/rattler/src/cli/auth/oauth.rs
```diff
@@ -270,6 +301,16 @@ async fn discover_endpoints(
     })
 }
 
+/// Return `url` with its trailing `/` flipped: append one if absent, strip
+/// one if present.
+fn toggle_trailing_slash(url: &str) -> String {
```
wolfv, line 306: I am not sure why we need the toggle function here?
And can we just ensure server side that we either use trailing slash or not?

PR #2399, opened 2026-05-01
fix: handle missing symlinks on Windows install path

### LINKED ISSUE #5984: Cannot install package with symlink on Windows
### Checks

- [x] I have checked that this issue has not already been reported.

- [x] I have confirmed this bug exists on the [latest version](https://github.com/prefix-dev/pixi/releases/latest) of pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:
```shell
❯ pixi init
✔ Created C:\ProgramData\DevDrive\temp\sl\pixi.toml

❯ pixi add quarto==1.9.37
Error:   × failed to link quarto-1.9.37-h0e8a320_0.conda
  ├─▶ failed to link 'Library/share/quarto/extension-subtrees/julia-engine/CLAUDE.md'
  ├─▶ could not open source file
  ╰─▶ The system cannot find the file specified. (os error 2)
```

`pixi.toml`/`pyproject.toml` file that reproduces my issue:
```toml
[workspace]
authors = ["Martijn Visser <mgvisser@gmail.com>"]
channels = ["conda-forge"]
name = "sl"
platforms = ["win-64"]
version = "0.1.0"

[tasks]

[dependencies]
```
`pixi info` output:
```plaintext
System
------------
       Pixi version: 0.67.2
        TLS backend: rustls
           Platform: win-64
   Virtual packages: __win=10.0.26200=0
                   : __cuda=13.1=0
                   : __archspec=1=skylake
          Cache dir: C:\ProgramData\DevDrive\cache\rattler
       Auth storage: C:\Users\visser_mn\.rattler\credentials.json
   Config locations: C:\Users\visser_mn\.pixi\config.toml

Global
------------
            Bin dir: C:\Users\visser_mn\.pixi\bin
    Environment dir: C:\Users\visser_mn\.pixi\envs
       Manifest dir: C:\Users\visser_mn\.pixi\manifests\pixi-global.toml

Workspace
------------
               Name: sl
            Version: 0.1.0
      Manifest file: C:\ProgramData\DevDrive\temp\sl\pixi.toml
       Last updated: 01-05-2026 10:42:08

Environments
------------
        Environment: default
           Features: default
           Channels: conda-forge
   Dependency count: 0
   Target platforms: win-64
    Prefix location: C:\ProgramData\DevDrive\temp\sl\.pixi\envs\default
```

Other files (e.g. script files, source files, etc.):

### Issue description

The latest version of quarto contains a symlink that prevents installation on Windows, as we get a "failed to link" error as shown above.

Originally submitted as https://github.com/conda-forge/quarto-feedstock/issues/97 and worked around in https://github.com/PumasAI/QuartoNotebookRunner.jl/pull/412.

I tried micromamba and it installs the same version of quarto without issue, preserving the symlink.

### Expected behavior

I expect to be able to install the package. Not sure what (if anything) should be done with symlinks for best interoperability.

baszalmstra:
## Summary

When a conda package contains symlinks, extracting them on Windows requires Developer Mode, admin privileges, or a Dev Drive. Previously `rattler_package_streaming` unconditionally skipped every symlink entry on Windows, but `rattler::install::link::link_file` still tried to link from the (now non-existent) source, failing the whole install with `could not open source file` / `os error 2`.

Fix has two halves, addressing both ends of the broken contract:

- **`rattler_package_streaming::tokio::shared::unpack_tar_archive`**: instead of unconditionally skipping symlinks on Windows, attempt to extract first and only fall back to skip-with-warning if creation fails. Users with the right setup (Developer Mode, admin, or a Dev Drive) now actually get the symlink, matching micromamba behaviour.
- **`rattler::install::link::link_file`**: if a `SoftLink` entry has no source on disk (because extraction skipped it), catch the resulting `NotFound` from `read_link`/`fs::copy` and return `Ok(None)` instead of crashing. The function now returns `Result<Option<LinkedFile>, _>` and both callers in `install/mod.rs` handle the `None` case. No upfront `symlink_metadata` syscall, the existing I/O surfaces `NotFound` already.

Fixes prefix-dev/pixi#5984.

## Note on API change

`pub fn link_file` returns `Result<Option<LinkedFile>, LinkFileError>` instead of `Result<LinkedFile, _>`. `Ok(None)` means "this entry was intentionally skipped (no error)". Callers in this repo are updated; external callers will need a one-line adjustment.

## Test plan

- [x] New unit test `test_missing_symlink_source_is_skipped` in `crates/rattler/src/install/link.rs` covers the link-side fix
- [x] Existing tests in `rattler_package_streaming` still pass
- [x] `cargo clippy` clean on both crates
- [x] **Manual verification on Windows reproduces and fixes the issue.** Built pixi against this branch and ran the exact repro from prefix-dev/pixi#5984:

  Before (released pixi 0.67.2):
  ```
  × failed to link quarto-1.9.37-h0e8a320_0.conda
  ├─▶ failed to link 'Library/share/quarto/extension-subtrees/julia-engine/CLAUDE.md'
  ├─▶ failed to copy file to destination
  ╰─▶ failed to copy file from F:\packages\pixi\pkgs\... (os error 2)
  ```

  After (pixi patched to use this branch):
  ```
  WARN skipping symlink entry 'Library/share/quarto/extension-subtrees/julia-engine/CLAUDE.md':
       source missing in package cache (likely skipped during extraction)
  ✔ Added quarto==1.9.37
  ```

  My machine hits the `FailedToLink(Copy, NotFound)` variant (no symlink privileges, so `allow_symbolic_links=false` falls through to copy). The original reporter hit `FailedToReadSymlink(NotFound)` (had symlink privileges, so the symlink branch ran); the or-pattern in the fix covers both.

### DIFF
--- a/crates/rattler_package_streaming/src/tokio/shared.rs
+++ b/crates/rattler_package_streaming/src/tokio/shared.rs
@@ -33,21 +33,26 @@ pub(super) async fn unpack_tar_archive<R: tokio::io::AsyncRead + Unpin>(
     while let Some(entry) = entries.next().await {
         let mut file = entry.map_err(ExtractError::IoError)?;
 
-        // On Windows, skip symlink entries as they require special privileges
-        if cfg!(windows) && file.header().entry_type().is_symlink() {
-            tracing::warn!(
-                "Skipping symlink in tar archive: {}",
-                file.path().map_err(ExtractError::IoError)?.display()
-            );
-            continue;
-        }
-
         // Unpack the file into the destination directory
         #[cfg_attr(not(unix), allow(unused_variables))]
-        let unpacked_path = file
-            .unpack_in_raw(&destination, &mut memo)
-            .await
-            .map_err(ExtractError::IoError)?;
+        let unpacked_path = match file.unpack_in_raw(&destination, &mut memo).await {
+            Ok(path) => path,
+            Err(e) => {
+                // On Windows, creating symlinks requires Developer Mode, admin
+                // privileges, or a Dev Drive. If unpacking a symlink fails (typically
+                // permission denied), warn and skip the entry rather than failing the
+                // whole extraction. Users with the right setup still get the symlink.
+                #[cfg(windows)]
+                if file.header().entry_type().is_symlink() {
+                    tracing::warn!(
+                        "skipping symlink in tar archive: {}: {e}",
+                        file.path().map_err(ExtractError::IoError)?.display()
+                    );
+                    continue;
+                }
+                return Err(ExtractError::IoError(e));
+            }
+        };
 
         // Preserve the executable bit on Unix systems
         #[cfg(unix)]
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -163,7 +163,7 @@ pub fn link_file(
     apple_codesign_behavior: AppleCodeSignBehavior,
     modification_time: filetime::FileTime,
     external_symlink_policy: ExternalSymlinkPolicy,
-) -> Result<LinkedFile, LinkFileError> {
+) -> Result<Option<LinkedFile>, LinkFileError> {
     let source_path = package_dir.join(&path_json_entry.relative_path);
 
     let destination_path = target_dir.path().join(&destination_relative_path);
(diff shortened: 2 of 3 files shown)


crates/rattler_package_streaming/src/tokio/shared.rs
```diff
@@ -33,21 +33,26 @@ pub(super) async fn unpack_tar_archive<R: tokio::io::AsyncRead + Unpin>(
...
+            Ok(path) => path,
+            Err(e) => {
+                // On Windows, creating symlinks requires Developer Mode, admin
+                // privileges, or a Dev Drive. If unpacking a symlink fails (typically
+                // permission denied), warn and skip the entry rather than failing the
+                // whole extraction. Users with the right setup still get the symlink.
+                #[cfg(windows)]
+                if file.header().entry_type().is_symlink() {
+                    tracing::warn!(
+                        "skipping symlink in tar archive: {}: {e}",
```
wolfv, line 48: Should we mention dev mode on Windows?

wolfv:
This is somewhat unrelated but should our "Paths V2" also store symlink destinations? So that we are not bound on extracting symlinks, or could fix it later if symlink creation failed?

wolfv (approved):
Looks good - just needs a rebase / merge

PR #2409, opened 2026-05-06
fix: copy symlinked files when symbolic linking is disabled

baszalmstra:
### Description

When a package contains a symbolic link entry and symbolic linking is disabled at install time (`allow_symbolic_links = false`), the install fell through to `copy_to_destination`, which calls `fs::copy` on the cached symlink. That fails when the symlink target is only valid relative to the install prefix rather than the cache directory, so the file was never copied.

This change adds `copy_symlink_target_to_destination`, which canonicalizes the cached symlink to the real file inside the cache and copies that file. The `link_file` dispatch now picks it for the `SoftLink && !allow_symbolic_links` branch, and `symlink_to_destination`'s fallback also routes through it (the source is still a symlink there).

### How Has This Been Tested?

Added a unit test `test_copy_symlink_target_to_destination` that creates a real file plus a relative symlink in a fake cache, runs `copy_symlink_target_to_destination`, and asserts the destination is a regular file with the target's contents. The test is `#[ignore]`d on Windows because creating symlinks there needs admin or developer mode (same constraint as the existing symlink tests in this file).

Existing `cargo test -p rattler --lib install::link` tests pass, except for two pre-existing Windows-only failures that also need symlink-creation privileges and are unrelated to this change.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/rattler/src/install/link.rs
+++ b/crates/rattler/src/install/link.rs
@@ -294,6 +294,8 @@ pub fn link_file(
             target_dir.path(),
             external_symlink_policy,
         )?
+    } else if path_json_entry.path_type == PathType::SoftLink {
+        copy_symlink_target_to_destination(&source_path, &destination_path)?
     } else {
         copy_to_destination(&source_path, &destination_path)?
     };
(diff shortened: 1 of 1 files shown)


ruben-arts (approved):
Sure

PR #2417, opened 2026-05-08
chore: bump pixi lock-file to v7

baszalmstra:
### Description

Bumps the pixi lock-file to v7. This removes the need to bump the pixi-lock on every release.

### How Has This Been Tested?

CI

### AI Disclosure
No AI was harmed in the process.

### DIFF
--- a/.github/workflows/e2e-s3-tests.yml
+++ b/.github/workflows/e2e-s3-tests.yml
@@ -45,7 +45,7 @@ jobs:
         with:
           environments: minio
 
-      - run: pixi run e2e-s3-minio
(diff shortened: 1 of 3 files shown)


baszalmstra:
@pavelzw Something weird is happening. It looks like the source packages are built twice! Once in the setup-pixi job and once when running the task. I cant quite figure out why though. When running 

```sh
> pixi install -e minio --locked
> pixi run e2e-s3-minio
```

It only installs the packages once. 

We changed the layout of the `.pixi` folder in the latest release of Pixi. Could Pixi-setup maybe clean up folders or something?

PR #2420, opened 2026-05-10
fix(hardening): Look for TOCTOU issues in the code base

hunger:
### Description

I want to get to a point where I can share package caches between users. For that the backend code needs to be more hardened to all kinds of mischief people can come up with.

Here is a first round of TOCTOU and similar issues in Claude found in rattler, trying to close the window where an attacker could potentially sneak in a symlink into unexpected places or get a file descriptor to something that will then get more restrictive permissions applied.

The biggest addition is a rattler_fs_safety crate that tries to have some common and safe filesystem helpers.

The rest is pretty small and and self-contained changes, often using rattler_safe_fs.

### How Has This Been Tested?

Tests

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/rattler/src/install/entry_point.rs
+++ b/crates/rattler/src/install/entry_point.rs
@@ -5,9 +5,9 @@ use rattler_conda_types::{
     prefix_record::{PathType, PathsEntry},
     Platform,
 };
-use rattler_digest::HashingWriter;
-use rattler_digest::Sha256;
-use std::{fs::File, io, io::Write, path::Path};
+use rattler_digest::{compute_bytes_digest, Sha256};
+use rattler_fs_safety::atomic_write;
+use std::{io, path::Path};
 
 use super::Prefix;
 
@@ -56,7 +56,7 @@ pub fn create_windows_python_entry_point(
     )?;
     let script_contents =
         python_entry_point_template(target_prefix, true, entry_point, python_info);
-    let (hash, size) = write_and_hash(&script_path, script_contents)?;
+    let (hash, size) = write_and_hash(&script_path, script_contents, None)?;
 
     // Construct a path to where we will create the python launcher executable.
     let relative_path_script_exe = python_info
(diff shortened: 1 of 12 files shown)


crates/rattler/src/install/entry_point.rs
```diff
@@ -188,12 +184,19 @@ pub fn python_entry_point_template(
...
+    path: &Path,
+    contents: impl AsRef<[u8]>,
+    mode: Option<u32>,
+) -> io::Result<(Output<Sha256>, usize)> {
     let bytes = contents.as_ref();
-    let mut writer = HashingWriter::<_, Sha256>::new(File::create(path)?);
-    writer.write_all(bytes)?;
-    let (_, hash) = writer.finalize();
+    atomic_write(path, bytes, mode)?;
+    let hash = compute_bytes_digest::<Sha256>(bytes);
```
pavelzw, line 199: Won't this result in performance degradation?

crates/rattler_fs_safety/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+repository.workspace = true
+license.workspace = true
+readme.workspace = true
+
+[dependencies]
+# Capability-based filesystem access. The helpers in this crate
+# operate against an open `Dir` capability, so symlink swaps of
+# intermediate path components can't redirect us between calls.
+# cap-std covers Linux, macOS, FreeBSD, and Windows — the full
+# intersection of rattler's CI matrix.
```
pavelzw: This is too much text

crates/rattler_fs_safety/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+cap-std = { workspace = true }
+# Exposes the `OpenOptionsFollowExt` trait so `open_no_follow_in`
+# can push `O_NOFOLLOW` (and the Windows reparse-point equivalent)
+# into the kernel open syscall itself, instead of doing a separate
+# lstat-then-open pair that would race a symlink swap.
+cap-fs-ext = "3.4"
```
pavelzw: Why is this not a workspace dependency?

hunger:
Looks like I need to apply more windows-love :-)

pavelzw (commented):
haven't looked at everything so far. only a small review from my phone

crates/rattler/src/install/entry_point.rs
hunger: Am I missing a trick here? I thought it would be the same in this case where we write all data in one go anyway.

crates/rattler_fs_safety/Cargo.toml
hunger: Fixed in the last push.

crates/rattler_fs_safety/Cargo.toml
hunger: Removed the comments from Cargo.toml.

crates/rattler_fs_safety/Cargo.toml
hunger: It is now, resolving.

hunger:
Most of this has been covered upstream and I need to merge the two somehow.

hunger:
I won't get around to do this, let's just close it.

PR #2421, opened 2026-05-11
refactor: reuse pep508_rs Display via GivenVerbatimUrl

baszalmstra:
### Description

I noticed we use a custom implementation of pypi requirement serialization. This replaces it with an implementation that defers to the `pep508_rs` implementation as much as possible. 

I added a small `GivenVerbatimUrl` newtype that implements `pep508_rs::Pep508Url`. Its `Display` returns the verbatim `given()` string only when the URL scheme is `file`, falling back to the standard `VerbatimUrl` Display otherwise. This lets us reuse pep508_rs's own `Requirement` Display impl rather than maintaining a parallel copy.

Behavior is tightened intentionally: only `scheme == "file"` URLs use the verbatim string; https/git URLs now go through the normalized URL Display, matching upstream pep508_rs. Relative path dependencies in `uv.lock`-style entries continue to round-trip unchanged.

`GivenVerbatimUrl::wrap_requirement` destructures the source `Requirement`, so a future field addition in pep508_rs becomes a compile error.

This doesnt fix any particular issue but just removes custom implementation code.

### How Has This Been Tested?

CI

### AI Disclosure

Tools: Claude (Claude Code)

### Checklist:

### DIFF
--- a/crates/rattler_lock/src/parse/models/v7/given_verbatim_url.rs
+++ b/crates/rattler_lock/src/parse/models/v7/given_verbatim_url.rs
@@ -8,7 +8,7 @@ use pep508_rs::{Pep508Url, Requirement, VerbatimUrl, VerbatimUrlError, VersionOr
 /// A [`VerbatimUrl`] newtype whose [`Display`] preserves the original (relative)
 /// string for file URLs, falling back to the absolute URL otherwise.
 ///
-/// This is used so we can reuse pep508_rs's own [`Display`] impl for
+/// This is used so we can reuse `pep508_rs`'s own [`Display`] impl for
 /// [`Requirement`] when serializing the lockfile, while still keeping
 /// relative path dependencies intact.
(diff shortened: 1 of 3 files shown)


baszalmstra:
@tdejager This no longer serializes the "given" url when its not a path. Can you see any reason why we would want to store the given instead of the canonical url in the lockfile?

baszalmstra:
Turns out this fixes https://github.com/prefix-dev/pixi/issues/6062 too

tdejager:
> @tdejager This no longer serializes the "given" url when its not a path. Can you see any reason why we would want to store the given instead of the canonical url in the lockfile?

Can't recall, no. I can only remember the uv solver in some cases giving the wrong urls because r they were replaced but this mostl has to do with path dependencies and something we fixed in pixi at the time.

hunger (approved):
Looks sensible to me.

PR #2426, opened 2026-05-12
feat: add trusted publisher middleware in rattler_networking

nichmor:
### Description

Adds a trusted publisher middleware.

```rust
let tp = TrustedPublishingMiddleware::new(
          channel_url.clone(),
          TrustedPublishingOptions::for_prefix_dev(),
          mint_client,
      );

let gateway_client = ClientBuilder::new(reqwest::Client::new())
          .with_arc(Arc::new(tp))
          .build();
```

Then, if you run it under a registered trusted publisher, the middleware
will attach the bearer token in the request

```rust
let channel = Channel::from_str(
          "https://prefix.dev/my-org-private-channel",
          &ChannelConfig::default_with_root_dir(std::env::current_dir()?),
      )?;
      let spec = MatchSpec::from_str("numpy", ParseStrictness::Strict)?;

      let records = gateway
          .query(vec![channel], vec![Platform::Linux64], vec![spec])
          .recursive(true)
          .await?;
```

### AI Disclosure

### Checklist:


baszalmstra (commented):
Wouldnt it make more sense to make this part of rattler_networking?

wolfv:
I think there are some files missing.
I am also wondering if we should ping the endpoint, and cache if we don't get a positive response? Minting new JWT token for every request might get expensive otherwise. Or users should still opt in or something.

nichmor:
> Wouldnt it make more sense to make this part of rattler_networking?

the `trusted_publishing` is part of the rattler_networking. or you mean something else?

baszalmstra:
You also added it to the gateway. We dont need that do we? We can just keep it as a middleware only?

wolfv:
Looks like it could work. The only thing that I would keep in mind is that the tokens are only valid for ~15 min (IIRC) but I think that should be fine for the use case.

crates/rattler_repodata_gateway/src/gateway/builder.rs
```diff
@@ -133,15 +138,44 @@ impl GatewayBuilder {
         self
     }
 
+    /// Configure [trusted publishing](rattler_networking::trusted_publishing)
+    /// for `channel_url`.
+    #[must_use]
+    pub fn with_trusted_publishing(
```
baszalmstra: Im not sure if it makes sense to add this to the gateway. Why not add it directly to the client that you can pass in?

crates/rattler_networking/src/trusted_publishing.rs
```diff
@@ -0,0 +1,748 @@
...
+                            if token.is_fresh(SystemTime::now()) =>
+                        {
+                            return Some(token.token.clone());
+                        }
+                        TrustedPublishingCache::Disabled => return None,
+                        TrustedPublishingCache::Empty | TrustedPublishingCache::Token(_) => {}
+                    }
+                }
+
+                let bootstrap_client = ClientWithMiddleware::from(
```
baszalmstra: We should pass this into the middleware struct itself instead of recreating. That allows reusing the client connection pool. We can also avoid the panic here.

crates/rattler_networking/src/trusted_publishing.rs
```diff
@@ -0,0 +1,748 @@
...
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+impl Middleware for TrustedPublishingMiddleware {
+    async fn handle(
+        &self,
+        mut req: reqwest::Request,
+        extensions: &mut http::Extensions,
+        next: Next<'_>,
+    ) -> reqwest_middleware::Result<reqwest::Response> {
+        if req.headers().get(reqwest::header::AUTHORIZATION).is_none()
+            && req.url().host_str() == Some(self.host.as_str())
```
baszalmstra: Im not sure if this is enough because you have to configure this on a per channel basis. Not on a per-host basis.

crates/rattler_networking/src/trusted_publishing.rs
```diff
@@ -0,0 +1,748 @@
...
+            audience: DEFAULT_AUDIENCE.to_string(),
+            mint_path: DEFAULT_MINT_PATH.to_string(),
+            gitlab_id_token_env: DEFAULT_GITLAB_ID_TOKEN_ENV.to_string(),
+        }
+    }
+}
+
+/// Represents the CI provider being used for trusted publishing.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CiProvider {
```
baszalmstra: Can we use [ambient-id](https://github.com/astral-sh/ambient-id) instead of building our own?

baszalmstra (changes requested):
There is also an expiry for tokens. Is that something we support?

typos.toml
```diff
@@ -18,3 +18,4 @@ ignore-hidden = false
 strat = "strat"
 haa = "haa"
 intoto = "intoto"
+certifi = "certifi"
```
nichmor, line 21: for some reason my lefthook failed like this:
```

❯ │ 🥊 lefthook v2.1.6  hook: pre-commit │                                                                                                           
  ╰──────────────────────────────────────╯                                                                                                           
  error: `certifi` should be `certify`, `certifies`, `certified`                                                                                     
     ╭▸ ./crates/rattler_solve/tests/sorting.rs:89:9                                                                                                 
     │                                                                                                                                               
  89 │ #[case::certifi("certifi >=2016.9.26", SolveStrategy::Highest)]                                                                               
     ╰╴        ━━━━━━━                                                                                                                               
  error: `certifi` should be `certify`, `certifies`, `certified`                                                                                     
     ╭▸ ./crates/rattler_solve/tests/sorting.rs:89:18                                                                                                
     │                                                                                                                                               
  89 │ #[case::certifi("certifi >=2016.9.26", SolveStrategy::Highest)]   
```

crates/rattler_networking/src/trusted_publishing.rs
```diff
@@ -0,0 +1,709 @@
...
+
+/// Returns the short-lived token to use against `server_url`.
+///
+/// Delegates OIDC ID-token retrieval to `ambient-id`; this function owns the
+/// mint exchange with `server_url`.
+pub async fn get_token(
+    client: &ClientWithMiddleware,
+    server_url: &Url,
+    options: &TrustedPublishingOptions,
+) -> Result<TrustedPublishingToken, TrustedPublishingError> {
```
baszalmstra: Can we let this return `None` if we couldnt determine a token?

crates/rattler_networking/src/trusted_publishing.rs
```diff
@@ -0,0 +1,709 @@
...
+    let detector = ambient_id::Detector::new_with_client(client.clone());
+    let oidc_token = match detector.detect(&options.audience).await? {
+        Some(token) => token,
+        None => return Err(TrustedPublishingError::NoOidcToken),
+    };
+
+    let publish_token = get_publish_token(oidc_token.reveal(), server_url, client, options).await?;
```
baszalmstra: Lets not call reveal until we really need the revealed token! So pass IdToken into this funciton.

crates/rattler_networking/src/trusted_publishing.rs
```diff
@@ -0,0 +1,709 @@
...
+/// Returns [`TrustedPublishResult::Skipped`] when no CI provider is detected
+/// (the common case outside CI). Errors during the flow are wrapped in
+/// [`TrustedPublishResult::Ignored`] so callers can fall back to other auth
+/// sources without unwinding.
+pub async fn check_trusted_publishing(
+    client: &ClientWithMiddleware,
+    server_url: &Url,
+    options: &TrustedPublishingOptions,
+) -> TrustedPublishResult {
+    if !on_supported_ci() {
```
baszalmstra: I think it would be better if we let ambient_id decide whether we should skip or not. By returning `None` from get_token.

baszalmstra (changes requested):

crates/rattler_networking/src/trusted_publishing.rs
```diff
@@ -0,0 +1,656 @@
...
+/// Refresh minted JWT tokens before they expire to avoid sending a token that
+/// becomes invalid while a request is in flight.
+const TOKEN_REFRESH_MARGIN: Duration = Duration::from_secs(60);
+
+/// Default audience for the OIDC ID token. Matches prefix.dev's expectation.
+pub const DEFAULT_AUDIENCE: &str = "prefix.dev";
+
+/// Default path on the server where the OIDC ID token is exchanged for a
+/// bearer token.
+pub const DEFAULT_MINT_PATH: &str = "/api/oidc/mint_token";
```
baszalmstra: These can go.

pixi.lock
baszalmstra, line 1: Please revert this change and update your pixi version. ;)

baszalmstra (approved):

PR #2436, opened 2026-05-18
fix: render condition as defined in the CEP

wolfv:
### Description

As per the CEP, matchspecs in conditions need slightly adjusted rendering: 

- no spaces
- only `{name}{op}{version}` allowed
- otherwise, need to use bracket syntax

This PR adjusts the rendering and adds more tests. This is required by `pixi` and when producing valid repodata.

### How Has This Been Tested?

Added more tests.

### AI Disclosure

Tools: Claude

### Checklist:


baszalmstra (changes requested):
I think we should ensure proper quoting here.

wolfv:
@baszalmstra can you elaborate a little bit? Is proper quoting to alternate between `"` vs `'` for you, or to escape quotes with `\"`?

baszalmstra:
Yeah thats a bit ambigous, I think using `'` in the inner matchspec looks nicer but if we have nesting in nesting fall back to `\"` again?

PR #2447, opened 2026-05-19
feat: only walk active extras when fetching repodata

baszalmstra:
### Description

The gateway used to walk every extra of every package it pulled in, which made even small solves (e.g. `black` from conda-pypi) cascade through ~500k records. This PR scopes extra traversal to the extras that are actually activated by the top-level specs, while still handling the case where a deeply nested package activates a new extra on an already-fetched package.

Changes:
- `PackageName::name_and_extras_from_matchspec_str` parses `pkg[extras=[a,b]]` and returns the activated extras alongside the normalized name.
- `PackageRecords` now splits unique deps into `unique_base_deps` and `unique_extra_deps` (per extra), so the gateway can pick which lists to walk.
- The query executor tracks active extras per package, walks only base deps plus active extras when queueing, and re-walks cached records when an extra is activated late. Works for both Input and Transitive paths and respects the existing spec filter on Input.

Effect: `rattler create --dry-run -c conda-pypi black` drops from ~500k to ~1.5k records (327x reduction). Multi-channel solve with conda-forge still picks the right `black` version.

### How Has This Been Tested?

- New unit tests in `subdir.rs` for `extract_unique_deps_split` covering base/extra dedup, base-wins across records (both orders), fully-subsumed extras, and the empty-records case.
- New rstest cases in `package_name.rs` for `name_and_extras_from_matchspec_str` including empty extras list, trailing comma, multiple bracket sections, and malformed input.
- New gateway integration tests in `mod.rs` exercising inactive vs active extras, late activation via Transitive, chained activation on the same package, Input scoping, and Input late activation with spec filter.
- End-to-end verified with `rattler create --dry-run -c conda-pypi black` and `-c conda-pypi -c conda-forge black`.

### AI Disclosure

Tools: Claude Code (Opus)

### Checklist:

### DIFF
--- a/crates/rattler_conda_types/src/package_name.rs
+++ b/crates/rattler_conda_types/src/package_name.rs
@@ -175,7 +175,12 @@ impl PackageName {
         let options = crate::ParseMatchSpecOptions::default().with_experimental_extras(true);
         match crate::MatchSpec::from_str(spec, options) {
             Ok(parsed) => (name, parsed.extras.unwrap_or_default()),
-            Err(_) => (name, Vec::new()),
+            Err(err) => {
+                // Failure mode is silent under-fetching of an extra's deps;
+                // log so it shows up under RUST_LOG=debug when investigating.
+                tracing::debug!("failed to parse matchspec '{spec}' for extras extraction: {err}",);
+                (name, Vec::new())
+            }
         }
     }
 }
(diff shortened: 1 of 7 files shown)


wolfv (approved):
LGTM

PR #2454, opened 2026-05-21
fix: match repodata records by file name

### LINKED ISSUE #1721: `file_name` not correctly set when using `{ path = "..." }` in `pixi.toml`
```toml
[project]
channels = ["conda-forge"]
name = "example"
platforms = ["linux-64", "linux-aarch64", "win-64", "osx-64", "osx-arm64"]

[dependencies]
my-webserver = { path = "noarch/my-webserver-0.1.0-pyh4616a5c_0.conda" }

```

```yaml
- conda: /Users/user/example/noarch/my-webserver-0.1.0-pyh4616a5c_0.conda
  subdir: noarch
  depends:
  - python >=3.12
  - fastapi
  - python
  constrains:
  - pydantic >=2,<3
  file_name: file:///Users/user/example/noarch/my-webserver-0.1.0-pyh4616a5c_0.conda
  timestamp: 1729264746409
```

The `file_name` should be only `my-webserver-0.1.0-pyh4616a5c_0.conda`

puneetdixit200:
### Description

Fixes #1721

`fn` constraints were parsed into `MatchSpec`, but were not checked when matching `MatchSpec` or `NamelessMatchSpec` against a `RepoDataRecord`. That allowed a path-style spec such as `fn=file:///.../package.conda` to behave like a name-only match instead of selecting by the record's archive filename.

This adds `RepoDataRecord` filename matching and treats URL/path `fn` values as their basename, so a local path dependency can match `package-...conda` while still rejecting a different archive filename.

### How Has This Been Tested?

Red check before implementation:
- `cargo test -p rattler_conda_types test_file_name_matches_repodata_identifier -- --nocapture` failed because `fn=other-1.0-py37_0.conda` still matched the record.

Green checks after implementation:
- `cargo test -p rattler_conda_types test_file_name_matches_repodata_identifier -- --nocapture`
- `cargo test -p rattler_conda_types --lib`
- `cargo fmt --all -- --check`
- `cargo clippy -p rattler_conda_types --all-targets -- -D warnings`
- `git diff --check`

### AI Disclosure

Tools: OpenAI GPT-5.5 was used to inspect the matchspec/repodata matching path, add the regression test and implementation, and draft this PR text. The changes and tests were reviewed and run locally.

### Checklist:

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -830,6 +830,10 @@ impl Matches<GenericVirtualPackage> for MatchSpec {
 }
 
 fn file_name_matches(spec_file_name: &str, record_file_name: &str) -> bool {
+    // Conda matches MatchSpec `fn` against PackageRecord.fn, which is the
+    // archive file name. RepoDataRecord stores the same value in its archive
+    // identifier, so callers should pass identifier.to_file_name() instead of
+    // deriving it from the record URL.
     spec_file_name_basename(spec_file_name) == record_file_name
 }
 
@@ -1255,6 +1259,8 @@ mod tests {
             channel: None,
         };
 
+        // Match against the archive identifier, mirroring conda's PackageRecord.fn
+        // semantics without parsing the record URL during matching.
         let match_spec = MatchSpec::from_str("mamba[fn=mamba-1.0-py37_0.conda]", Strict).unwrap();
         let nameless_spec = match_spec.clone().into_nameless().1;
         assert!(match_spec.matches(&repodata_record));
--- a/crates/rattler_conda_types/src/match_spec/mod.rs
+++ b/crates/rattler_conda_types/src/match_spec/mod.rs
@@ -10,6 +10,7 @@ use rattler_digest::{Md5, Sha256, parse_digest_from_hex};
 use rattler_digest::{Md5Hash, Sha256Hash, serde::SerializableHash};
 use serde::{Deserialize, Deserializer, Serialize};
 use serde_with::{serde_as, skip_serializing_none};
+use std::borrow::Cow;
 use std::fmt::{Debug, Display, Formatter};
 use std::hash::Hash;
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


crates/rattler_conda_types/src/match_spec/mod.rs
```diff
@@ -817,6 +830,28 @@ impl Matches<GenericVirtualPackage> for MatchSpec {
     }
 }
 
+fn file_name_matches(spec_file_name: &str, record_file_name: &str) -> bool {
+    spec_file_name_basename(spec_file_name) == record_file_name
+}
+
+fn spec_file_name_basename(value: &str) -> Cow<'_, str> {
+    if let Ok(url) = Url::parse(value)
```
baszalmstra: Parsing a url here will be very expensive.

baszalmstra (changes requested):
is this also what conda does?

puneetdixit200:
I checked conda's current `MatchSpec`/`PackageRecord` path. `MatchSpec.match()` iterates every match component and compares it against the same field on the package record, and `fn` is a `PackageRecord` field. When a record only has a URL, conda derives `fn` from `Channel(url).package_filename`; on the spec side, `MatchSpec.fn` normalizes `fn`/`url` through `basename(...)`.

So yes, this is intended to match conda behavior: a spec like `fn=file:///.../pkg.conda` is compared as the archive filename, not treated as a name-only match.

Refs I checked:
- https://github.com/conda/conda/blob/3037a5b5b0cc58e0856c758f763a4cf804775927/conda/models/match_spec.py#L377-L401
- https://github.com/conda/conda/blob/3037a5b5b0cc58e0856c758f763a4cf804775927/conda/models/match_spec.py#L640-L647
- https://github.com/conda/conda/blob/3037a5b5b0cc58e0856c758f763a4cf804775927/conda/models/records.py#L177-L188

puneetdixit200:
Addressed in the follow-up commits. The current implementation no longer parses a URL for `fn` matching; it strips query/fragment text and takes the basename with string slicing. I also added a comment documenting that this mirrors conda's `MatchSpec.fn`/`PackageRecord.fn` filename semantics.

puneetdixit200:
Addressed the URL parsing concern in the current head: `file_name_matches` now strips query/fragment and takes the basename with string operations instead of calling `Url::parse` during matching. I also reran `cargo test -p rattler_conda_types test_file_name_matches_repodata_identifier` locally and it passes.

PR #2455, opened 2026-05-21
feat: nicer OAuth callback page with per-project override

wolfv:
## Summary

Screenshot 2026-07-09 at 13 46 27

- Replace the bare `<html><body><h1>…` browser response shown after the OAuth redirect with a self-contained card layout (light/dark aware via `prefers-color-scheme`, inline SVG icons, HTML-escaped error detail, `Cache-Control: no-store`, `noindex`).
- Add `OAuthConfig::callback_page: Option<CallbackPageRenderer>` so downstream tools (pixi, rattler-build) can supply their own branded page. The renderer is `Box<dyn Fn(bool, &str) -> String + Send + Sync>` and receives `(success, plain-text detail)`.
- Make `default_callback_page` and `html_escape` `pub` so overriders can reuse the default or safely interpolate the detail.

## Example override

```rust
let config = OAuthConfig {
    // ...
    callback_page: Some(Box::new(|success, detail| {
        if success {
            include_str!("pixi_success.html").to_string()
        } else {
            format!("<!doctype html>...<pre>{}</pre>...", html_escape(detail))
        }
    })),
};
```

When `None`, falls back to the new default page.

### DIFF
--- a/crates/rattler/src/cli/auth.rs
+++ b/crates/rattler/src/cli/auth.rs
@@ -380,6 +380,7 @@ async fn login(
                 scopes,
                 redirect_uri,
                 user_agent: args.user_agent,
+                callback_page: None,
             };
 
             let auth = oauth::perform_oauth_login(config).await?;
--- a/crates/rattler/src/cli/auth/oauth.rs
+++ b/crates/rattler/src/cli/auth/oauth.rs
@@ -399,7 +399,12 @@ async fn auth_code_flow(
 
     // Verify CSRF state
     if callback.state != *csrf_token.secret() {
-        send_callback_response(&callback.stream, false, "CSRF state mismatch", callback_page);
+        send_callback_response(
+            &callback.stream,
+            false,
+            "CSRF state mismatch",
+            callback_page,
+        );
         return Err(OAuthError::CsrfMismatch);
     }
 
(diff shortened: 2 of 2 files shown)


tdejager:
Note that in pixi we use the device-flow authentication now by default, so you will not see this page with pixi.

wolfv:
Can you remind me why we do that?
With this code we could also make a Pixi branded "success" page :)

tdejager:
A couple of benefits:

* We are following the OAuth Device Authorization Grant RFC. The RFC was specifically created for devices and CLI applications where a browser callback isn’t always practical.
* which means: It doesn’t require running a browser on the machine where you’re logging in, so it works over SSH and other headless or remote environments.
* It doesn’t rely on opening a localhost callback server, which makes it work in more environments out of the box.
* GitHub takes the same approach with their CLI, I know that's an example you look at :)

W.r.t. the success page, the server serves it now. Since we know the oidc client-id we can still customize the styling, and all the assets already live there, so we don’t have to bundle them into the executable for it to look nice.

For me the strongest argument is that the device flow exists for exactly this use case. Rather than defaulting to a custom localhost callback flow, we default to the standardized flow that was designed for CLI applications.

tdejager:
Another benefit everyone has control over the success page on their server as opposed to needing to embed in their binary that uses rattler.

wolfv:
Sounds good!

PR #2462, opened 2026-05-27
feat: support CEP-42 in gateway

baszalmstra:
### Description

Implements [CEP-42] channel relations in the gateway. Channels can declare `base` and `overrides` relations in their `repodata.json` / sharded index, and the gateway now follows those relations recursively to expand the user's channel list and resolve a single priority order across all reachable channels.

`RepoDataQuery` and `NamesQuery` both gain two builder methods:

- `.channel_relations(mode)` with three modes:
  - `Disabled`: ignore declared relations, behave as today.
  - `Warn` (default): follow relations, log warnings for cycles or failed fetches of discovered channels, never fail the query.
  - `Strict`: follow relations, surface cycles as `GatewayError::ChannelRelationsError`.
- `.channel_relations_max_depth(depth)` overrides the default recursion depth (10).

Channel discovery overlaps with package fetching, so a query that follows a chain of related channels does not pay a serial penalty. Custom (non-channel) sources are not subject to CEP-42 ordering and keep their user-specified position. The user's explicit channel order always wins when it conflicts with a declared relation.

[CEP-42]: https://github.com/conda/ceps/blob/main/cep-0042.md

### How Has This Been Tested?

40 tests cover the new behavior:

- 27 unit tests on the resolution algorithm in isolation (priority ordering, base/overrides semantics, user-vs-relation conflicts, cycle breaking, max-depth limits, custom source placement).
- 3 integration tests on the per-subdir relations accessor.
- 10 end-to-end gateway tests against a local channel server, covering: base expansion ordering, `Disabled` mode short-circuits, `Strict` mode errors on a cycle, `Warn` mode tolerates a cycle, `max_depth` truncates recursion, a missing discovered subdir becomes an empty bucket, custom sources stay last after reorder, and the same three default/Disabled/Strict behaviors for `NamesQuery`.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5461,6 +5461,7 @@ dependencies = [
  "memmap2",
  "parking_lot",
  "pin-project-lite",
+ "proptest",
  "rattler_cache",
  "rattler_conda_types",
  "rattler_config",
--- a/crates/rattler-bin/src/commands/create.rs
+++ b/crates/rattler-bin/src/commands/create.rs
@@ -210,6 +210,11 @@ pub async fn create(opt: Opt, offline: bool) -> miette::Result<()> {
     .into_diagnostic()
     .context("failed to load repodata")?;
 
+    // Surface any non-fatal CEP-42 channel-relation problems.
+    for warning in &repo_data.warnings {
+        eprintln!("warning: {warning}");
+    }
+
     // Determine the number of records
     let total_records: usize = repo_data.iter().map(RepoData::len).sum();
     println!(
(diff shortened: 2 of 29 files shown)


wolfv:
I am wondering a few things:

- Should we go through the reporter so that the users can decide how / if they want to surface these errors to their CLI clients? Maybe that would be cleaner than a setting? Especially when interleaving with progress bars and such.
- I think certain clients might want to use "auto-"configuration (ie. actually _use_ the information). It migth not be somethign that we need to implement today but it would probably be good to think about something like having a function on the repodata gateway that lets you first fetch all repodata relations, then act on them (use, warn, ...?). Or a setting that says: "follow repodata relations".

wolfv (approved):
The only thing I am wondering:

- can we make warnings a must_use thing? So that API users would have to actively ignore it
- or should we add it to the gateway reporter?

Issue #2059, opened 2026-02-18
Virtual Filesystem Layer for Zero-Copy Environments


baszalmstra:
## Summary

Introduce a virtual filesystem library in rattler that serves package files directly from the package cache, as if they were installed into a target prefix — enabling near-instant, zero-copy environment creation.

Building on [#1162](https://github.com/conda/rattler/pull/1162) and [#1182](https://github.com/conda/rattler/pull/1182), the core library would handle path resolution and on-the-fly prefix rewriting. Platform-specific "bridges" then expose this as a real mountable filesystem.

I think we can call this "ephemeral" or "dematerialized" environments. 

## Proposed Architecture

The idea is to write a platform-agnostic library (`rattler-vfs`) responsible for 

- **Path resolution** — mapping virtual paths (e.g. `/env/lib/libz.so`) to real files in the cache.
- **Prefix rewriting** — transparently rewriting hardcoded prefixes so files appear to belong to the virtual environment (building on #1162 / #1182).
- **File metadata & directory listings** — providing stat-like metadata for lookups and listings.

This library is then used by "bridges" that adapt the core into real filesystem mounts using for instance NFS, Fuse, or [ProjFS](https://learn.microsoft.com/en-us/windows/win32/projfs/projected-file-system). Since all major platforms support mounting NFS that might be a good first candidate. This also has the benefit that this could work over the network! Research into the other implementations is required as well to determine whether the API should provide certain functionalities for certain bridges.

## Use Cases

Sure, but WHY?

### Ephemeral environments
`pixi exec` and `rattler-build` can create environments in milliseconds instead of copying/linking files to disk. Environment "creation" becomes a metadata operation — mount, run, tear down.

### Network-shared environments
A single server with a warm cache serves environments to remote clients over NFS. Especially useful for CI/CD clusters and HPC setups where shared network filesystems are the norm — no per-client downloads or extraction needed.

### Reduced disk usage
Multiple environments share a single cache with no file duplication, which matters on resource-constrained systems or when managing many environments.

## Benchmarking

Obviously there must also be a downside to this. Besides the complexity of implementing this I think another big downside is that all filesystem operations will go through user-space. We will have to benchmark some real world scenarios to determine if this poses a significant issue.

SanchitKS12:
Hello @baszalmstra ,I am interested in working on this issue and would like to focus specifically on the performance and benchmarking aspects of the proposed VFS-based environments.

In particular, I’d like to:

[...]

PR #2463, opened 2026-05-28
feat: rattler exec implementation

Dagmar-Dinjens:
### Description

Through working on #2059 it is nice to have a baseline working rattler exec to start working the zero-copy environments into. 

This implementation is greatly influenced by pixi exec and is in the benchmark which I held the performance for rattler exec against. 

### How Has This Been Tested?

I used pixi's unit tests, wrote a benchmark for comparison and did some e2e testing in both a release and a debug build:
image

works like pixi's
```
rattler exec [OPTIONS] [COMMAND] ...
```

### AI Disclosure

Tools: Claude

### Checklist:


baszalmstra:
Can you look at the CI issues and fill in the template completely.

crates/rattler-bin/src/commands/create.rs
```diff
@@ -420,7 +420,7 @@ fn print_transaction(
 
 /// Displays a spinner with the given message while running the specified
 /// function to completion.
-fn wrap_in_progress<T, F: FnOnce() -> T>(msg: impl Into<Cow<'static, str>>, func: F) -> T {
+pub fn wrap_in_progress<T, F: FnOnce() -> T>(msg: impl Into<Cow<'static, str>>, func: F) -> T {
```
baszalmstra: Maybe better to move this into its own module.

baszalmstra (approved):

crates/rattler-bin/Cargo.toml
```diff
@@ -70,7 +70,11 @@ rattler_upload = { workspace = true }
...
 reqwest-middleware = { workspace = true }
-tokio = { workspace = true, features = ["rt-multi-thread", "macros"] }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "signal"] }
 tracing-subscriber = { workspace = true, features = ["env-filter", "fmt"] }
 itertools = { workspace = true }
 url = { workspace = true }
+sha2.workspace = true
+tracing.workspace = true
+dunce.workspace = true
+regex.workspace = true
```
pavelzw:
```suggestion
sha2 = { workspace = true }
tracing = { workspace = true }
dunce = { workspace = true }
regex = { workspace = true }
```

crates/rattler-bin/src/commands/exec.rs
```diff
@@ -0,0 +1,509 @@
...
+        .chain(with_specs.iter())
+        .filter_map(|s| s.name.as_exact().map(|n| n.as_normalized().to_string()))
+        .collect();
+
+    if !display_names.is_empty() {
+        let env_name = format!(
+            "temp:{}",
+            display_names.iter().cloned().collect::<Vec<_>>().join(",")
+        );
+        extra_env.insert("RATTLER_ENVIRONMENT_NAME".into(), env_name.clone());
```
pavelzw: this will not work with https://github.com/starship/starship/blob/4f879b5aa42290557cc60a1c464a7526a7fbdab7/src/modules/pixi.rs#L15. might want to use `PIXI_ENVIRONMENT_NAME` here, even though it's not really pixi

crates/rattler-bin/src/commands/exec.rs
```diff
@@ -0,0 +1,509 @@
...
+
+    let download_client = Client::builder()
+        .build()
+        .expect("failed to build HTTP client");
+    let download_client = reqwest_middleware::ClientBuilder::new(download_client.clone())
+        .with_arc(Arc::new(
+            AuthenticationMiddleware::from_env_and_defaults().into_diagnostic()?,
+        ))
+        .with(rattler_networking::OciMiddleware::new(download_client))
+        .build();
```
pavelzw: didn't we have some default client in some way in rattler_networking?

i think we are using s3middleware in other parts of rattler-bin as well. we should use it here too

crates/rattler-bin/src/commands/exec.rs
```diff
@@ -0,0 +1,509 @@
...
+) -> miette::Result<PathBuf> {
+    let channel_urls: Vec<String> = channels.iter().map(|c| c.base_url.to_string()).collect();
+    let env_hash = compute_env_hash(specs, &channel_urls, platform);
+
+    let dir_name = match dir_prefix {
+        Some(ref p) => format!("{}-{}", p, &env_hash[..8]),
+        None => env_hash[..16].to_string(),
+    };
+
+    let prefix = cache_dir.join(EXEC_ENVS_DIR).join(&dir_name);
```
pavelzw, line 199: this means that `pixi exec` and `rattler exec` will share the same cache? might lead to weirdness down the line when the behavior of pixi differs from the behavior of rattler-bin but should be okay...

crates/rattler-bin/src/commands/exec.rs
```diff
@@ -0,0 +1,509 @@
...
+    fn multiple_explicit_specs_have_no_prefix() {
+        let prefix = exec_dir_prefix(&[spec("foo"), spec("bar")], Some("cmd"), false);
+        assert_eq!(prefix, None);
+    }
+
+    #[test]
+    fn guess_sanitizes_illegal_chars() {
+        let s = guess_package_spec("my/cmd!");
+        assert_eq!(s.name.as_exact().unwrap().as_normalized(), "my-cmd-");
+    }
```
pavelzw: is this behavior from pixi? seems a bit magic to me and i can't think of a reason to include it. are there use cases where this is helpful?

pavelzw (commented):
mostly looks good. some comments

crates/rattler-bin/src/commands/exec.rs
Dagmar-Dinjens: The pixi exec also does this but indeed has no test for it, I just assumed there was one as the pixi side does this.

crates/rattler-bin/src/commands/exec.rs
Dagmar-Dinjens: Having in mind that the rattler Exec will down the line use a virtual layer for the cache this won't become an issue I assume.

crates/rattler-bin/src/commands/exec.rs
pavelzw: @baszalmstra not sure who implemented this in pixi, do you have more context whether it's needed?

baszalmstra: I dont remember the details, fine to remove that here.

baszalmstra (approved):
Looks good as a first implementation!

baszalmstra:
@pavelzw Lets improve this later when we find weird things. Merging for now.

PR #2465, opened 2026-06-01
feat(rattler-bin): Add inject-into-prefix and remove-from-prefix subcommands

pavelzw:
### Description

this makes `pixi-inject` irrelevant. will add a deprecation warning there as well https://github.com/pavelzw/pixi-inject/pull/62

### AI Disclosure

Tools: codex

```
add a `rattler inject` and `rattler remove-from-prefix` command.

  # add a conda package to a prefix
  rattler inject my-package-0.0.1-h123456.conda --skip-compatibility-checks # can be a remote url as well

  # remove a single conda package from a prefix
  # maybe support matchspecs here as well?
  rattler remove-from-prefix my-package-0.0.1-h123456 --skip-compatibility-checks
```

In the end decided remote url is not that relevant since there is `rattler download` and I would be using this feature with local files only anyway 

### Checklist:

---

usage:

```
rattler remove-from-prefix --prefix ./.pixi/envs/default libexpat --skip-compatibility-checks
rattler inject-into-prefix --prefix ./.pixi/envs/default my-package.conda --skip-compatibility-checks
```


crates/rattler-bin/src/commands/prefix.rs
```diff
@@ -0,0 +1,417 @@
...
+
+    if !opt.skip_compatibility_checks {
+        let records = installed_packages
+            .iter()
+            .map(|record| &record.repodata_record.package_record)
+            .chain(std::iter::once(&package_record))
+            .collect::<Vec<_>>();
+        PackageRecord::validate(records)
+            .into_diagnostic()
+            .context("injected package is not compatible with the prefix")?;
```
JaninaWibkerQC:
```suggestion
            .context("injecting this package would make the prefix incompatible")?;
```

pavelzw:
```
❯ target/debug/rattler remove-from-prefix -p .pixi/envs/default xz-gpl-tools
Error:   × removing this package would make the prefix incompatible
  ╰─▶ package 'xz=5.8.1=h9a6d368_2' has dependency 'xz-gpl-tools 5.8.1 h9a6d368_2', which is not in the environment

❯ target/debug/rattler inject-into-prefix -p .pixi/envs/default polars-1.41.2-pyh58ad624_0.conda
Error:   × injecting this package would make the prefix incompatible
  ╰─▶ package 'polars=1.41.2=pyh58ad624_0' has dependency 'polars-runtime-32 ==1.41.2', which is not in the environment
```

crates/rattler-bin/src/commands/prefix.rs
```diff
@@ -0,0 +1,706 @@
...
+    skip_compatibility_checks: bool,
+}
+
+pub async fn inject(opt: InjectOpt) -> miette::Result<()> {
+    let target_prefix = std::path::absolute(opt.target_prefix).into_diagnostic()?;
+    let packages = opt
+        .packages
+        .into_iter()
+        .map(|package_path| {
+            let package_path = std::path::absolute(package_path).into_diagnostic()?;
```
baszalmstra: Should this be canonicalize instead?

crates/rattler-bin/src/commands/prefix.rs
```diff
@@ -0,0 +1,706 @@
...
+    let transaction = Transaction::from_current_and_desired(
+        installed_packages.clone(),
+        desired_records,
+        None,
+        None,
+        Platform::current(),
+    )
+    .into_diagnostic()?;
+
+    let driver = InstallDriver::builder()
```
baszalmstra: Can you construct an `Installer` instead? Thats completely optimized to do this work.

crates/rattler-bin/src/commands/prefix.rs
```diff
@@ -0,0 +1,706 @@
...
```
baszalmstra: Same! I think using an `Installer` would be much better.

crates/rattler-bin/src/commands/prefix.rs
```diff
@@ -0,0 +1,706 @@
...
+    Ok(())
+}
+
+fn package_record_from_archive(path: &Path) -> miette::Result<PackageRecord> {
+    let archive_type = CondaArchiveType::try_from(path)
+        .ok_or_else(|| miette::miette!("unsupported package archive: {}", path.display()))?;
+
+    match archive_type {
+        CondaArchiveType::TarBz2 => rattler_index::package_record_from_tar_bz2(path),
+        CondaArchiveType::Conda => rattler_index::package_record_from_conda(path),
```
baszalmstra: Feels like we could also add a function there to read from an arbitrary archive type.

baszalmstra (changes requested):

crates/rattler-bin/src/commands/prefix.rs
pavelzw: `Installer` before only used `get_or_fetch_from_url_with_retry` so i needed to also add one small adjustment in `rattler` to account for local paths

crates/rattler/src/install/installer/mod.rs
```diff
@@ -832,21 +832,30 @@ async fn populate_cache(
...
-        .await
-        .map_err(|e| InstallerError::FailedToFetch(record.identifier.to_string(), e))
+    let reporter = reporter.map(|(reporter, cache_index)| {
+        Arc::new(CacheReporterBridge {
+            reporter,
+            cache_index,
+        }) as _
+    });
+
+    if let Ok(path) = record.url.to_file_path() {
```
baszalmstra: I think you also have to check here whether the url has schema `file://` otherwise a localhost path will also return a path which is not what we want.

baszalmstra (changes requested):
This looks much cleaner! One small issue left but otherwise looks good!

crates/rattler/src/install/installer/mod.rs
pavelzw: [`138af2f` (this PR)](https://github.com/conda/rattler/pull/2465/commits/138af2f81d67e430113945586a128b21f819858a)

baszalmstra (approved):

PR #2486, opened 2026-06-04
fix: return 404 for sharded repodata when using oci

baszalmstra:
### Description

Using an `oci://` channel with the gateway could fail because the gateway looked for sharded repodata, which OCI registries do not provide. This change makes `oci://` channels fall back to regular repodata, so they resolve correctly out of the box (for example with sharded repodata enabled, as pixi does by default).

### How Has This Been Tested?

Ran `rattler create --dry-run -c oci://ghcr.io/channel-mirrors/conda-forge black` and confirmed it resolves `black` and its dependencies from the OCI channel.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/rattler_networking/src/oci_middleware.rs
+++ b/crates/rattler_networking/src/oci_middleware.rs
@@ -53,6 +53,17 @@ impl OciMiddleware {
     }
 }
 
+/// The file name of the sharded repodata index.
+const REPODATA_SHARDS_FILENAME: &str = "repodata_shards.msgpack.zst";
+
+/// Returns `true` if the URL refers to a sharded repodata index, which OCI
+/// registries do not serve.
+fn is_sharded_repodata_url(url: &Url) -> bool {
+    url.path_segments()
+        .and_then(|mut s| s.next_back())
+        .is_some_and(|name| name == REPODATA_SHARDS_FILENAME)
+}
+
 /// The action to perform on the OCI registry
 pub enum OciAction {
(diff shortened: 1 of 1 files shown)


wolfv (approved):

crates/rattler_networking/src/oci_middleware.rs
```diff
@@ -306,6 +317,18 @@ impl Middleware for OciMiddleware {
             return next.run(req, extensions).await;
         }
 
+        // OCI registries (conda-oci-mirror layout) only serve `repodata.json`,
+        // never sharded repodata. Return a 404 for sharded repodata requests so
+        // that the gateway falls back to the regular `repodata.json` path
+        // instead of trying (and failing) to fetch a shard index that does not
+        // exist.
+        if is_sharded_repodata_url(req.url()) {
```
wolfv, line 325: It is a bit weird that the OCI registry would not return a 404 anyways for files / layers that do not exist? But sure - we can do it this way for now.

baszalmstra: the problem is that the OCI layer just passes along the oci:// url because it doesnt recognize the structure. Then the next layer (reqwest itself) doesnt know how to deal with the oci:// url.

baszalmstra: You were right, this is a problem in pixi instead.

PR #2498, opened 2026-06-10
feat(lock): store binary conda package locations verbatim

hunger:
Change CondaBinaryData.location to Verbatim so a relative 'conda:' path round-trips through the lock file.

Relates to: prefix-dev/pixi#6322

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/rattler_lock/src/parse/models/v7/conda_package_data.rs
+++ b/crates/rattler_lock/src/parse/models/v7/conda_package_data.rs
@@ -14,7 +14,7 @@ use serde_with::serde_as;
 use url::Url;
 
 use crate::{
-    CondaPackageData, ConversionError, UrlOrPath, VariantValue,
+    CondaPackageData, ConversionError, UrlOrPath, VariantValue, Verbatim,
     conda::CondaBinaryData,
     utils::derived_fields::{self, LocationDerivedFields},
 };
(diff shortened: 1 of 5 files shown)


hunger:
@Bas: I would like to have the verbatim data for this, so that we can round-trip relative channel URLs through pixi. I do not plan to support those officially, but we should not break them when a user sticks them into their files.

crates/rattler_lock/src/parse/models/v7/conda_package_data.rs
```diff
@@ -50,8 +50,12 @@ fn skip_variant_serialization(input: &Option<Cow<'_, BTreeMap<String, VariantVal
 #[derive(Serialize, Deserialize, Eq, PartialEq)]
 pub(crate) struct CondaPackageDataModel<'a> {
     /// The location of the package. This can be a URL or a path.
+    ///
+    /// Stored verbatim so a relative path round-trips: the `given` string is
+    /// re-emitted on serialization, while the inner value is left for the
+    /// consumer to resolve against the lock file's base directory.
     #[serde(rename = "conda")]
-    pub location: UrlOrPath,
+    pub location: Cow<'a, Verbatim<UrlOrPath>>,
```
baszalmstra, line 58: Cant a Path already be relative though?

hunger: Of course.

But one of the first things we do is make it an absolute path and pass that around. I do not want to change that and still keep the "relative" information around.

baszalmstra: But the serialization also would then also only store either relative or absolute so you would have to reconstruct it anyway.

hunger:
I don't think I will bother fixing this going forward, so I'll just close this.

PR #2500, opened 2026-06-11
feat: Add JSON output support to search command

pavelzw:
### Description

Adds a `--json` flag to the `search` command that outputs results in JSON format, grouped by platform/subdir. This matches the output format of `pixi search --json` for consistency across tools.

When `--json` is enabled:
- Results are grouped by platform (subdir) in an `IndexMap`
- Records within each platform are sorted in descending order
- Output is formatted as pretty-printed JSON
- Informational messages (search query, channels) are suppressed
- The `--json` flag conflicts with `--limit`, `--limit_packages`, and `--all` flags

### Changes

- Added `--json` flag to search command options with appropriate conflicts
- Conditionally suppress informational output when JSON mode is enabled
- Group and sort search results by platform before JSON serialization
- Added dependencies: `indexmap` (with serde feature) and `serde_json`

### How Has This Been Tested?

The change is straightforward and follows existing patterns in the codebase. Existing tests continue to pass, and the JSON output can be verified by running:
```
rattler search <package> --json
```

### AI

i used fable 5 for this with the following prompt (having `rattler` and `pixi` repositories available for context)

```
add rattler search --json (same format as pixi search --json)
```

### Checklist:
- [x] Changes follow existing code patterns and conventions

https://claude.ai/code/session_013rT68LK9jWcabzYxjNyCQC


pavelzw:
one alternative to this could be (maybe explore in future PR) to have a generic search implementation (similar to `auth`) that can then be shared by both pixi and rattler. i think the only pixi-specific thing in pixi search is that it uses the channels (and platforms(?)) from `pixi.toml` by default?

baszalmstra (approved):

PR #2504, opened 2026-06-11
feat!: challenge-reactive AuthChallengeMiddleware with prefix.dev OIDC wiring

nichmor:
## Summary

Replaces `TrustedPublishingMiddleware` (added in #2426, no consumers yet) with a `WWW-Authenticate`-reactive `AuthChallengeMiddleware` driven by pluggable `AuthFlow` strategies, and wires it up so that prefix.dev private-channel reads authenticate automatically from CI with zero configuration. This is the client-side half of fixing prefix-dev/pixi#6318 (OIDC repository access never auto-authenticates for private channel *reads*); the server-side half is prefix.dev sending `WWW-Authenticate` challenges on anonymous private-channel access.

- **`challenge_middleware` (new module):** owns the HTTP mechanics: tolerant RFC 7235 `WWW-Authenticate` parsing via the `http-auth` crate (a header value it rejects contributes no challenges rather than failing the set), `BearerToken` (redacted `Debug`, sensitive headers), per-origin token caches with JWT-`exp`-aware refresh, negative caching, and single-flight acquisition (concurrent challenged requests share one mint), and the `AuthChallengeMiddleware` itself: on a challenge it consults its registered `AuthFlow`s in order and replays the request exactly once with the first token produced.
- **`AuthFlow` trait:** the pluggable extension point (`acquire_token(url, challenges) -> Result<Option<BearerToken>, AuthFlowError>`). `Ok(None)` = "flow not applicable" (the origin is negative-cached once all flows decline); errors are logged and never fail the request, so the caller always observes the server's original response. The middleware invokes flows for every challenged URL; flows that forward credentials gate on the URL's origin themselves.
- **`TrustedPublishingFlow`:** wraps the existing `ambient-id` CI detection + mint-endpoint exchange. Responds only to `Bearer` challenges; performs no origin gating of its own (documented as such).
- **`PrefixAuthAmbientFlow`:** the origin gate (per review feedback): delegates to `TrustedPublishingFlow` if and only if the challenged URL is `https` on `prefix.dev` or a true subdomain. Keyed on the request URL alone, so server-controlled challenge params (e.g. a spoofed `realm="prefix.dev"`) cannot lure CI credentials; lookalike hosts (`evil-prefix.dev`, `prefix.dev.evil.com`, trailing-dot) fail closed.
- **`AuthChallengeMiddleware::default()`:** registers `PrefixAuthAmbientFlow`. Consumers (rattler-bin here, pixi next) add one unconditional middleware and get challenge-reactive prefix.dev auth with no per-channel wiring.
- **`TrustedPublishingOptions::for_host` / `for_server`:** derive the mint endpoint from the channel/upload host. Key finding from live testing: prefix.dev deployments validate GitHub OIDC tokens against the **fixed audience `prefix.dev`** (not per-host), which `for_server` encodes.
- **`rattler_upload`:** derives the trusted-publishing audience from the upload host instead of hardcoding it.
- **`rattler-bin`:** the CLI client stack registers `AuthChallengeMiddleware::default()`.
- **Breaking:** `TrustedPublishingMiddleware` and its support machinery are removed; `TrustedPublishingToken` remains as a deprecated alias of `BearerToken`. `check_trusted_publishing`/`get_token` are unchanged.

## Design decisions reviewers may want to weigh in on

- **The `WWW-Authenticate` header is the trigger, not the status code** (works for 401 and prefix.dev's 403; a stray header on a 2xx would cause one replay). Strictly challenge-reactive: no speculative minting for public channels.
- **Negative caching is process-lifetime and per origin:** when every flow declines an origin, that origin is disabled until restart. Rationale: no retry storms, and behavior is identical to not having the middleware. A TTL/bounded-retry variant is a possible follow-up.
- **Trust gating lives in the flows, not the middleware** (per review feedback): the middleware is origin-agnostic with per-origin token caches (scheme + host + effective port, so a token minted for one origin is never replayed to another and https-to-http replay is impossible); each credential-forwarding flow validates the challenged URL itself. reqwest's redirect policy additionally strips `Authorization` on cross-origin redirects.
- **Hardening:** tokens are validated as header values *before* caching (a malformed token falls through to the next flow instead of poisoning the origin), and `TrustedPublishingFlow::new` normalizes a missing leading `/` on `mint_path`.
- **Fixed `prefix.dev` audience** for all prefix.dev deployments (incl. beta), discovered during live e2e bring-up; per-host audiences are rejected by the server.

## Test Plan

- [x] 60+ tests in `rattler_networking` (challenge parsing: multi-challenge, multi-header, per-value tolerance; middleware behaviors: challenge, mint and replay, per-origin token & negative caches, single-flight mint under concurrent challenges, flow ordering and fall-through, Authorization precedence, replay-once cap, stale-token reacquire, error swallowing, inert-by-default for untrusted origins; `PrefixAuthAmbientFlow` trust-gate suite incl. realm-spoofing and lookalike hosts; GitLab-env mint exchange via `temp-env` + mock mint endpoint)
- [x] Composed end-to-end test: `AuthChallengeMiddleware` + `TrustedPublishingFlow` + axum server hosting both the protected route and the mint endpoint, proving challenge, OIDC detect, mint, replay, and cross-request caching
- [x] `rattler_upload` suite passes unmodified (proactive upload path untouched)
- [x] Workspace `pixi run cargo-clippy` (`-D warnings`) and `cargo doc` clean
- [x] **Verified live against beta.prefix.dev (2026-06-12)** with an out-of-repo e2e harness (GitHub OIDC in CI): the private-channel read path (anonymous 403 + challenge, OIDC mint, authenticated replay) is green end to end. The upload leg currently 403s on beta for minted tokens that can read; under server-side investigation, independent of this client change.

### AI Disclosure

Tools: Claude Code (Fable 5)

### DIFF
--- a/crates/rattler-bin/src/commands/client.rs
+++ b/crates/rattler-bin/src/commands/client.rs
@@ -70,7 +70,7 @@ pub fn create_client_with_middleware_for_channels(
         {
             continue;
         }
-        let Some(options) = TrustedPublishingOptions::for_host(url) else {
+        let Some(options) = TrustedPublishingOptions::for_server(url) else {
             continue;
         };
         let mut server = url.clone();
--- a/crates/rattler-bin/src/commands/client.rs
+++ b/crates/rattler-bin/src/commands/client.rs
@@ -1,13 +1,39 @@
 use std::{collections::HashMap, sync::Arc};
 
 use miette::{Context, IntoDiagnostic};
-use rattler_networking::{AuthenticationMiddleware, AuthenticationStorage};
+use rattler_conda_types::Channel;
+use rattler_networking::{
+    AuthChallengeMiddleware, AuthenticationMiddleware, AuthenticationStorage,
+    trusted_publishing::{TrustedPublishingFlow, TrustedPublishingOptions},
+};
 use reqwest::Client;
+use url::Url;
 
 pub const USER_AGENT: &str = concat!("rattler/", env!("CARGO_PKG_VERSION"));
 
+/// Hosts for which the CLI is willing to perform CI OIDC trusted publishing.
+/// Restricting this keeps the CLI from volunteering CI identity tokens to
+/// arbitrary channel hosts.
+fn is_prefix_dev_host(host: &str) -> bool {
+    host == "prefix.dev" || host.ends_with(".prefix.dev")
+}
+
 /// Creates an HTTP client with the middleware stack used by the CLI for remote fetches.
 pub fn create_client_with_middleware() -> miette::Result<reqwest_middleware::ClientWithMiddleware> {
+    create_client_with_middleware_for_channels(&[])
+}
+
+/// Like [`create_client_with_middleware`], but additionally layers one
+/// [`AuthChallengeMiddleware`] per unique `https` channel host that matches
+/// the prefix.dev host policy. On a `WWW-Authenticate` challenge from such a
+/// host, the middleware acquires a token via CI OIDC trusted publishing
+/// (audience = the channel host) and replays the request.
+///
+/// This is the reference wiring for challenge-reactive private-channel
+/// reads (see prefix-dev/pixi#6318).
+pub fn create_client_with_middleware_for_channels(
+    channels: &[Channel],
+) -> miette::Result<reqwest_middleware::ClientWithMiddleware> {
     let download_client = Client::builder()
         .no_gzip()
         .user_agent(USER_AGENT)
@@ -18,11 +44,40 @@ pub fn create_client_with_middleware() -> miette::Result<reqwest_middleware::Cli
     let authentication_storage =
         AuthenticationStorage::from_env_and_defaults().into_diagnostic()?;
 
-    let client = reqwest_middleware::ClientBuilder::new(download_client.clone())
-        .with_arc(Arc::new(AuthenticationMiddleware::from_auth_storage(
-            authentication_storage.clone(),
-        )))
-        .with(rattler_networking::OciMiddleware::new(download_client));
+    let mut client =
+        reqwest_middleware::ClientBuilder::new(download_client.clone()).with_arc(Arc::new(
+            AuthenticationMiddleware::from_auth_storage(authentication_storage.clone()),
+        ));
+
+    // The mint exchange must not itself go through AuthChallengeMiddleware
+    // (it would recurse), so it uses a plain client.
+    let mint_client = reqwest_middleware::ClientBuilder::new(download_client.clone()).build();
+
+    let mut seen_hosts = std::collections::HashSet::new();
+    for channel in channels {
+        let url: &Url = channel.base_url.as_ref();
+        if url.scheme() != "https" {
+            continue;
+        }
+        let Some(host) = url.host_str() else {
+            continue;
+        };
+        if !is_prefix_dev_host(host) || !seen_hosts.insert(host.to_string()) {
+            continue;
+        }
+        let Some(options) = TrustedPublishingOptions::for_host(url) else {
+            continue;
+        };
+        let mut server = url.clone();
+        server.set_path("/");
+        server.set_query(None);
+        client = client.with_arc(Arc::new(AuthChallengeMiddleware::new(
+            server,
+            Arc::new(TrustedPublishingFlow::new(options, mint_client.clone())),
+        )));
+    }
+
+    let client = client.with(rattler_networking::OciMiddleware::new(download_client));
     #[cfg(feature = "s3")]
     let client = client.with(rattler_networking::S3Middleware::new(
         HashMap::new(),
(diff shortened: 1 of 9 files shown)


docs/superpowers/specs/2026-06-10-auth-challenge-middleware-design.md
```diff
@@ -0,0 +1,197 @@
+# Auth Challenge Middleware — Design
+
+**Date:** 2026-06-10
+**Crate:** `rattler_networking`
+**Motivation:** [pixi#6318](https://github.com/prefix-dev/pixi/issues/6318) — OIDC "repository access" never auto-authenticates for private channel *reads*. The existing `TrustedPublishingMiddleware` (added in rattler#2426) is channel-scoped, proactively mints, and has no consumers. This design replaces it with a host-scoped, challenge-reactive, pluggable middleware.
```
wolfv: I think this file doesn't need to be committed, same as the other plan file.

crates/rattler-bin/src/commands/client.rs
```diff
@@ -18,11 +47,44 @@ pub fn create_client_with_middleware() -> miette::Result<reqwest_middleware::Cli
...
+            continue;
+        };
+        let mut server = url.clone();
+        // Cosmetic: the middleware scopes on scheme+host+port and ignores
+        // path/query; normalizing just keeps Debug output tidy.
+        server.set_path("/");
+        server.set_query(None);
+        client = client.with_arc(Arc::new(AuthChallengeMiddleware::new(
+            server,
+            Arc::new(TrustedPublishingFlow::new(options, mint_client.clone())),
```
baszalmstra: I would just make a `PrefixAuthAmbientFlow` that wraps a `TrustedPublihsingFlow` that we unconditionally add. I would make the AuthChallengeMiddleware not dependent on server. I would just only give it A list of TrustedPublihsingFlows. When the middleware detects a 401 with a WWW-AUthentication header, it asks each registered flow if it can mint a token. The PrefixAuthAmbientFlow only tries to mint a token if we are in CI and the bearer challange lists a prefix domain. 

You then add a AuthChallengeMiddleware::default(). That just constructs an instance with default providers, we can then use that in pixi directly without any additional configuration.

crates/rattler_networking/src/challenge_middleware.rs
```diff
@@ -0,0 +1,1001 @@
...
+///
+/// Implementations decide which challenges they support (e.g. only scheme
+/// `Bearer`) and how to acquire the token (OIDC exchange, device flow, ...).
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+pub trait AuthFlow: Send + Sync + fmt::Debug {
+    /// Respond to `challenges` received from `url`.
+    ///
+    /// Return `Ok(None)` when this flow does not apply (e.g. unsupported
+    /// scheme, or not running in a CI environment) — the middleware caches
```
baszalmstra: You can make the comments terser and remove emdashes.

crates/rattler_networking/src/challenge_middleware.rs
```diff
@@ -0,0 +1,1001 @@
...
+
+fn attach_bearer(req: &mut reqwest::Request, header: reqwest::header::HeaderValue) {
+    req.headers_mut()
+        .insert(reqwest::header::AUTHORIZATION, header);
+}
+
+#[cfg_attr(not(target_arch = "wasm32"), async_trait::async_trait)]
+#[cfg_attr(target_arch = "wasm32", async_trait::async_trait(?Send))]
+impl Middleware for AuthChallengeMiddleware {
+    async fn handle(
```
baszalmstra, line 393: When doing many requests, this can be called many times concurrently. We should make sure that we only ever mint the token once. Is that properly handled here?

baszalmstra (changes requested):

crates/rattler_networking/src/challenge_middleware.rs
```diff
@@ -0,0 +1,1207 @@
...
+            challenge.params.insert(key, val);
+        }
+    }
+    challenges
+}
+
+/// Parse one `key=value` or `key="quoted value"` auth-param. Returns `None`
+/// for non-params (e.g. token68 blobs like `YII=`, which have an empty
+/// "value" after the trailing `=`).
+fn parse_param(s: &str) -> Option<(String, String)> {
```
baszalmstra: There is a lot custom parsing code in here which seems like something we are not alone in. There is actually a crate for this: https://crates.io/crates/http-auth . Lets use that instead!

crates/rattler_networking/src/challenge_middleware.rs
```diff
@@ -0,0 +1,1077 @@
...
+    Empty,
+    /// Every flow declined or failed; stop asking.
+    Disabled,
+    /// A previously acquired token.
+    Token(CachedToken),
+}
+
+/// Per-origin cache slot plus the gate serializing its acquisition.
+#[derive(Debug, Default)]
+struct OriginEntry {
```
baszalmstra, line 202: Can we use futures Shared instead?

crates/rattler_networking/src/challenge_middleware.rs
```diff
@@ -0,0 +1,1077 @@
...
+        let url = req.url().clone();
+        let response = next.clone().run(req, extensions).await?;
+
+        let challenges = parse_challenges(response.headers());
+        if challenges.is_empty() {
+            return Ok(response);
+        }
+        let Some(mut retry_req) = retry_req else {
+            return Ok(response);
+        };
```
baszalmstra: Should we not just immediately revert to `next` if the clone of the request failed? And should we warn on that case? Now that is silently dropped. WDYT?

baszalmstra (approved):

PR #2507, opened 2026-06-12
fix: include sha256 in cache key for file:// packages

baszalmstra:
### Description

Packages installed from a local `file://` path could be served stale from the package cache. When two different packages shared the same name, version, and build string, the first one cached was reused for all later installs, even though the content differed. This change makes the cache recognize that the package changed and re-extract it, so installs from local paths always get the correct content.

This fixes a bug in pixi which was uncovered when trying to update rattler in pixi. 

### How Has This Been Tested?

Added a regression test that caches a local package, then requests a different package with the same name, version, and build string, and verifies the cache replaces the stale entry instead of serving it. Ran the full test suites for the `rattler_cache` and `rattler` crates, all passing.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/rattler/src/install/installer/mod.rs
+++ b/crates/rattler/src/install/installer/mod.rs
@@ -851,7 +851,7 @@ async fn populate_cache(
         })?;
 
         cache
-            .get_or_fetch_from_path(&path, reporter)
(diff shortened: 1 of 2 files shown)


Hofer-Julian (approved):
Looks good!

PR #2511, opened 2026-06-15
fix(rattler_index): use atomic writes for repodata.json to support memory-mapped files

wolfv:
### Description

This PR fixes an issue on Windows where `index_fs` fails when reindexing a channel whose `repodata.json` is currently memory-mapped by the repodata gateway.

On Windows, attempting to truncate a memory-mapped file in place fails with `ERROR_USER_MAPPED_FILE` (os error 1224). The fix implements atomic writes by:

1. Writing to a temporary directory (`.tmp`) on the same volume as the channel root
2. Renaming the temp file over the target file instead of truncating in place
3. The rename succeeds even while the file is memory-mapped, as long as it's opened with `FILE_SHARE_DELETE` (which the gateway does)

The `.tmp` directory is placed inside the channel root to ensure it's on the same volume/network share, keeping the rename atomic. It's automatically ignored when enumerating platform subdirectories since `.tmp` doesn't parse as a valid `Platform`.

This change is applied to both `index_fs_with_channel_metadata` and `ensure_channel_initialized_fs_with_channel_metadata` functions.

### How Has This Been Tested?

Added a regression test `test_index_fs_over_memory_mapped_repodata` that:
1. Generates initial `repodata.json` via `index_fs`
2. Memory-maps the file using `SparseRepoData::from_file` (same as the gateway does)
3. Runs `index_fs` again while the mapping is still active
4. Verifies the operation succeeds and the atomic write directory is created

The test exercises the actual regression on Windows and ensures the functionality works correctly on other platforms as well.

### Checklist

https://claude.ai/code/session_01FEVryK7L3mvcahvkxpVq6t

### DIFF
--- a/crates/rattler_index/tests/integration/atomic_write.rs
+++ b/crates/rattler_index/tests/integration/atomic_write.rs
@@ -1,0 +1,69 @@
+use rattler_conda_types::{Channel, ChannelConfig, Platform};
+use rattler_index::{IndexFsConfig, PackageRevisionAssignment, index_fs};
+use rattler_repodata_gateway::sparse::SparseRepoData;
+
+/// Regression test for indexing a channel whose `repodata.json` is currently
+/// memory-mapped, as the repodata gateway does via
+/// [`SparseRepoData::from_file`].
+///
+/// On Windows, re-writing a memory-mapped file in place fails with
+/// `ERROR_USER_MAPPED_FILE` (os error 1224). `index_fs` now writes through an
+/// atomic temp dir on the same volume and renames over the target, so the
+/// mapped file is replaced rather than truncated. Combined with the gateway
+/// opening the file with `FILE_SHARE_DELETE`, the rename succeeds while the
+/// mapping is still live.
+///
+/// On other platforms the rename is harmless and the test simply asserts that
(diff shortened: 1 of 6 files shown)


crates/rattler_index/tests/integration/atomic_write.rs
```diff
@@ -0,0 +1,69 @@
+use rattler_conda_types::{Channel, ChannelConfig, Platform};
+use rattler_index::{IndexFsConfig, PackageRevisionAssignment, index_fs};
+use rattler_repodata_gateway::sparse::SparseRepoData;
+
+/// Regression test for indexing a channel whose `repodata.json` is currently
+/// memory-mapped, as the repodata gateway does via
+/// [`SparseRepoData::from_file`].
+///
+/// On Windows, re-writing a memory-mapped file in place fails with
```
baszalmstra: can you make the vomments terse?

baszalmstra (changes requested):

crates/rattler_index/tests/integration/atomic_write.rs
wolfv: Done

baszalmstra (approved):

PR #2517, opened 2026-06-17
feat: support OAuth2 authentication in rattler-upload for non-prefix backends

### LINKED ISSUE #2324: rattler-upload does not support oauth2 authentication
### Checklist

### What is the idea?

rattler-upload supports authentication methods but does not support oauth2 yet. The logic for refresh tokens also lives in the AuthenticationMiddleware which upload doesnt have access to. I think we should refactor this slightly so that tokens can also be refreshed outside the middleware. 

### Why is this needed?

_No response_

### What should happen?

_No response_

### Additional Context

_No response_

mvanhorn:
### Description

The Quetz, Anaconda, Cloudsmith, and Artifactory upload backends did not accept OAuth2 credentials: they looked up authentication via the non-refreshing `storage.get_by_url(...)` and only matched their native token variant (`CondaToken`/`BearerToken`), returning a wrong-authentication-type error for an `Authentication::OAuth` entry.

The prefix.dev path already solves this in `prefix.rs::fetch_token_from_storage`, which calls the async `storage.get_by_url_refreshed(url).await` (transparently invoking `maybe_refresh_oauth` when the access token is near expiry) and accepts `Authentication::OAuth { access_token, .. }`.

This PR switches the four non-prefix backends to the same `get_by_url_refreshed(...).await` lookup and adds an `Authentication::OAuth { access_token, .. }` match arm that yields the (refreshed) access token, mirroring `prefix.rs`. Each backend continues to send the resulting token over its existing transport, so the same OAuth credential now works uniformly across all upload backends and is automatically refreshed before use.

Fixes #2324

### How Has This Been Tested?

- `cargo test -p rattler_upload` (23 tests pass).
- Added `test_cloudsmith_upload_oauth_from_storage_success`: stores an `Authentication::OAuth` credential (with a non-expired `access_token`, so no refresh network call fires) for the Cloudsmith host in a `MemoryStorage` backend, runs the upload against a mock server, and asserts the server receives the OAuth access token.
- The pre-existing `WrongAuthenticationType` / `MissingApiKey` error paths are unchanged and still covered by their existing tests (e.g. `test_cloudsmith_upload_missing_api_key`).
- `cargo fmt --check` and `cargo check -p rattler_upload` are clean.

### AI Disclosure

Tools: Codex

### Checklist:

### DIFF
--- a/crates/rattler_upload/src/upload/mod.rs
+++ b/crates/rattler_upload/src/upload/mod.rs
@@ -91,8 +91,12 @@ pub async fn upload_package_to_quetz(
 ) -> miette::Result<()> {
     let token = match quetz_data.api_key {
         Some(api_key) => api_key,
-        None => match storage.get_by_url(Url::from(quetz_data.url.clone())) {
+        None => match storage
+            .get_by_url_refreshed(Url::from(quetz_data.url.clone()))
+            .await
+        {
             Ok((_, Some(Authentication::CondaToken(token)))) => token,
+            Ok((_, Some(Authentication::OAuth { access_token, .. }))) => access_token,
             Ok((_, Some(_))) => {
                 return Err(miette::miette!("A Conda token is required for authentication with quetz.
                         Authentication information found in the keychain / auth file, but it was not a Conda token"));
@@ -145,7 +149,10 @@ pub async fn upload_package_to_artifactory(
 ) -> miette::Result<()> {
     let token = match artifactory_data.token {
         Some(t) => t,
-        _ => match storage.get_by_url(Url::from(artifactory_data.url.clone())) {
+        _ => match storage
+            .get_by_url_refreshed(Url::from(artifactory_data.url.clone()))
+            .await
+        {
             Ok((_, Some(Authentication::BearerToken(token)))) => token,
             Ok((
                 _,
@@ -159,6 +166,7 @@ pub async fn upload_package_to_artifactory(
                 );
                 password
             }
+            Ok((_, Some(Authentication::OAuth { access_token, .. }))) => access_token,
             Ok((_, Some(_))) => {
                 return Err(miette::miette!("A bearer token is required for authentication with artifactory.
                             Authentication information found in the keychain / auth file, but it was not a bearer token"));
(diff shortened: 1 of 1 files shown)


baszalmstra:
Do you have a way to test this properly against these other backends? Like quetz doesnt support OAuth2 authentication in the first place does it?

mvanhorn:
Honest answer: no, I don't have live accounts on these backends - the coverage is unit-level (auth store lookup and header shape, mirroring prefix.rs), not end to end.

And on quetz specifically you're right: its OAuth2 support is for web login (GitHub/Google), while the upload API authenticates with quetz API keys, so an OAuth access token from the rattler auth store wouldn't authenticate an upload there anyway. Happy to drop quetz from this PR.

The defensible case is Artifactory, where access tokens are consumed as standard bearer tokens on upload endpoints. Cloudsmith and anaconda.org are deployment-dependent and I can't demonstrate them live either.

Two ways I could narrow this - your call: (a) keep only the Artifactory OAuth arm and drop the other three, or (b) keep the get_by_url_refreshed unification everywhere (it fixes stale-token refresh regardless of auth type) but only add the OAuth match arm for Artifactory. I'd lean (b).

baszalmstra:
I would only accept this PR if we can demonstrate that this functionality works. So adding implementations for backends that dont support oauth in the first place makes no sense to me.

mvanhorn:
Fair, and I went and checked each backend rather than guessing. You are right about two of the four and I will drop them.

quetz: agreed, and I should have dropped it back in June. Its OAuth2 is web login, uploads authenticate with quetz API keys, so an OAuth access token from the auth store would never authenticate an upload there.

anaconda: also wrong, and this one is a real bug in the PR rather than a scope question. `anaconda.rs` builds `Authorization: token {token}`, the binstar API-token scheme. Feeding an OAuth access token into that slot produces `Authorization: token <access_token>`, which is not a bearer credential and anaconda.org will not accept it. Dropping this arm too.

That leaves the two I think do hold up:

cloudsmith: our client sends `X-Api-Key: <token>`. Cloudsmith's OIDC docs describe passing a JWT from their token-exchange endpoint in exactly that header, working across all their API endpoints including the format-specific ones, with a 2 hour lifetime. So the credential shape this PR produces is the one Cloudsmith documents for uploads.

artifactory: JFrog access tokens are consumed as `Authorization: Bearer <token>`, and Artifactory supports conda repositories for deploy. Our artifactory path already accepts `BearerToken`, so the OAuth arm only routes an OAuth access token into that same transport.

On demonstrating it: I do not have a live Cloudsmith org or a JFrog instance with a conda repo, and I would rather not claim an end to end run I have not done. So, two questions:

1. Does rattler have, or would you accept, a CI fixture or test deployment for these upload paths? If there is something I can point at, I will redo this against it.
2. If not, would you take the narrowed PR (cloudsmith and artifactory only) on the documented credential shapes plus the mock-server tests, or would you rather this stay closed until someone with real accounts can verify it?

Either answer is fine. I would rather land the smaller correct thing than argue for the bigger one.

baszalmstra:
I dont appreciate pasting the output from claude directly here as a reply to my comments. This is against our AI policy.

mvanhorn:
Understood, and sorry. I should not have pasted assistant-written text into that reply. The conclusions were mine, the wording was not, and that is on me.

PR #2522, opened 2026-06-19
fix: coalesce concurrent OAuth token refresh

tdejager:
### Description

This PR makes OAuth token refresh in `rattler_networking` coalesce concurrent refresh attempts for the same stored credential.

The motivation is a somewhat theoretical / hard-to-reproduce failure mode we suspect can happen when multiple requests start at the same time with an expired OAuth access token. For example, pixi may request both `noarch/repodata_shards.msgpack.zst` and `<platform>/repodata_shards.msgpack.zst` for a single channel. If both requests notice that the access token is expired, they can both try to use the same refresh token.

That is problematic when the authorization server rotates refresh tokens: the first refresh succeeds and stores the rotated token, while the second refresh can fail with `invalid_grant` because it is still using the old refresh token. Before this change, that failure path could still attach the expired access token to the original request.

I do not have a reliable reproducer for the actual pixi failure, but this matches the kind of occasional 401-on-repodata behavior we have seen. Although I suspect a server-side change (prefix.dev) is required as well, probably when the refresh token TTL has also passed.

This PR:
- adds a per-`AuthenticationStorage` / per-auth-key refresh gate
- re-reads the stored credential after waiting for the refresh gate, so concurrent requests can reuse a freshly stored token
- avoids attaching an expired access token when refresh failed
- adds a timeout to refresh requests so a stuck token endpoint does not block indefinitely
- keeps the refresh outcome internal to `rattler_networking` instead of adding new public API
- adds tests for rotating refresh tokens, failed refresh, and middleware-level behavior

No linked issue.

### How Has This Been Tested?

```shell
pixi run cargo-fmt
pixi run -- cargo nextest run -p rattler_networking oauth_refresh authentication_middleware::tests::expired_oauth_with_failed_refresh_sends_no_authorization_header authentication_middleware::tests::concurrent_oauth_refresh_is_coalesced_by_authentication_middleware
pixi run -- cargo nextest run -p rattler_networking
```

All rattler_networking tests pass locally.

### AI Disclosure

Tools: pi / ChatGPT / Opus

Prompt:

```plaintext
Investigate a suspected pixi/rattler OAuth refresh race that can lead to 401s on public prefix.dev repodata shard index requests, then implement a rattler-side fix. Use jj for the checkout. Keep the change focused, add tests, and avoid adding speculative public API.
```

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -860,6 +860,7 @@ checksum = "31b698c5f9a010f6573133b09e0de5408834d0c82f8d7475a89fc1867a71cd90"
 dependencies = [
  "axum-core",
  "bytes",
+ "form_urlencoded",
  "futures-util",
  "http 1.4.2",
  "http-body 1.0.1",
@@ -875,6 +876,7 @@ dependencies = [
  "serde_core",
  "serde_json",
  "serde_path_to_error",
+ "serde_urlencoded",
  "sync_wrapper",
  "tokio",
  "tower",
--- a/crates/rattler_networking/Cargo.toml
+++ b/crates/rattler_networking/Cargo.toml
@@ -91,7 +91,7 @@ tokio = { workspace = true, features = ["macros", "time"] }
 # `stream` (test-only) lets us build an unclonable streaming-body request to
 # exercise the challenge-middleware's no-replay path.
 reqwest = { workspace = true, features = ["json", "form", "stream"] }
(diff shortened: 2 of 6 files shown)


nichmor (approved):
by human reading I think it make sense!

baszalmstra (approved):

PR #2536, opened 2026-06-25
feat(networking): surface AuthenticationExpired for unrecoverable OAuth logins

nichmor:
### Description

Builds on the OAuth refresh handling already in `main` (`OAuthRefreshOutcome` /
`OAuthRefreshFailure`, which drops an expired access token when it cannot be
refreshed and sends the request unauthenticated so the auth-challenge
middleware can take over).

This adds a typed **`AuthenticationExpired { host }`** error for the case where
the stored login itself is no longer usable. In `AuthenticationMiddleware`:

- after a refresh attempt, inspect `refresh_result.failure()` and treat only
  `MissingRefreshToken` and `ReauthenticationRequired` as "the login is dead"
  (transient refresh problems keep the existing log-and-continue behavior);
- only when the dead credential was actually dropped (so the request goes out
  unauthenticated) remember the host;
- if that request still comes back `401`/`403`, return
  `AuthenticationExpired { host }` so callers can prompt the user to
  re-authenticate (e.g. re-run their login command).

`AuthenticationExpired` is re-exported from the crate root.

### How Has This Been Tested?

- `pixi run cargo-fmt` and `pixi run -- cargo clippy -p rattler_networking --all-targets` are clean.
- Relies on the existing `oauth_refresh` / `authentication_middleware` unit
  tests in `main` for the refresh-outcome paths.

```
pixi run -- cargo nextest run -p rattler_networking
```

> Opened as a **draft**: no test yet asserts the new `AuthenticationExpired`
> path end to end — happy to add one before marking ready.

### AI Disclosure

Tools: Claude (Claude Code, Opus 4.8)

```plaintext
what we have done in this branch?
can you create a draft pr?
ok, just make this branch to be exactly the same as main for now
can you push it again? I've made some changes there
```

### Checklist:

### DIFF
--- a/crates/rattler_networking/src/authentication_middleware.rs
+++ b/crates/rattler_networking/src/authentication_middleware.rs
@@ -11,9 +11,21 @@ use reqwest_middleware::{Middleware, Next};
 use url::Url;
 
 use crate::{
-    Authentication, AuthenticationStorage, authentication_storage::AuthenticationStorageError,
-    oauth_refresh,
+    Authentication, AuthenticationStorage,
+    authentication_storage::AuthenticationStorageError,
+    oauth_refresh::{self, OAuthRefreshFailure},
 };
+
+/// Error returned when a request to `host` failed authentication while the
+/// stored OAuth credential for that host was expired and could not be
+/// refreshed. Surfaced so callers can tell the user to authenticate again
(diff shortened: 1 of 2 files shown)


wolfv:
@nichmor should we add a test?

PR #2537, opened 2026-06-25
feat: add offline mode to rattler CLI

baszalmstra:
## Summary

- Add an offline middleware in `rattler_networking` that rejects outgoing requests before they reach the rest of the client stack.
- Add a global `--offline` flag to the `rattler` CLI and pass it explicitly into commands that create HTTP clients.
- Use cached repodata only for solve/create/search/exec while offline, and avoid auth/upload network calls.

## Validation

- `cargo fmt --all`
- `cargo check -p rattler_networking -p rattler-bin`
- `cargo test -p rattler_networking offline_middleware --lib`
- `git diff --check`

## AI Disclosure

Tools: pi coding agent

### DIFF
--- a/crates/rattler-bin/src/commands/auth.rs
+++ b/crates/rattler-bin/src/commands/auth.rs
@@ -3,6 +3,8 @@ use rattler::cli::auth;
 
 pub type Opt = auth::Args;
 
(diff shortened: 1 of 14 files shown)


pavelzw (approved):
Cc @delsner we can use this in pixi-pack

PR #2538, opened 2026-06-26
feat: Support remote URLs in inject, fetch-file, and inspect commands

pavelzw:
<details>
<summary>🤖 yap</summary>

### Description

Makes the single-archive `rattler` subcommands symmetric: each accepts **either a local file path or a remote URL** for a conda package (`.conda` / `.tar.bz2`).

| Command | Before | After |
|---|---|---|
| `inspect` | URL only | path or URL |
| `fetch-file` | URL only | path or URL |
| `inject-into-prefix` | local path only | path or URL |

Already symmetric on `main`, so unchanged in behaviour: `extract`, `compare-packages`, `whoneeds`.

Implementation:

- New shared `commands/package_source.rs`: `PackageSource::{Url, Path}` with `parse` (converts `file://` URLs to paths and treats a single-letter scheme as a Windows drive letter — `Url::parse` otherwise reads `C:\pkgs\foo.conda` as scheme `c`), `open` (via `PackageArchive::from_url` / `from_path`), and `client_for` (only builds an HTTP client when some source is remote, so local-only invocations never touch the network).
- `inspect` / `fetch-file`: open the archive through `PackageArchive`, so local and remote share one code path. `fetch-file` uses `PackageArchive::read_file`.
- `inject-into-prefix`: resolves local archives with `rattler_package_streaming::fs::repodata_record_from_package_archive`. Remote archives are streamed to a temp dir, resolved the same way, recorded under their **remote URL** (not the temp path), and pre-seeded into the shared `PackageCache` so the installer does not download them a second time. Now honours `--offline`.
- `compare-packages` switched to the shared module, removing its private copy of the same parsing logic.

### How Has This Been Tested?

- `cargo clippy -p rattler-bin --all-targets -- -D warnings -Dclippy::dbg_macro` (CI flags): clean.
- `cargo test -p rattler-bin`: 20 passed, including the `inject` tests (local archives, duplicate / already-installed rejection) and new `package_source` tests for URL vs. path classification (http/https/s3, relative and absolute paths, Windows drive letters, `file://`).

### Checklist:

</details>

prompt:

In rattler bin Give me an overview of subcommands that only support local file system conda packages and which only support remote conda packages. Is it feasible to support both?

image

lets add url support to inject-into-prefix and local fs support for inspect + fetch-file

Please rebase the pr with main, handle conflicts and check for eventual new commands that we want to apply this patch to as well

https://claude.ai/code/session_01QkUDTwKbtUcLCwrnuzwYn5

### DIFF
--- a/crates/rattler-bin/src/commands/compare_packages.rs
+++ b/crates/rattler-bin/src/commands/compare_packages.rs
@@ -15,7 +15,8 @@ use rattler_package_streaming::{
     archive::{ArchiveEntryKind, PackageArchive, Section},
 };
 use sha2::{Digest, Sha256};
-use url::Url;
+
+use super::package_source::{PackageSource, client_for};
 
 /// Compare two conda packages and report the differences.
 ///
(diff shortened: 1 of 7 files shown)


baszalmstra (approved):

wolfv:
@pavelzw ready to undraft?

PR #2543, opened 2026-06-30
feat: Implementation of the NFS backend for the Rattler-VFS crate

Dagmar-Dinjens:
### Description

Within this PR I added the first functionality for the NFS backend of the VFS crate.

The crate is now able to setup an NFS-server and a mount to dynamically attach and use this as a cache for the 

you can run this crate by running cargo run and passing in a pixi lock file, 
```
cargo run -- --PIXI_LOCK /path/to/pixi.lock
```

Screenshot 2026-06-30 at 13 15 53

when its ran it will tell you where it's at and the cache for the specific lock file should be virtually mounted for the destination.

Files within here can be used, even executables and prefix replacements are adapted where needed on the fly.

Screenshot 2026-06-30 at 13 14 03

This are the first steps towards realising #2059 

### How Has This Been Tested?

There are tests for the logic within the test folder, 
the e2e test should be run with a pixi.lock within the data folder and the /cache folder. this can be added through initialising a pixi environment within this folder and then running the tests

### AI Disclosure
Tools: Claude

### Checklist:


crates/rattler_vfs/Cargo.toml
```diff
@@ -0,0 +1,42 @@
...
+async-trait = "0.1"
+clap = { workspace = true, features = ["derive"] }
+libc = "0.2"
+
+memchr = { workspace = true }
+memmap2 = { workspace = true }
+sha2 = { workspace = true }
+
+nfs3_server = "0.5"
+nfs3_types = "0.3.4"
```
baszalmstra, line 22: Use workspace dependencies.

crates/rattler_vfs/src/main.rs
```diff
@@ -0,0 +1,140 @@
...
+    pub pixi_lock: PathBuf,
+    pub cache_origin: PathBuf,
+    pub mount_dir: PathBuf,
+    pub mount_type: MountBackend,
+    pub environment_name: String,
+    pub download: bool,
+}
+
+fn handle_input_arguments() -> anyhow::Result<MountArgs> {
+    let matches = Command::new("mount")
```
baszalmstra, line 53: Please use clap derive for this.

crates/rattler_vfs/src/backends/mod.rs
```diff
@@ -0,0 +1,19 @@
+#[cfg(unix)]
+pub mod nfs;
+#[cfg(unix)]
```
baszalmstra, line 3: Can we add these `cfg`s to the files themselves?

crates/rattler_vfs/src/backends/mod.rs
```diff
@@ -0,0 +1,19 @@
...
```
baszalmstra, line 1: No windows support?

baszalmstra (commented):
- At least the public types need way better documentation.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
+use crate::metadata::{CustomPrefixPlaceholder, FSMetadata};
```
baszalmstra, line 1: Id like to see top level module documentation that explains what this crate does, and how to use its API.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+pub mod codesign;
+pub mod metadata;
+pub mod mount;
+pub mod prefix_replacement;
+pub mod tests;
+pub mod virtual_fs_core;
+
+pub use mount::{MountBackend, MountSession};
+
+pub async fn mount_environment(
```
baszalmstra, line 23: I dont think this high-level function is a good entrypoint. It requires a path to a lock-file which you might not have yet. I think a better API design would be to pass in a PackageCache and the packages that should be served from it. I think ideally using a newtype. This function is also missing documentation.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+) -> anyhow::Result<Box<dyn MountSession>> {
+    let package_refs = solve_environment(&pixi_lock, &environment_name)?;
+
+    let package_dirs = resolve_package_dirs(package_refs, cache_origin, download_if_missing)?;
+
+    let mut metadata = vec![FSMetadata::new_directory(PathBuf::from("."), 0)];
+    let mut directory_indices = HashMap::new();
+    directory_indices.insert(PathBuf::from("."), 0);
+
+    for package_dir in package_dirs {
```
baszalmstra, line 39: There is room for parallelization here.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+                            )
+                            .await?,
+                        );
+                    }
+                    anyhow::Ok(dirs)
+                })
+            })
+            .join()
+            .map_err(|err| anyhow!("package cache resolution thread panicked {err:?}"))?
+    })
```
baszalmstra, line 84: Why run this on a new thread? The function is already async and its calling other async functions.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+            UrlOrPath::Path(path) => {
+                return Err(anyhow!(
+                    "package '{}' lockfile location is a local path ({}); cannot download",
+                    package_ref.name(),
+                    path
+                ));
+            }
+        };
+
+        println!("Resolving {} from {}", package_ref.name(), url);
```
baszalmstra, line 110: This is logging. Use the `tracing` crate for that.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+) -> anyhow::Result<PathBuf> {
+    let package_data = package_ref
+        .as_binary_conda()
+        .ok_or_else(|| anyhow!("only binary conda packages can be mounted"))?;
+    let cache_key = CacheKey::from(&package_data.package_record);
+
+    if download_if_missing {
+        let url = match &package_data.location {
+            UrlOrPath::Url(url) => url.clone(),
+            UrlOrPath::Path(path) => {
```
baszalmstra, line 101: This seems trivial to support.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+pub use mount::{MountBackend, MountSession};
+
+pub async fn mount_environment(
+    pixi_lock: PathBuf,
+    cache_origin: PathBuf,
+    mount_dir: PathBuf,
+    backend: MountBackend,
+    environment_name: String,
+    download_if_missing: bool,
+) -> anyhow::Result<Box<dyn MountSession>> {
```
baszalmstra, line 30: Dont use anyhow, use thiserror to create a typed error.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+pub mod metadata;
+pub mod mount;
+pub mod prefix_replacement;
+pub mod tests;
+pub mod virtual_fs_core;
+
+pub use mount::{MountBackend, MountSession};
+
+pub async fn mount_environment(
+    pixi_lock: PathBuf,
```
baszalmstra, line 24: Some of these should be passed by reference to avoid cloning.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+        package_dir,
+    )?)
+}
+
+pub fn path_parse(
+    paths_json: PathsJson,
+    package_dir: PathBuf,
+    env_paths: &mut Vec<FSMetadata>,
+    directory_indices: &mut HashMap<PathBuf, usize>,
+    _mount_point: &Path,
```
baszalmstra, line 175: If this is not used, can we remove it?

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,233 @@
...
+
+pub fn get_paths_json(package_dir: &Path) -> anyhow::Result<PathsJson> {
+    Ok(PathsJson::from_package_directory_with_deprecated_fallback(
+        package_dir,
+    )?)
+}
+
+pub fn path_parse(
+    paths_json: PathsJson,
+    package_dir: PathBuf,
```
baszalmstra, line 172: Are these values consumed? Otherwise maybe pass them by reference.

crates/rattler_vfs/src/metadata.rs
```diff
@@ -0,0 +1,143 @@
...
+        }
+    }
+
+    pub fn new_directory(prefix_path: PathBuf, parent: usize) -> Self {
+        FSMetadata::FSDirectory(FSDirectory::new(prefix_path, parent))
+    }
+
+    pub fn new_file(
+        file_name: OsString,
+        parent: usize,
```
baszalmstra, line 102: This is an index, instead of passing around random anonymous `usize` types, I prefer to use a new type. This makes it much harder to mix up different indexes.

crates/rattler_vfs/src/mount.rs
```diff
@@ -0,0 +1,53 @@
...
+#[cfg(unix)]
+use crate::backends::nfs::NfsProvider;
+use crate::{metadata::FSMetadata, virtual_fs_core::VirtualFSCore};
+
+pub trait MountSession: Send + Sync {
+    fn unmount(self: Box<Self>) -> anyhow::Result<()>;
+}
+
+#[derive(Debug, Clone, Copy)]
+pub enum MountBackend {
```
baszalmstra, line 13: This requires more public facing documentation.

crates/rattler_vfs/src/mount.rs
```diff
@@ -0,0 +1,53 @@
...
+        match self {
+            #[cfg(unix)]
+            MountBackend::Nfs => {
+                let fs = Arc::new(VirtualFSCore::new(metadata, mount_point.clone()));
+                NfsProvider::mount(fs, mount_point).await
+            }
+            // A windows machine would not be able to setup the server part of this implementation
+            // When it would be mounting to a separate server this could work with nfs on windows
+            #[cfg(not(unix))]
+            MountBackend::Nfs => {
```
baszalmstra, line 33: Why not? Windows also supports nfs?

crates/rattler_vfs/src/mount.rs
```diff
@@ -0,0 +1,53 @@
...
+        }
+    }
+}
+
+/// Picks the backend to mount for, because NFS should work on most systems this is chosen as the default.
+impl From<&str> for MountBackend {
+    fn from(_value: &str) -> Self {
+        MountBackend::Nfs
+    }
+}
```
baszalmstra, line 45: This cannot be infallible. If I pass in `bla` now it would also just become nfs.

crates/rattler_vfs/src/mount.rs
```diff
@@ -0,0 +1,53 @@
...
+#[derive(Debug, Clone, Copy)]
+pub enum MountBackend {
+    Nfs,
+}
+
+impl MountBackend {
+    #[cfg_attr(not(unix), allow(unused_variables))]
+    pub async fn mount(
+        &self,
+        metadata: Vec<FSMetadata>,
```
baszalmstra, line 21: I think it would be good to add a wrapper type around Vec<FSMetadata>.

crates/rattler_vfs/src/prefix_replacement.rs
baszalmstra, line 1: Is this implementation different from what is inside the `rattler` crate? Why do we have to copy this here?

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+use crate::codesign;
+
+use memmap2::Mmap;
+
+use crate::metadata::{FSFile, FSMetadata};
+
+pub struct VirtualFSCore {
+    /// The filesystem tree in which the files are stored
+    metadata: Vec<FSMetadata>,
+    /// Which
```
baszalmstra, line 27: ?

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+};
+
+#[cfg(target_os = "macos")]
+use crate::codesign;
+
+use memmap2::Mmap;
+
+use crate::metadata::{FSFile, FSMetadata};
+
+pub struct VirtualFSCore {
```
baszalmstra, line 24: Does this need to be pub facing?

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+    /// The filesystem tree in which the files are stored
+    metadata: Vec<FSMetadata>,
+    /// Which
+    mount_point: PathBuf,
+    /// Current files
+    open_files: Mutex<HashMap<u64, Mmap>>,
+    open_handles: Mutex<HashMap<usize, u64>>,
```
baszalmstra, line 31: Instead of a HashMap, we should probably use a Vec with a freelist so that reads are `O(1)`. Also an RWLock or even a lock-free datastructure might be more suitable here since reads probably occur more often than writes.

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+    next_fh: AtomicU64,
+    /// Lazily-populated cache for Mach-O binaries after binary prefix replacement
+    /// and in-place re-signing. Keyed by metadata index. macOS only.
+    #[cfg(target_os = "macos")]
```
baszalmstra, line 35: This doesnt need to be guarded at compile time does it? We should resign only if the target platform for the mount requires it.

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+        let parent_dir = self.metadata.get(parent_idx)?.as_directory()?;
+
+        parent_dir.children.iter().find_map(|child_idx| {
+            let child = &self.metadata[*child_idx];
+            (child.file_name() == name).then_some(*child_idx)
+        })
+    }
+
+    /// `ino` is a 1-based inode number.
+    pub fn getattr(&self, ino: usize) -> anyhow::Result<VirtualAttr> {
```
baszalmstra, line 102: Again we should use an explicit error type instead of anyhow. No anyhow in libraries.

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+        let idx = self
+            .ino_to_index(ino)
+            .ok_or_else(|| anyhow!("invalid ino {ino}"))?;
+
+        let entry = &self.metadata[idx];
+
+        #[cfg(unix)]
+        let (uid, gid) = unsafe { (libc::getuid(), libc::getgid()) };
+        #[cfg(not(unix))]
+        let (uid, gid) = (0u32, 0u32);
```
baszalmstra, line 112: Should this not be part of metadata? Or be configured, so we only do this once.

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+        let mmap = open_files.get(&fh).ok_or_else(|| anyhow!("invalid fh"))?;
+        if offset >= mmap.len() {
+            return Ok(vec![]);
+        }
+        let end = offset.saturating_add(size).min(mmap.len());
+        Ok(mmap[offset..end].to_vec())
+    }
+
+    /// `ino` is a 1-based inode number. Returns the symlink target bytes.
+    pub fn readlink(&self, ino: usize) -> anyhow::Result<Vec<u8>> {
```
baszalmstra, line 296: I think this requires clarification to indicate the byte encoding. In your current implementation its Utf8.

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+            Ok(target.as_os_str().as_bytes().to_vec())
+        }
+        #[cfg(not(unix))]
+        {
+            Ok(target.to_string_lossy().into_owned().into_bytes())
+        }
+    }
+
+    /// `ino` is a 1-based inode number.
+    pub fn readdir(&self, ino: usize) -> anyhow::Result<Vec<DirectoryEntry>> {
```
baszalmstra, line 322: I think returning a custom iterator here which borrows instead of clones data might be better here. But we can do that later.

crates/rattler_vfs/src/virtual_fs_core.rs
```diff
@@ -0,0 +1,346 @@
...
+            open_files: Mutex::new(HashMap::new()),
+            open_handles: Mutex::new(HashMap::new()),
+            next_fh: AtomicU64::new(1),
+            #[cfg(target_os = "macos")]
+            codesign_cache: Mutex::new(HashMap::new()),
+        }
+    }
+
+    /// Converts a 1-based inode number (as exposed to NFS/callers) to a 0-based metadata index.
+    fn ino_to_index(&self, ino: usize) -> Option<usize> {
```
baszalmstra, line 71: I think for this type it might make sense to use a new type for `ino` instead of using a usize.

typos.toml
```diff
@@ -19,3 +19,4 @@ strat = "strat"
 haa = "haa"
 intoto = "intoto"
 certifi = "certifi"
+placehold = "placehold"
```
baszalmstra, line 22: I dont know where `placehold` is used but I would prefer to just use `placeholder` spelled out in the code instead of adding an exclusion here.

crates/rattler_vfs/src/codesign.rs
```diff
@@ -0,0 +1,206 @@
+use sha2::{Digest, Sha256};
```
baszalmstra, line 1: Wolf also has an implementation: https://github.com/conda/rattler/pull/2554 

Maybe we can use that? Or how is this different?

crates/rattler_vfs/src/backends/nfs.rs
```diff
@@ -0,0 +1,278 @@
...
+pub struct NfsProvider;
+
+pub struct NfsSession {
+    mount_point: PathBuf,
+    server_thread: std::thread::JoinHandle<anyhow::Result<()>>,
+}
+
+impl MountSession for NfsSession {
+    fn unmount(self: Box<Self>) -> anyhow::Result<()> {
+        let status = std::process::Command::new("umount")
```
baszalmstra, line 27: Instead of spawning a subprocess you can simply call the umount or umount2 syscall.

crates/rattler_vfs/src/backends/nfs.rs
```diff
@@ -0,0 +1,278 @@
...
+                    }
+                    Err(err) => {
+                        let _ = ready_tx.send(Err(err.to_string()));
+                        return Err(err.into());
+                    }
+                };
+
+                listener.handle_forever().await.map_err(Into::into)
+            })
+        });
```
baszalmstra, line 75: The function is already async, why not use `tokio::spawn`?

baszalmstra: Also tokio and compio are not really compatible, here you are using tokio while in `main.rs` you use compio, is that on purpose?

crates/rattler_vfs/src/backends/nfs.rs
```diff
@@ -0,0 +1,278 @@
...
+        Ok(())
+    }
+}
+
+#[async_trait]
+impl MountProvider for NfsProvider {
+    async fn mount(
+        fs: Arc<VirtualFSCore>,
+        mount_point: PathBuf,
+    ) -> anyhow::Result<Box<dyn MountSession>> {
```
baszalmstra, line 52: Again, use a custom error type.

crates/rattler_vfs/src/backends/nfs.rs
```diff
@@ -0,0 +1,278 @@
...
+                listener.handle_forever().await.map_err(Into::into)
+            })
+        });
+
+        ready_rx
+            .recv()
+            .map_err(|err| anyhow::anyhow!("NFS server thread exited before binding: {err}"))?
+            .map_err(|err| anyhow::anyhow!("failed binding NFS listener: {err}"))?;
+
+        eprintln!("NFS server listening on {NFS_ADDR}");
```
baszalmstra, line 82: Use `tracing`.

crates/rattler_vfs/src/backends/nfs.rs
```diff
@@ -0,0 +1,278 @@
...
+            // kernel NFS client can discover it on port 11111.
+            let rpcbind_running = rpcbind_register(NFS_PORT);
+            eprintln!(
+                "rpcbind registration: {}",
+                if rpcbind_running {
+                    "ok"
+                } else {
+                    "skipped (rpcbind not running)"
+                }
+            );
```
baszalmstra, line 111: Cant we bind to a random port?

crates/rattler_vfs/src/backends/nfs.rs
```diff
@@ -0,0 +1,278 @@
...
+
+        #[cfg(not(target_os = "linux"))]
+        if !status.success() {
+            return Err(anyhow::anyhow!(
+                "mount command failed with exit code {:?}",
+                status.code()
+            ));
+        }
+
+        eprintln!("mounted at {}", mount_point.display());
```
baszalmstra, line 124: tracing

crates/rattler_vfs/src/backends/nfs.rs
```diff
@@ -0,0 +1,278 @@
...
+    // discovers the ports via portmapper in userspace (binary-mode mount), so
+    // we do NOT specify port= or mountport= here.
+    let opts = "nolock,nfsvers=3,tcp";
+    let source = "127.0.0.1:/";
+    let target = mount_point;
+
+    // Try each candidate in order, stop at the first one that runs (even if it
+    // fails — we want that specific failure, not an exec-not-found error).
+    let candidates: &[&[&str]] = &[
+        &["sudo", "-n", "mount.nfs", "-o", opts, source],
```
baszalmstra, line 147: If this requires admin priviliges we should not manually call sudo, instead the application itself should return a good error message saying that it must run as sudo.

.gitignore
```diff
@@ -29,3 +29,4 @@ test-data/channels/conda-forge/
 
 # Local environment files
 *.local.*
+crates/rattler_vfs/src/tests/data/cache/
```
baszalmstra, line 32: There is also a gitignore folder in the rattler_vfs folder, can we remove this entry here?

baszalmstra (commented):
Since the nfs3 server crate uses tokio I dont think compio buys us anything at all.

baszalmstra:
Can we close this in favor of newer work?

PR #2552, opened 2026-07-02
fix: accept quoted extras lists and enforce CEP 44 group names

wolfv:
### Description

MatchSpec `extras` lists only accepted bare identifiers, so a Python-style quoted list failed while the unquoted form worked:

```
foobar[extras=[science, plot]]        # ✅ worked
foobar[extras=["science", "plot"]]    # ❌ "invalid value for 'item': invalid bracket"
```

The value parser captured the raw list content (`"science", "plot"`) and passed it to `parse_extras`, whose feature-name matcher (`[A-Za-z0-9_-]+`) then choked on the leading quote.

This PR:
- **Accepts optionally single/double-quoted list items**, so `["science", 'plot']` parses the same as `[science, plot]`. This is consistent with CEP 44, which describes the value as "a string or a list of string" and shows the scalar form quoted (`example[extras="group-name"]`).
- **Enforces the CEP 44 group-name grammar `[a-z0-9_.+-]{1,64}`** for every group name (bare and quoted), so we never accept — and later serialize — names that would be invalid metadata per spec. Non-conforming names now return a descriptive `ParseMatchSpecError::InvalidExtraName` instead of an opaque `invalid bracket`.

Behavior changes vs. before (all toward CEP conformance):

| Case | Before | After |
|------|--------|-------|
| Quoted list items `["a","b"]` | ❌ error | ✅ accepted |
| `.` / `+` (`foo.bar`, `c++`) — CEP-allowed | ❌ rejected (bare) | ✅ accepted |
| Uppercase / non-ASCII (`GPU`, `café`) — off-spec | ✅ accepted | ❌ rejected |
| Length > 64 — off-spec | ✅ accepted | ❌ rejected |

Note: `Numpy[extras=[GPU]]` previously yielded the extra `GPU`; it now yields no extra. This is correct per CEP (real group keys are lowercase, so `GPU` could never match a stored group), but it is a stricter, non-normalizing behavior change worth calling out.

### How Has This Been Tested?

- New `test_extras_cep44_grammar`: uppercase, non-ASCII, empty, the 64/65-char boundary, and every allowed symbol.
- New `test_quoted_extras`: double/single/mixed-quote lists, `.`/`+`, and rejection of quote-smuggled illegal chars (space, `/`, empty).
- Updated the `Numpy[extras=[GPU]]` case in `package_name.rs`.
- Full crate suite passes (`cargo nextest run -p rattler_conda_types`, 460 tests); `cargo clippy -p rattler_conda_types` clean; `cargo check --workspace` clean (new error variant breaks no exhaustive matches).

### AI Disclosure

Tools: Claude Code (Opus 4.8)

```plaintext
I just tried `foobar[extras=["bla", "baz"]]` but with the quotes in the list I get invalid bracket. Can you check if that is an issue in the parser, or by design? ... write a test case for this and then fix it or document it ... double check vs. the CEP ... We should also enforce these things per CEP (lowercase, 64 cap ...) otherwise we might produce metadata not up to spec!
```

### Checklist:

### DIFF
--- a/crates/rattler_conda_types/src/match_spec/parse.rs
+++ b/crates/rattler_conda_types/src/match_spec/parse.rs
@@ -2401,7 +2401,8 @@ mod tests {
         assert!(parse_extras("gpu, MKL").is_err());
 
         // Non-ASCII "alphanumerics" are rejected even though `char::is_alphanumeric` accepts them.
-        assert!(parse_extras("caf\u{e9}").is_err());
+        assert!(parse_extras("\u{e9}").is_err());
+        assert!(parse_extras("\u{3b1}").is_err());
 
         // Length: 1..=64 characters.
         assert!(parse_extras("").is_err());
--- a/crates/rattler_conda_types/src/match_spec/parse.rs
+++ b/crates/rattler_conda_types/src/match_spec/parse.rs
@@ -62,6 +62,12 @@ pub enum ParseMatchSpecError {
     #[error("invalid bracket key: {0}")]
     InvalidBracketKey(String),
 
+    /// An extras group name does not conform to the CEP 44 grammar.
+    #[error(
+        "invalid extras group name '{0}': names must match the pattern [a-z0-9_.+-] and be between 1 and 64 characters long"
+    )]
+    InvalidExtraName(String),
+
     /// Missing package name in match spec
     #[error("missing package name")]
     MissingPackageName,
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):

ForgottenProgramme:
Hey, @baszalmstra !
When is py-rattler 0.26 expected?

baszalmstra:
Waiting for more changes to gather since it will be a breaking change. I especially want to land #2627 before making a new release.

PR #2553, opened 2026-07-02
fix: make shard creation deterministic

russkel:
### Description

Sharded repodata (`repodata_shards.msgpack.zst` + the `shards/` directory) is written non-deterministically.

`index_subdir_inner` iterates the `registered_packages` map — an `ahash::HashMap`, seeded with a random `RandomState` per process — to build the per-shard `IndexMap`s. That random iteration order is preserved as each shard's serialization order, so a package name with more than one build serializes to a **different shard digest on every index run**. Because shards are content-addressed and written with `if_not_exists`, each reindex writes a fresh shard file and orphans the previous one, so `shards/` grows without bound even when no packages changed.

(`repodata.json` is unaffected — its packages are sorted at serialization time. Only the shards are order-sensitive.)

**Impact:** reindexing an unchanged channel churns shards. On a real channel we saw a `shards/` directory bloated to ~10–40× the live shard count, with every reindex adding a fresh batch of orphans (nothing GCs unreferenced shards). The new-orphans-per-run count closely matches `Σ (1 − 1/n!)` over multi-build names: single-build names never reshuffle, two-build names ~50% of the time, larger ones almost always.

**Fix:** sort `registered_packages` by filename before building the shards, using the `sort_unstable_by` idiom already present elsewhere in the file. Shard digests are now reproducible across runs.

### How Has This Been Tested?

Added `test_sharded_repodata_is_deterministic`. It synthesizes one package name with 10 builds (a single shard holding 10 records), indexes the channel, then reindexes the unchanged channel 3 times and asserts that `shards/` contains exactly one file. With 10 records, two independent random orderings coincide with probability 1/10! (~3e-7), so the test reliably catches the bug rather than flaking.

- **Before the fix:** fails — `4 shard files after reindexing an unchanged channel (expected 1)`.
- **After the fix:** passes.
- Full `rattler_index` integration suite: 16 passed, 0 failed. `cargo fmt` clean.

### AI Disclosure

Tools: Claude Code (Opus 4.8)

Prompt (paraphrased): investigated non-deterministic sharded-repodata output on a private channel, then asked Claude Code to locate the root cause in `rattler_index`, write a reproducing regression test, and implement the fix.

### Checklist:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5094,6 +5094,7 @@ dependencies = [
  "rattler_redaction",
  "rayon",
  "regex",
+ "rmp-serde",
  "rstest",
  "serde",
  "serde-untagged",
--- a/crates/rattler_conda_types/Cargo.toml
+++ b/crates/rattler_conda_types/Cargo.toml
@@ -81,6 +81,7 @@ hex-literal = { workspace = true }
 criterion = { workspace = true, features = ["html_reports"] }
 pathdiff = { workspace = true }
 dunce = { workspace = true }
+rmp-serde = { workspace = true }
 tools = { path = "../tools" }
 
 [[bench]]
--- a/crates/rattler_conda_types/src/repo_data/mod.rs
+++ b/crates/rattler_conda_types/src/repo_data/mod.rs
@@ -36,6 +36,7 @@ use crate::{
         TimestampMs, UrlWithTrailingSlash,
         serde::{
             DeserializeFromStrUnchecked, sort_index_map_alphabetically, sort_map_alphabetically,
+            sort_set_alphabetically,
         },
     },
 };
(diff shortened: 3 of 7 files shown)


wolfv:
Wow, thank you!

russkel:
> Wow, thank you!

Thank you for the great tools. We're part way through migrating our build/package infrastructure to pixi/rattler and we're big fans.

wolfv:
Hey @russkel Fable found a slight improvement to sort the packages already in the type - so that we always get deterministic ordering! Hope you don't mind!

russkel:
> Hey @russkel Fable found a slight improvement to sort the packages already in the type - so that we always get deterministic ordering! Hope you don't mind!

Looks much better! Unfortunately I don't know much Rust so I was unable to guide it much more than that. I did ask it to make sure it follows existing conventions and that's what it gave me ;).

baszalmstra (approved):

PR #2557, opened 2026-07-05
refactor: make the extension mechanism reliable and self-serving

wolfv:
### Description

This PR redesigns `rattler_config` so that pixi, rattler-build and rattler-index can genuinely share one configuration model and extend it with tool-specific keys. It is a follow-up to the design review requested for the shared-config effort (see prefix-dev/pixi#6144, which starts consuming the leaf types from this crate).

**Why the previous design wasn't good enough:**

1. **`#[serde(flatten)]` silently killed unknown-key warnings.** Pixi's UX relies on `serde_ignored` to warn about typos in `config.toml`. The moment a tool adopted `ConfigBase<Extension>` (one flattened field), that detection died — the crate's own `link_config.rs` doc comment admitted this. This was the single biggest blocker for pixi ever adopting `ConfigBase`.
2. **Validation never ran.** `ConfigBase::validate()` unconditionally returned `Ok(())`, so `load_from_files` validated nothing — not the nested sections, not the extension.
3. **The `Config` trait asked too much of extension authors**: hand-written `set()` per type (with inconsistent key-prefix conventions and a `strip_prefix().unwrap()` panic path), hand-maintained `keys()` listings that were wrong (`repodata.default`, `proxy.https` — not real TOML paths), and a `get_extension_name` method with a `"__NONE__"` hack for `()`.
4. **State leaks and small bugs**: `ProxyConfig::default()` read `HTTP(S)_PROXY`/`NO_PROXY` env vars, so saving a config after `config set` wrote the machine's proxy environment into the user's config file; `ProxyConfig::is_default` checked `https` twice and never `http`; `loaded_from` was never populated; the `load_config` free function neither merged nor validated.
5. **No shared file discovery** — every tool reinvented "where does config.toml live", and rattler-build had no default discovery at all.

**What this PR does:**

- Moves the shared fields into `CommonConfig` (deliberately flatten-free); `ConfigBase<T>` composes it with the flattened extension and `Deref`s to it, so field access (`config.default_channels`) keeps compiling everywhere.
- Adds `ConfigBase::from_toml_str`, which deserializes the document twice (common + extension) under `serde_ignored` and reports keys unknown to **both** — restoring typo warnings, at full depth (e.g. `repodata-config.disable-jlap` inside a known table).
- Slims the `Config` trait to a single required method: `merge_config`. `validate`, `is_default` and `keys` have sensible defaults; `get_extension_name` and `set` are gone. Writing an extension is now ~10 lines (see the crate docs example).
- Implements `set`/`unset` **generically** by round-tripping through TOML: every key — including extension keys — is editable with zero per-field code (~400 lines of hand-written `set()` deleted). Values parse as JSON with plain-string fallback (compatible with the existing `pixi config set` conventions), unknown keys are rejected via the same two-pass detection, and dotted segments can be quoted (`mirrors."https://conda.anaconda.org"`).
- Fixes validation to recurse into nested sections and the extension, fills `loaded_from`, and makes unset `tls-no-verify` representable (`None` instead of `Some(false)`).
- Makes `ProxyConfig::default()` empty and adds an explicit `ProxyConfig::from_env()`.
- Adds a `locations` module + `ConfigBase::load_from_default_locations(&["pixi", "rattler-build"])` for shared config discovery: `/etc/<tool>/config.toml`, `$XDG_CONFIG_HOME/<tool>/config.toml`, `$<TOOL>_HOME` or `~/.<tool>/config.toml` — mirroring pixi's existing search semantics, so cooperating tools can layer their config on top of pixi's.
- Replaces the `()` extension with `NoExtension` (the default type parameter of `ConfigBase`), removes the unwired `link_config` module in favor of flat top-level `allow-*-links` keys matching pixi's schema, and updates `rattler-index`.

Breaking changes are listed in the crate CHANGELOG. In-repo consumers (`rattler_index`, `rattler_repodata_gateway`, `rattler_networking`, `rattler_upload`) compile unchanged apart from the one-line `ConfigBase<()>` → `ConfigBase` swap in `rattler-index`.

Companion PRs: pixi (migration to the shared leaf types) and rattler-build (extension type + default config discovery) on the same branch name in their repos.

### How Has This Been Tested?

- `cargo test -p rattler_config`: 51 unit tests + doctest pass, including new tests for two-pass unknown-key reporting (typos, extension keys, deprecated keys inside known tables), generic `set` on extension keys, validation recursion, and location ordering.
- Snapshot tests reviewed: the only diffs are the two intended behavior changes (unset `tls-no-verify` no longer serialized; `set` normalizes S3 bucket order).
- `cargo check` on all in-repo consumers; `cargo clippy -p rattler_config --all-targets` is clean.

### AI Disclosure

Tools: Claude Code

```plaintext
I want rattler, rattler-build and pixi to share a configuration. We already have a PR in pixi
that uses the low-level rattler_config crate (prefix-dev/pixi#6144) which we also want to use
in rattler-build. However, the idea in rattler_config was to make it extensible so that Pixi
can add certain config keys that only make sense for Pixi and rattler-build the same.
Can you check if the design is good enough? If not, improve it (you can be radical, hehe).
```

### Checklist:

---
_Generated by [Claude Code](https://claude.ai/code/session_01EzYs4p6XJDj9QBcNB4ipR7)_


crates/rattler/Cargo.toml
```diff
@@ -13,6 +13,7 @@ readme.workspace = true
 
 [features]
 default = ['rustls']
+rattler_config = ['dep:rattler_config']
```
baszalmstra: Not needed.

crates/rattler/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
```
baszalmstra: Not needed, this is generated.

wolfv: Fixed, thx!

crates/rattler/Cargo.toml
wolfv: Fixed, thx!

baszalmstra (approved):

PR #2565, opened 2026-07-07
feat: Implementation offsets in paths.json

chrisburr:
### Description

Follow up to https://github.com/conda/rattler/pull/2032 that is needed for the (about to be opened) draft RattlerFS PR. See the original PR for details.

Closes #2032

### How Has This Been Tested?

See the testing in the coming RattlerFS and Pixi PRs.

### AI Disclosure

This is mostly the work of @Dagmar-Dinjens. The AI part has come as part of many many sessions working on the downstream parts of this so there isn't a clear prompt.

### Checklist:


baszalmstra:
I think this will require a Conda Enhancement Proposal (CEP), but I would be happy to merge this in rattler already. Do you think you can take care of writing a CEP?

chrisburr:
Yeah I can add that to my list, though it proabably won't happen for a week or two.

crates/rattler/src/install/link.rs
```diff
@@ -796,6 +872,57 @@ pub fn copy_and_replace_textual_placeholder(
...
+            target_platform,
+        );
+        // let replaced = first_line.replace(prefix_placeholder, target_prefix);
+        destination.write_all(new_shebang.as_bytes())?;
+        source_bytes = rest;
+    }
+
+    let mut last_match = 0;
+
+    for &offset in offsets {
```
baszalmstra, line 1106: These offsets do not take into account the change of the shebang. So the offsets might be wrong. This requires a test.

Fable original comment:

> `copy_and_replace_textual_placeholder_offsets` (`crates/rattler/src/install/link.rs:895`) strips the shebang line and rebases `source_bytes = rest`, but then applies `offsets` that are **absolute file positions**. Every offset is now wrong by the shebang length: the loop injects the prefix mid-body and, for offsets near EOF, `&source_bytes[last_match..offset]` panics with an out-of-range slice. Any package that ships the new `offsets` field *and* has a shebang'd text file installs corrupted or crashes the installer. This is in the production install path, gated only by whether `offsets` is populated. The one test combining shebang + offsets passes an **empty** offsets vec, so it never triggers. This is the most important finding — it's a regression in already-shipping code.

crates/rattler/src/install/link.rs
```diff
@@ -796,6 +872,57 @@ pub fn copy_and_replace_textual_placeholder(
...
+        destination.write_all(&source_bytes[last_match..offset])?;
```
baszalmstra: If the offsets are corrupted for whatever reason this could panic. 

Fable original comment: 

> `offsets` comes straight from a package's `paths.json` (attacker-controllable) and is never validated. Empty binary groups panic at `group.split_at(group.len() - 1)` (usize underflow); out-of-range/out-of-order offsets panic at `&source[src_start..src_end]`. A package with `"offsets": [[1000, 2000]]` on a 10-byte file kills the FUSE/NFS read thread → whole mount dies. The identical arithmetic in the install-time variants panics the installer. Needs a validation pass (sorted, in-bounds, spaced ≥ placeholder length) at build/mount time.

baszalmstra (changes requested):

chrisburr:
I'm about to push an amended version of this PR based on the review above and the CEP that I've prepared. Though it has exposed a messy edge case described here: https://github.com/conda/ceps/pull/179/changes#r3551264320

chrisburr:
> I think this will require a Conda Enhancement Proposal (CEP), but I would be happy to merge this in rattler already. Do you think you can take care of writing a CEP?

@baszalmstra How does this fit into the potentially [open question in the CEP](https://github.com/conda/ceps/pull/179/files/e66a34f499abe6dbe47bf5b6828c9cd65ac6d8d7#diff-bebfa3f1d70547b3695124927ba31659f76574696cb2ae65c971578b2069ff5f) about encodings? I don't mind iterating here a couple of times but I guess you don't want to deal with backwards compatibility issues if it changes too much.

baszalmstra:
@chrisburr Since this code is not breaking (its an addition) I would be happy to iterate here. As long as we mark the fields `experimental_` we warn people that this can break. Once the CEP is merged, we make the fields stable. 

(We only need to prefix the fields `experimental_` in rust, we can leave the serialized form as is).

chrisburr:
> (We only need to prefix the fields experimental_ in rust, we can leave the serialized form as is).

Doesn't it become breaking if it's in people's caches and a future version of rattler doesn't understand the outdated schema? I see how `experimental_offsets` would avoid that problem but not if it's only in rust.

baszalmstra:
Yeah that is true but as far as I understand we wont be adding it to any caches right now? 

We can also mark those fields in serde as non-failing. E.g. just use a default value if parsing fails.

chrisburr:
Indeed, I've added an issue for adding support for writing `paths.json` when the CEP is finished: https://github.com/conda/rattler/issues/2607

In the meantime I think this is ready.

crates/rattler_conda_types/src/package/paths.rs
```diff
@@ -217,17 +278,266 @@ pub struct PathsEntry {
...
+
+    /// Whether the group carried members other than `encoding` and `ranges`.
+    ///
+    /// The CEP defines exactly those two keys today; a future CEP may add
+    /// more. Until a consumer implements such a CEP it cannot know whether an
+    /// extra member changes the meaning of the group, so the group must be
+    /// treated like corrupt metadata (fall back to searching). Unrecognized
+    /// members are not preserved on re-serialization.
+    #[serde(skip)]
+    pub has_unknown_members: bool,
```
baszalmstra: Would it make sense to store these extra keys in here? So we can report an error?

crates/rattler_conda_types/src/package/paths.rs
```diff
@@ -217,17 +278,266 @@ pub struct PathsEntry {
...
+    }
+}
+
+/// Error describing why recorded offset metadata cannot be used.
+///
+/// Per the CEP, consumers that hit this SHOULD fall back to locating
+/// occurrences by searching the file contents (and MAY report a warning).
+#[derive(Debug, thiserror::Error)]
+#[error("{0}")]
+pub struct InvalidOffsetsError(String);
```
baszalmstra: I would prefer to use an enum.

crates/rattler/src/install/link.rs
```diff
@@ -650,6 +707,122 @@ pub fn copy_and_replace_placeholders(
...
+/// `paths.json` instead of searching the file contents.
+///
+/// Per the CEP, an installer applies exactly the groups whose encodings its own search-based
+/// replacement covers. rattler's search-based replacement covers UTF-8 only, so the UTF-8 group's
+/// ranges (selected and structurally validated by [`select_utf8_offset_ranges`]) are spliced by
+/// [`copy_and_replace_textual_placeholder_offsets`] or
+/// [`copy_and_replace_cstring_placeholder_offsets`]; groups for the other defined encodings are
+/// skipped, since their occurrences would not have been replaced by the search either. Valid
+/// metadata without a UTF-8 group means there is nothing to splice: the file is copied through
+/// unchanged apart from the shebang handling of text files.
```
baszalmstra: Is that because rattler doenst impement utf-16 replacement? We could implement that for binary files couldnt we?

crates/rattler/src/install/link.rs
```diff
@@ -871,6 +1225,133 @@ pub fn copy_and_replace_cstring_placeholder(
...
+                })?;
+            if &source_bytes[offset..end] != old_prefix {
+                return Err(OffsetReplaceError::inconsistent(format!(
+                    "placeholder bytes are not present at recorded offset {offset}"
+                )));
+            }
+            prev_end = end;
+        }
+        prev_end = nul_pos;
+    }
```
baszalmstra, line 1370: Maybe we should move this to a seperate function?

baszalmstra: Same for the text variant

baszalmstra (commented):
Can you make the comments less verbose, and generally less AIy, strip them from emdashes for instance. 

One thing I dont understand is that this CEP adds support for other encodings but it looks like this code only deals with UTF-8? Why not also add support for the other encodings?

chrisburr:
> One thing I don't understand is that this CEP adds support for other encodings but it looks like this code only deals with UTF-8? Why not also add support for the other encodings?

That changes the path which scans for the prefix which felt like it was out of scope at this point, especially as it's a much less trivial scan than the current one used in rattler. It should definitely be a follow up.

PR #2566, opened 2026-07-07
feat: add rattler_vfs crate

chrisburr:
### Description

This builds on top of https://github.com/conda/rattler/pull/2565 and the work on @Dagmar-Dinjens to add a `rattler_fs` crate. See https://github.com/conda/rattler/pull/1182 for the original concept. Basically the idea is to serve data directly from the rattler cache directory using a virtual filesystem, therefore avoiding the overhead of actually making environments. The usual install-time tricks (prefix-subsitution, python entrypoints, shebang-fixing, apple's code signing, ...) are applied on-the-fly as the data is read.

I see several motivations for this but the main ones are:

* Using environments in large scale distributed systems without the overhead of installing environments while being able to make use of a pre-existing rattler cache directory.
* Running on systems with slow filesystems (e.g. NFS) where environment creation is slow due to limited IOPS.
* Avoiding the disk space overhead of having many pixi environments locally, many of which are used infreqently.

This PR includes 3 VFS backends for RattlerFS which are supported on the following platforms:

| Platform | FUSE | NFS | ProjFS |
| -------- | ---- | --- | ------ |
| Linux    | ✅ | ✅ | N/A |
| macOS    | ✅ | ✅ | N/A |
| Windows  | N/A | ❌ | ✅ |

We default to FUSE on Linux, NFS on macOS and ProjFS on Windows which allows all versions to run in user space assuming the VFS drivers are available. Windows is notably less extensively tested than the others. I also looked into FSKit on macOS but it seems that it's not yet suitable for something like RattlerFS.

This PR also contains the concept of a writable overlay for FUSE and NFS which then allows for edits to be made on top of the read-only underlying environment. This also makes it possible to pip install packages on top, though in principle RattlerFS could be expanded to support wheels natively rather than needing the overlay for that case. I've laid the ground work for this but didn't finish it.

The main component here is the `rattler_fs` crate which is used by the draft Pixi PR (https://github.com/prefix-dev/pixi/pull/6548). There is a `rattler mount` command added to `rattler_bin` with an optional feature. I could see this as being useful in it's own right by systems which want to mount an environment from a lock file but don't want all of the the workspace functionality of pixi.

### How Has This Been Tested?

There are End-To-End tests included which have been useful during development to make sure all three platforms continue to work. Also see https://github.com/prefix-dev/pixi/pull/6548.

### AI Disclosure

There have been far too many sessions involved in the creation of this to track exactly what tools have been used and which parts are hand written vs AI generated. It's more AI-generated than I would like but equally it contains far more functionality and edge case handling than I would have managed to find time for alone. This project represents most of my rust experience so I'm not in a position to evaluate the quality of the code but I think it's in a reasonable state all things considered.

### Checklist:


baszalmstra:
This is pretty amazing work! Thanks @chrisburr ! Coincidentally @Dagmar-Dinjens is also working on this in https://github.com/conda/rattler/pull/2543.

Ill be going over the code!

chrisburr:
I guess you meant @Dagmar-Dinjens? (Hi 👋! I'm glad to see the internship worked out!) I'd missed https://github.com/conda/rattler/pull/2543 and I fear that PR is proabably orthogonal to this one...

If you'd like to chat about this PR on a call I'm happy to do so as it might be easier for getting up to speed as there is a lot going on here.

crates/rattler_vfs/src/overlay_fs/inode.rs
```diff
@@ -0,0 +1,82 @@
...
+};
+
+/// Inodes at or above this value belong to the upper (overlay) layer.
+/// Anything below is owned by the lower read-only VFS.
+pub(crate) const UPPER_INODE_BASE: u64 = u64::MAX / 2;
+
+/// Bidirectional inode ↔ path mapping for upper-layer entries.
+pub(crate) struct UpperInodeMap {
+    path_to_ino: HashMap<PathBuf, u64>,
+    ino_to_path: HashMap<u64, PathBuf>,
```
baszalmstra, line 26: Since values are assigned sequentially, this could just be a Vec

crates/rattler_vfs/src/overlay_fs/inode.rs
```diff
@@ -0,0 +1,82 @@
...
+    next_ino: AtomicU64,
```
baszalmstra, line 27: The rest of this struct requires locking for concurrency, so what does an atomic u64 buy us here.

crates/rattler_vfs/src/overlay_fs/inode.rs
```diff
@@ -0,0 +1,82 @@
...
+impl UpperInodeMap {
+    pub(crate) fn new() -> Self {
+        Self {
+            path_to_ino: HashMap::new(),
+            ino_to_path: HashMap::new(),
+            next_ino: AtomicU64::new(UPPER_INODE_BASE),
+        }
+    }
+
+    pub(crate) fn get_or_assign(&mut self, virtual_path: PathBuf) -> u64 {
```
baszalmstra, line 39: Instead of u64, maybe we can use a newtype?

crates/rattler_vfs/src/codesign.rs
baszalmstra, line 1: Maybe we can use https://github.com/conda/rattler/pull/2554 once it lands? Or is this implementation still very different from that?

crates/rattler_vfs/Cargo.toml
```diff
@@ -0,0 +1,66 @@
...
+reqwest-middleware = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "io-std", "io-util", "signal"] }
+tracing = { workspace = true }
+
+# NFS transport (optional)
+nfs3_server = { version = "0.10", optional = true }
+nfs3_types = { version = "0.5", optional = true }
+
+[target.'cfg(target_os = "windows")'.dependencies]
+windows = { version = "0.62", features = ["Win32_Storage_ProjectedFileSystem", "Win32_Foundation", "Win32_System_LibraryLoader"] }
```
baszalmstra: Lets use workspace dependencies.

crates/rattler_vfs/build.rs
```diff
@@ -0,0 +1,12 @@
...
+    // Delay-load ProjectedFSLib.dll so the binary can start even when the
+    // Windows Projected File System optional feature is not installed.  Without
+    // this, the DLL is in the normal import table and the Windows loader kills
+    // the process before main() when the DLL is absent.  With delay-load the
+    // DLL is only resolved on the first ProjFS API call, giving us a chance to
+    // check availability and return a clear error.
+    if std::env::var("CARGO_CFG_TARGET_OS").unwrap_or_default() == "windows" {
+        println!("cargo::rustc-link-arg=/DELAYLOAD:projectedfslib.dll");
+        println!("cargo::rustc-link-lib=delayimp");
+    }
```
baszalmstra, line 11: The downside of this approach is that we have no way of detecting whether the library is available on the user system. This means you get a really bad error with projectedfslib.dll is missing on windows. 

Perhaps we should look at dynamic loading the library so we can report if the library is not installed on the user machine?

crates/rattler-bin/Cargo.toml
```diff
@@ -23,17 +23,21 @@ native-tls = [
...
   "rattler_repodata_gateway/rustls",
   "rattler_networking/rustls",
   "rattler_cache/rustls",
+  "rattler_fs?/rustls",
 ]
 s3 = ["rattler_networking/s3", "rattler_upload/s3"]
 gcs = ["rattler_networking/gcs"]
 oauth = ["rattler/oauth"]
+# Opt-in `rattler mount` subcommand.
+mount = ["dep:rattler_fs", "dep:rattler_lock"]
```
baszalmstra: What is the downside of always including this feature?

crates/rattler-bin/src/commands/mount.rs
```diff
@@ -0,0 +1,127 @@
...
+    /// Where the virtual environment should appear.
+    pub mount_point: PathBuf,
+
+    /// Environment name in the lock file.
+    #[clap(long, default_value = DEFAULT_ENVIRONMENT_NAME)]
+    pub environment: String,
+
+    /// Persistent writable overlay directory. If omitted, the mount is read-only.
+    /// Required for ProjFS (which is always writable).
+    #[clap(long)]
```
baszalmstra, line 32: I think we can annotate this field with require-if to require that it is specified if transport is ProjFS.

crates/rattler-bin/src/commands/mount.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub overlay: Option<PathBuf>,
+
+    /// Transport backend. Defaults to the platform's preferred transport.
+    #[clap(long, value_enum, default_value_t = TransportArg::Auto)]
```
baszalmstra, line 36: If Auto just switches based on the os, we can also make the default Os specific here instead of Auto.

crates/rattler-bin/src/commands/mount.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub transport: TransportArg,
+
+    /// Allow other users to access the mount (FUSE only; requires
+    /// `user_allow_other` in `/etc/fuse.conf`).
+    #[clap(long)]
```
baszalmstra, line 41: We should mark this too as only allowed when transport is fuse.

crates/rattler-bin/src/commands/mount.rs
```diff
@@ -0,0 +1,127 @@
...
+use rattler_lock::{DEFAULT_ENVIRONMENT_NAME, LockFile};
+
+/// Mount a lockfile as a virtual conda environment.
+#[derive(Debug, Parser)]
+pub struct Opt {
+    /// Lock file to mount (e.g. `pixi.lock`).
+    pub lock_file: PathBuf,
+
+    /// Where the virtual environment should appear.
+    pub mount_point: PathBuf,
```
baszalmstra, line 24: Instead of two positional arguments I would prefer to have one at most. That avoids misuse.

crates/rattler-bin/src/commands/mount.rs
```diff
@@ -0,0 +1,127 @@
...
+pub async fn mount(opt: Opt) -> Result<()> {
+    let lockfile = LockFile::from_path(&opt.lock_file).into_diagnostic()?;
+    let platform = Platform::current();
+    let env_hash =
+        compute_env_hash(&lockfile, &opt.environment, platform).map_err(anyhow_to_miette)?;
+
+    let cache_dir = default_cache_dir().map_err(anyhow_to_miette)?.join("pkgs");
+    let package_cache = PackageCache::new(cache_dir);
+
+    let mount_point = opt.mount_point.canonicalize().into_diagnostic()?;
```
baszalmstra: Canonicalize requires that the directory already exists. I think you want absolutize instead and create the directory if it doesnt exist yet. And error if the path points to an existing file.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+
+/// Transport backend for the virtual filesystem.
+///
+/// See the [crate-level docs](crate#platform-support) for why each platform
+/// has a different default.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Transport {
+    /// Auto-detect the best backend for the current platform: FUSE on Linux,
+    /// NFS on macOS, `ProjFS` on Windows.
+    Auto,
```
baszalmstra, line 341: Instead of an enum value, how about we use an associate function that returns the best Transport layer?

crates/rattler_fs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+}
+
+/// Whether the mount is read-only or writable, and where the writable
+/// overlay lives.
+///
+/// Replaces the old `overlay_dir: Option<PathBuf>` convention with an explicit
+/// distinction between read-only and writable mounts. `ProjFS` does not support
+/// read-only mode (it lacks a pre-creation notification, so new files can
+/// always be created via the virtualization root) — passing [`Mode::ReadOnly`]
+/// to a `ProjFS` mount returns a clear error.
```
baszalmstra: This talks about "previous" code which is not relevant anymore.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+    ) -> Self {
+        Self {
+            mount_point,
+            mode: Mode::Writable { overlay_dir },
+            transport,
+            env_hash,
+            allow_other: false,
+            overlay_mismatch: OverlayMismatch::Error,
+        }
+    }
```
baszalmstra, line 539: Instead of having multiple constructors I think it would be nicer to have a simple constructor and multiple `with_` functions to customize it futher. Ideally we also have `set_` methods that take a `&mut self` and return `&mut Self`.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+/// - Construct a [`PackageCache`] (commonly via
+///   [`rattler_cache::default_cache_dir()`]). Decoupling the cache from this
+///   function lets pixi share its own cache and lets tests use a temp dir.
+pub async fn build_metadata_tree(
+    lockfile: &LockFile,
+    environment_name: &str,
+    platform: Platform,
+    package_cache: &PackageCache,
+    mount_point: &Path,
+) -> anyhow::Result<MetadataTree> {
```
baszalmstra, line 638: Maybe we can make this an associate method on `MetadataTree` instead of a free function? Ideally we also make it an optional gated by `rattler_lock`.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+
+    let (mut env_paths, mut directory_indices) = new_empty_tree();
+    let mount_str = mount_point.to_string_lossy().to_string();
+
+    // Build a single lazily-initialized HTTP client for the whole package loop.
+    // `LazyClient::default()` forces construction eagerly, which on macOS walks
+    // the keychain via `rustls_native_certs` and takes several seconds. Using
+    // `LazyClient::new` defers that work until the first cache miss, so
+    // warm-cache mounts skip it entirely.
+    let client = LazyClient::new(reqwest_middleware::ClientWithMiddleware::default);
```
baszalmstra, line 670: This is something that should be passed in instead of constructed here.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+/// and constructs the virtual directory tree with noarch Python path rewriting
+/// and entry point generation.
+///
+/// Caller responsibilities:
+/// - Parse the lock file once via [`LockFile::from_path`].
+/// - Pick a [`Platform`] (usually [`Platform::current()`]).
+/// - Construct a [`PackageCache`] (commonly via
+///   [`rattler_cache::default_cache_dir()`]). Decoupling the cache from this
+///   function lets pixi share its own cache and lets tests use a temp dir.
+pub async fn build_metadata_tree(
```
baszalmstra, line 632: It would be nice if we can attach a progress reporter.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+        }
+
+        tracing::debug!("parsed {} metadata entries", env_paths.len());
+    }
+
+    Ok(MetadataTree(env_paths))
+}
+
+/// Mount a pre-built metadata tree. Returns a handle that unmounts on drop.
+pub async fn mount(metadata: MetadataTree, config: &MountConfig) -> anyhow::Result<MountHandle> {
```
baszalmstra, line 783: I think I would prefer this to be an associate function on MetadataTree as well.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+                .collect();
+            let handle = unsafe {
+                windows::Win32::System::LibraryLoader::LoadLibraryW(windows::core::PCWSTR(
+                    dll.as_ptr(),
+                ))
+            };
+            if handle.is_err() {
+                return Err(MountError::ProjFsDllMissing.into());
+            }
+        }
```
baszalmstra, line 810: This didnt work on my machine. When running `rattler mount` without the feature enabled I got an error (from windows) that the projectedfslib.dll was missing.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+}
+
+/// Mount a pre-built metadata tree. Returns a handle that unmounts on drop.
+pub async fn mount(metadata: MetadataTree, config: &MountConfig) -> anyhow::Result<MountHandle> {
+    let transport = config.transport.resolve();
+
+    // ProjFS-specific pre-flight checks: DLL availability, mode validity,
+    // overlay state. Done before VFS construction so we don't waste offset
+    // computation if ProjFS isn't installed or the user passed Mode::ReadOnly.
+    #[cfg(target_os = "windows")]
```
baszalmstra, line 789: I think this functionality would be better to have in projfs specific module.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+
+// ---------------------------------------------------------------------------
+// Internal: transport-specific mount helpers
+// ---------------------------------------------------------------------------
+
+/// Mount via FUSE, with optional writable overlay.
+///
+/// The overlay is retried once if the env hash mismatches (environment updated).
+#[cfg(any(target_os = "linux", feature = "fuse"))]
+fn mount_fuse(vfs: VirtualFS, config: &MountConfig) -> anyhow::Result<fuser::BackgroundSession> {
```
baszalmstra, line 1007: I would prefer to move this function to a fuse specific module.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+            )?)
+        }
+    }
+}
+
+/// Mount via NFS, with optional writable overlay.
+///
+/// The overlay is retried once if the env hash mismatches (environment updated).
+#[cfg(feature = "nfs")]
+async fn mount_nfs(
```
baszalmstra, line 1063: I would prefer to move this to a nfs specific module.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+    {
+        let mnt = config.mount_point.display().to_string();
+        // Probe passwordless sudo first — `mount -t nfs` needs CAP_SYS_ADMIN
+        // which isn't available in unprivileged user namespaces, so there's no
+        // userspace fallback we can reach for. Fail loudly instead of letting
+        // sudo prompt interactively (terrible UX in `pixi run`).
+        //
+        // TODO(bind-mount): investigate `unshare -Urm` + `mount --bind` as a
+        // rootless alternative transport on Linux. That would work in rootless
+        // containers where neither FUSE nor sudo is available.
```
baszalmstra, line 1135: But if the user is already running the application using sudo, mounting could just work no?

crates/rattler_vfs/src/lib.rs
baszalmstra, line 1: I think it would be nice to have modules (directories with mod.rs) for the individual implementations which contain the specifics for each transport.

crates/rattler_vfs/src/metadata_tree.rs
```diff
@@ -0,0 +1,162 @@
...
+            parent,
+            children: vec![],
+        }
+    }
+}
+
+#[derive(Debug)]
+pub struct FileNode {
+    pub file_name: OsString,
+    pub parent: usize,
```
baszalmstra, line 40: Instead of "random" `usize` I would prefer a newtype to avoid mixing these indexes up with something else.

crates/rattler_vfs/src/metadata_tree.rs
```diff
@@ -0,0 +1,162 @@
...
+    pub cache_base_path: Arc<Path>,
```
baszalmstra, line 41: Maybe we can use an enum for a path from cache *or* a virtual file?

test-data/rattler-fs/pixi.toml
```diff
@@ -0,0 +1,10 @@
+[workspace]
```
baszalmstra: Cant we create a specific environment in the pixi.toml at the root for the e2e test? That also contains the task itself.

Cargo.toml
```diff
@@ -74,12 +74,14 @@ dirs = "6.0.0"
 dbus-secret-service-keyring-store = { version = "1.0.0", default-features = false }
 dunce = "1.0.5"
 enum_dispatch = "0.3.13"
+env_logger = "0.11.9"
```
baszalmstra: How is this used?

crates/rattler-bin/src/commands/mount.rs
```diff
@@ -0,0 +1,127 @@
...
+        }
+    }
+}
+
+/// Convert any `anyhow::Error` (used by rattler_fs) into a miette diagnostic.
+fn anyhow_to_miette(e: anyhow::Error) -> miette::Report {
+    miette::miette!("{e}")
+}
+
+pub async fn mount(opt: Opt) -> Result<()> {
```
baszalmstra, line 75: I really miss more information here. I think having a tracing subscriber and more tracing output in the library would help! Also more progress would be nice!

crates/rattler_vfs/src/fuse_adapter.rs
```diff
@@ -0,0 +1,497 @@
...
+            Self::Mapped(mmap) => Some(mmap.as_ref()),
+            Self::Passthrough { .. } | Self::VfsManaged(_) | Self::InodeRead(_) => None,
+        }
+    }
+}
+
+/// FUSE transport adapter, generic over any `VfsOps` implementation.
+pub struct FuseAdapter<T: VfsOps> {
+    vfs: T,
+    open_files: Mutex<HashMap<u64, OpenFileContent>>,
```
baszalmstra, line 92: I wonder if we can use a more concurrency friendly datastructure here. Like an RWLock. We can also use a Vec with a freelist instead of a hashmap maybe?

crates/rattler_vfs/src/nfs_adapter.rs
```diff
@@ -0,0 +1,691 @@
+//! NFS transport adapter.
```
baszalmstra, line 1: Out of curiosity, why do we use nfs3 instead of nfs4?

crates/rattler_vfs/src/vfs_ops.rs
```diff
@@ -0,0 +1,210 @@
...
+    pub ino: u64,
+    pub kind: FileKind,
+    pub name: OsString,
+}
+
+/// Transport-agnostic filesystem operations.
+///
+/// Read operations are required. Write operations default to `EROFS` (read-only
+/// filesystem), allowing read-only implementations to skip them.
+pub trait VfsOps: Send + Sync + 'static {
```
baszalmstra, line 149: Should we not implement a custom error type? Why always return i32?

baszalmstra (commented):
I did a thorough review and tested the implementation on my windows machine. I think the code is already in a good state and apart from some tweaks here and there I would be fine if we merge this soon and then iterate on it in followup PRs. 

Unfortunately I think this far surpasses the work that @Dagmar-Dinjens has pushed so far. @chrisburr I propose that we get this merged and let @Dagmar-Dinjens continue with followups? Would be happy to jump on a call to discuss this in more detail if that helps

I left way more comments than what I think is actually required to merge this. I think a lot of them we can address in follow-up PRs. 

My main blockers are:

- I think we should name the crate `rattler_vfs` instead of `rattler_fs`. But I can be persuaded.
- We should not yet publish this crate automatically. This requires annotating the Cargo.toml with `package.publish = false`.

Btw why is windows nfs not supported?

baszalmstra:
Also some notes from fable worth looking at:

# Review: conda/rattler#2566 — `feat: add rattler_fs crate`

## What it does

Adds a new `rattler_fs` crate that serves conda environments as virtual filesystems (FUSE on Linux/macOS, a localhost NFS server, ProjFS on Windows) directly from the package cache, applying install-time transforms — prefix replacement, Python entry points, shebang fixes, macOS ad-hoc code signing — **on the fly during reads** instead of materializing files on disk. A writable overlay layer (whiteouts + copy-up) lets users edit files and `pip install` on top of the read-only view. Supporting changes: `paths.json` gains optional `offsets`/`shebang_length` fields, `link.rs` gains offset-based replacement variants, plus a `mount` CLI command, an e2e nushell harness, and a CI workflow.

## Blocking issues

### 1. The `link.rs` change corrupts real installs (not just mounts)
~~`copy_and_replace_textual_placeholder_offsets` (`crates/rattler/src/install/link.rs:895`) strips the shebang line and rebases `source_bytes = rest`, but then applies `offsets` that are **absolute file positions**. Every offset is now wrong by the shebang length: the loop injects the prefix mid-body and, for offsets near EOF, `&source_bytes[last_match..offset]` panics with an out-of-range slice. Any package that ships the new `offsets` field *and* has a shebang'd text file installs corrupted or crashes the installer. This is in the production install path, gated only by whether `offsets` is populated. The one test combining shebang + offsets passes an **empty** offsets vec, so it never triggers. This is the most important finding — it's a regression in already-shipping code.~~
https://github.com/conda/rattler/pull/2565/changes#r3544983267

### 2. Untrusted `paths.json` offsets panic inside filesystem callbacks
~~`offsets` comes straight from a package's `paths.json` (attacker-controllable) and is never validated. Empty binary groups panic at `group.split_at(group.len() - 1)` (usize underflow); out-of-range/out-of-order offsets panic at `&source[src_start..src_end]`. A package with `"offsets": [[1000, 2000]]` on a 10-byte file kills the FUSE/NFS read thread → whole mount dies. The identical arithmetic in the install-time variants panics the installer. Needs a validation pass (sorted, in-bounds, spaced ≥ placeholder length) at build/mount time.~~
https://github.com/conda/rattler/pull/2565/changes#r3544991157

### 3. Overlay copy-up has three independent data-corruption bugs
~~All in `overlay_fs.rs::copy_to_upper`, all untested, all hit by ordinary `pip install`/`mv` on a conda env:~~
- ~~**Symlinks are copied up as empty regular files** — only `Directory` is special-cased; a symlink's `content_source` returns ENOENT, falls to `read` (returns `vec![]` for symlinks), and `fs::write` creates a 0-byte file that permanently shadows the link. Conda envs are full of `libfoo.so → libfoo.so.1`.~~
- ~~**Directory rename loses lower children** — the `upper_path.exists()` early-return fires before the directory-recursion check, so a partially-materialized dir isn't recursed; renaming it drops any lower-only children, then a whiteout hides the originals.~~
- ~~**Exec bit dropped on copy-up** — `fs::write` creates 0644 with no permission copy, so `chmod`/`mv`/truncate on `bin/python` yields a non-executable copy.~~

~~Plus a **lookup/opaque ordering bug** (`overlay_fs.rs:308`): `lookup` returns ENOENT for opaque dirs before checking the upper layer, so a freshly created file in a reinstalled package dir is visible via `readdir` but not `stat`/`open` — breaking the exact pip uninstall/reinstall pattern the feature targets.~~
https://github.com/conda/rattler/issues/2582

### 4. NFS server has no per-user access control (security)
~~Two independent findings converge here. The server binds `127.0.0.1` (remote hosts excluded — good), but that bind is the *only* gate: NFS `AUTH_UNIX` is unauthenticated, the adapter never consults caller credentials, and access via raw RPC bypasses the mount point's Unix permission bits. **Any other local user can speak NFS3 to the port and read (and, on writable mounts, write/delete) another user's environment, forging any uid.** Compounding it, NFS wire filenames aren't sanitized: `CREATE(dir, "../../../home/user/.bashrc")` escapes the overlay directory via `overlay_dir.join(...)`. A loopback-TCP NFS server is inherently shared among local users; at minimum this needs prominent documentation, ideally peer-credential rejection.~~
https://github.com/conda/rattler/issues/2579

### 5. Packaging will break the release pipeline
`rattler_fs` has no `description` and no `publish = false`, and `release-plz release` runs on every push to main → publish fails (crates.io requires a description). Separately, `rattler-bin` (a published crate) now has `rattler_fs = { path = "../rattler_fs", optional = true }` with no version → `cargo publish` of rattler-bin fails. Convention is to add the crate to the pinned internal-crate list in the workspace `Cargo.toml` and use `workspace = true`.

## Major issues (fix before real-world use)

- **Mount-path-longer-than-placeholder panics the read path.** `binary_ranged_read` has `assert!(new_prefix.len() <= old_prefix.len())`; on Linux it's called directly, so mounting a short-placeholder package (e.g. the 32-char `/opt/anaconda1anaconda2anaconda3`) at a longer path aborts the FUSE thread. The installer returns a graceful `InvalidData` error for the same case.
- ~~**VFS diverges from the installer in ways that break environments:**~~ https://github.com/conda/rattler/issues/2580
- ~~**macOS codesign can't be byte-identical.** Install shells out to `/usr/bin/codesign` (rebuilds the whole superblob, changes size/bytes); `adhoc_resign` only patches page hashes and only the first CodeDirectory (dual-signed / SHA-1-first binaries fail or go stale). This breaks the parity premise for signed binaries and is undocumented/untested.~~
- ~~**codesign slice panics on malformed Mach-O** — inconsistent `n_code_slots` vs `code_limit` produces `&data[4096..100]` → panic in the read handler. ~~ https://github.com/conda/rattler/issues/2571~~
- ~~**ProjFS enumeration sort uses `str::cmp`** instead of `PrjFileNameCompare` (Windows case-insensitive collation); mis-ordering can make ProjFS drop or duplicate directory entries.~~ https://github.com/conda/rattler/issues/2581
- ~~**ProjFS `/DELAYLOAD` never reaches the binary** — `build.rs`'s `rustc-link-arg` applies only to rattler_fs's own targets, not to `rattler.exe`. rattler-bin has no build.rs, so the Windows binary still hard-imports `projectedfslib.dll` and dies at startup on machines without ProjFS — defeating the `ProjFsDllMissing` graceful-degradation path.~~ https://github.com/conda/rattler/issues/2572

## Performance (will dominate real-world experience)

- ~~**O(N²) ranged reads.** Both `text_ranged_read` and `binary_ranged_read` walk the output stream byte-by-byte from position 0 on every call. Reading a 500 MB `.so` through FUSE in 128 KB chunks is ~terabyte-scale byte work. Offsets are sorted → binary-search to the start position and `extend_from_slice` the spans.~~ https://github.com/conda/rattler/issues/2583
- ~~**Eager full-file scan at mount time.** `offsets` is a new rattler-only extension absent from essentially all real packages, so `VirtualFS::with_platform` falls back to `fs::read` on *every* prefix-bearing file (all the `.so`s and binaries), serially, on the caller's thread — defeating the "instant no-copy mount" goal.~~ https://github.com/conda/rattler/issues/2584
- ~~**Overlay metadata: full-JSON rewrite + fsync per op** under the state mutex → O(N²) for a bulk `pip uninstall`, blocking all concurrent lookups.~~ https://github.com/conda/rattler/issues/2585

## Minor / cleanup

Accidental unused workspace deps (`env_logger`, and the obscure `mem = "0.5.0"` — looks like a mangled `memchr` edit, remove it); `memchr` added to `rattler_conda_types/Cargo.toml` but unused there; CI path filter omits `crates/rattler-bin/**` and root `pixi.toml` (so a `mount.rs`-only PR skips the e2e); action pins lag repo convention (checkout v4 vs v6); mutex `.lock().unwrap()` poisoning throughout turns one panic into a mount-wide outage; single non-looping `read()` treats a short read as EOF; stat-failure fabricates a 0-byte file instead of EIO; overlay copy-up isn't crash-safe (no tmp+rename+fsync on the data path, unlike the metadata path); `.rattler_fs_state.json` is reachable and writable from inside the mount; `PathType::Directory` entries modeled as file leaves give inconsistent stat/readdir; `VirtualFS::new` is `pub` but takes a `pub(crate)` type; `MountHandle::Fuse(fuser::...)` leaks `fuser` into the public API (SemVer hazard).

## Test coverage — the core weakness

~~The unit and parity suites are good but cover only internally-consistent happy paths. **Nothing** exercises: shebang files (finding 1 & shebang divergence), malformed offsets (finding 2), overlay symlinks / dir-rename-with-children / exec-bit / lookup-in-opaque-dir (finding 3), adversarial NFS filenames (finding 4), mount-path-longer-than-placeholder, Windows entry-point layout, clobbering, dual-CD codesign, or any of the three adapters at unit level. The e2e nushell harness is rigorous (I had an agent verify its pass/fail detection empirically) but validates behavior, not these boundaries. Adding cases in these gaps is where the real hardening is.~~
https://github.com/conda/rattler/issues/2586

## Recommendation

**Request changes.** The concept and structure are excellent and worth landing, but I'd block on: (1) the `link.rs` shebang+offsets install corruption, (2) untrusted-offset panic validation, (3) the three overlay copy-up corruptions + opaque-lookup bug, (4) the NFS access-control/path-traversal exposure (at least documented, ideally fixed), and (5) the release-pipeline breakage. The mount-path panic, VFS/installer divergences (shebang, clobber, Windows entry points, pypi), and the two structural performance issues should follow before this is used on real environments at scale. Given the surface area, splitting the `paths.json`/`link.rs` offset work into its own reviewed PR (it's independently useful and independently risky) would de-risk the rest.

chrisburr:
> @chrisburr I propose that we get this merged and let @Dagmar-Dinjens continue with followups? Would be happy to jump on a call to discuss this in more detail if that helps

Yes, I am very happy to have @Dagmar-Dinjens pick this up again, especially as I'd like to spend some time on the package cache itself to help with distribution via CVMFS which is orhtogonal but useful for this.

If you want to arrange a chat, the conda-forge Zulip is probably the most reliable way for you to get my attention.

> I left way more comments than what I think is actually required to merge this. I think a lot of them we can address in follow-up PRs.

How do you want to proceed on that side? I'll work my way though replying and implement some of them? Or do you want to use the comments as a chance to onboard @Dagmar-Dinjens?

> I think we should name the crate rattler_vfs instead of rattler_fs. But I can be persuaded.

I'm not sure how many people know VFS as a concept. That said, the name has been in my brain for a long time so the problem might just be that it's different from what I've become used to. I don't have a strong opinion either way.

> We should not yet publish this crate automatically. This requires annotating the Cargo.toml with package.publish = false.

👍

> Btw why is windows nfs not supported?

Initially I tried to use only NFS and just ProjFS for later but I ran into a bunch of limitations. My memory is a bit fuzzy as this was a couple of months ago I think the main ones were:

* it couldn't satisfy the `pixi` use case in of mounting the environment in `%CWD%/.pixi/...` without hacky tricks
* limited to ~20 mounts before you run out of drive letters

crates/rattler_vfs/src/nfs_adapter.rs
chrisburr: I think the best supported crate I found only had v3 and I wasn't aware of anything about v4 really mattering.

crates/rattler_vfs/src/overlay_fs/inode.rs
baszalmstra: https://github.com/conda/rattler/issues/2569

crates/rattler_vfs/src/overlay_fs/inode.rs
baszalmstra: https://github.com/conda/rattler/issues/2570

crates/rattler-bin/Cargo.toml
chrisburr: I think I just did this out of an abundance of caution given it was a massive PR. An earlier version of the code had a `ratterfs` binary but it became clear that it was duplicating a ton of what was in `rattler-bin` and made `rattler-fs` far heavier than it needed to be for the `pixi` use case.

I can remove it.

crates/rattler_vfs/src/codesign.rs
baszalmstra: https://github.com/conda/rattler/issues/2571

crates/rattler-bin/src/commands/mount.rs
chrisburr: How would you handle this? Default to `pixi.lock` with an `--lock-file` flag to override?

crates/rattler_vfs/build.rs
baszalmstra: Also found by fable:

> ProjFS /DELAYLOAD never reaches the binary — build.rs's rustc-link-arg applies only to rattler_fs's own targets, not to rattler.exe. rattler-bin has no build.rs, so the Windows binary still hard-imports projectedfslib.dll and dies at startup on machines without ProjFS — defeating the ProjFsDllMissing graceful-degradation path.

https://github.com/conda/rattler/issues/2572

crates/rattler-bin/Cargo.toml
baszalmstra: @chrisburr if you can take a look at this question?

crates/rattler_vfs/Cargo.toml
baszalmstra: @Dagmar-Dinjens If you can look into this perhaps.

crates/rattler-bin/src/commands/mount.rs
baszalmstra: https://github.com/conda/rattler/issues/2573

crates/rattler-bin/src/commands/mount.rs
baszalmstra: https://github.com/conda/rattler/issues/2573

crates/rattler-bin/src/commands/mount.rs
baszalmstra: https://github.com/conda/rattler/issues/2573

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2574

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2575

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2575

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2575

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2575

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2575

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2572

crates/rattler_vfs/src/codesign.rs
chrisburr: At a glance I don't think there is any reason not to. When I wrote this I thought it might be worth having as the default path.

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2576

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2576

crates/rattler_vfs/src/lib.rs
baszalmstra: https://github.com/conda/rattler/issues/2576

crates/rattler_vfs/src/lib.rs
baszalmstra: @chrisburr Maybe you can take a look at this?

crates/rattler_vfs/src/lib.rs
baszalmstra: See: https://github.com/conda/rattler/issues/2576

crates/rattler_vfs/src/metadata_tree.rs
baszalmstra: https://github.com/conda/rattler/issues/2570

crates/rattler_vfs/build.rs
chrisburr: This is what I do isn't it?

https://github.com/chrisburr/rattler/blob/742a2d818d6c3451703d930139cfefde2e837e07/crates/rattler_fs/src/lib.rs#L792-L811

test-data/rattler-fs/pixi.toml
baszalmstra: This seems unused. Or am I missing something @chrisburr ?

Cargo.toml
baszalmstra: I think this can just go @chrisburr ?

crates/rattler-bin/src/commands/mount.rs
baszalmstra: https://github.com/conda/rattler/issues/2577

crates/rattler_vfs/src/vfs_ops.rs
baszalmstra: https://github.com/conda/rattler/issues/2578

crates/rattler_vfs/src/lib.rs
chrisburr: Yeah it could. I went around in circles for this for quite a while before deciding that regardless FUSE is probably the happy path on Linux.

test-data/rattler-fs/pixi.toml
chrisburr: It's used here: https://github.com/chrisburr/rattler/blob/742a2d818d6c3451703d930139cfefde2e837e07/scripts/e2e/rattler-fs-mount.nu#L30

Well the lock file which is made from this, but I thought it was good to have the source of the lock file.

chrisburr: This is the environment that the e2e test mounts, rather than the environment that it runs in.

Cargo.toml
chrisburr: I think this is left over from an earlier version of this before I figured out how to interface pixi more cleanly.

crates/rattler_vfs/src/fuse_adapter.rs
chrisburr: Probably, this predates fuser supporting threads but it seems like that has changed.

baszalmstra:
> How do you want to proceed on that side? I'll work my way though replying and implement some of them? Or do you want to use the comments as a chance to onboard @Dagmar-Dinjens?

I have created issues for all the non-blocking items in this PR (prefixed with vfs:). I also updated the fable review. So all the items that are still open should be tackled in this PR. I tagged @Dagmar-Dinjens in some of them, I hope she can take care of those? @chrisburr Would you be able to take care of the remaining ones?

Once merged we can work on the individual issues. We can simply assign people to make sure we dont work on the same things. 

Does that help/make sense?

> I'm not sure how many people know VFS as a concept. That said, the name has been in my brain for a long time so the problem might just be that it's different from what I've become used to. I don't have a strong opinion either way.

So I asked ChatGPT for the deciding vote:

> I’d vote rattler_vfs.
>
> Why:
> 
> rattler_fs sounds like a broad filesystem utility crate for Rattler: path helpers, atomic writes, cache layout helpers, symlink handling, file locking, etc.
> 
> This implementation is much more specific: it presents conda environments as a virtual filesystem, backed by FUSE/NFS/ProjFS, with on-the-fly prefix replacement, generated files, code signing handling, and overlays. The PR itself describes it as serving data “using a virtual filesystem” and having “3 VFS backends.” The changed files also include vfs_ops.rs and virtual_fs.rs, which makes vfs the more precise abstraction.

@chrisburr Does that resonate with you?

crates/rattler-bin/Cargo.toml
baszalmstra: Yeah, the rattler binary is to showcase all capabilities anyway. I think it would be perfectly fine to include it by default.

test-data/rattler-fs/pixi.toml
baszalmstra: Ah that makes sense. Perhaps you can add a README or comment in the file to explain that?

chrisburr:
That's a compelling arugment for `rattler_vfs` and it's growing on me over time. I've made the change 👍

> We should not yet publish this crate automatically. This requires annotating the Cargo.toml with package.publish = false.

This is in contradiction with having `rattler_vfs` be a non-optional feature in `rattler_bin`: https://github.com/conda/rattler/pull/2566#discussion_r3542729981

For now I've gone with removing the feature and publishing the crate.

The biggest open question from my perspective is https://github.com/conda/ceps/pull/179 and especially the handling of non-UTF-8 strings.

crates/rattler_vfs/src/nfs_adapter.rs
chrisburr: Another thing to add here is https://github.com/facebook/sapling/blob/61b3cde3bffdd7553094a9b4738af369abdf935b/eden/fs/docs/macOS.md

In general I found EdenFS's documentation to be a useful reference. For example the [Windows](https://github.com/facebook/sapling/blob/61b3cde3bffdd7553094a9b4738af369abdf935b/eden/fs/docs/Windows.md) docs were what made me finally give up on trying to support the read-only flag for ProjFS.

crates/rattler_vfs/src/lib.rs
```diff
@@ -0,0 +1,1787 @@
...
+    // `soft` with a bounded timeout so a dead userspace NFS server (e.g. the
+    // sidecar crashed) surfaces EIO to clients instead of wedging them in
+    // uninterruptible D-state on every access, which a hard mount would. The
+    // server is always local, so the usual soft-mount data-loss caveat is moot:
+    // if it dies, the mount is gone regardless. timeo is in deciseconds.
+    let mut opts = format!(
+        "noacl,nolock,soft,timeo=100,retrans=3,vers=3,tcp,port={port},mountport={port},rsize=1048576"
+    );
+    if read_only {
+        opts.push_str(",ro");
```
Dagmar-Dinjens, line 1107: As the read-only mount serves an immutable package set and nothing will change for as long as the mount is active, we could let the client cache attributes aggressively instead of periodically re-validating them with fresh GETTR/LOOKUP round trips. It helps longer running mounts with serving repeat metadata from its cache.
```suggestion
        opts.push_str(",actimeo=3600,ro");
```

chrisburr:
Thanks for the fixes @Dagmar-Dinjens, I've rebased and built back up to https://github.com/chrisburr/pixi-rattlerfs and everything seems to be in order.

PR #2568, opened 2026-07-08
perf: speed up CUDA virtual package detection

### LINKED ISSUE #6775: Release binary does not detect __cuda_arch on Jetson Orin
### Checks

- [x] I have checked that this issue has not already been reported.
- [x] I have confirmed this bug exists on the latest version of Pixi, using `pixi --version`.

### Reproducible example

Commands I ran and their output:

```shell
$ pixi info
System
------------
       Pixi version: 0.76.1
        TLS backend: rustls
           Platform: linux-aarch64
   Virtual packages: __unix=0=0
                   : __linux=5.15.148=0
                   : __glibc=2.35=0
                   : __cuda=12.6=0
                   : __archspec=1=neoverse_n1
          Cache dir: /home/bdai/.cache/rattler/cache
       Auth storage: /home/bdai/.rattler/credentials.json
   Config locations: No config files found

Global
------------
            Bin dir: /home/bdai/.pixi/bin
    Environment dir: /home/bdai/.pixi/envs
       Manifest dir: /home/bdai/.pixi/manifests/pixi-global.toml
```

Pixi was installed from the official release binary and updated using `pixi self-update`.

### Issue description

I am trying to detect Jetson Orin devices using rich platforms. Pixi correctly detects `__cuda=12.6`, but it does not detect `__cuda_arch=8.7`.

The CUDA compute capability would be useful for matching this hardware because `8.7` identifies this family of NVIDIA devices.

Based on a discussion with one of the developers, `__cuda_arch` is detected by dynamically loading the CUDA driver. The released Linux binary targets `musl`, preventing it from loading the driver shared object and causing the detection to fail.

The following override provides a workaround while preserving the automatically detected CUDA version:

```shell
$ CONDA_OVERRIDE_CUDA_ARCH=8.7 pixi info
System
------------
       Pixi version: 0.76.1
        TLS backend: rustls
           Platform: linux-aarch64
   Virtual packages: __unix=0=0
                   : __linux=5.15.148=0
                   : __glibc=2.35=0
                   : __cuda=12.6=0
                   : __cuda_arch=8.7=0
                   : __archspec=1=neoverse_n1
          Cache dir: /home/bdai/.cache/rattler/cache
       Auth storage: /home/bdai/.rattler/credentials.json
   Config locations: No config files found

Global
------------
            Bin dir: /home/bdai/.pixi/bin
    Environment dir: /home/bdai/.pixi/envs
       Manifest dir: /home/bdai/.pixi/manifests/pixi-global.toml
```

### Expected behavior

The official Pixi release binary should detect the CUDA compute capability and expose it as a virtual package without requiring an override:

```plaintext
__cuda_arch=8.7=0
```

baszalmstra:
### Description

Detecting `__cuda` and `__cuda_arch` takes about 1.5 seconds on Windows when the GPU has been idle, which is the normal case for a command line tool. Measured with CUDA 13.0 and compute capability 8.9, each in a fresh process:

| driver state | detection |
|---|---|
| idle, or first use after boot | 1.5 s |
| used seconds earlier | 460 ms |
| through the cache | 1.0 ms |

Nearly all of it is one step. Loading NVML takes 0.5 ms and enumerating devices adds 24 ms. The rest is the driver version query, which loads the CUDA driver library and starts the user-mode driver to read the version out of it. If the driver powered down in the meantime it has to come back up first, and that is where the 1.5 seconds goes. No detection API avoids this, so the only fix is to not do it again.

So this caches the result on disk. A cache hit never touches the driver, which also leaves an idle GPU asleep instead of waking it on every invocation.

The cache is keyed on the boot session, a fingerprint of the installed driver and one of the visible GPUs, so reboots, driver updates and plugging in an eGPU all invalidate it. A TTL covers whatever the fingerprints miss. Callers pass the cache directory through a new `cache_dir` argument, and `None` disables it. That is a breaking change to the detection API. The Python bindings take the same argument.

It also fixes three things in the detection itself:

- Detection uses NVML instead of libcuda, so `CUDA_VISIBLE_DEVICES` no longer affects the result. Under a job scheduler or in CI, where that variable is often set, `__cuda` and `__cuda_arch` could come out wrong or disappear entirely.
- `__cuda_arch` now works on musl, where it was always absent.
- `__cuda` is detected even when the driver fails to initialize, since the version query no longer needs initialization.

Fixes https://github.com/prefix-dev/pixi/issues/6775

### How Has This Been Tested?

- Unit tests for the cache roundtrip and for invalidation on reboot, driver change and TTL, alongside the existing detection tests.
- Benchmarks per detection path, as ignored tests. The numbers above come from running them on Windows with an NVIDIA GPU.
- `cargo clippy` and `cargo fmt` pass, the crate builds for `x86_64-unknown-linux-musl` and both Windows targets, and `cargo check` passes for `py-rattler`.
- Checked the nvidia-smi fallback against NVML on hardware: both report CUDA 13.0 and compute capability 8.9.

### AI Disclosure

Tools: Claude Code

### Checklist:


crates/rattler_virtual_packages/src/cuda.rs
```diff
@@ -109,206 +177,338 @@ pub fn cuda_version() -> Option<Version> {
...
+            Err(err) => {
+                tracing::trace!(library_path = *path, error = %err, "failed to load NVML library");
+            }
+        }
+    }
 
-    // Initialize the CUDA library
-    if unsafe { cu_init(0) } != 0 {
+    let Some(library) = library else {
+        tracing::debug!("could not load NVML library from any known path");
```
wolfv, line 415: Should we fall back to try `nvidia-smi`?

wolfv:
CUDA detection logic looks good to me overall 👍 

I can’t fully validate the DLL/loading behavior across all platforms, but the general approach makes sense. The boot cache also looks great, especially for Windows startup time. 

One small thing I was wondering about: for the fallback paths, can we make sure we consistently try the nvidia-smi-based detection where possible? From the discussion it sounds like the current MUSL/fallback behavior is similar to what cuda-oxide does, which seems reasonable. 

I also briefly wondered whether we could extract some constants directly from the driver with something like goblin, even on MUSL systems, but for __cuda_arch we need to load/query the CUDA side anyway, so that probably doesn’t buy us much here. 

So from my side: looks good, modulo the fallback consistency check.

baszalmstra:
Could you maybe recheck, I added a lot of robustness checks

hunger:
On Linux you will get problems with this:

Containers share a kernel and with that their boot id. You might want to hash that together with the machine ID (if available). At least on systemd-based systems you should be able to tell different containers apart from each other.

hunger:
Actually: Thinking about that, the graphics card is managed by the kernel, which is the same for all containers anyway. So this should be fine (together with the device-node check for the graphics cards).

wolfv:
Tested on Fedora with CUDA GPU: works both with MUSL and regular glibc build! Code looks good, too.

wolfv (approved):

PR #2588, opened 2026-07-08
Optimize macOS code signing with single-pass patch+sign for thin Mach-O binaries

wolfv:
### Description

This PR optimizes the code signing process for macOS binaries during package installation by implementing a single-pass streaming approach for thin Mach-O binaries.

**Key improvements:**

1. **Single-pass patch+sign+hash for thin Mach-O binaries**: Instead of patching the binary, then re-signing it (requiring a full re-read and re-hash), thin Mach-O binaries now flow through a streaming signer in one pass. The patched bytes are signed and hashed on-the-fly, eliminating the extra I/O overhead.

2. **In-process code signing**: Replaced reliance on spawning `/usr/bin/codesign` with the `arwen_codesign` crate for in-process ad-hoc signing. This is orders of magnitude faster and works cross-platform (e.g., signing macOS binaries from Linux).

3. **Smarter signing decisions**: Added `binary_content_will_change()` to detect whether prefix replacement actually modifies the binary content. If the placeholder doesn't occur in the binary, the original linker signature is preserved, avoiding unnecessary re-signing.

4. **Fallback support**: On macOS hosts, `/usr/bin/codesign` remains available as a fallback for fat binaries and any edge cases the in-process signer cannot handle.

**Technical changes:**

- Refactored `link_file()` to compute whether re-signing is needed upfront
- Added `patch_and_sign_single_pass()` for the optimized streaming path
- Updated `apple_codesign.rs` to use `arwen_codesign` for in-process signing
- Added comprehensive tests for thin and fat Mach-O binaries with and without placeholder replacement
- Added test fixtures (`test_exe_linker_signed`, `test_exe_fat`) for macOS code signing tests

Fixes #<issue_number>

### How Has This Been Tested?

- Added three new unit tests in `link.rs`:
  - `test_macho_binary_is_resigned_after_prefix_replacement`: Verifies thin Mach-O binaries are re-signed after prefix replacement and the hash/size are correct
  - `test_macho_binary_without_placeholder_is_not_resigned`: Confirms that when the placeholder doesn't occur, the original linker signature is preserved
  - `test_fat_macho_binary_is_resigned`: Validates fat (universal) binaries are properly re-signed with all architecture slices
- Tests run on any host platform (signing happens in-process)
- Existing tests continue to pass

### Checklist:

https://claude.ai/code/session_01PWnq7uDMrjDi3q3b9P4qn3


Cargo.toml
```diff
@@ -28,6 +28,8 @@ ambient-id = { version = "0.0.11", default-features = false }
 anyhow = "1.0.98"
 apple-native-keyring-store = { version = "1.0.0", default-features = false }
 archspec = "0.2.0"
+# TODO: switch to the crates.io release once arwen-codesign >= 0.1.0 is published
+arwen-codesign = { git = "https://github.com/wolfv/arwen", branch = "claude/arwen-codesign-rattler-frrg3u" }
```
baszalmstra, line 32: If this is released Id love to pull this in!

PR #2590, opened 2026-07-10
feat(rattler_git): unify git cache with pixi's implementation

wolfv:
### Description

pixi's `pixi_git` crate and `rattler_git` share a common ancestor (both derived from uv-git) but have drifted apart. This PR ports everything `pixi_git` has grown back into `rattler_git`, so pixi, rattler-build and rattler can all use **one low-level git cache code base**:

**Ported from pixi:**
- **Git LFS support**: new tri-state `CheckoutOptions::lfs`:
  - `Some(true)` — fetch LFS objects into the git database (`git lfs fetch <url> <rev>` + `git lfs fsck --objects` validation) and run the smudge filter during checkout so pointers materialise into real content. Cached databases are only reused when their LFS artifacts validate, and `Fetch::lfs_ready()` reports the outcome.
  - `Some(false)` — force-skip the smudge filter (`GIT_LFS_SKIP_SMUDGE=1`), the previous `rattler_git` behavior and the default.
  - `None` — leave `GIT_LFS_SKIP_SMUDGE` untouched so ambient git config applies (pixi's default when `PIXI_GIT_LFS` is unset).
- **`LazyClient`** instead of an eager `ClientWithMiddleware`, so the HTTP client is only initialized if the GitHub fast path actually runs. It mirrors `rattler_networking::LazyClient` but lives in this crate to keep `rattler_git` dependency-light (pulling all of `rattler_networking` for one 60-line type also breaks downstream `[patch.crates-io]` testing, since released rattler crates pin `rattler_networking` exactly). `impl Into<LazyClient>` keeps existing callers source-compatible.
- **Cloneable errors**: `GitError` and `OidParseError` are now `Clone` (Arc-wrapped inner errors) so fetch results can be cached/shared (pixi memoizes them in its compute engine).
- **Graceful GitHub fast path**: non-success statuses (403 rate limit, 404, 422, 5xx) fall back to a regular `git fetch` instead of failing the whole fetch.
- **`GIT_TERMINAL_PROMPT=0`** on fetches, so unreachable remotes fail fast instead of hanging on a credential prompt.
- **`RepositoryUrl` canonicalizes on `Deserialize`**.
- Generic `lfs_enabled_from_env(var_name)` helper for the tri-state env-var convention (pixi passes `PIXI_GIT_LFS`).

**Kept from rattler_git** (things pixi lacked and now gains): submodule support with relative-URL resolution against the real source URL, `protocol.file.allow=always` for file-based submodules, and corrupted-cache detection/re-clone.

Companion PRs: prefix-dev/rattler-build (adopts built-in LFS, drops its hand-rolled `git lfs` pull) and prefix-dev/pixi (replaces the `pixi_git` implementation with a thin shim over this crate).

Note: this is a breaking change for `rattler_git` (new `CheckoutOptions` field, `LazyClient` parameter, `Clone`-able errors), so it should be released as a minor bump (0.2).

### How Has This Been Tested?

- `cargo test -p rattler_git` — existing tests plus pixi's LFS integration tests ported over with a self-contained fixture (build a repo with `.gitattributes` `filter=lfs`, then assert: pointers stay pointers with `lfs: Some(false)`, blobs materialise with `Some(true)`, and warm-cache fetches validate LFS artifacts without network). Tests skip themselves when `git-lfs` is not installed.
- `cargo clippy -p rattler_git --all-targets` and `cargo fmt` are clean.
- The companion pixi branch compiles and passes its command-dispatcher integration test suite against this branch via `[patch.crates-io]`; the companion rattler-build branch passes its source-cache test suite (30 tests, including a new end-to-end LFS test).

### AI Disclosure

Tools: Claude Code

```plaintext
We have source / git caches in rattler, rattler-build and Pixi. I think Pixi
"git cache" has drifted a bit from rattler's. Can you port everything needed
in Pixi to rattler, so that we can use one low level code base for this?
Then open a draft PR to Pixi and rattler with the changes, and possibly
rattler-build if needed. Things that might be different are related to lfs,
submodules, ..
```

### Checklist:

https://claude.ai/code/session_013EJTWLNjUFMPPPoT8746Me

---
_Generated by [Claude Code](https://claude.ai/code/session_013EJTWLNjUFMPPPoT8746Me)_


crates/rattler_git/src/client.rs
```diff
@@ -0,0 +1,68 @@
+//! A lazily-initialized HTTP client.
```
baszalmstra: Why not use the one from rattler_networking?

crates/rattler_git/Cargo.toml
```diff
@@ -4,7 +4,7 @@ version = "0.1.3"
 edition.workspace = true
 license.workspace = true
 repository.workspace = true
-description = "Git operations support for rattler-build"
+description = "Git operations support for rattler-build and pixi"
```
baszalmstra:
```suggestion
description = "Git operations support for rattler based crates"
```

crates/rattler_git/src/git.rs
```diff
@@ -811,25 +1011,26 @@ fn github_fast_path(
 
     runtime.block_on(async move {
         tracing::debug!("Attempting GitHub fast path for: {url}");
-        let mut request = client.get(&url);
+        let mut request = client.client().get(&url);
         request = request.header("Accept", "application/vnd.github.3.sha");
-        request = request.header("User-Agent", "pixi");
+        request = request.header("User-Agent", "rattler");
```
baszalmstra: I think we should only do this if the client doesnt already have an agent configured.

baszalmstra (changes requested):

crates/rattler_git/src/client.rs
wolfv: Reconsidered this: all intended consumers already use rattler_networking, so duplicating LazyClient is not worthwhile. Updated the PR to depend on and re-export rattler_networking::LazyClient, and removed the local implementation.

baszalmstra (approved):

PR #2595, opened 2026-07-13
feat(rattler_upload): support basic Artifactory auth again

ytausch:
### Description

https://github.com/prefix-dev/rattler-build/issues/1254 and https://github.com/prefix-dev/rattler-build/pull/1280 deprecated support for HTTP Basic Auth for uploading to Artifactory, assuming that, in every situation, the HTTP Basic Auth password can just be passed as a Bearer token.

However, there are corporate IT situations where, in CI, only a username and password for a local Artifactory user are available. The password of local Artifactory users cannot be passed as a bearer token to Artifactory (error: `401 Unauthorized, Token failed verification: signature`). This also makes sense conceptually, since, unlike tokens, two users can have the same password.

### How Has This Been Tested?

I tested this change on Windows with the Artifactory instance of a corporate IT network. Since I only have access to a local username and password in a CI system unavailable for testing, I tested HTTP Basic auth not in the situation described above, but with my username and an Artifactory token as password:

1. Navigate to the Artifactory instance
2. Click User Menu -> Edit Profile
3. Generate an Identity Token
4. Run `rattler.exe upload artifactory --url https://artifactory.example.com/artifactory -c my-channel --username USERNAME --password TOKEN ./package.conda`
5. Observe successful upload.

### AI Disclosure

Tools: ChatGPT Codex

```plaintext
Please find out why upload_package_to_artifactory only works with Bearer tokens. If you don't find any good reason, please make it support basic auth as well.
```
(after this, some minor manual rework followed, I don't share the prompts here for brevity)

### Breaking Change

Marked as breaking because public `ArtifactoryData` changes from

```rust
pub token: Option<String>
```

to 

```rust
pub authentication: Option<ArtifactoryAuthentication>
```

### Checklist:

Cc @pavelzw

### DIFF
--- a/crates/rattler_upload/src/upload/opt.rs
+++ b/crates/rattler_upload/src/upload/opt.rs
@@ -185,16 +185,16 @@ pub struct ArtifactoryOpts {
     #[arg(short, long = "channel", env = "ARTIFACTORY_CHANNEL")]
     pub channels: String,
 
-    /// Your Artifactory username for HTTP basic authentication. Requires a password.
-    #[arg(long, env = "ARTIFACTORY_USERNAME")]
+    /// Your Artifactory username for HTTP basic authentication.
+    #[arg(long, env = "ARTIFACTORY_USERNAME", requires = "password")]
     pub username: Option<String>,
 
-    /// Your Artifactory password for HTTP basic authentication. Requires a username.
-    #[arg(long, env = "ARTIFACTORY_PASSWORD")]
+    /// Your Artifactory password for HTTP basic authentication.
+    #[arg(long, env = "ARTIFACTORY_PASSWORD", requires = "username")]
     pub password: Option<String>,
 
-    /// Your Artifactory bearer token. Takes precedence over username/password authentication.
-    #[arg(short, long, env = "ARTIFACTORY_TOKEN")]
+    /// Your Artifactory token for bearer authentication.
+    #[arg(short, long, env = "ARTIFACTORY_TOKEN", conflicts_with_all = ["username", "password"])]
     pub token: Option<String>,
 }
 
@@ -204,12 +204,7 @@ pub enum ArtifactoryAuthentication {
     /// Authenticate with a bearer token.
     Token(String),
     /// Authenticate with HTTP basic authentication.
-    Basic {
-        /// The username to use for basic authentication.
-        username: String,
-        /// The password to use for basic authentication.
-        password: String,
-    },
+    Basic { username: String, password: String },
 }
 
 #[derive(Debug)]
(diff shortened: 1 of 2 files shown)


pavelzw:
> Marked as breaking

conventional commits breaking `!` is not necessary here, release-plz will look at the API anyway and figure out on its own if something is breaking or not. only if there is a breaking behavior change, the `!` is needed

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -183,52 +185,66 @@ pub struct ArtifactoryOpts {
...
-    /// Your Artifactory password
-    #[arg(long, env = "ARTIFACTORY_PASSWORD", hide = true)]
+    /// Your Artifactory password for HTTP Basic authentication. Requires a username.
+    #[arg(long, env = "ARTIFACTORY_PASSWORD")]
     pub password: Option<String>,
 
-    /// Your Artifactory token
+    /// Your Artifactory bearer token. Takes precedence over username/password authentication.
     #[arg(short, long, env = "ARTIFACTORY_TOKEN")]
     pub token: Option<String>,
```
pavelzw: in clap, it's possible to make other arguments required/force other args to not be set

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -183,52 +185,66 @@ pub struct ArtifactoryOpts {
...
+pub enum ArtifactoryAuthentication {
+    /// Authenticate with a bearer token.
+    Token(String),
+    /// Authenticate with HTTP Basic authentication.
+    Basic {
+        /// The username to use for basic authentication.
+        username: String,
+        /// The password to use for basic authentication.
+        password: String,
+    },
```
pavelzw: capitalization between basic authentication and Basic authentication. keep consistent please

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -183,52 +185,66 @@ pub struct ArtifactoryOpts {
...
+            channels,
+            username,
+            password,
+            token,
+        } = value;
+
+        match (username, password, token) {
+            (_, _, Some(token)) => Ok(Self::new(url, channels, Some(token))),
+            (Some(username), Some(password), None) => {
+                Ok(Self::new(url, channels, None).with_basic_auth(username, password))
```
pavelzw: this looks a bit complicated.
`ArtifactoryData`'s `new` also now doesn't make sense anymore. if you redesign that, it should be a bit easier...

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -183,52 +185,66 @@ pub struct ArtifactoryOpts {
...
             }
-            _ => None,
-        };
-        Ok(Self::new(value.url, value.channels, token))
+            (Some(_), None, None) => Err(miette::miette!(
+                "Artifactory username provided without a password"
```
pavelzw: when you use the clap args (comment above), you can make this an `unreachable!`

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -183,52 +185,66 @@ pub struct ArtifactoryOpts {
...
+            )),
+            (None, Some(_), None) => Err(miette::miette!(
+                "Artifactory password provided without a username"
```
pavelzw: when you use the clap args (comment above), you can make this an `unreachable!`

crates/rattler_upload/src/upload/opt.rs
```diff
@@ -238,9 +254,41 @@ impl ArtifactoryData {
...
+        self
+    }
+}
+
+#[cfg(test)]
+mod artifactory_tests {
+    use super::{ArtifactoryAuthentication, ArtifactoryData, ArtifactoryOpts};
+
+    #[test]
+    fn username_and_password_are_kept_as_basic_auth() {
```
pavelzw: let's just make this a regular test that does one basic auth and one bearer auth check instead of letting this be a regression test

crates/rattler_upload/src/upload/opt.rs
ytausch: The codebase spells it lowercase otherwise. I think uppercase is the most "correct" spelling, but in order to keep the diff small, I made it lowercase in c721f85492567d3a0199fe183bd82811b4d438de.

crates/rattler_upload/src/upload/opt.rs
ytausch: fixed in 7b4db9c1a00af90d100a132a289b956f88ac2c73

crates/rattler_upload/src/upload/opt.rs
ytausch: I used `expect` in 7b4db9c1a00af90d100a132a289b956f88ac2c73

crates/rattler_upload/src/upload/opt.rs
ytausch: 7b4db9c1a00af90d100a132a289b956f88ac2c73

pavelzw (approved):

PR #2603, opened 2026-07-14
ci: report breaking API changes on pull requests

baszalmstra:
### Description

Breaking API changes currently only show up when `release-plz` prepares a release. This adds the same feedback directly to pull requests.

The workflow checks changed public crates against the PR base. It adds or updates a comment when `cargo-semver-checks` finds a breaking change, then removes that comment when the issue is fixed. Tooling failures remain visible as failed checks and do not clear an existing warning.

The check runs with read-only permissions. A separate `workflow_run` workflow handles comments so this also works safely for fork PRs.

The workflow logic lives in Nushell scripts. With `cargo-semver-checks` 0.48 installed, the same check and Markdown can be reproduced locally:

```shell
pixi run -e semver-check breaking-changes --base-ref main
```

### How Has This Been Tested?

- `pixi run actionlint`
- `pixi run -e lint dprint check .github/workflows/breaking-changes.yml .github/workflows/breaking-changes-comment.yml scripts/check-breaking-changes.nu scripts/update-breaking-change-comment.nu scripts/test-breaking-changes.nu`
- `pixi run -e semver-check test-breaking-changes`
- `pixi run cargo-fmt`
- Ran the local command with a temporary removal of `simple_spawn_blocking::Cancelled`; it selected that crate, detected `struct_missing`, and generated `breaking-changes.md`
- Ran the local command without public crate changes and confirmed it exits successfully without generating a comment
- Ran the updated `Check semver` workflow on this PR; it passed and produced the expected success artifact

`pixi run cargo-clippy` was also attempted. It currently fails on existing Windows-only `collapsible_if` warnings in `rattler_menuinst`, outside this workflow-only change.

### AI Disclosure

Tools: ChatGPT using the pi coding agent

### Checklist:


.github/workflows/breaking-changes-comment.yml
```diff
@@ -0,0 +1,229 @@
...
+              "the pull request's base revision.\n\n"
+              "<details>\n"
+              "<summary>Details</summary>\n\n"
+              f"{details}\n\n"
+              "</details>\n"
+          )
+          Path("semver-comment.md").write_text(body, encoding="utf-8")
+          PY
+
+      - name: Create or update failure comment
```
pavelzw: will this work with a github token from a fork? i doubt it would work without `pull_request_target`...

baszalmstra: Clanker says: the current design should work:                                                                                                                                                                                    
                                                                                                                                                                                                               
 - The `pull_request` workflow runs fork code with a read-only token.                                                                                                                                                 
 - The separate `workflow_run` workflow runs from the default branch and receives a new token for the base repository with `pull-requests: write`.
 - GitHub explicitly supports write-capable tokens in `workflow_run` workflows even when the triggering workflow was unprivileged.                                                                                     
                                                                                                                                                                                                                     
Using `pull_request_target` would be less safe here because `cargo-semver-checks` compiles PR-controlled build scripts and proc macros.

pavelzw: > receives a new token for the base repository

yes, so this token cannot write a comment in the PR on the conda/rattler repo, no?

baszalmstra: The base repository is `conda/rattler`. The PR and its comments also live in `conda/rattler`; only the head branch comes from the fork. The `workflow_run` therefore gets a `conda/rattler` token with `pull-requests: write`, which can update the PR comment.

pavelzw: i would still be surprised if this really worked given my past experience when i fiddled around with similar things...

happy to merge it in and test it out though.

but i don't really like that there is so much shell scripting going on inside of the workflow files. i would prefer some clean and readable nushell files that one can also reproduce locally

baszalmstra: I refactored the whole thing. It now uses a nushell script that you can also run locally.

.github/workflows/breaking-changes.yml
```diff
@@ -0,0 +1,60 @@
...
+        with:
+          fetch-depth: 0
+          lfs: false
+          persist-credentials: false
+          submodules: recursive
+
+      - name: Install Nushell
+        uses: hustcer/setup-nu@ccd5bb5426b05a32009c2ba967946231f3919c97 # v3.25
+        with:
+          version: "0.106.1"
```
pavelzw: we could theoretically also use setup-pixi, i think nushell is part of our pixi.toml...

.github/workflows/breaking-changes-comment.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
+        with:
+          lfs: false
+          persist-credentials: false
+          ref: ${{ github.event.repository.default_branch }}
+
+      - name: Install Nushell
+        uses: hustcer/setup-nu@ccd5bb5426b05a32009c2ba967946231f3919c97 # v3.25
+        with:
+          version: "0.106.1"
```
pavelzw: we could theoretically also use setup-pixi, i think nushell is part of our pixi.toml...

pavelzw (approved):
still unsure whether this even works permission-wise but feel free to try it out on `main`

.github/workflows/breaking-changes.yml
baszalmstra: Done!

pavelzw (approved):

pixi.toml
```diff
@@ -82,6 +82,21 @@ cargo-clippy = "cargo clippy --all-targets --workspace -- -D warnings -Dclippy::
 [environments]
 lint = { features = ["lint"], solve-group = "default" }
 
+#------------------------------
+# Breaking API change detection
+#------------------------------
+[feature.semver-check.dependencies]
+nushell = ">=0.106.1,<0.107"
```
pavelzw:
```suggestion
nushell = "*"
```

pavelzw (approved):

pavelzw:
some sort of progress indicator would be nice here, especially if it takes as long as it does 

image

PR #2605, opened 2026-07-17
fix(rattler_index): re-index packages whose file size changed under the same name

tenzinplatter:
### Problem

Incremental indexing (`index` without `--force`) decides what to hash purely by **filename**: it reuses every record already present in the previous `repodata.json` and only reads packages whose filename is new (`packages_to_add = uploaded − registered`). If a package is rebuilt and re-uploaded under the **same** filename, its bytes change but the old `sha256`/`size` are kept in `repodata.json`.

Clients then download the new bytes, hash them, and abort with a hash mismatch, e.g.:

```
hash mismatch when extracting <channel>/noarch/<pkg>.conda:
expected 519001f5…, got a434986e…
```

This assumes `name-version-build.<ext>` is immutable, which holds for well-behaved channels but breaks silently for any pipeline that republishes a rebuilt artifact under the same name (`.conda`/`.tar.bz2` archives aren't reproducible). Only a full `--force` reindex recovers.

### Fix

When reusing records from the previous `repodata.json`, also drop any package whose **current file size differs from the recorded size**, so it falls into `packages_to_add` and gets re-read and re-hashed. The listing already populates `content_length` for every entry, so this adds no extra stat/round-trips.

Size is the only signal available without re-reading every package; a rebuild that happens to produce the exact same byte count still requires `--force`. This is a strict improvement over the current filename-only behaviour and preserves incremental performance (unchanged packages are still reused).

### Test

`test_incremental_reindexes_replaced_package` indexes a package, replaces it in place with different bytes under the same filename, reindexes incrementally, and asserts `repodata.json`'s `size`/`sha256` now reflect the new bytes. It fails on `main` and passes with this change.

### DIFF
--- a/crates/rattler_index/src/lib.rs
+++ b/crates/rattler_index/src/lib.rs
@@ -767,21 +767,26 @@ async fn index_subdir_inner(
             }
         };
 
-    // List all the packages in the subdirectory.
-    let uploaded_packages: HashSet<DistArchiveIdentifier> = op
+    // List all the packages in the subdirectory, keeping each file's size. The
+    // listers populate content_length as part of the listing, so this needs no
+    // extra stat round-trips.
+    let uploaded_sizes: HashMap<DistArchiveIdentifier, u64> = op
         .list_with(&format!("{}/", subdir.as_str()))
         .await?
         .iter()
         .filter_map(|entry| {
-            if entry.metadata().mode().is_file() {
-                let filename = entry.name().to_string();
+            let meta = entry.metadata();
+            if meta.mode().is_file() {
                 // Check if the file is an archive package file.
-                DistArchiveIdentifier::try_from_filename(&filename)
+                DistArchiveIdentifier::try_from_filename(entry.name())
+                    .map(|id| (id, meta.content_length()))
             } else {
                 None
             }
         })
         .collect();
+    let uploaded_packages: HashSet<DistArchiveIdentifier> =
+        uploaded_sizes.keys().cloned().collect();
 
     tracing::debug!(
         "Found {} already uploaded packages in subdir {}.",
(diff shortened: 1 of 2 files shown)


crates/rattler_index/src/lib.rs
```diff
@@ -809,6 +814,37 @@ async fn index_subdir_inner(
...
+    // Re-index packages whose file no longer matches the size recorded in the
+    // previous repodata. `.conda`/`.tar.bz2` archives aren't reproducible, so a
+    // package rebuilt and republished under the same filename has different
+    // bytes; without this the stale record's sha256/size are kept and clients
+    // hit a hash mismatch on download. Dropping the mismatched entry here moves
+    // it into `packages_to_add` below, which re-reads and re-hashes it.
+    //
+    // Size is the only signal available without re-reading every package: a
+    // rebuild that lands on the exact same byte count still slips through, and
+    // `--force` remains the way to rebuild repodata unconditionally.
```
pavelzw, line 826: at least on S3, you might be able to get the hash of the file in some way. have you looked into this?

i am not really a fan of this approach as this won't cover all bases but just a subset where the size changes (which tbf in reality is probably most cases)

crates/rattler_index/src/lib.rs
```diff
@@ -809,6 +814,37 @@ async fn index_subdir_inner(
...
+        .filter(|(id, pkg)| match (uploaded_sizes.get(id), pkg.record.size) {
+            (Some(current), Some(recorded)) => *current != recorded,
+            (Some(_), None) => true, // no recorded size to trust
+            (None, _) => false,      // absent from the channel: handled above
+        })
+        .map(|(id, _)| id.clone())
+        .collect::<Vec<_>>();
+
+    if !stale.is_empty() {
+        tracing::info!(
```
pavelzw, line 838: this should be harsher than info imo; people should not do this, especially since it's not guaranteed that this will always trigger when the file sizes are the same but only the contents differ

```suggestion
        tracing::warn!(
```

crates/rattler_index/tests/integration/basic_indexing.rs
```diff
@@ -378,6 +378,95 @@ async fn test_index_repodata_revision_from_index_json() {
...
+        write_tar_bz2_package(
+            writer,
+            &build_dir,
+            &[info_dir.join("index.json"), build_dir.join("payload.txt")],
+            CompressionLevel::Default,
+            None,
+            None,
+        )
+        .unwrap();
+    };
```
pavelzw, line 414: i'm not sure whether this is the simplest way to write a conda package in a unit test. is there no helper function for this?

pavelzw (commented):
i don't really like this fix as it's just a workaround that doesn't cover all bases; so people could still screw up by uploading different packages that happen to be of the same size as before.

at least for S3, have you looked into getting the hash from the S3 metadata?

tenzinplatter:
Sorry this was meant to be against our fork for now! Sorry to waste your time, once we have been using it and cleaned it up a bit more I can reopen this

PR #2609, opened 2026-07-20
feat: let callers restrict which candidates a solve may pick

Hofer-Julian:
## Description

The motivation is pixi's improved offline support (prefix-dev/pixi#3482): restrict solving to conda packages that are already available locally, so a solve that succeeds can be installed without downloading anything. Pixi builds the exclusion map from the package cache and hands it to the solver.

### AI Disclosure

Tools: Claude Code (Opus 4.8)

### Checklist:


baszalmstra:
Since this is still in draft, I assume you are still working on this. I think there will be a lot of overhead from the many string allocations. Maybe we can use an `Arc<str>`?

Hofer-Julian:
> Since this is still in draft, I assume you are still working on this. I think there will be a lot of overhead from the many string allocations. Maybe we can use an `Arc<str>`?

Good idea. And yes, I am still iterating on this. Just wanted to already make sure CI is green on all platforms

baszalmstra (changes requested):
I don't think this is ready to merge yet. The split between `CacheIndex` and solver-side exclusions makes sense, but there are a few correctness issues in the new code.

## Correctness

### Excluding a pinned candidate drops the pin

[`pinned_packages` is passed as `locked_records`](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_solve/src/resolvo/mod.rs#L1058-L1062), but [`candidates.locked` is only set when the candidate is not excluded](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_solve/src/resolvo/mod.rs#L586-L597).

That silently removes a hard pin. The solver can select another version instead of reporting the contradictory constraints.

I think the pin and exclusion should be applied independently. Resolvo can then report the conflict as unsatisfiable.

### Archive deduplication happens too early

[`.conda`/`.tar.bz2` deduplication runs before the new exclusions are applied](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_solve/src/resolvo/mod.rs#L398-L470).

Consider a package available in both formats where only the `.tar.bz2` is cached. The deduplication keeps the preferred `.conda`, discards the cached tarball, and only then excludes the `.conda`. The solver now downgrades or fails even though a valid candidate is available locally.

[The new test currently encodes this behaviour](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_solve/tests/backends/main.rs#L883-L920) by expecting a lower version. I think it should instead select the same-version `.tar.bz2`.

### A cold shard is treated as an empty shard

[Cache-only mode returns an empty record set when the shard is not cached](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_repodata_gateway/src/gateway/sharded_subdir/tokio/mod.rs#L193-L219).

A missing shard means that we do not know its contents. It does not mean the package does not exist. This can also hide a package that is present in the package cache but whose repodata shard was never fetched.

Do we really want this to change the meaning of `UseCacheOnly` and `ForceCacheOnly` for every caller? I think the cache miss needs to remain distinct, with a fallback to cached full repodata where possible.

### `CacheIndex` does not mirror actual layered-cache lookup

[The index keeps only the first SHA for a path segment](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_cache/src/package_cache/mod.rs#L428-L445).

If the first layer contains SHA A and a later layer contains the requested SHA B, the index reports B as unavailable. Normal cache lookup can continue past A and use B.

The index should retain all relevant layer entries and report a match when any usable layer contains the package.

There are two other mismatches here:

- Missing, unreadable, or truncated metadata becomes `None`, which [is treated as a matching hash](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_cache/src/package_cache/cache_lock.rs#L32-L50).
- [`contains_record` always uses URL-origin hashing](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler_cache/src/package_cache/mod.rs#L87-L108), while the installer [uses path-origin hashing for `file:` URLs](https://github.com/conda/rattler/blob/b348c8d2f68fd12efecf71008980a86c47dfeaf9/crates/rattler/src/install/installer/mod.rs#L956-L980).

Both can incorrectly exclude packages that are available locally, or include packages that still require validation or downloading.

The new URL-bearing errors should also redact channel credentials before formatting them.

## Performance

I don't think the current PR demonstrates that the new path is cheap enough.

`Arc<str>` avoids allocating the reason repeatedly, but the caller still creates an owned `Url` and hash-table entry for every excluded candidate. In offline mode that is normally almost the entire candidate set. Resolvo then stores another excluded entry for each of them.

`CacheIndex::contains_record` also allocates strings for the name, version, build and rendered cache path for every candidate. Origin-aware caches additionally clone and hash the URL.

A borrowed candidate filter seems like a better API:

```rust
fn exclusion_reason(&self, record: &RepoDataRecord) -> Option<&str>;
```

Resolvo could call that while interning records. That avoids building the exclusion map, copying URLs and doing a separate pass. A default-deny policy with a small allowlist would also better match offline solving, where cached packages are usually the smaller set.

The empty-map overhead for existing solves is probably small, but it is still unmeasured. I would add benchmarks for empty, 90% excluded and 99% excluded cases, plus cache indexing and bulk membership lookup.

## Test coverage

The cache and solver tests currently exercise their parts separately. There is no end-to-end test for:

```text
CacheIndex -> exclusions -> solve -> install
```

That test should assert that the fetch callback is never called.

I would also add focused tests for:

- excluded plus hard-pinned candidates;
- cached `.tar.bz2` with an excluded equivalent `.conda`;
- conflicting hashes across cache layers;
- missing or invalid metadata;
- `file:` URLs with origin caching;
- both cache-only modes with a request counter.

Two related issues already existed before this PR:

- cache metadata is written before a replacement download succeeds;
- `UseCacheOnly` can still perform a request for a cold or corrupt sharded index.

I would not call those PR regressions. However, the new offline path relies on both behaviours, so the end-to-end tests should cover them.

Hofer-Julian:
Good catches, all addressed. The fixes are separate commits on top so they are easy to re-review.

**Pinned + excluded**: Pin and exclusion are now applied independently, so this fails as unsatisfiable when the package is needed. The error names the pin but not the reason: resolvo skips exclusion reasons in the root-level conflict rendering. Fixed in prefix-dev/resolvo#255, the message includes the reason once we bump to the next resolvo release.

**Dedup**: Caller exclusions now feed the same `excluded` flag as `exclude_newer`. The test asserts the same-version `.tar.bz2` instead of the downgrade.

**Cold shard**: Agreed, this shouldn't change for every caller. Brought back the knob as `SourceConfig::missing_shards_are_empty`, default off. The default is the old error again, now typed as `ShardNotCached`. Pixi opts in.

**CacheIndex**: The index now keeps every layer's sha and matches if any layer is compatible, and `file:` URLs are hashed by path like the installer does. Both have tests.

Missing metadata as `None` stays on purpose: entries from before hash recording have no readable hash either, and treating `None` as absent would drop every older cache from offline availability. Presence never promised validity, the entry is validated on use like `get_or_fetch` would. Added a test so it reads as intent.

**Redaction**: Done.

**Performance**: Added `exclusions_bench`. The exclusions never remove anything the solution needs, so every run solves identically and only the bookkeeping differs:

| Case | build map | solve |
|---|---|---|
| `python=3.9`, empty | 39 µs | 5.0 ms |
| `python=3.9`, 99% excluded | 145 µs | 3.5 ms |
| `quetz`, empty | 1.2 ms | 283 ms |
| `quetz`, 99% excluded | 7.6 ms | 302 ms |

The map is one to two orders of magnitude below the solve, so I'd rather keep it than give up `Clone`/`Debug`/serde on `SolverTask` for a callback.

**Tests**: Added the end-to-end test (index the cache, derive exclusions, solve, install everything from the cache with a fetch callback that errors when called). The cold-shard tests run under both cache-only modes with a request counter asserting the network is never touched.

baszalmstra:
Resolvo released with your fix: 0.12.0.

baszalmstra (changes requested):
Most of the earlier issues are fixed now. The benchmark also makes the exclusion-map overhead look acceptable, so I do not think performance is a reason to block this anymore.

There is still one issue that breaks the offline guarantee.

## `CacheIndex` can report a package as locally available even though installation fetches it

[`CacheIndex::contains_key`](https://github.com/conda/rattler/blob/7f679fb6238e5e5224f569a413c0a2c9adf20102/crates/rattler_cache/src/package_cache/mod.rs#L143-L152) now returns true when any layer has a matching hash. That is the right intent, but it does not match `PackageCache::get_or_fetch` for a fresh cache object.

`index()` scans the directories but does not populate the per-layer `packages` `DashMap`. [`try_validate`](https://github.com/conda/rattler/blob/7f679fb6238e5e5224f569a413c0a2c9adf20102/crates/rattler_cache/src/package_cache/mod.rs#L273-L281) therefore returns `PackageNotFound` for the existing entries. [`get_or_fetch`](https://github.com/conda/rattler/blob/7f679fb6238e5e5224f569a413c0a2c9adf20102/crates/rattler_cache/src/package_cache/mod.rs#L530-L566) then falls through to `validate_or_fetch` on the first writable layer.

With the setup from [the new layered-cache test](https://github.com/conda/rattler/blob/7f679fb6238e5e5224f569a413c0a2c9adf20102/crates/rattler_cache/src/package_cache/mod.rs#L1919-L1974), layer 1 contains SHA A and layer 2 contains the valid SHA B. The index accepts B, but a fresh cache skips layer 2 and invokes the fetch callback for layer 1.

Could we extend that test to construct a fresh layered cache, build the index, and then call `get_or_fetch` with a failing fetch callback? The callback should not run. We either need lookup to lazily validate on-disk entries in every layer, or the index needs to follow the actual lookup semantics.

## Non-blocking

The `excluded_candidates` docs say that the exclusion reason is reported when it causes the solve to fail. The [pinned + excluded test](https://github.com/conda/rattler/blob/7f679fb6238e5e5224f569a413c0a2c9adf20102/crates/rattler_solve/tests/backends/main.rs#L1063-L1069) documents that this is not true yet. I would either bump Resolvo once the upstream fix is available, or narrow that part of the docs.

Hofer-Julian:
**Layered lookup**: Confirmed, and it ran deeper than the index: a fresh `PackageCache` object could not see on-disk entries in read-only layers at all, `try_validate` only trusted its own in-memory bookkeeping. Lookup now lazily validates on-disk entries in every layer, like `validate_or_fetch` already did for the writable one. The layered test ends with the `get_or_fetch` you asked for: fresh cache object, failing fetch callback, served from the later layer.

**Docs**: Narrowed. The pinned-conflict case doesn't include the reason yet, that needs prefix-dev/resolvo#255. I'll bump resolvo here once that is released.

baszalmstra (approved):

PR #2613, opened 2026-07-22
feat: add iOS and Android subdirs and __ios/__android virtual packages

wolfv:
### Description

Adds conda subdirs and virtual packages for **iOS** and **Android**, mapping the [PyPI](https://pypi.org) wheel-tag model onto conda's.

A PyPI wheel tag packs three axes into one string — arch, ABI, and min-OS-version — e.g. `ios_13_0_arm64_iphoneos` ([PEP 730](https://peps.python.org/pep-0730/), iOS) or `android_21_arm64_v8a` ([PEP 738](https://peps.python.org/pep-0738/), Android). conda splits those axes:

- **arch + ABI → subdir**
- **min-iOS-version / Android API level → virtual package** (`__ios` / `__android`)

Putting the version axis in a virtual package means compatibility falls out of the normal solver, exactly like `__glibc` and `__osx` — no new matching logic. pip had to [special-case](https://github.com/pypa/pip) the min-version-in-tag; for us a package built for `__ios >=13.0` just satisfies a solve targeting iOS 15.

#### New subdirs

Following conda convention, `x86_64`/`x86` are spelled `-64`/`-32` (like `linux-64`, `osx-64`); `arm64`/`aarch64`/`armv7a` are spelled out (like `linux-aarch64`).

| PyPI tag | conda subdir | virtual package |
|---|---|---|
| `ios_13_0_arm64_iphoneos` | `ios-arm64` | `__ios >=13.0` |
| `ios_13_0_arm64_iphonesimulator` | `iossimulator-arm64` | `__ios >=13.0` |
| `ios_13_0_x86_64_iphonesimulator` | `iossimulator-64` | `__ios >=13.0` |
| `android_21_arm64_v8a` | `android-aarch64` | `__android >=21` |
| `android_21_armeabi_v7a` | `android-armv7a` | `__android >=21` |
| `android_21_x86_64` | `android-64` | `__android >=21` |
| `android_21_x86` | `android-32` | `__android >=21` |

Each conda subdir is single-arch by construction, so there's no fat/multi-arch case to model (PyPI explicitly rejected a universal2-style iOS wheel), and the device/simulator split is just separate subdirs. Every subdir keeps to a single `<os>-<arch>` dash so tools that split on `-` keep working — the simulator distinction is folded into the os token (`iossimulator`) rather than added as a third segment.

#### New virtual packages

`__ios` and `__android` are added as **stubs**: `current()` returns `None` because rattler never runs on the target device. They are produced when cross-compiling to an `ios-*` / `android-*` subdir (via `detect_for_platform`, falling back to version `0`) or overridden with `CONDA_OVERRIDE_IOS` / `CONDA_OVERRIDE_ANDROID`. They classify as `__unix` but deliberately not as `__osx` / `__linux`, since iOS/Android use different C libraries.

#### Naming calls for the CEP (none load-bearing)

- `aarch64` (Android) vs `arm64` (iOS): matches `linux-aarch64` vs `osx-arm64` respectively — inherits the existing conda inconsistency either way.
- `iossimulator` os token for the simulator subdirs (keeps a single dash; alternatives considered were `iossim` and `ios_sim`).
- whether to carry `iossimulator-64` / `android-32` long-term (Intel Macs / 32-bit Android are fading) — PEP 730/738 kept them.

### How Has This Been Tested?

- `cargo nextest run -p rattler_conda_types -p rattler_virtual_packages` — added `test_ios_android_platform` (subdir round-trip, arch split, unix-not-osx/linux classification) and `test_ios_android_virtual_packages` (cross-compile emits `__ios`/`__android` + `__unix` but not `__osx`/`__linux`; version overrides work).
- `cargo clippy` and `cargo fmt` clean; `py-rattler` bindings compile.

### AI Disclosure

Tools: Claude Code (Opus 4.8)

Prompt:
```plaintext
Implement new subdirs mapping PyPI iOS/Android wheel tags onto conda: the
arch+ABI axis becomes the subdir, the min-version/API-level axis becomes a
virtual package (__ios / __android). Add ios-arm64, ios-arm64-simulator,
ios-x86_64-simulator, android-aarch64, android-armv7a, android-x86_64,
android-x86, and at least stub virtual packages for __ios and __android.
```
(Follow-ups: x86_64/x86 subdirs renamed to conda's `-64`/`-32` convention; iOS simulator subdirs renamed to `iossimulator-<arch>` to keep a single dash.)

### Checklist:


crates/rattler_conda_types/src/platform.rs
```diff
@@ -74,6 +83,8 @@ pub enum Arch {
     Arm64,
     ArmV6l,
     ArmV7l,
+    // armv7a is used for android (the `armeabi-v7a` ABI)
```
baszalmstra: Maybe explain how its different.

crates/rattler_conda_types/src/platform.rs
```diff
@@ -39,6 +39,15 @@ pub enum Platform {
...
     OsxArm64,
 
+    IosArm64,
+    IosSimulatorArm64,
+    IosSimulator64,
+
+    AndroidAarch64,
+    AndroidArmV7a,
+    Android64,
+    Android32,
```
baszalmstra, line 67: Perhaps it makes sense to add some of the design rationale from the PR description to a doc comment in this module or of this type or just as comments above these enums? Either way, it would be good to keep the rationale in the code somewhere. 

Probably also good to describe that Android is also just Linux, but because it uses a very different c library that we cant mix it with standard linux-.. because that might not properly declare the libc constraint.

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -175,6 +175,12 @@ pub enum VirtualPackage {
     /// Available when running on `OSX`
     Osx(Osx),
 
+    /// Available when targeting `iOS`
+    Ios(Ios),
```
baszalmstra, line 182: Maybe we should add `#[non_exhaustive]` on this and the `VirtualPackages` struct to avoid API breakages in the future. What do you think?

crates/rattler_conda_types/src/platform.rs
```diff
@@ -257,6 +270,26 @@ impl Platform {
         matches!(self, Platform::Osx64 | Platform::OsxArm64)
     }
 
+    /// Returns true if the platform is an iOS based platform (device or
+    /// simulator).
+    pub const fn is_ios(self) -> bool {
```
baszalmstra, line 349: One downside of this now is that in the code we sometimes check that the platform is_osx and we special case behavior on it. Link in linking we sign the binaries again. This now doesnt happen for ios where this should also happen. Maybe do a scan through the code where we special case osx and see if we should do the same for ios.

baszalmstra (changes requested):
`VirtualPackageOverrides` in the python bindings has not been extended. 

This should also add support for `Platform::current`.

```rust
#[cfg(target_os = "ios")]
{
    // Mac Catalyst (`*-apple-ios-macabi`) also reports `target_os = "ios"`,
    // but produces binaries that run on macOS. No conda subdir exists for it.
    #[cfg(target_abi = "macabi")]
    return Platform::Unknown;

    #[cfg(all(target_arch = "aarch64", target_abi = "sim"))]
    return Platform::IosSimulatorArm64;

    #[cfg(all(target_arch = "aarch64", not(any(target_abi = "sim", target_abi = "macabi"))))]
    return Platform::IosArm64;

    // The only x86_64 iOS target (`x86_64-apple-ios`) is the simulator.
    #[cfg(all(target_arch = "x86_64", not(target_abi = "macabi")))]
    return Platform::IosSimulator64;

    #[cfg(not(any(target_arch = "aarch64", target_arch = "x86_64")))]
    compile_error!("unsupported ios architecture");
}

#[cfg(target_os = "android")]
{
    #[cfg(target_arch = "aarch64")]
    return Platform::AndroidAarch64;

    // armv7-linux-androideabi and thumbv7neon-linux-androideabi → armeabi-v7a
    #[cfg(target_arch = "arm")]
    return Platform::AndroidArmV7a;

    #[cfg(target_arch = "x86")]
    return Platform::Android32;

    #[cfg(target_arch = "x86_64")]
    return Platform::Android64;

    #[cfg(not(any(/* the four above */)))]
    compile_error!("unsupported android architecture"); // e.g. riscv64-linux-android
}
```

wolfv:
Both done. The pyo3 layer already exposed ios/android but the pure-Python VirtualPackageOverrides wrapper didn't — it now takes them in the constructor and as properties. (Side note: the Python wrapper also doesn't expose win/linux overrides, but that pre-dates this PR — happy to add them in a follow-up.)
`Platform::current()` now uses your suggested cfg tree, including Mac Catalyst → `Platform::Unknown`. 

Verified by compiling rattler_conda_types for aarch64-apple-ios, aarch64-apple-ios-sim, x86_64-apple-ios, aarch64-linux-android, and armv7-linux-androideabi.

crates/rattler_conda_types/src/platform.rs
wolfv: Expanded the comment: armv7l is the uname -m value used for linux-armv7l, while armeabi-v7a is Android's name for its 32-bit ARMv7-A ABI — same ISA generation, but softfp calling convention and Bionic instead of glibc, so binaries aren't interchangeable between the two.

crates/rattler_conda_types/src/platform.rs
wolfv: Good idea. Added the rationale as comments on the enum: the iOS variants carry the wheel-tag mapping (PEP 730/738 pack arch+ABI+min-OS-version into one tag; conda splits them into subdir + __ios/__android virtual package, plus the single-dash iossimulator reasoning), and the Android variants get a dedicated paragraph explaining that Android runs a Linux kernel but links Bionic, so it deliberately isn't is_linux() — linux-* packages declare their libc requirement via __glibc, which Bionic can't satisfy, and mixing subdirs would leave that constraint silently violated.

crates/rattler_virtual_packages/src/lib.rs
wolfv: Done!

crates/rattler_conda_types/src/platform.rs
wolfv: Good catch — I scanned all is_osx() call sites:

- rattler/src/install/link.rs (codesigning): now also runs for is_ios(). iOS binaries are Mach-O too, and on arm64 an invalidated signature doesn't just warn — the binary is killed on load, so device and simulator prefixes need the ad-hoc re-sign after prefix replacement just as much as macOS.
- link.rs shebang length limits: Android now uses the Linux limit (127, same kernel), iOS the macOS limit (512, same XNU kernel).
- rattler_menuinst: left as-is — desktop menu shortcuts don't apply to ios-*/android-* target prefixes. 
- Archspec::from_platform: already handles iOS/Android in this PR.

wolfv:
Thanks for the review!!

baszalmstra (approved):

PR #2614, opened 2026-07-22
feat: add `auth token` subcommand

flferretti:
### Description

This PR adds a `pixi auth token <HOST>` / `rattler auth token <HOST>` subcommand to have an easy way to recover the raw token. There's no way to pipe a stored prefix.dev/anaconda.org credential into another tool (e.g. `curl -H "Authorization: Bearer $(... auth token repo.prefix.dev)"`) without manually digging through the keyring or auth file.

`auth token <HOST>`:
- Resolves the host the same way outbound requests do (exact match, then wildcard `*.<domain>` walking up subdomains) via `AuthenticationStorage::get_by_url_refreshed`/`get_by_url`.
- Refreshes an expired OAuth access token before printing it (same refresh path already used for real requests), unless `--offline` is set, in which case it falls back to an unrefreshed read rather than erroring.
- Prints the bearer/conda/OAuth-access token on success.
- Errors with `NotLoggedIn` if nothing is stored for the host, or `NotAToken` if what's stored isn't a bearer-style secret (basic auth / S3 credentials).

### How Has This Been Tested?

Added unit tests in `crates/rattler/src/cli/auth.rs`:
- `token_prints_bearer_token_for_matching_host`: a token stored under `*.prefix.dev` is returned for a `repo.prefix.dev` lookup.
- `token_unknown_host_returns_not_logged_in`
- `token_basic_auth_credentials_are_not_a_token`

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/rattler/src/cli/auth.rs
+++ b/crates/rattler/src/cli/auth.rs
@@ -124,6 +124,12 @@ struct LogoutArgs {
 }
 
 #[derive(Parser, Debug)]
+struct TokenArgs {
+    /// The host to print the stored token for (e.g. prefix.dev)
+    host: String,
+}
+
+#[derive(Parser, Debug)]
 struct StatusArgs {
     /// Show endpoint URLs, client ID, and other IdP-introspection fields
     /// that are only useful for debugging.
@@ -143,6 +149,8 @@ enum Subcommand {
     Login(LoginArgs),
     /// Remove authentication information for a given host
     Logout(LogoutArgs),
+    /// Print the stored authentication token for a given host
+    Token(TokenArgs),
     /// Show stored authentication entries and non-secret token metadata
     Status(StatusArgs),
 }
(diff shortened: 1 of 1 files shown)


baszalmstra:
Im not sure that we should do this. I dont think we should use pixi/rattler as a means to store and retrieve authentication tokens.

wolfv:
I think this is inspired by gh, right? I see that it's useful for e.g. talking to the graphql API

flferretti:
> Im not sure that we should do this. I dont think we should use pixi/rattler as a means to store and retrieve authentication tokens.

That's fair, I understand. Would you recommend any other alternative?

baszalmstra:
> Would you recommend any other alternative?

Can you explain what your usecase is?

flferretti:
> I think this is inspired by gh, right? I see that it's useful for e.g. talking to the graphql API

Yes, exactly, `gh` or `hf`. My own main use case is passing the token into a Google Cloud startup script. The VM needs to authenticate against a private channel non-interactively at boot, and this would give a clean way to pull the credential into a shell variable there rather than reaching into the keychain/auth file by hand. It could also be handy for talking to GraphQL API directly

baszalmstra:
Ok I can see why that is useful. Especially with OAuth. Merging! Thanks for taking the time to make a PR!

flferretti:
> Ok I can see why that is useful. Especially with OAuth. Merging! Thanks for taking the time to make a PR!

Thank you @baszalmstra and @wolfv!

PR #2615, opened 2026-07-23
feat: add Azure Blob Storage channel support

### LINKED ISSUE #2285: Add Azure Blob Middleware to rattler_networking
### Checklist

### What is the idea?

Add support for Azure Blob Storage as a channel to pull/push.

### Why is this needed?

Anyone/companies that happen to use Azure as their cloud infrastructure provider. They would be able to host channels within the ecosystem.

### What should happen?

Implement Azure Blob Storage as a middleware a la S3.

### Additional Context

_No response_

tenzinplatter:
### Description

Adds **Azure Blob Storage** as a first-class conda channel backend across rattler,
addressed through `az://` URLs — the Azure counterpart to the existing S3 (`s3://`)
and GCS support. Channels hosted in a blob container can be **fetched, solved,
indexed, and uploaded to** with no separate endpoint/config block.

Closes #2285.

#### URL model

A channel is addressed by its full blob host:

```
az://<account>.blob.core.windows.net/<container>/<prefix>
```

The account, endpoint, container and root prefix are all derived from the URL, so
the channel is spelled the same way it appears on the wire. Sovereign clouds and
the Azurite emulator work automatically because the endpoint lives in the host —
no allow-list of accounts or endpoints is baked in.

#### What's included

- **`rattler_azure`** (new crate): `az://` URL parsing into typed
  `AzureCoordinates { account, container }`, `AzureCredentials`
  (account key / SAS token), and optional `az login` → short-lived
  user-delegation SAS minting for write paths (behind the `clap` feature).
- **`rattler_networking`**: `AzureMiddleware` — rewrites `az://` → `https://`
  (a plain scheme swap) and signs via reqsign's `DefaultCredentialProvider`
  chain (env vars → workload/managed identity → `az login`). Public/anonymous
  containers work with zero credentials; a *configured-but-broken* credential is
  a hard error rather than a silent anonymous downgrade.
- **`rattler_repodata_gateway`**: accepts the `az://` channel scheme for
  (sharded) repodata fetch.
- **`rattler_index`**: `azblob` indexing backend (opendal), CLI subcommand.
- **`rattler_upload`**: `az://` upload path with concurrent block upload.
- **`rattler-bin`**: registers `AzureMiddleware` behind an `azure` feature.
- **`py-rattler`**: `AzureMiddleware` binding, included in the default client stack.

#### Authentication

| Source | How |
| --- | --- |
| Account key | `--account-key` / `AZURE_STORAGE_KEY` |
| SAS token | `--sas-token` / `AZURE_STORAGE_SAS_TOKEN` |
| `az login` | `--azure-cli` mints a short-lived user-delegation SAS (write paths) |
| Ambient (read) | reqsign chain: env → managed/workload identity → `az login` |

#### Security / trust model

The URL host is trusted verbatim and becomes the request target, so `az://`
**userinfo is rejected**. This is a pure safety gate that removes no
functionality: Azure Blob has no `user:pass@` auth mode — credentials come from
Shared Key headers, SAS query strings, or AAD bearer tokens, and the storage
account is the host's first label (`account.blob.core.windows.net`), never
userinfo. So userinfo in an `az://` URL can do exactly one thing: hide the real
host — `az://trusted.blob.core.windows.net@attacker.example/…` targets
`attacker.example`. Rejecting it closes that host-spoofing vector while leaving
every legitimate URL untouched. (Scoped to `az://`; `https://` channels that
legitimately use Basic-auth userinfo are unaffected.)

On Windows the Azure CLI is resolved via `which` and invoked directly rather than
through `cmd /C`, avoiding a metacharacter-injection surface.

#### Known limitations (documented in-code)

- opendal 0.57's azblob backend honours `if_not_exists` only on the single-shot
  Put Blob path (≤ 10 MiB), not the multi-block path used for larger packages,
  so the upload overwrite guard does an explicit pre-write `stat` to cover all
  sizes (with a small, documented stat→write TOCTOU window).
- A user-delegation SAS minted against a flat container is *container-scoped*,
  not prefix-scoped; a short TTL bounds the blast radius.

### How Has This Been Tested?

- Unit tests across `rattler_azure`, `rattler_networking` (URL rewrite, userinfo
  rejection, credential-source detection, unsigned fallback, SAS pass-through),
  and `rattler_upload`.
- Azurite-backed integration tests for the index path.
- Live-Azure gated tests (`az login` AAD fetch) against a real account.
- `py-rattler`: unit test asserting `AzureMiddleware` is in the default client stack.
- Full workspace `cargo build`, `cargo clippy -D warnings`, `cargo fmt`, and
  `cargo nextest` all green.
- **Running in production** on our internal build farm, where it indexes and
  serves conda packages from an Azure Blob-hosted channel.

> **Platform caveat:** all testing has been on **Linux** only — I have no access
> to Windows or macOS machines. In particular the Windows Azure CLI handling (the
> `which`-based `az`/`az.cmd` resolution) is written but has not been exercised on
> a real Windows host; a review pass or CI on those platforms would be welcome.

### AI Disclosure

Tools: Claude Code (Claude Opus 4.8)

### Checklist:


baszalmstra:
Thanks for working on this. Adding Azure support across fetching, indexing, and uploading is valuable, and the shared `rattler_azure` crate seems like a sensible direction.

I do think a few issues should be addressed before merging:

1. **Credentials can be sent to arbitrary hosts.** Both the URL parsing and `AzureMiddleware` trust any dotted hostname. For example, `az://victimaccount.attacker.example/container` could mint credentials for `victimaccount` and send them to `attacker.example`. Ambient or generated credentials should only be used for known Azure Blob suffixes. Custom endpoints should require explicit trusted configuration.

2. **The default `--azure-cli` upload path appears broken.** The generated SAS has `cw` permissions, but the upload first calls `stat`, which requires read permission. Either the token needs `rcw`, or the preflight read should be replaced with a conditional write.

3. **Credential detection duplicates reqsign’s logic without matching it.** `azure_credential_source_present` skips standard VM-managed identity, misses some supported environment variables, and checks variables reqsign does not use. There is also a mismatch between `AZURE_STORAGE_KEY` for writes and `AZURE_STORAGE_ACCOUNT_KEY` for reads. An explicit anonymous/default-credentials mode would be easier to understand and maintain.

4. **Concurrent indexing can lose updates.** Azure disables `PreconditionChecks`, making the existing read-modify-write flow last-writer-wins. Two indexers can therefore silently remove each other’s packages from repodata. Azure ETags, leases, or another locking mechanism should be used here.

5. **The canonical `az://` URL is lost too early.** `parse_channel_url` returns an `https://` URL, so `[index-config."az://..."]` entries never match. A small `AzureChannelUrl` type retaining both the channel URL and the wire URL could centralize host validation and remove duplicated parsing rules.

The tests are a useful start, but they do not cover the most important failure modes. The PR mentions Azurite-backed index tests, but I could not find them. Upload tests use the memory backend and a small file, so they do not exercise Azure authentication, multipart uploads, the `>10 MiB` overwrite path, or concurrent indexing. Standard Azurite compatibility also looks incomplete because the middleware forces HTTPS while Azurite normally uses HTTP path-style URLs.

A few smaller points:

* The synthetic `max-age=60` changes cache semantics for every origin, not just Azure.
* `AzureCredentials` redacts `Debug`, but `AzureAuthSource` and `AzureCredentialsOpts` can still expose secrets.
* The Azure feature cannot be fully disabled because `rattler-bin` enables `rattler_upload`’s default features.
* Multi-upload continues all uploads after an error instead of failing fast.

Overall, I think the architecture is heading in the right direction. Tightening the URL and credential trust model, adding concurrency protection, and testing against a real Azure-compatible backend would make this much safer and easier to maintain.

tenzinplatter:
Thanks for the thorough review — all five points hold up, and working through them led to a design that's simpler than what's on the branch. Before I start on the rework I'd like to check the shape with you.

### 2. Broken `--azure-cli` upload path, confirmed and agreed

Reproduced: the minted SAS has `cw` but the overwrite guard `stat`s first, which needs `r` — every non-forced upload 403s. Fix is minting `rcw` (container-scoped and short-TTL anyway; dropping the `stat` would reopen the >10 MiB overwrite hole you noted).

### 1 + 3. Credential trust and detection — proposal: anonymous-by-default

These two collapse into one mechanism. Instead of an allow-list plus detection heuristics:

- **Every `az://` fetch is unsigned by default.** Public containers work zero-config, and there's no IMDS probe to time out — which was the entire reason `azure_credential_source_present()` existed. That function (and the `azureProfile.json` sniffing) gets deleted, so there's no reqsign mirror left to drift.
- **Credentials attach only via an explicit, user-scoped grant**, mirroring `[s3-options]`:

  ```toml
  [azure-options."mycompany.blob.core.windows.net"]
  auth = true                 # run reqsign's DefaultCredentialProvider for this host

  [azure-options."127.0.0.1:10000"]   # custom endpoint / Azurite: entry doubles as trust grant
  auth = true
  scheme = "http"             # az:// rewrites to this scheme (default https)
  path-style = true           # account = first path segment (default host-style)
  ```

- A configured-but-broken credential is then *always* a hard error — auth was explicitly requested, so there's no silent-anonymous downgrade path left at all.
- Your exfiltration scenario dies structurally rather than by allow-list: `az://victim.attacker.example/...` gets an unsigned request no matter what ambient credentials exist, because no grant covers that host. Custom endpoints need an entry, which is itself the trust declaration. The load-bearing condition is that the grant table stays *user-scoped* — if a consumer lets project manifests contribute entries, a checked-in manifest can self-grant auth for a host it also controls and the whole thing collapses. Worth being explicit about that, because the stakes are higher than for a signature-based scheme: reqsign's default chain can resolve to an Entra ID token that goes out as a plain `Authorization: Bearer …`, replayable against any account that identity can reach rather than scoped to the single request it was minted for. I assume there is some similar attack vector for s3 but I haven't checked.
- The one UX wrinkle: Azure answers anonymous reads of private containers with 404 (it hides existence), so "forgot to grant auth" looks like not-found. We'd attach a hint to that failure pointing at `[azure-options]`.

Does anonymous-by-default sit right with you as the default polarity? The alternative (auth-by-default gated to an official-suffix allow-list) also closes the exfiltration vector — official-suffix hosts are Microsoft-operated, so a credential leaked toward an attacker's *account* never actually reaches the attacker's *server* — but it has to keep the credential-source detection (or pay the IMDS probe timeout on every credential-less fetch), which is the drift problem from point 3, and it bakes in a suffix list that sovereign-cloud/custom-endpoint users then need an escape hatch from. Anonymous-by-default needs neither: no detection, no hardcoded suffix list at all — the grant table is the entire trust model.

### Azurite and the test claims

You're right that the description overclaimed — Azurite-backed tests existed earlier in the branch and were dropped when the URL model moved to full-host `az://` URLs, which Azurite's path-style addressing didn't fit. The `scheme`/`path-style` options above restore Azurite support properly, and the integration tests come back with it (index round-trip + fetch, run against Azurite in CI). The >10 MiB multi-block overwrite path gets a real test there too. PR description will be corrected either way.

### 4. Concurrent indexing — your preference?

Agreed the read-modify-write is last-writer-wins today. ETag CAS (`If-Match`) seems clearly better than leases here — leases need an acquire/renew/release protocol, deadlock the channel if an indexer crashes holding one (or require a renewal loop for runs longer than the 60s maximum), and opendal can't pass lease headers anyway; CAS is stateless, crash-safe, and semantically identical to the existing S3 `PreconditionChecks` path. The blocker is opendal: azblob doesn't expose `if_match` on writes. I've scoped the upstream contribution and it's relatively small — the capability flag plus the `If-Match` header on Put Blob and on the Put Block List commit (the s3 service is an exact template; core plumbing already exists), and it incidentally fixes azblob dropping `if_not_exists` on the multi-block path, which is the same bug this PR works around with the pre-write `stat`. I have the patch drafted and would be willing to drive it upstream if that fits the design you want here.

Question: are you OK with this PR landing with the Azure path at `PreconditionChecks::Disabled` plus a documented pointer at the upstream work, or would you rather see an interim raw-REST conditional write for the repodata files until the opendal release catches up?

### 5. Canonical URL — adopting your suggestion

An `AzureChannelUrl { canonical: az://…, wire: http(s)://… }` type in `rattler_azure`, constructed once at parse time. Config lookup (`[index-config."az://…"]`) matches on the canonical form; requests use the wire form; host validation lives in one place instead of being duplicated into the middleware.

### The smaller points

- `max-age=60` synthesis: gated to `az://` channels only — agreed it shouldn't touch other origins.
- `Debug` on `AzureAuthSource` / `AzureCredentialsOpts`: redacted like `AzureCredentials`.
- `rattler-bin` → `rattler_upload` gets `default-features = false` so `azure` is genuinely disableable.
- Multi-upload: fail-fast.
- `ServerType::Azure` semver: happy to take the `rattler_upload` 0.10 → 0.11 bump.

If the design above looks right I'll edit: ~push the rework as reviewable per-step commits.~ start on the rework.

tenzinplatter:
https://github.com/apache/opendal/pull/7990
^azure changes in opendal

wolfv:
Hi @tenzinplatter sorry for the slow process. This is a pretty darn big PR! :) 

We would love to get it in but ideally you could split this PR into some meaningful smmaler chunks we could review it more easily.

For example, one per crate touched? And maybe a single one for the new `rattler_azure` crate? I know this is asking some extra work but it would really simplify our lives. 

Thank you so much for the huge contribution, by the way!

tenzinplatter:
Of course, would be happy to split it out, I've just been working on improving the design locally so once I get that to a place I am happy with I will split it out and open some new prs. Thanks to you guys for taking on the review of such a huge change aswell, I've been enjoying working on this

wolfv:
Great to hear this! Thanks again for your work here :)

tenzinplatter:
Hi @wolfv @baszalmstra — update on the split. The first three chunks are up:

- #2727 — the new rattler_azure crate on its own: the shared vocabulary (endpoint keys, channel URLs, location matching, credentials) with no wiring into other crates yet
- #2728 — the azure-options config table in rattler_config (stacked on #2727, so its diff includes that commit until it merges)
- #2729 — SAS-signature masking in rattler_redaction (independent, reviewable on its own)

These also carry the design rework I mentioned - the addressing bool is gone in favour of endpoint keys, and grants are per-container.

The remaining chunks are carved out locally and I'll put them up as the stack progresses: the rattler_networking middleware (+ py-rattler bindings), then rattler_index/rattler_upload with the Azurite e2e CI, and finally the gateway/CLI integration.

PR #2617, opened 2026-07-23
fix: Implement Flexible Channel Priority

### LINKED ISSUE #976: Feature: Support flexible priority in solve task
### Checklist

### What is the idea?

Following https://github.com/conda/rattler/pull/598, it would be very useful if rattler/resolvo also supported flexible priority: `With flexible channel priority,
the solver may reach into lower priority channels to fulfill
dependencies, rather than raising an unsatisfiable error`

### Why is this needed?

The main use case I see for this is user requesting for a package A==2.0.0 and getting an error because this newer version of the package is not present in the first priority channel.  

### What should happen?

Frankly, I'm not sure I have enough knowledge of the internals to be very opinionated here. But I'd like to see `A==2.0.0` be satisfied if possible, given the users _acknowledgement_ that they're searching for a specific version

### Additional Context

_No response_

andrejvelichkovski:
### Description

Fixes https://github.com/conda/rattler/issues/976

This PR introduces a new `ChannelPriority` option: `Flexible`. It mimics the implementation that Conda has for this solve priority, namely it sorts the options by channels, and then prefers options coming from higher channels. In case the first channel doesn't have the option, it still has a fallback to lower priority channels.

### How Has This Been Tested?

I have performed few different tests:
1. I have added automated unit tests that should simulate the test behaviour, extending the existing test suite
2. I have imported and manually verified that the solver prefers packages coming from the first defined channels.

### AI Disclosure

This PR doesn't contain AI-generated content. However, AI was used to check and validate the code.

Tools: Claude Code for AI review

```plaintext
The code in this PR is supposed to implement the functionality for using the Flexible Channel Priority strategy when solving the dependencies of a requested Python Environment. It should mimic the existing Conda behaviour and fall back to other channels if the first channel doesn't have all the dependencies. Check for edge cases and places where the code might not be covering the correct functionality.
```

### Checklist:

- [x] I have made corresponding changes to the documentation - No documentation changes for this PR


crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -444,6 +461,12 @@ impl<'a> CondaDependencyProvider<'a> {
                 let candidates = records.entry(package_name).or_default();
                 candidates.candidates.push(solvable_id);
 
+                // Record the priority rank of this channel (first-seen order).
+                let next_rank = channel_order.len() as u32;
+                channel_order
+                    .entry(record.channel.clone())
```
baszalmstra: This clones the channel for every record which seems like a waste. Maybe we can pull a record into a separate struct that includes the precomputed channel rank; that way, comparison becomes much cheaper and uses way less memory.

crates/rattler_solve/src/resolvo/mod.rs
```diff
@@ -591,9 +614,18 @@ impl<'a> CondaDependencyProvider<'a> {
             strategy,
             direct_dependencies,
             dependency_overrides: override_map,
+            channel_priority,
+            channel_order,
         })
     }
 
+    /// Returns the priority rank for a channel, where a lower rank indicates a
```
baszalmstra, line 666: Doc comment is wrong on two counts: records with channel: `None` do get a rank if they appear in repodata (e.g. direct-URL records, `None` is a perfectly good key), so they don't "sort last"; and locked/favored records are never inserted into `channel_order`, so if their channel string doesn't byte-match a repodata channel they silently rank `u32::MAX`. This ties into the pre-existing // TODO: Normalize these channel names at line 367,  flexible inherits the same string-equality fragility as strict ("…/conda-forge" vs "…/conda-forge/" would get two ranks).

baszalmstra (changes requested):
Nice one! This looks correct. I did leave questions with regards to performance. 

Can you add tests for `libsolv_c` for flexible as well? We have an extensive test framework that you can use which works with both solver implementations.

Also missing: a dependency-driven fallback test (the actual motivating scenario of #976) and a flexible run with reversed channel order mirroring `channel_priority_strict`.

crates/rattler_solve/src/resolvo/mod.rs
andrejvelichkovski: This is a great idea! I missed that this will actually be quite compute intensive if we have large repo data files.

If I understand correctly your suggestion, we can create new struct `RankedRecord` which consists of `record: Record` and `channel_rank: u32`. This way we don't need to keep the expensive hashing to look inside the map and we optimize the lookup since we can compare on `channel_rank`. I was thinking this would mean we need to change all the places where we use `Record` to use `RankedRecord` instead.

Alternatively, I was thinking we can keep the HashMap that maps Channel -> Rank but instead of storing strings we can store only references to the string, which should make the look ups more optimized, and if we combine with `contains_key` I think it will avoid actually doing the full clone.

baszalmstra: Yeah whatever works!

crates/rattler_solve/src/resolvo/mod.rs
andrejvelichkovski: Thanks for bringing this up! I have now simplified this in [bdf4bee](https://github.com/conda/rattler/pull/2617/commits/bdf4bee820279e60da4ba9ddceacaa00d5d50256)

andrejvelichkovski:
> Nice one! This looks correct. I did leave questions with regards to performance.
> 
> Can you add tests for `libsolv_c` for flexible as well? We have an extensive test framework that you can use which works with both solver implementations.
> 
> Also missing: a dependency-driven fallback test (the actual motivating scenario of #976) and a flexible run with reversed channel order mirroring `channel_priority_strict`.

Thanks for the detailed review @baszalmstra! I now addressed the comments:

For the performance issue, I decided to keep the HashMap but replaced it to avoid the memory-expensive cloning of the channel url and instead only perform it when we actually see a new channel.

For the tests, I now added each test for both Resolvo and LibsolvC implementations, and I also added the dependency fallback test that you mentioned.

Looking forward to your comments!

crates/rattler_solve/tests/backends/main.rs
```diff
@@ -1686,6 +1686,156 @@ fn channel_priority_disabled() {
...
+        read_conda_forge_sparse_repo_data(),
+        read_pytorch_sparse_repo_data(),
+    ];
+    solve_to_get_channel_of_spec::<rattler_solve::resolvo::Solver>(
+        "pytorch-cpu",
+        "https://conda.anaconda.org/conda-forge/",
+        repodata,
+        ChannelPriority::Flexible,
+    );
+}
```
baszalmstra: We have a macro in the tests that is evaluated for both resolvo and libsolv. This includes the tests that should run for both solvers. We can use that instead of duplicating the tests.

andrejvelichkovski: Ah, missed that we have `solver_backend_tests` macro which does the duplicate invocation for us! I now moved the tests there, thanks!

baszalmstra (approved):

baszalmstra:
Thank you for this PR!

andrejvelichkovski:
> Thank you for this PR!

Thanks for quickly reviewing and giving useful feedback @baszalmstra!

PR #2618, opened 2026-07-23
feat(exec): mount temporary environments via rattler_vfs

Dagmar-Dinjens:
> 🚧 **Draft — stacked PR **
> This is stacked on **#2566** (`feat: add rattler_vfs crate`), 

### Description

Adds a virtual-filesystem backend to `rattler exec`: instead of extracting and linking a temporary environment to disk, the environment is solved once and served lazily through a `rattler_vfs` mount (NFS on macOS/Linux). Files are read on demand from the shared package cache, so creating and running a throwaway env is dramatically cheaper — no extract/link step — while the on-disk footprint stays at a few kilobytes (just the cached lock file + mount point).

- `perf(exec): skip force-unmount when the prefix is not mounted` — the stale-mount cleanup before mounting ran an unconditional `umount` subprocess (and printed a spurious "not currently mounted" error) on every invocation. It's now guarded by a mount-point check (compare the prefix's `st_dev` against its parent; a stat error still triggers cleanup so dead-mount recovery is preserved). Removes a fixed per-run cost.

Benchmarked with `hyperfine` against `pixi exec` across cold / fresh-env / warm cache states (macOS, arm64):

- **cold** and **fresh-env** (env created each run): VFS wins by large margins (ripgrep ~4–9×, python+numpy ~20–40×) since nothing is extracted or linked.
- **warm** (env reused): VFS is slower than a native on-disk env (ripgrep ~57 ms vs ~21 ms; python+numpy ~440 ms vs ~90 ms) — this is userspace-NFS per-RPC round-trip overhead, expected for a lazy mount and the subject of future work (persistent/reused mounts).

The force-unmount guard tightened warm run-to-run timing for lightweight commands (ripgrep warm 62.2 ± 3.5 ms → 56.9 ± 1.7 ms over repeated runs).

### How Has This Been Tested?

- `cargo nextest run -p rattler-bin` — all tests pass.
- `cargo fmt --check` and `cargo clippy` clean on the touched crates.
- Manual end-to-end: `rattler exec -s python=3.12 -s numpy -- python -c 'import numpy'` on macOS; verified the mount is created/torn down, no spurious `umount` output, and (separately) mount options via `nfsstat -m`.
- `hyperfine` benchmark comparing `rattler exec` vs `pixi exec` across cold/fresh-env/warm as described above.

### AI Disclosure

Tools: Claude Code (Claude Opus 4.8)

```plaintext
Implement performance/consistency improvements to `rattler exec`'s VFS mount
path: guard the pre-mount force-unmount so it only runs when the prefix is
actually mounted (avoiding a per-run umount subprocess), then rebuild,
benchmark (hyperfine) and profile (samply) to verify the impact, and prepare
a PR that stacks cleanly on the rattler_vfs and paths.json PRs.
```

### Checklist:


Hofer-Julian:
This seems to be slower on my machine

I tried the following:

```
hyperfine --prepare "rm -rf /tmp/cache" "RATTLER_CACHE_DIR=/tmp/cache pixi exec tokei" "RATTLER_CACHE_DIR=/tmp/cache pixi run rattler exec tokei"
```

This is the result, showing that `rattler exec` is twice as slow as `pixi exec`:

```
Benchmark 1: RATTLER_CACHE_DIR=/tmp/cache pixi exec tokei
  Time (mean ± σ):      1.077 s ±  0.055 s    [User: 0.408 s, System: 0.201 s]
  Range (min … max):    1.031 s …  1.183 s    10 runs

Benchmark 2: RATTLER_CACHE_DIR=/tmp/cache pixi run rattler exec tokei
  Time (mean ± σ):      2.435 s ±  0.081 s    [User: 1.024 s, System: 1.017 s]
  Range (min … max):    2.323 s …  2.593 s    10 runs

Summary
  RATTLER_CACHE_DIR=/tmp/cache pixi exec tokei ran
    2.26 ± 0.14 times faster than RATTLER_CACHE_DIR=/tmp/cache pixi run rattler exec tokei
```

PR #2621, opened 2026-07-26
fix(shell): escape env var values in activation script

Ashish-Kumar-Dash:
### Description

  Only the `Bash` shell escaped environment-variable values before writing them into activation scripts. `Zsh`, `Xonsh`, `Fish` and `PowerShell` interpolated the raw value inside a double-quoted string, and `NuShell` escaped backslashes but not double quotes.

  A value containing a literal `"` therefore terminated the string early, producing a broken activation script, or command injection for a value like `a";cmd;"b`. Package `activate.d` scripts can set arbitrary env-var values, so this is user reachable.

  Verified before the fix (value `a"b`):

  | Shell | Output |
  |-------|--------|
  | bash | `export FOO='a"b'` ✅ |
  | zsh | `export FOO="a"b"` ❌ |
  | fish | `set -gx FOO "a"b"` ❌ |
  | xonsh | `$FOO = "a"b"` ❌ |
  | powershell | `${Env:FOO} = "a"b"` ❌ |
  | nushell | `$env.FOO = "a"b"` ❌ |

  Adds `escape_double_quoted` (backslash-style, for zsh/xonsh/fish/nushell) and `escape_powershell_double_quoted` (backtick-style), applied at the five `set_env_var` call sites.

  **Scope:** variable-expansion semantics (`$`, `$(...)`) are unchanged, matching Bash's existing double-quoted branch. `cmd.exe` (`@SET "VAR=value"`) is also unescaped but left out, its quote handling is a separate, thornier problem.

  ### How Has This Been Tested?

  New regression test `test_set_env_var_escapes_special_chars` asserts the exact escaped output of each shell for value `a"b\c`. `cargo test -p rattler_shell` passes (30 tests); no existing snapshot changed, confirming benign values are untouched. `cargo clippy -p rattler_shell --all-targets -- -D warnings` is clean.
  
 ### AI Disclosure
    
  Tools: Claude 

  ### Checklist:

  ### Notes
  - The commit also backticks cmd.exe/PowerShell in one doc comment (line 595) to clear a pre-existing doc_markdown clippy error.

### DIFF
--- a/crates/rattler_shell/src/shell/mod.rs
+++ b/crates/rattler_shell/src/shell/mod.rs
@@ -472,6 +472,7 @@ pub struct Zsh;
 impl Shell for Zsh {
     fn set_env_var(&self, f: &mut impl Write, env_var: &str, value: &str) -> ShellResult {
         validate_env_var_name(env_var)?;
+        let value = escape_double_quoted(value);
         Ok(writeln!(f, "export {env_var}=\"{value}\"")?)
     }
 
@@ -533,6 +534,7 @@ pub struct Xonsh;
 impl Shell for Xonsh {
     fn set_env_var(&self, f: &mut impl Write, env_var: &str, value: &str) -> ShellResult {
         validate_env_var_name(env_var)?;
+        let value = escape_double_quoted(value);
         Ok(writeln!(f, "${env_var} = \"{value}\"")?)
     }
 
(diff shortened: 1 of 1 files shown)


baszalmstra (approved):

baszalmstra:
thanks!

PR #2627, opened 2026-07-28
feat: Accept sparse repodata in solve

soapy1:
### Description

This change introduces the ability to pass sparse repodata to the `rattler.solve` function. This change moves towards unifying the py-rattler solve apis (`solve` and `solve_with_sparse_repodata`).

### How Has This Been Tested?

* [x] unit tests
* [x] conda-rattler-solve-example: this is intended to be used in conda-rattler-solver, for example https://github.com/conda/conda-rattler-solver/compare/main...soapy1:add-installed-records-to-solve-pool-2?expand=1
* [x] benchmarks https://github.com/soapy1/rattler/pull/1 
* [x] small example: consider the sample script that exercises using sparse repodata and a channel with the `rattler.solve` funciton
```
import asyncio
from rattler import solve, SparseRepoData, Channel

async def main() -> None:
    # Setup sparse repodata object to represent a slice of repodata
    # originating from a local channel that has been sharded
    chn = Channel("http://127.0.0.1:40093/noarch")
    repo = SparseRepoData(chn, "noarch", "./sparse_repodata.json")

    solved_records = await solve(
        # Sources (channels or custom RepoDataSource) to use for solving
        sources=[repo, "conda-forge"],
        # The specs to solve for
        specs=["tzdata", "foo ==1"],
    )
    print("solved required dependencies:")
    for rec in solved_records:
        print(f"  * {rec.name.normalized} v{rec.version}")


if __name__ == "__main__":
    asyncio.run(main())
```

<details>
<summary>with the sample sparse repodata file called `./sparse_repodata.json`</summary>
```
{
  "repodata_version": 2,
  "info": {
    "subdir": "noarch",
    "repodata_revisions": {
      "v3": {}
    },
    "base_url": "http://127.0.0.1:40093/noarch"
  },
  "packages": {},
  "packages.conda": {
    "foo-1-0_a.conda": {
      "name": "foo",
      "version": "1",
      "build": "0_a",
      "build_number": 0,
      "depends": [
        "bar"
      ],
      "constrains": [
        "splat<3"
      ],
      "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
    },
    "bar-1-0_a.conda": {
      "name": "bar",
      "version": "1",
      "build": "0_a",
      "build_number": 0,
      "depends": [],
      "constrains": [
        "splat<3"
      ],
      "sha256": "e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855"
    }
  },
  "v3": {
    "tar.bz2": {},
    "conda": {},
    "whl": {}
  }
}```
</details>

produces the result:
```
$ python solve.py
solved required dependencies:
  * foo v1
  * bar v1
  * tzdata v2026c
```


### AI Disclosure
Tools: Claude

### Checklist:

<!-- Just as a reminder, everyone in all conda org spaces (including PRs)
     must follow the Conda Org Code of Conduct (link below).

     Finally, once again, thanks for your time and effort. If you have any
     feedback in regards to your experience contributing here, please
     let us know!

     Helpful links:
       - Conda Org COC: https://github.com/conda/.github/blob/main/CODE_OF_CONDUCT.md
       - Contributing docs: https://github.com/conda/.github/blob/main/CONTRIBUTING.md -->

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/source.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/source.rs
@@ -8,7 +8,7 @@ use super::{
     GatewayError,
     subdir::{PackageRecords, SubdirClient, extract_unique_deps_split},
 };
-use crate::Reporter;
+use crate::{Reporter, sparse::SparseRepoData};
 
 /// A source of repodata records for a specific subdirectory.
 ///
@@ -39,7 +39,7 @@ pub trait RepoDataSource: Send + Sync {
 /// A source of repodata, either a channel or a custom source.
 ///
 /// This enum allows the [`Gateway::query()`](super::Gateway::query) method
-/// to accept both traditional channels and custom repodata sources.
+/// to accept both traditional channels custom repodata sources and sparse repodata.
 #[derive(Clone)]
 pub enum Source {
     /// A traditional conda channel (expanded to all requested platforms).
@@ -47,6 +47,9 @@ pub enum Source {
 
     /// A custom repodata source (provides records for requested platforms).
     Custom(Arc<dyn RepoDataSource>),
+
+    /// A sparse repodata source (provides records for requested platforms from sparse repodata).
+    SparseRepoData(Arc<SparseRepoData>),
 }
 
 impl From<Channel> for Source {
@@ -58,6 +61,12 @@ impl From<Channel> for Source {
 impl From<Arc<dyn RepoDataSource>> for Source {
     fn from(source: Arc<dyn RepoDataSource>) -> Self {
         Source::Custom(source)
+    }
+}
+
+impl From<Arc<SparseRepoData>> for Source {
+    fn from(source: Arc<SparseRepoData>) -> Self {
+        Source::SparseRepoData(source)
     }
 }
 
--- a/crates/rattler_repodata_gateway/src/gateway/source.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/source.rs
@@ -48,8 +48,9 @@ pub enum Source {
     /// A custom repodata source (provides records for requested platforms).
     Custom(Arc<dyn RepoDataSource>),
 
-    /// A sparse repodata source (provides records for requested platforms from sparse repodata).
-    SparseRepoData(Arc<SparseRepoData>),
+    /// A sparse repodata source (provides records for requested platforms from sparse
+    /// repodata). Each entry represents a different subdir.
+    SparseRepoData(Vec<Arc<SparseRepoData>>),
 }
 
 impl From<Channel> for Source {
@@ -66,7 +67,13 @@ impl From<Arc<dyn RepoDataSource>> for Source {
 
 impl From<Arc<SparseRepoData>> for Source {
     fn from(source: Arc<SparseRepoData>) -> Self {
-        Source::SparseRepoData(source)
+        Source::SparseRepoData(vec![source])
+    }
+}
+
+impl From<Vec<Arc<SparseRepoData>>> for Source {
+    fn from(sources: Vec<Arc<SparseRepoData>>) -> Self {
+        Source::SparseRepoData(sources)
     }
 }
 
--- a/py-rattler/src/repo_data/sparse.rs
+++ b/py-rattler/src/repo_data/sparse.rs
@@ -2,7 +2,9 @@ use std::{path::PathBuf, sync::Arc};
 
 use pyo3::{Bound, PyRef, PyResult, Python, pyclass, pymethods};
 
+use rattler_conda_types::{PackageName, Platform, RepoDataRecord};
 use rattler_repodata_gateway::sparse::{PackageFormatSelection, SparseRepoData};
+use rattler_repodata_gateway::{GatewayError, RepoDataSource};
 
 use crate::channel::PyChannel;
 use crate::match_spec::PyMatchSpec;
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: py-rattler/src/repo_data/sparse.rs [hunk 1]
```diff
@@ -25,14 +27,81 @@ pub struct PySparseRepoData {
...
+        let name = name.clone();
+        tokio::task::spawn_blocking(move || {
+            let lock = inner.read();
+            let Some(sparse) = lock.as_ref() else {
+                return Err(GatewayError::Generic(
+                    "I/O operation on closed file.".to_string(),
+                ));
+            };
+            sparse
+                .load_records(&name, PackageFormatSelection::PreferCondaWithWhl)
```
soapy1: I'm not sure about this package format selection here. In `rattler.solve_with_sparse_repodata` we have a [`package_format_selection`](https://github.com/conda/rattler/blob/main/py-rattler/rattler/solver/solver.py#L154) field that the user can set. 
Should this pr be updated so that `rattler.solve` also supports this parameter?

baszalmstra: I think that would be worth a seperate PR but yeah adding the package_format_selection to the gateway query (and thus the rattler.solve function) makes a lot of sense to me.

baszalmstra: Something along the lines of:

```rust
gateway
    .query(sources, platforms, specs)
    .package_format_selection(PackageFormatSelection::PreferCondaWithWhl)
    .execute()
    .await?;
```

baszalmstra:
This already looks really good, but I'd be curious to see if there are performance gaps between the two implementations. Did you benchmark any of this?

soapy1:
> I'd be curious to see if there are performance gaps between the two implementations. Did you benchmark any of this?

I did some quick benchmarks in the conda-rattler-solver context. Results are here https://github.com/conda/conda-rattler-solver/pull/101#issuecomment-5076012593. There are some caveat's with these measurements (eg. hyperfine is just measuring how long it takes to run a command, the measurement includes much more than just the solve time, results will vary between machines, etc.). These results suggest that using this pr is <1s faster for a large package like `vaex`.

I wrote some quick [pytest benchmarks](https://github.com/soapy1/rattler/pull/1). They show the opposite result, that `rattler.solve_with_sparse_repodata` is marginally faster for small repodata and small packages. 

This difference doesn't inspire much confidence. I'd be interested to run pytest benchmark style test on some more realistic size solve.

soapy1:
Ok, I've added one more test to [my little benchmarks branch](https://github.com/soapy1/rattler/pull/1). It addiitonally compares using `rattler.solve` with a sparse repodata source implemented in python. As [expected](https://github.com/conda/rattler/blob/main/py-rattler/src/repo_data/source.rs#L24), this is quite a bit slower than using `rattler.SparseRepoData` directly. So, I think this change provides an improvement to users trying to do solves with sparse repodata. And there is still opportunity to make `rattler.solve` more performant (but maybe that is better addressed in a more perfomance oriented pr).

@baszalmstra does this investigation resolve your questions around performance gaps?

py-rattler/src/repo_data/sparse.rs
danyeaw: I agree that a follow up PR would be good. Should we change the type to `PreferConda` for now and then users can still pass in `solve_with_sparse_repodata(..., PREFER_CONDA_WITH_WHL)`?

baszalmstra:
@soapy1 Looking at your benchmarks I think a 1.5x slowdown is still significant. I think we should be able to improve this significantly. When using channels directly internally the repodata gateway also just use a sparse repodata. Given that you already provide it a sparse repodata should therefor be very fast.

crates/rattler_repodata_gateway/src/gateway/source.rs
```diff
@@ -39,14 +39,17 @@ pub trait RepoDataSource: Send + Sync {
...
 #[derive(Clone)]
 pub enum Source {
     /// A traditional conda channel (expanded to all requested platforms).
     Channel(Channel),
 
     /// A custom repodata source (provides records for requested platforms).
     Custom(Arc<dyn RepoDataSource>),
+
+    /// A sparse repodata source (provides records for requested platforms from sparse repodata).
+    SparseRepoData(Arc<SparseRepoData>),
```
soapy1: looks like this is a breaking change that can be avoided by adding `#[non_exhaustive]`. I think we probably don't want to add that, because every case should always proabably be handled. But not sure if that's the rust way?

soapy1:
@baszalmstra, I've updated this pr to add sparse repodata as source for the gateway. I've also updated the benchmark notes https://github.com/soapy1/rattler/pull/. The benchmarks are still pretty jittery, like each run has pretty different results. So, looking to make a few improvements there too.

crates/rattler_repodata_gateway/src/gateway/source.rs
baszalmstra: I think we want to represent this as a `Vec` of `Arc<SparseRepoData>` where each entry represents a different subdir. Otherwise we also mess with the channel priority. E.g. one subdir will have higher priority than another.

baszalmstra (changes requested):
This looks good to me! I think one way to improve the performance even further is if we implement a constructor for SparseRepoData that uses in memory bytes instead of going through the filesystem. The Rust implementation already supports that.

crates/rattler_repodata_gateway/src/gateway/source.rs
soapy1: Good idea. I tried to add a test for this channel priority issue in praticular, but couldn't really find a nice way to represent this. Open to suggestions if you have any thoughts.

soapy1:
> This looks good to me! I think one way to improve the performance even further is if we implement a constructor for SparseRepoData that uses in memory bytes instead of going through the filesystem. The Rust implementation already supports that.

Tried this out, but looks like this doesn't really improve things in the benchmarks https://github.com/soapy1/rattler/pull/1#discussion_r3769593755 ([diff](https://github.com/soapy1/rattler/pull/1/changes/f98b54ec3d05418b45a824cd5a2ffe42699bcca3)).

baszalmstra (approved):

baszalmstra:
Looks great! Thanks @soapy1

PR #2630, opened 2026-07-30
feat(repodata_gateway): read virtual package plugin registrations

hunger:
Channels can register virtual package detection plugins in `info.virtual_package_plugins`, mapping the name of the package providing the plugin to the virtual packages it provides. Keying by plugin lets one `cuda-detect` cover both `__cuda` and `__cuda_arch` instead of registering the same detector twice. Parsed from both `repodata.json` and the sharded index; names are deserialized unchecked so one malformed entry cannot make a channel unusable.

The gateway exposes the registrations two ways:

- `Gateway::virtual_package_plugins(channel, platform)` mirrors `Gateway::channel_relations` and needs no specs, which matters because the plugin package names only exist inside the metadata being fetched.
- `RepoDataQueryOutput::virtual_package_plugins` lists them per subdir in resolved channel-priority order, so channels a solve discovers through CEP-42 are covered too.

Registrations are reported exactly as declared: duplicate claims on the same virtual package are left for the caller to resolve, and no plugin is fetched or executed. Registrations stay per subdir rather than collapsed per channel, matching how CEP-42 relations are handled.

`rattler virtual-packages -c <channel>` prints them for manual testing against the new `test-data/channels/virtual-package-plugins` fixture.

The design, including the parts that are not implemented yet -- the plugin protocol, execution, caching and the trust model -- is written up in `crates/rattler_repodata_gateway/docs/virtual-package-plugins.md`.

All of it sits behind the `experimental-virtual-package-plugins` feature; with the feature off the public API and the serialized output are unchanged. Cargo features are additive, so every crate constructing `ChannelInfo` or `ShardedSubdirInfo` must enable it in lockstep, hence the forwarding entries in `rattler_index` and `rattler-bin`.

Also switches the sharded-index test mock to `rmp_serde::to_vec_named`, matching what the indexer writes. Positional encoding misaligns as soon as a skipped field precedes a present one.

This is just an initial attempt at the data structure in repo-data only and just enough extra code to get something out of the `rattler` binary.

The preliminary spec is in crates/rattler_repodata_gateway/docs/virtual-package-plugins.md, based on @wolfv suggestion with changes from claude to update it with the decisions taken in this change.

We can merge this into main: Everything is hidden behind `experimental-virtual-package-plugins` features.

### How Has This Been Tested?

By running `cargo run --bin rattler --features experimental-virtual-package-plugins -- virtual-packages -c ./test-data/channels/virtual-package-plugins` and with unit tests.

### AI Disclosure

- [ ] This PR contains AI-generated content.
  - [ ] I have tested any AI-generated content in my PR.
  - [ ] I take responsibility for any AI-generated content in my PR.

Tools: Claude

### Checklist:

### DIFF
--- a/crates/rattler-bin/Cargo.toml
+++ b/crates/rattler-bin/Cargo.toml
@@ -34,6 +34,10 @@ rustls = [
 s3 = ["rattler_networking/s3", "rattler_upload/s3"]
 gcs = ["rattler_networking/gcs"]
 oauth = ["rattler/oauth"]
+experimental-virtual-package-plugins = [
+  "rattler_repodata_gateway/experimental-virtual-package-plugins",
+  "rattler_index/experimental-virtual-package-plugins",
+]
 
 [dependencies]
 anyhow = { workspace = true }
--- a/crates/rattler-bin/src/commands/virtual_packages.rs
+++ b/crates/rattler-bin/src/commands/virtual_packages.rs
@@ -1,12 +1,30 @@
 use miette::IntoDiagnostic;
 use rattler_conda_types::GenericVirtualPackage;
+#[cfg(feature = "experimental-virtual-package-plugins")]
+use rattler_conda_types::Platform;
 use rattler_virtual_packages::VirtualPackageOverrides;
 
 /// Print detected virtual packages.
 #[derive(Debug, clap::Parser)]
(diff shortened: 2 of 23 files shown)


hunger:
Let's use the stacked PR instead :-)

PR #2633, opened 2026-07-31
feat: add `lfs` field to git source locations in the lock file

Hofer-Julian:
### Description

This allows pixi to record whether a git dependency needs git LFS files directly in the lock file instead of encoding it in the stored git URL.

- Add `lfs: Option<bool>` to `GitSourceLocation`, serialized on v7 git source locations and omitted when absent, so existing lock files render byte-identically
- Skip an absent `lfs` in the `Hash` impl so source identifier hashes of existing lock files stay stable
- The v6 parser always leaves it `None` since the format predates the field

### How Has This Been Tested?

- Added tests
- Use in https://github.com/prefix-dev/pixi/pull/6722

### AI Disclosure

Tools: Claude

```plaintext
[About pixi round-tripping the LFS flag through a query pair on the locked git
URL:] this looks odd. I think we should upstream to rattler-lock instead
```

### Checklist:


hunger (approved):
The code changes look OK to me.

PR #2634, opened 2026-08-01
fix: report download completion before extraction

wolfv:
## Summary

- measure progress as compressed archive bytes are consumed by the extractor, including the compressed total when known
- emit `DownloadReporter::on_download_complete` when the streamed response body reaches EOF instead of after package extraction
- report completion exactly once when the reader is consumed or dropped
- clarify that extraction may still be running when the callback fires

Package downloads are streamed directly into the extractor. Previously, progress was counted when reqwest yielded a complete network frame and completion was delayed until extraction completed. This could make consumers display chunky or frozen download progress during long extractions.

The reporting reader now counts the exact bytes handed to the extractor. This provides determinate combined download/extraction progress against the compressed archive size. Response completion remains separate from cache/extraction completion, so consumers can show an extraction spinner for any work remaining after the archive reaches 100%.

## Tests

- `cargo test --manifest-path crates/rattler_package_streaming/Cargo.toml --lib --features reqwest`
- `cargo clippy --manifest-path crates/rattler_package_streaming/Cargo.toml --all-targets --features reqwest -- -D warnings`
- `cargo fmt --check --manifest-path crates/rattler_package_streaming/Cargo.toml`

### DIFF
--- a/crates/rattler_package_streaming/src/lib.rs
+++ b/crates/rattler_package_streaming/src/lib.rs
@@ -141,9 +141,12 @@ pub struct ExtractResult {
 pub trait DownloadReporter: Send + Sync {
     /// Called when the download starts.
     fn on_download_start(&self);
-    /// Called when the download makes progress.
+    /// Called as downloaded archive bytes are consumed by the extractor.
+    ///
+    /// `total_bytes` is the compressed archive size when it is known.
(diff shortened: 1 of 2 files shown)


baszalmstra (approved):

baszalmstra:
The downside is that now the download is reported as finished but the extraction is still happening in the background and the user has no feedback at all.

PR #2639, opened 2026-08-04
feat: add CEP-6 channel notice support

wolfv:
### Description

Add end-to-end support for [CEP-6 channel notices](https://github.com/conda/ceps/blob/main/cep-0006.md).

- add shared `ChannelNotice`, `ChannelNoticeLevel`, and `ChannelNotices` types
- optionally fetch `notices.json` through the repodata gateway
- return notices on query results and stream them through `Reporter::on_channel_notice`
- cache notices until their earliest `expires_at`, retry failures after a bounded interval, and coalesce concurrent fetches
- treat notice failures as non-fatal, isolate malformed entries, filter expired notices, and cap responses at 1 MiB
- publish `notices.json` from `rattler-index` configuration for filesystem and S3 channels
- publish notices only after indexing succeeds and honor S3 precondition checks
- expose notice configuration and structured results in the Python and JavaScript bindings
- retain `expired_at` as a read alias while emitting CEP-6's standard `expires_at`
- update the source-package build backend to `pixi-build-rust` 0.5 and exempt that backend from the workspace's seven-day package cutoff; Pixi 0.76 requires build API v6, so source-package CI cannot initialize the older backend

Example `rattler-index` configuration:

```toml
[[index-config.notices]]
id = "security-1"
message = "Please update the affected package"
level = "critical"
created_at = "2025-01-01T12:00:00Z"
expires_at = "2025-02-01T12:00:00Z"
```

### How Has This Been Tested?

- `cargo test -p rattler_repodata_gateway --all-features channel_notice`
- `cargo test -p rattler_index --test test_index test_index_writes_channel_metadata`
- `cargo clippy -p rattler_repodata_gateway -p rattler_index --all-features -- -D warnings`
- `cargo check --manifest-path py-rattler/Cargo.toml`
- `pixi run -m py-rattler/pixi.toml lint-python`
- `pixi run -m py-rattler/pixi.toml type-check`
- `pixi run -m py-rattler/pixi.toml test -- tests/unit/test_gateway.py -q`
- `cargo check --manifest-path js-rattler/Cargo.toml`
- `npm test -- --runInBand src/Gateway.test.ts`
- `npx tsc --noEmit`
- `wasm-pack build --mode normal --debug --target nodejs`

Coverage includes local and HTTP channels, query results and reporter callbacks, malformed entries, expiration refresh, request coalescing, response-size limits, indexing output, and Python/JavaScript configuration.

### AI Disclosure

Tools: OpenAI Codex

Prompt:
```plaintext
Can you implement channel notices? I think it should be enabled on the repodata gateway, and if enabled and any notices found returned as part of the reporter, or as part of the result ... or both.

Please also make a follow up commit implementing support to rattler index!

Anything else we should do on this? Can you also do an adversarial review?

OK yes, please do that.
```

### Checklist:


baszalmstra (changes requested):
Most of the properties of what to query are on the query itself, not on the gateway. I think it would be much nicer if we can set the `with_channel_notices` on the gateway.query instead of on the gateway itself.

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -284,8 +303,39 @@ impl RepoDataQuery {
...
+            })
+        })?;
+        // Fetch again with the complete channel set. Explicit channels hit the
+        // cache while this adds channels discovered through CEP-42.
+        let notices = if channel_notices {
+            gateway
+                .get_channel_notices(channels.iter(), reporter.as_deref())
+                .await
+        } else {
+            Vec::new()
```
baszalmstra: Id like to make this part of the subdir/channel discovery of the internal executor instead of something that we do after the executor has run. This otherwise also wont work for CEP-42 channel relations.

baszalmstra (changes requested):

baszalmstra (approved):

PR #2641, opened 2026-08-05
feat(rattler_conda_types): convert to and from `semver::Version`

### LINKED ISSUE #2629: rattler_conda_types: conversion to/from `semver::Version`
At https://github.com/prefix-dev/pixi/pull/6537#discussion_r3540219404 we are converting from https://docs.rs/semver/1.0.28/semver/struct.Version.html via `to_string` and `from_str`. It may be nice to implement a more robust conversion.

baszalmstra:
### Description

Converting a `semver::Version` into a conda `Version` goes through `to_string` and `from_str` today, see the discussion in prefix-dev/pixi#6537. Callers end up handling a parse error that cannot really happen, and the result is whatever semver decides to print instead of something we pick.

This adds `From<semver::Version> for Version` plus the fallible way back. Two things I decided, both documented and covered by tests:

- Pre-releases are separated with a `.` instead of the `-` that semver uses. A `-` breaks conda package filenames, which use it to split name, version and build string. Both compare equal in conda.
- Conda and semver order identifiers differently in a couple of spots, so the ordering is not identical in every case. Those cases are documented and pinned by a test.

`semver` is an optional dependency, so nothing changes if you do not use it.

Fixes #2629

### How Has This Been Tested?

Unit tests in `crates/rattler_conda_types/src/version/semver.rs` cover both directions, the rejected cases, round trips and ordering. The conversion is also asserted to produce the same `Version` as parsing the semver string, so it cannot drift from the parser.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/rattler_conda_types/src/version/semver.rs
+++ b/crates/rattler_conda_types/src/version/semver.rs
@@ -1,0 +1,512 @@
+//! Conversions between a conda [`Version`] and a [`semver::Version`].
+
+use std::fmt::Write;
+
+use semver::{BuildMetadata, Prerelease, Version as SemverVersion};
+use thiserror::Error;
+
+use super::{Component, ComponentVec, SegmentVec, Version, flags::Flags, segment::Segment};
+
+/// The index of the first local segment is stored in 7 bits, so at most this
+/// many segments fit in front of it.
+const MAX_LEADING_SEGMENTS: usize = 127;
+
+/// Maximum number of components a single [`Segment`] can hold.
+const MAX_COMPONENTS_PER_SEGMENT: usize = (1 << 13) - 1;
+
+/// An error that can occur when converting a conda [`Version`] into a
+/// [`semver::Version`].
+#[derive(Debug, Error)]
+pub enum VersionToSemverError {
+    /// The version has an epoch. semver does not know about those.
+    #[error("a semver version cannot express the epoch of `{0}!`")]
+    EpochNotSupported(u64),
+
+    /// The version has more than three release numbers.
+    #[error("a semver version can only hold three release numbers (`major.minor.patch`)")]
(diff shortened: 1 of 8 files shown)


crates/rattler_conda_types/src/version/semver.rs
```diff
@@ -0,0 +1,512 @@
...
+    }
+
+    /// Where conda and semver ordering disagree. There is no way around these.
+    #[test]
+    fn test_ordering_divergence() {
+        // semver orders numeric identifiers below alphanumeric ones, conda
+        // orders numbers above strings.
+        let numeric = Version::from(SemverVersion::parse("1.0.0-alpha.1").unwrap());
+        let alphanumeric = Version::from(SemverVersion::parse("1.0.0-alpha.beta").unwrap());
+        assert!(numeric > alphanumeric);
```
pavelzw: maybe also assert that the numeric `SemverVersion` is < than the alphanumeric `SemverVersion` here?

pavelzw (approved):

crates/rattler_conda_types/src/version/semver.rs
baszalmstra: Good catch, that assert was only showing the conda half so it did not actually demonstrate the divergence. Added the semver side, and did the same for the `post` case so both read the same way:

```rust
let numeric = SemverVersion::parse("1.0.0-alpha.1").unwrap();
let alphanumeric = SemverVersion::parse("1.0.0-alpha.beta").unwrap();
assert!(numeric < alphanumeric);
assert!(Version::from(&numeric) > Version::from(&alphanumeric));
```

Both hold, so the test now pins the disagreement from both sides instead of just asserting what conda does.

PR #2646, opened 2026-08-05
feat: Introduce default versions for virtual packages

pavelzw:
i noticed that `rattler solve --platform <different-platform-than-host>` results in `__linux=0=0` which leads to weird solving errors by default. wdyt about having default versions?

```
❯ rattler solve rattler --platform linux-64 python
Solving for platform: linux-64
Loaded 3247 records in 61.649958ms
Virtual packages:
  - __unix=0=0
  - __linux=0=0
  - __glibc=0=0
  - __archspec=1=x86_64

Error:   × Cannot solve the request because of: rattler * cannot be installed because there are no viable options:
  │ └─ rattler 0.1.0 | 0.1.1 | ... | 0.2.7 would require
  │    └─ __glibc >=2.17,<3.0.a0, for which no candidates were found.
  │
```

prompt

```
rattler solve uses wrong virtual packages by default when doing cross-platform solves.

❯ rattler solve pnpm --platform linux-64
Solving for platform: linux-64
Loaded 4061 records in 215.789625ms
Virtual packages:
  - __unix=0=0
  - __linux=0=0
  - __glibc=0=0
  - __archspec=1=x86_64


it should specify certain versions by default, for example for __linux and __glibc.
pixi does that already, can you fetch those default versions there and put them into rattler somewhere and use them here?
```

i think we could use those versions in `pixi` as well in their defaults, wdyt?

<details>
<summary>🤖 yap</summary>

### Description

This PR introduces a new `defaults` module that provides default virtual package versions matching those used by pixi when detecting packages for platforms other than the current host. Previously, cross-compilation would use hardcoded version 0 for most packages, which doesn't reflect real-world defaults.

**Changes:**
- Added `crates/rattler_virtual_packages/src/defaults.rs` with functions for default versions:
  - `default_glibc_version()` → 2.28 (RHEL 8 / Debian 10)
  - `default_linux_version()` → 4.18 (RHEL 8 kernel)
  - `default_windows_version()` → 10.0
  - `default_mac_os_version()` → 13.0 (for macOS platforms)
- Updated `VirtualPackages::detect_for_platform()` to use these defaults instead of hardcoded zeros
- Updated documentation to reference the new defaults module and clarify version behavior
- Added comprehensive test `test_cross_platform_default_versions()` to verify defaults are applied correctly when cross-compiling

The defaults now align with pixi's behavior, ensuring consistent virtual package resolution across tools in the conda ecosystem.

### How Has This Been Tested?

- Added unit test `test_cross_platform_default_versions()` that verifies:
  - Linux defaults (kernel 4.18, glibc 2.28) are used when targeting Linux from non-Linux platforms
  - macOS default (13.0) is used when targeting macOS from non-macOS platforms
  - Windows default (10.0) is used when targeting Windows from non-Windows platforms
  - Host platform detection takes precedence (test skips checks for current platform)
- Existing tests continue to pass

### Checklist:

https://claude.ai/code/session_01WyyLK4E1qspcGuhDgoqS7K

</details>

### DIFF
--- a/crates/rattler_virtual_packages/src/defaults.rs
+++ b/crates/rattler_virtual_packages/src/defaults.rs
@@ -1,0 +1,39 @@
+//! Default virtual package versions to use when the actual version cannot be
+//! detected from the host system, for example when detecting virtual packages
+//! for a platform other than the current one (see
+//! [`crate::VirtualPackages::detect_for_platform`]).
+//!
+//! The values match the defaults used by
+//! [pixi](https://github.com/prefix-dev/pixi) when no system requirements are
+//! specified.
+
+use rattler_conda_types::{Platform, Version};
+
+/// The default `glibc` version to use when the version cannot be detected.
+///
+/// This is the `glibc` version that ships with RHEL 8 and Debian 10.
+pub fn default_glibc_version() -> Version {
+    "2.28".parse().unwrap()
+}
+
+/// The default Linux kernel version to use when the version cannot be
+/// detected.
+///
+/// This is the kernel version that ships with RHEL 8.
+pub fn default_linux_version() -> Version {
+    "4.18".parse().unwrap()
+}
+
+/// The default Windows version to use when the version cannot be detected.
+pub fn default_windows_version() -> Version {
+    "10.0".parse().unwrap()
+}
+
+/// Returns the default macOS version to use when the version cannot be
+/// detected, or `None` if the given platform is not a macOS platform.
+pub fn default_mac_os_version(platform: Platform) -> Option<Version> {
+    match platform {
+        Platform::Osx64 | Platform::OsxArm64 => Some("13.0".parse().unwrap()),
+        _ => None,
+    }
+}
--- a/crates/rattler_virtual_packages/src/defaults.rs
+++ b/crates/rattler_virtual_packages/src/defaults.rs
@@ -2,10 +2,6 @@
 //! detected from the host system, for example when detecting virtual packages
 //! for a platform other than the current one (see
 //! [`crate::VirtualPackages::detect_for_platform`]).
-//!
-//! The values match the defaults used by
-//! [pixi](https://github.com/prefix-dev/pixi) when no system requirements are
-//! specified.
 
 use rattler_conda_types::{Platform, Version};
 
--- a/crates/rattler_virtual_packages/src/lib.rs
+++ b/crates/rattler_virtual_packages/src/lib.rs
@@ -332,8 +332,15 @@ impl VirtualPackages {
             let win = if platform.is_windows() {
                 // Check override first, fall back to the default version
                 virtual_packages.win.or_else(|| {
+                    let version = defaults::default_windows_version();
+                    log_default_virtual_package(
+                        "__win",
+                        platform,
+                        &version,
+                        Windows::DEFAULT_ENV_NAME,
+                    );
                     Some(Windows {
-                        version: Some(defaults::default_windows_version()),
+                        version: Some(version),
                     })
                 })
             } else {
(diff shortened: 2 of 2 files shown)


crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -328,31 +330,31 @@ impl VirtualPackages {
         } else {
             // When cross-compiling, respect overrides but fall back to defaults
             let win = if platform.is_windows() {
-                // Check override first, fall back to default (no version)
-                virtual_packages
-                    .win
-                    .or_else(|| Some(Windows { version: None }))
```
pavelzw, line 334: why was this `None` before while in all other places we used `0`? maybe we should make it non-optional?

crates/rattler_virtual_packages/src/defaults.rs
```diff
@@ -0,0 +1,39 @@
...
+}
+
+/// Returns the default macOS version to use when the version cannot be
+/// detected, or `None` if the given platform is not a macOS platform.
+pub fn default_mac_os_version(platform: Platform) -> Option<Version> {
+    match platform {
+        Platform::Osx64 | Platform::OsxArm64 => Some("13.0".parse().unwrap()),
+        _ => None,
+    }
+}
```
swenger, line 35: Does this match what pixi uses for non-host platforms?

swenger: Ah yes, at least the comment in crates/rattler_virtual_packages/src/lib.rs says so :)

crates/rattler_virtual_packages/src/defaults.rs
```diff
@@ -0,0 +1,39 @@
+//! Default virtual package versions to use when the actual version cannot be
+//! detected from the host system, for example when detecting virtual packages
+//! for a platform other than the current one (see
+//! [`crate::VirtualPackages::detect_for_platform`]).
+//!
+//! The values match the defaults used by
+//! [pixi](https://github.com/prefix-dev/pixi) when no system requirements are
+//! specified.
```
pavelzw:
```suggestion
```

we should then also use this file in pixi

crates/rattler_virtual_packages/src/defaults.rs
pavelzw, line 1: we should use those in a follow-up PR in pixi

crates/rattler_virtual_packages/src/lib.rs
```diff
@@ -307,14 +308,15 @@ impl VirtualPackages {
...
+    /// defaults are used if no override is provided (see the [`defaults`] module, the versions
+    /// match the defaults used by pixi):
     ///
-    /// - **Windows** (`__win`): No version specified
-    /// - **Linux** (`__linux`): Version 0
-    /// - **OSX** (`__osx`): Version 0
+    /// - **Windows** (`__win`): Version 10.0
+    /// - **Linux** (`__linux`): Version 4.18
+    /// - **OSX** (`__osx`): Version 13.0
     /// - **iOS** (`__ios`): Version 0 (minimum supported iOS version)
```
baszalmstra: Maybe we can just point to the defaults module? TO make sure this doesnt go out of date.

pavelzw: [`7333876` (this PR)](https://github.com/conda/rattler/pull/2646/commits/7333876672e2e2c6aa2f8b714a581e9e767c3da6)

baszalmstra (approved):

crates/rattler_virtual_packages/src/lib.rs
pavelzw: @baszalmstra do you have opinions on this?

baszalmstra: I dont remember no. I think it might be because the version was added later.

PR #2677, opened 2026-08-12
feat(py): expose pip injection for solve APIs

### LINKED ISSUE #2676: Expose Python-pip patching in sparse repodata solves
### What happened?

`solve_with_sparse_repodata` cannot apply the record patch used by conda to add `pip` as a dependency of Python 2 and 3 packages.

The underlying sparse repodata implementation already accepts a record patch function and documents adding `pip` to Python as its primary use case. The Python binding currently passes `None` unconditionally.

Expose an `add_pip_as_python_dependency: bool = False` argument on `solve_with_sparse_repodata`. When enabled, patch Python 2 and 3 records before recursive dependency discovery so that `pip` is loaded and solved as a dependency.

The default must remain `False` so existing py-rattler callers are unchanged.

This enables conda/conda-rattler-solver#122 without rewriting full repodata files or applying the policy as a global root requirement. A global conditional root is not equivalent because it also applies to retained installed Python records.

### What did you expect to happen?

- The option defaults to disabled.
- When enabled, Python 2 and 3 records gain a `pip` dependency.
- Python 4 records remain unchanged.
- `pip` is discovered recursively and included in the solution.
- Locked installed records are not patched retroactively.

### Additional Context

The existing hook is `SparseRepoData::load_records_recursive(..., patch_function, ...)` in `rattler_repodata_gateway`.

jezdez:
### Description

Expose an `add_pip_as_python_dependency` option on `solve` and `solve_with_sparse_repodata`. The option applies conda's existing pip-injection rule to Python versions matching `2.*` or `3.*` before dependency discovery. It defaults to `False`.

Gateway solves use a query-local record patch after records are read from cache and before recursive dependency discovery. Replacements are not written back to the cache. Sparse solves continue to use the existing record patch hook. Locked and pinned packages remain unchanged.

This lets conda preserve its current behavior across solver implementations without rewriting repodata or changing installed prefix records.

Fixes #2676

### How Has This Been Tested?

Unit coverage exercises both solve paths, enabled and disabled behavior, the Python version boundary, recursive pip discovery, query-local cache behavior, and locked package behavior.

### AI Disclosure

Tools: Codex

```plaintext
Expose add_pip_as_python_dependency on both solver APIs. Implement gateway behavior with a query-local record patch that does not mutate cached records. Cover enabled and disabled behavior, the Python version boundary, recursive discovery, cache isolation, and locked record behavior.
```

### Checklist:

### DIFF
--- a/py-rattler/CHANGELOG.md
+++ b/py-rattler/CHANGELOG.md
@@ -7,10 +7,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Added
-
-- Add an `add_pip_as_python_dependency` option to `solve_with_sparse_repodata`.
-
 ## [0.25.0] - 2026-06-09
 
 ### Changed
--- a/py-rattler/CHANGELOG.md
+++ b/py-rattler/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add an `add_pip_as_python_dependency` option to `solve_with_sparse_repodata`.
+
 ## [0.25.0] - 2026-06-09
 
 ### Changed
--- a/crates/rattler_repodata_gateway/src/gateway/mod.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/mod.rs
@@ -1425,6 +1425,120 @@ mod test {
         )
     }
 
+    #[tokio::test]
+    async fn test_record_patch_is_query_local() {
+        fn names(output: &crate::RepoDataQueryOutput) -> std::collections::BTreeSet<String> {
+            output
+                .iter()
+                .flat_map(RepoData::iter)
+                .map(|record| record.package_record.name.as_normalized().to_string())
+                .collect()
+        }
+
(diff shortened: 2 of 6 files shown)


py-rattler/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Add an `add_pip_as_python_dependency` option to `solve_with_sparse_repodata`.
+
```
baszalmstra: Please remove, this is generated on release.

baszalmstra (changes requested):
I don't particularly like this addition because I feel it is way to specific and quite hacky. But I also don't really see another way to achieve this.

One thing missing is that this now only works with `solve_with_repodata` and now with the `solve` function. I think we need to integrate this into the gateway as well to make sure this problem wont surface again when we deprecate `solve_with_repodata`. I think adding a function to the gateway query to add a patch method accepting a closure that returns an `Option<RepodataRecord>` to ensure we dont cache patched repodata would suffice.

py-rattler/CHANGELOG.md
jezdez: Removed.

baszalmstra (approved):

PR #2687, opened 2026-08-17
feat(js): support custom fetch and record queries in the gateway

Hofer-Julian:
### Description

Renovate is already using rattler-js, but not for repodata fetching. That's because our API wasn't powerful enough for Renovate's needs.
Renovate needs two things the bindings don't offer:
- the full version list of a package (`simpleSolve` returns a single solved set), and
- a way to send the gateway's HTTP through Renovate's own stack for authentication, proxies, caching, and test mocking.

### How Has This Been Tested?

- Offline jest tests drive a gateway through a custom fetch, plus a live query against prefix.dev (let me know if you want to rather not have online tests)
- Consumed from my local Renovate work

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/js-rattler/crate/gateway.rs
+++ b/js-rattler/crate/gateway.rs
@@ -1,6 +1,9 @@
 use std::{collections::HashMap, path::PathBuf, str::FromStr};
 
-use rattler_conda_types::{Channel, ChannelNoticeLevel, Platform};
+use rattler_conda_types::{
+    Channel, ChannelNoticeLevel, MatchSpec, ParseMatchSpecOptions, Platform, RepoDataRecord,
+    RepodataRevision,
+};
 use rattler_repodata_gateway::{
     ChannelConfig, Gateway, GatewayWarning, SourceConfig, fetch::CacheAction,
 };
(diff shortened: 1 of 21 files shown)


js-rattler/crate/gateway.rs
```diff
@@ -172,4 +180,54 @@ impl JsGateway {
             .map(|name| name.as_source().to_string())
             .collect())
     }
+
+    /// Queries the given channels and platforms for records matching the
+    /// given match specs. Returns the matching records as plain objects in
+    /// the same shape as they appear in `repodata.json`, extended with the
+    /// `fn`, `url` and `channel` fields.
+    #[wasm_bindgen(js_name = "queryRecords")]
+    pub async fn query_records(
```
baszalmstra: Can we name this the same as the Rust and the rust bindings. E.g. `query`?

js-rattler/crate/gateway.rs
```diff
@@ -172,4 +180,54 @@ impl JsGateway {
...
+                    &s,
+                    ParseMatchSpecOptions::lenient().with_repodata_revision(RepodataRevision::V3),
+                )
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let output = self
+            .inner
+            .query(channels, platforms, specs)
+            .recursive(false)
```
baszalmstra: We should definitely add this as a parameter.

js-rattler/crate/gateway.rs
```diff
@@ -172,4 +180,54 @@ impl JsGateway {
...
+            .execute()
+            .await?;
+        emit_gateway_warnings(output.warnings);
```
baszalmstra, line 304: Should we return these as well? Similar to the query method?

baszalmstra (changes requested):
Nice! I think we need to change the API slightly to make it mimic the Gateway::query method much more.

js-rattler/crate/gateway.rs
Hofer-Julian: https://github.com/conda/rattler/pull/2688 introduces a callback for that. Should I change it to return warnings instead?

baszalmstra: I think we should have both, just like the rust implementation.

- warnings are emitted to the configured reporter.
- warnings are returned as part of the result.

Hofer-Julian: Done

js-rattler/src/Gateway.ts
```diff
@@ -66,6 +74,62 @@ export type GatewayOptions = {
...
+ * this field is ignored.
+ *
+ * @public
+ */
+export type GatewayQueryResult = RepoDataRecordJson[] & {
+    /** The matching records. This aliases the result array for symmetry. */
+    records: RepoDataRecordJson[];
+    /** Non-fatal warnings encountered during the query. */
+    warnings: string[];
 };
```
baszalmstra, line 131: This is odd. Its both an array, but it also has records which are the same?

baszalmstra (changes requested):

js-rattler/src/Gateway.ts
Hofer-Julian: Nice find, I let the clanker fix it

baszalmstra (approved):

PR #2697, opened 2026-08-18
feat(rattler_conda_types): Implment virtual package plugins

hunger:
... in the rattler_conda_types layer.

All changes are behind the `experimental-virtual-package-plugins` feature flag.

### How Has This Been Tested?

With unit tests and within the wider set of changes that sits on top of this.

### AI Disclosure

Tools: Claude

### Checklist:

### DIFF
--- a/crates/rattler_conda_types/src/repo_data/mod.rs
+++ b/crates/rattler_conda_types/src/repo_data/mod.rs
@@ -24,6 +24,8 @@ use serde_with::{
 use thiserror::Error;
 use url::Url;
 
+#[cfg(feature = "experimental-virtual-package-plugins")]
+use crate::utils::serde::DeserializeVirtualPackagePlugins;
 use crate::{
     Arch, Channel, Flag, MatchSpec, Matches, NoArchType, PackageName, PackageUrl,
     ParseMatchSpecError, ParseStrictness, Platform, RepoDataRecord, VersionWithSource,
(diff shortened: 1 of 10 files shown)


crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -1341,6 +1364,358 @@ mod test {
         }
     }
 
+    #[cfg(feature = "experimental-virtual-package-plugins")]
+    #[test]
+    fn test_virtual_package_plugins() {
```
wolfv, line 1369: Looks good! I like the amoutn of tests. Maybe move to type + tests to a separate file to split concerns?

crates/rattler_conda_types/src/utils/serde.rs
```diff
@@ -229,6 +229,447 @@ impl Serialize for TimestampMs {
...
+/// A registration whose plugin name is invalid is dropped entirely; an invalid
+/// virtual package name is dropped from the plugin that claimed it. Each is
+/// reported as a warning. A malformed name is a mistake by one channel, so it
+/// must not fail the surrounding repodata, but it must not be handed on to a
+/// consumer that would resolve it to an executable either.
+#[cfg(feature = "experimental-virtual-package-plugins")]
+pub struct DeserializeVirtualPackagePlugins;
+
+#[cfg(feature = "experimental-virtual-package-plugins")]
+impl<'de> DeserializeAs<'de, crate::repo_data::VirtualPackagePlugins>
```
wolfv, line 244: Not sure if we really need such complicated deserializers. Would love to understand better why - afaik we don't have it for all other types.

hunger: Claude found it in an adversarial review: msgpacks are self-describing and if a server serves something with a deeply nested structure, then the client will try to read that and exhaust its stack space. That is a valid DOS attack, so I asked it to fix the issue, even though we happily run into the same (and similar) issues elsewhere.

I'll remove it again: IMHO we should harden rattler against this, but probably in a hardening patch and not piecemeal like it is done here.

crates/rattler_conda_types/src/repo_data/mod.rs
hunger: I did not do that in #2701 yet.

crates/rattler_conda_types/src/utils/serde.rs
hunger: This was dropped (as far as possible) in #2701.

PR #2698, opened 2026-08-19
feat: Add RepoDataRecord.from_package_archive method to allow installing local packages without a channel

JeanChristopheMorinPerso:
### Description

This I think is the last missing piece before I can move even more of my workflows to the rattler python API.

Essentially, I want to do `rattler.install(<local package>)`.

This PR adds a new static method to `RepoDataRecord` called `from_package_archive`. It takes a path to a local package and returns a repodata record that can then be passed to `rattler.install` directly.

I have decided to make it asynchronous because other similar looking functions are all async.

### How Has This Been Tested?

Ran the tests and also used the function in some of my scripts.

```python
record = await rattler.RepoDataRecord.from_package_archive("/pth/to/file")
await rattler.install([record], prefix="/tmp/asd/myprefix")
```

### AI Disclosure

Tools: [Kilo](https://kilo.ai/)

```plaintext
I want to implement https://github.com/conda/rattler/discussions/2460. Please propose a possible API for both rust and python.
```

And I worked from there. It initially suggested naming the method `from_package_file` but I decided to instead name it `from_package_archive` to better align with the rest of the codebase. It also made weird choices while implementing the rust function. I directed it to look at https://github.com/conda/rattler/blob/ee385d71af0dca916c6ec36dec438b1c16e3564f/crates/rattler_repodata_gateway/src/gateway/direct_url_query.rs#L72-L95 to take inspiration. The last mile was to make it asynchronous.

### Checklist:


crates/rattler_package_streaming/src/local_package.rs
JeanChristopheMorinPerso, line 1: I'm not too sure if this should go here or somewhere else. I'm open to ideas and suggestions.

py-rattler/src/error.rs
```diff
@@ -74,6 +75,8 @@ pub enum PyRattlerError {
     #[error(transparent)]
     ExtractError(#[from] ExtractError),
     #[error(transparent)]
+    LocalPackageRecordError(#[from] LocalPackageRecordError),
```
JeanChristopheMorinPerso, line 78: I'm curious, should we expose every error individually or is it fine to wrap all of them in one exception class like this?

crates/rattler_package_streaming/src/local_package.rs
```diff
@@ -0,0 +1,191 @@
...
+
+        let index_json: IndexJson = read_package_file(package_path)?;
+
+        // Prefer deriving the identifier straight from the filename, since
+        // that is the canonical form. But local files aren't always named
+        // following the `name-version-build.ext` convention (e.g. a
+        // package downloaded and renamed by a user), so fall back to
+        // building the identifier from the archive's own `index.json`
+        // metadata instead of failing outright.
+        let identifier = CondaArchiveIdentifier::try_from_path(package_path).map_or_else(
```
JeanChristopheMorinPerso: This logic was taken from https://github.com/conda/rattler/blob/ee385d71af0dca916c6ec36dec438b1c16e3564f/crates/rattler_repodata_gateway/src/gateway/direct_url_query.rs#L132-L141

I wasn't sure if I needed to replicate that or not, but I figured it didn't hurt to replicate the behavior.

py-rattler/tests/unit/test_repo_data_record.py
```diff
@@ -0,0 +1,54 @@
...
+@pytest.mark.asyncio
+async def test_from_package_archive_conda(test_data_dir: str) -> None:
+    path = Path(test_data_dir) / "clobber" / "clobber-fd-1-0.1.0-h4616a5c_0.conda"
+
+    record = await RepoDataRecord.from_package_archive(path)
+
+    assert isinstance(record, RepoDataRecord)
+    assert record.name.normalized == "clobber-fd-1"
+    assert record.channel is None
+    assert record.url == f"file://{path.resolve()}"
```
baszalmstra: This is wrong on Windows and for paths requiring URL escaping. I think this should work:

```suggestion
    assert record.url == path.resolve().as_uri()
```

crates/rattler_package_streaming/src/local_package.rs
```diff
@@ -0,0 +1,191 @@
...
+/// # async fn main() {
+/// use rattler_package_streaming::local_package::repodata_record_from_package_archive;
+///
+/// let record = repodata_record_from_package_archive("/path/to/local/package.conda")
+///     .await
+///     .unwrap();
+/// # }
+/// ```
+pub async fn repodata_record_from_package_archive(
+    package_path: impl AsRef<Path> + Send + 'static,
```
baszalmstra: Lets remove the `Send + 'static` bounds from this argument and convert to an owned `PathBuf` before the spawn blocking. That makes the API slightly more ergonomic.

py-rattler/rattler/repo_data/record.py
```diff
@@ -14,6 +16,35 @@ def __init__(self, package_record: PackageRecord, file_name: str, url: str, chan
         )
         self._record = record
 
+    @staticmethod
+    async def from_package_archive(path: os.PathLike[str]) -> RepoDataRecord:
```
baszalmstra:
```suggestion
    async def from_package_archive(path: str | os.PathLike[str]) -> RepoDataRecord:
```

crates/rattler_package_streaming/src/local_package.rs
baszalmstra: I think we should always build the archive identifier from the `index.json` . That is authoritative.  Locally on disk the filename might be different which would lead to discrepancies.

crates/rattler_package_streaming/src/local_package.rs
baszalmstra: `local_package` might be slightly inconsistent with the rest of the crate, which is mostly organized by I/O mechanism. Because this function operates on a local filesystem archive, I would put it in the `fs` module.

py-rattler/src/error.rs
baszalmstra: I dont have a strong opinion. This works for me. We could also give it its own python error type. Up to you!

baszalmstra (changes requested):
Thanks for working on this!

py-rattler/tests/unit/test_repo_data_record.py
JeanChristopheMorinPerso: Oh right, I always forget that `.as_uri()` exists. Though, it's deprecated as of python 3.14: https://docs.python.org/3/library/pathlib.html#pathlib.Path.as_uri...

crates/rattler_package_streaming/src/local_package.rs
JeanChristopheMorinPerso: As I continue to learn rust, are asking for this:
```rust
pub async fn repodata_record_from_package_archive(
    package_path: impl AsRef<Path>,
) -> Result<RepoDataRecord, LocalPackageRecordError> {

    let package_path = package_path.as_ref().to_path_buf();

    simple_spawn_blocking::tokio::run_blocking_task(move || {
        // ...
    })
    .await
}
```

baszalmstra (approved):

PR #2699, opened 2026-08-19
feat: Expose package format selection to `pyrattler.solve` api

soapy1:
### Description

Follow up to https://github.com/conda/rattler/pull/2627 to address the comment: https://github.com/conda/rattler/pull/2627#discussion_r3668956334.

This change exposes the `package_format_select` option to the `pyrattler.solve` api. Other notable changes:
* surfaces `package_format_select` option on pyrattler `Gateway.query` 
* support selecting package by format for sharded subdirs

### How Has This Been Tested?

* [x] unit tests
* [x] try it out
```
import asyncio
import os
from rattler import solve, SparseRepoData, Channel, PackageFormatSelection, MatchSpec, GenericVirtualPackage, PackageName, Version

async def main() -> None:
    # Setup sparse repodata object to represent a slice of repodata
    chn = Channel("with-wheels")

    data_dir = os.path.join( "/home/sophia/projects/rattler/test-data/")  # choose your path to rattler
    noarch_path = os.path.join(data_dir, "channels/with-wheels/noarch/repodata.json")
    noarch_data = SparseRepoData(
        channel=chn,
        subdir="noarch",
        path=noarch_path,
    )
    linux_64_path = os.path.join(data_dir, "channels/with-wheels/linux-64/repodata.json")
    linux_64_data = SparseRepoData(
        channel=chn,
        subdir="linux-64",
        path=linux_64_path,
    )

    solved_data = await solve(
        [noarch_data, linux_64_data],
        [MatchSpec("starlette")],
        virtual_packages=[
            GenericVirtualPackage(PackageName("__unix"), Version("15"), "0"),
            GenericVirtualPackage(PackageName("__linux"), Version("0"), "0"),
        ],
        package_format_selection=PackageFormatSelection.PREFER_CONDA_WITH_WHL,
    )
    print("solved required dependencies:")
    for rec in solved_data:
        print(f"  * {rec.name.normalized} v{rec.version}")


if __name__ == "__main__":
    asyncio.run(main())

```

### AI Disclosure

Tools:Claude

### Checklist:

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
@@ -91,7 +91,7 @@ impl SubdirClient for LocalSubdirClient {
                 )
             };
             let records = sparse_repodata
-                .load_records(&name, PackageFormatSelection::PreferConda)
+                .load_all_records_by_name(&name)
                 .map_err(io_error)?;
             let removed = sparse_repodata
                 .load_removed(Some(&name))
--- a/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
@@ -114,7 +114,7 @@ impl SubdirClient for LocalSubdirClient {
     fn package_names(&self) -> Vec<String> {
         let sparse_repodata: Arc<SparseRepoData> = self.sparse.clone();
         sparse_repodata
-            .package_names(PackageFormatSelection::PreferCondaWithWhl)
+            .package_names(PackageFormatSelection::PreferConda)
             .map(std::convert::Into::into)
             .collect()
     }
--- a/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
@@ -114,7 +114,7 @@ impl SubdirClient for LocalSubdirClient {
     fn package_names(&self) -> Vec<String> {
         let sparse_repodata: Arc<SparseRepoData> = self.sparse.clone();
         sparse_repodata
-            .package_names(PackageFormatSelection::Both)
+            .package_names(PackageFormatSelection::PreferCondaWithWhl)
             .map(std::convert::Into::into)
             .collect()
     }
--- a/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
@@ -114,7 +114,7 @@ impl SubdirClient for LocalSubdirClient {
     fn package_names(&self) -> Vec<String> {
         let sparse_repodata: Arc<SparseRepoData> = self.sparse.clone();
         sparse_repodata
-            .package_names(PackageFormatSelection::PreferConda)
+            .package_names(PackageFormatSelection::Both)
             .map(std::convert::Into::into)
             .collect()
     }
--- a/crates/rattler_repodata_gateway/src/gateway/mod.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/mod.rs
@@ -2292,6 +2292,47 @@ mod test {
         assert_eq!(all_records.len(), expected_count);
     }
 
+    #[rstest]
+    #[case::default_prefers_conda(PackageFormatSelection::PreferConda, 2)]
+    #[case::only_conda(PackageFormatSelection::OnlyConda, 2)]
+    #[case::only_tar_bz2(PackageFormatSelection::OnlyTarBz2, 0)]
+    #[case::both(PackageFormatSelection::Both, 2)]
+    #[tokio::test]
+    async fn test_package_format_selection_custom_source(
+        #[case] selection: PackageFormatSelection,
+        #[case] expected_count: usize,
+    ) {
+        let gateway = Gateway::new();
+
+        // Create a mock source with some records
+        let mut mock_source = MockRepoDataSource::new();
+        mock_source.add_record(
+            Platform::Linux64,
+            make_test_record("testpkg", "1.0.0", "linux-64"),
+        );
+        mock_source.add_record(
+            Platform::Linux64,
+            make_test_record("testpkg", "2.0.0", "linux-64"),
+        );
+
+        let source: Arc<dyn super::RepoDataSource> = Arc::new(mock_source);
+
+        // Query using the custom source
+        let records = gateway
+            .query(
+                vec![super::Source::Custom(source.clone())],
+                vec![Platform::Linux64],
+                vec![PackageName::from_str("testpkg").unwrap()].into_iter(),
+            )
+            .recursive(false)
+            .package_format_selection(selection)
+            .await
+            .unwrap();
+
+        let all_records: Vec<_> = records.iter().flat_map(RepoData::iter).collect();
+        assert_eq!(all_records.len(), expected_count);
+    }
+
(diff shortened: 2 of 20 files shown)


crates/rattler_repodata_gateway/src/gateway/query.rs
soapy1, line 570: `Custom` (and `Channel`) sources don't currently do surface the ability to filter by package format selection. Demo for this https://github.com/conda/rattler/compare/main...soapy1:rattler:solve-with-package-format-selection?expand=1

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -547,11 +562,16 @@ impl QueryExecutor {
...
+                            package_format_selection,
                         );
                         (kind, fut)
                     }
                     Source::Custom(custom_source) => {
-                        let client = CustomSourceClient::new(custom_source, platform);
+                        let client = CustomSourceClient::new(
+                            custom_source,
+                            platform,
+                            package_format_selection,
```
soapy1: A different approach to surfacing this option is to make `fetch_package_records` accept `package_format_selection` as an option. I think both of these break backwards compatibility in different ways. Not sure what the right way to compare these two solutions is.

baszalmstra (changes requested):
I dont think this is the right approach. This PR creates and caches subdirs + format selection. I feel format selection is just a view of all the data so I would much prefer we retain a single subdirclient and let it return a filtered set of records. That way we ensure that if you change the format selection for a second query you do get a lot of already cached entries.

crates/rattler_repodata_gateway/src/gateway/mod.rs
```diff
@@ -1887,6 +1888,73 @@ mod test {
         }
     }
 
+    /// Filters and, where applicable, deduplicates `records` by `selection`.
+    /// A real `RepoDataSource` is expected to honor `package_format_selection`
+    /// itself (the gateway applies no fallback filtering for custom sources),
+    /// so `MockRepoDataSource` implements the same selection contract here to
+    /// exercise it in tests.
+    fn filter_by_package_format(
```
soapy1, line 1896: This is just for the `MockRepoDataSource` so we can kick the tires on package format selection on custom repodata sources.

crates/rattler_repodata_gateway/src/gateway/source.rs
```diff
@@ -27,6 +30,7 @@ pub trait RepoDataSource: Send + Sync {
         &self,
         platform: Platform,
         name: &PackageName,
+        package_format_selection: PackageFormatSelection,
```
soapy1, line 33: This kinda seems like a big api change to me?

crates/rattler_repodata_gateway/src/gateway/source.rs
soapy1, line 40: should this also be updated to take package format selection as a parameter?

baszalmstra: Yep I think that makes sense.

crates/rattler_repodata_gateway/src/gateway/source.rs
baszalmstra: It is, but I also dont really see a way around it.

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/mod.rs
```diff
@@ -88,38 +90,113 @@ async fn decode_zst_bytes_async<R: AsRef<[u8]> + Send + 'static>(
...
-
-            let channel_str = channel_base_url.url().clone().redact().to_string();
-            let base_url_str = base_url.as_str();
-            let mut records: Vec<Arc<RepoDataRecord>> = packages
-                .map(|(file_name, package_record)| {
+    let parse = move || {
+        let shard = rmp_serde::from_slice::<Shard>(bytes.as_ref())
+            .map_err(|e| std::io::Error::new(std::io::ErrorKind::InvalidData, e.to_string()))
+            .map_err(FetchRepoDataError::IoError)?;
+
```
baszalmstra: This now reparses the entire shard if we ask for a different format selection.

baszalmstra (changes requested):

crates/rattler_repodata_gateway/src/gateway/sharded_subdir/mod.rs
soapy1: One strategy to fix this is to cache the parsed shards on the `ShardedSubdir`. Now, if caching is enabled, rattler will try to read from the cache of shards. If nothing is found, it will load the shards from the shard cache path, or download the shard data, then save this information to the parsed shard cache.

baszalmstra (changes requested):
This effectively loads all repodata in memory twice, once when the shard is parsed and once when its cloned back to the user.

The current implementation always parses everything from the shard through serde. I dont think we currently have a way around this because there is no such thing as serde_json::RawValue for serde_rmp.

We now clone `shard`, keeping one cached and one we extract the records from.

Lets do this differently. Lets:

1. Always parse the entire shard into repodatarecords and caching them (use Arc).
2. Filter the records based on the package selection.

I apologize for sending you on these different tangents, I know this is close to what you had previously.

crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
```diff
@@ -91,7 +91,7 @@ impl SubdirClient for LocalSubdirClient {
                 )
             };
             let records = sparse_repodata
-                .load_records(&name, PackageFormatSelection::PreferConda)
+                .load_all_records_by_name(&name)
```
soapy1, line 94: We'll load all the records here, since we'll choose which records to show later down in the process. But don't want to remove records here prematurely.  This means we don't get to use package format selection from the sparse repodata implementation.

crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
```diff
@@ -114,7 +114,7 @@ impl SubdirClient for LocalSubdirClient {
     fn package_names(&self) -> Vec<String> {
         let sparse_repodata: Arc<SparseRepoData> = self.sparse.clone();
         sparse_repodata
-            .package_names(PackageFormatSelection::PreferConda)
+            .package_names(PackageFormatSelection::PreferCondaWithWhl)
```
soapy1: to do, get package names by package format selection.

crates/rattler_repodata_gateway/src/sparse/mod.rs
```diff
@@ -72,6 +72,9 @@ pub enum PackageFormatSelection {
 
     /// Both .tar.bz2 and .conda packages are used
     Both,
+
+    /// All packages are included
+    All,
```
soapy1, line 77: Add the ability to choose all package formats so that filtering on package format can be done at a different stage. In particular for querying a subdir.

crates/rattler_repodata_gateway/src/gateway/local_subdir.rs
baszalmstra: Im still not quite happy with this because it will always eagerly load everything and only then filter elements.

For shards there is no way around it, we always have to load the entire shard. But for sparse records its kind of a waste.

baszalmstra (changes requested):
Its getting there but Im still not quiet happy. Do you mind if I amend your PR?

soapy1:
>  Do you mind if I amend your PR?

Much appreciated, go ahead. Thanks.

PR #2727, opened 2026-08-27
feat: add rattler_azure crate

tenzinplatter:
### Description

First PR in a series splitting up #2615 (Azure Blob Storage channel support) into reviewable pieces, as discussed there.

This introduces the new `rattler_azure` crate on its own, with no wiring into any other crate yet. It holds the shared Azure vocabulary and helpers the follow-up PRs (config, networking middleware, index, upload) will build on:

- Validated types for Azure addressing: `AzureChannelUrl` (`az://` URLs), `AzureEndpointKey` (host-style vs path-style account addressing), `AzureHost`, `AccountName`, `ContainerName`, `AzureScheme`
- `AzureLocation` / `locate()` — matching a channel URL against configured endpoints
- `AzureEndpointOptions` + `Auth` — per-container grant configuration (serde behind the `serde` feature)
- `AzureCredentials` (account key / SAS, wrapped in `secrecy`)
- `azblob_config()` — opendal azblob operator configuration (behind the `opendal` feature)
- `clap` feature: CLI credential options and resolution, including user-delegation SAS minting via the `az` CLI

Workspace changes are limited to adding the crate and the `secrecy` workspace dependency. Feature flags: `default = []`, `clap`, `opendal`, `serde`.

Follow-up PRs will add, in order: `rattler_config` azure-options, SAS signature redaction, the `rattler_networking` middleware (+ py-rattler bindings), `rattler_index`/`rattler_upload` support, and finally gateway/CLI integration.

### How Has This Been Tested?

- `cargo nextest run -p rattler_azure --all-features` — 57/57 pass
- `cargo clippy -p rattler_azure --all-features -- -D warnings` and `cargo fmt --check` — clean
- End-to-end behaviour (Azurite emulator tests and real-Azure fetch tests) is exercised on the full branch behind #2615 and will land with the follow-up PRs that use this crate.

### AI Disclosure
- [x] This PR contains AI-generated content.
  - [x] I have tested any AI-generated content in my PR.
  - [x] I take responsibility for any AI-generated content in my PR.

Tools: Claude Code

```plaintext
Start by getting an overview of these changes, then we want to split this pr into multiple, starting with a pr focusing on introducing the rattler_azure crate
```

### Checklist:
- [x] I have performed a self-review of my own code
- [x] I have added sufficient tests to cover my changes.

PR #2728, opened 2026-08-27
feat: add azure-options to rattler_config

tenzinplatter:
### Description

Second PR in the series splitting up #2615 (Azure Blob Storage channel support). **Stacked on #2727** — the diff includes that PR's commit until it merges; review only the `rattler_config` changes here.

This adds the `azure-options` table to `rattler_config`, letting users configure Azure endpoints and per-container auth grants:

```toml
[azure-options."mycompany.blob.core.windows.net"]
scheme = "https"

[azure-options."mycompany.blob.core.windows.net".auth]
releases = true
public = false

# Path-style addressing (Azurite, proxies): account as first path segment
[azure-options."127.0.0.1:10000/devstoreaccount1"]
scheme = "http"
```

- Entries are keyed by `rattler_azure::AzureEndpointKey` — a host (host-style, real Azure) or `host[:port]/account` (path-style). A document naming the same host both ways is rejected at parse time, since the two readings would silently shadow each other.
- Each entry carries an optional `scheme` and an `auth` table of per-container grants (`<container> = true/false`); an unlisted container is not granted.
- Across config layers the endpoint map extends, but each endpoint's entry is replaced wholesale — so a higher layer can revoke a grant by omission.
- `azure_options` is wired into `CommonConfig` (defaults, merge, validate, keys), and `edit::save()` now validates before writing.

### How Has This Been Tested?

- `cargo nextest run -p rattler_config --all-features` — 90/90 pass, including e2e fixture parsing (`kitchen-sink.toml`, `override-layer.toml`), layer-merge semantics, edit round-trips, and insta snapshots
- `cargo clippy -p rattler_config --all-features -- -D warnings` and `cargo fmt --check` — clean

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/test-data/sparse/generate.py
+++ b/test-data/sparse/generate.py
@@ -34,9 +34,7 @@ def tar_zst(entries, level=3):
 
 
 def index_json(name):
-    return json.dumps(
-        {"name": name, "version": "1.0.0", "build": "0", "build_number": 0}
-    ).encode()
+    return json.dumps({"name": name, "version": "1.0.0", "build": "0", "build_number": 0}).encode()
 
 
 def write_conda(stem, pkg_entries, info_entries, force_zip64=False):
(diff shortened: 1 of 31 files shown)


test-data/sparse/generate.py
```diff
@@ -34,9 +34,7 @@ def tar_zst(entries, level=3):
 
 
 def index_json(name):
-    return json.dumps(
-        {"name": name, "version": "1.0.0", "build": "0", "build_number": 0}
-    ).encode()
+    return json.dumps({"name": name, "version": "1.0.0", "build": "0", "build_number": 0}).encode()
```
tenzinplatter, line 37: These `test-data` changes are just from running `pixi run lint`, can take this out if you'd rather as it is unrelated

PR #2733, opened 2026-08-27
fix(rattler_networking): report sharded repodata as a 404 instead of erroring

### LINKED ISSUE #2726: `oci://` channels are unusable with sharded repodata: ghcr.io token 403 aborts the solve instead of falling back to repodata.json
### Checklist

### What happened?

**Versions:** `rattler_networking` 0.30.6, `rattler_repodata_gateway` 0.33.0, reproduced
through pixi 0.77.1 on macOS (osx-arm64).

With sharded repodata enabled (pixi's default), any `oci://` channel — or any `oci://`
entry in `[mirrors]` — fails the solve outright with a 403 from the ghcr.io token
endpoint. The gateway's existing "sharded repodata is missing, fall back to
repodata.json" path is never reached.

**Reproducer 1 — `oci://` as a channel:**

```console
$ pixi search black --channel oci://ghcr.io/channel-mirrors/conda-forge --platform osx-arm64
Using channels: oci://ghcr.io/channel-mirrors/conda-forge/
Error:   × Reqwest error: HTTP status client error (403 Forbidden) for url (https://ghcr.io/token?service=ghcr.io&scope=repository%3Achannel-mirrors%2Fconda-
  │ forge%2Fnoarch%2Frepodata_shards.msgpack.zst%3Apull)
  ╰─▶ HTTP status client error (403 Forbidden) for url (https://ghcr.io/token?service=ghcr.io&scope=repository%3Achannel-mirrors%2Fconda-forge%2Fnoarch%2Frepodata_shards.msgpack.zst%3Apull)

```

**Reproducer 2 — `oci://` as a mirror:**

```toml
# $PIXI_HOME/config.toml
[mirrors]
"https://conda.anaconda.org/conda-forge" = ["oci://ghcr.io/channel-mirrors/conda-forge"]
```

```console
$ pixi global install yt-dlp
  × Reqwest error: HTTP status client error (403 Forbidden) for url (https://ghcr.io/token?service=ghcr.io&scope=repository%3Achannel-mirrors%2Fconda-forge%2Fosx-arm64%2Frepodata_shards.msgpack.zst%3Apull)
```

### Root cause

Three things compose into a hard failure.

**1. `OCIUrl::new` has no branch for sharded repodata filenames.**
`oci_middleware.rs` (0.30.6, lines 474-545) only recognises `*.conda`, `*.tar.bz2` and
`repodata.json*`. There is no `else`, so `repodata_shards.msgpack.zst` is passed through
verbatim and becomes the OCI *repository* name, with an empty `media_type`:

```
repository:channel-mirrors/conda-forge/osx-arm64/repodata_shards.msgpack.zst:pull
```

That repository does not exist — the OCI layout publishes `repodata.json` as the
repository and carries the compressed variants as layers, which is exactly what the
`filename.starts_with("repodata.json")` branch exists for.

**2. ghcr.io answers "repository does not exist" with 403, not 404.**

| token scope | status |
| --- | --- |
| `channel-mirrors/conda-forge/noarch/repodata.json:pull` | 200 |
| `channel-mirrors/conda-forge/osx-arm64/repodata.json:pull` | 200 |
| `channel-mirrors/conda-forge/osx-arm64/repodata.json.zst:pull` | 403 |
| `channel-mirrors/conda-forge/osx-arm64/repodata_shards.msgpack.zst:pull` | 403 |
| `channel-mirrors/conda-forge/osx-arm64/this-does-not-exist-xyz:pull` | 403 |

```json
{"errors":[{"code":"DENIED","message":"requested access to the resource is denied"}]}
```

**3. A 403 from the token exchange is a fatal middleware error, and the gateway only
treats 404/501 as "no sharded repodata".**

`get_token` calls `error_for_status()`, so the 403 becomes
`OciMiddlewareError::Reqwest`, and `lookup_error_to_response` softens only two variants:

```rust
OciMiddlewareError::LayerNotFound => Ok(create_404_response(...)),
OciMiddlewareError::ManifestRequestFailed(StatusCode::NOT_FOUND) => Ok(create_404_response(...)),
_ => Err(reqwest_middleware::Error::Middleware(error.into())),   // 403 lands here
```

Meanwhile `rattler_repodata_gateway` 0.33.0, `gateway/sharded_subdir/mod.rs:22-24`:

```rust
pub(super) fn is_missing_sharded_repodata_status(status: StatusCode) -> bool {
    status == StatusCode::NOT_FOUND || status == StatusCode::NOT_IMPLEMENTED
}
```

Only those two statuses become `SubdirNotFoundError`, which `subdir_builder.rs:67-72`
catches to fall back to `repodata.json`. Every other error propagates and kills the solve.

### Expected behaviour

Requesting sharded repodata from an OCI registry should degrade to `repodata.json`, the
same way a 404 or 501 from an HTTP channel does. `[mirrors]` entries and `oci://`
channels should both work with pixi's default configuration.

### Suggested fix

**(a) Preferred — make `OCIUrl::new` reject filenames it cannot map to an OCI artifact,
as a 404 response rather than passing them through as a repository name.** OCI channel
layouts do not publish sharded repodata at all, so `repodata_shards.msgpack.zst` (and
`*.msgpack.zst` generally) is statically known to be absent. This is registry-agnostic,
costs no network round trip, and cannot mask a genuine authorization problem.

**(b) Secondary — treat an anonymous token exchange that returns 401/403 with no stored
credentials as "artifact not available" (404).** This would also cover other registries
that report absence as denial, but it is riskier: with credentials present, a real
permission error must stay an error, otherwise private-registry misconfiguration silently
degrades into confusing "package not found" failures. If applied, it should be scoped to
the no-credentials case only.

Also worth considering: `is_missing_sharded_repodata_status` could accept 403, but I'd
argue against it — the gateway is the wrong layer to reinterpret registry-specific
semantics, and it would weaken the signal for real HTTP channels.

### Related

- #2486 (closed, **not merged**) — attempted the same fallback, but keyed on the channel
  URL scheme. That approach cannot fix the `[mirrors]` case: mirror substitution happens
  in `MirrorMiddleware`, so the gateway only ever sees `https://conda.anaconda.org/...`
  and has no idea the request will be served by an OCI registry.
- #2651 (merged, present in 0.30.6) — made a *missing manifest* a clean 404. The failure
  here happens earlier, during the token exchange, so it is not covered.
- #2624 — adjacent problems in the same middleware.

### Workaround

```toml
[repodata-config]
disable-sharded = true
```

XhstormR:
### Description

Fixes #2726

`OCIUrl::new` recognises `*.conda`, `*.tar.bz2` and `repodata.json*` filenames and passes
everything else through as the OCI repository name. With sharded repodata enabled — the
default in `SourceConfig` and therefore in pixi — the gateway probes
`repodata_shards.msgpack.zst` first, so the repository becomes
`channel-mirrors/conda-forge/<subdir>/repodata_shards.msgpack.zst`. No registry publishes
that, and ghcr.io reports an absent repository from its token endpoint as `403 DENIED`
rather than `404`:

| token scope | status |
| --- | --- |
| `channel-mirrors/conda-forge/osx-arm64/repodata.json:pull` | 200 |
| `channel-mirrors/conda-forge/osx-arm64/repodata_shards.msgpack.zst:pull` | 403 |
| `channel-mirrors/conda-forge/osx-arm64/this-does-not-exist-xyz:pull` | 403 |

`get_token` turns the 403 into `OciMiddlewareError::Reqwest`, and `lookup_error_to_response`
only softens `LayerNotFound` and `ManifestRequestFailed(404)`, so the chain aborts with a
middleware error. On the gateway side `is_missing_sharded_repodata_status` accepts only 404
and 501, so the existing "sharded repodata is missing, fall back to `repodata.json`" branch
in `subdir_builder` is never reached. The result is that both `oci://` channels and `oci://`
entries in `[mirrors]` fail out of the box.

This answers the two sharded-repodata URL shapes used by the gateway locally: the exact
`repodata_shards.msgpack.zst` index and `shards/<sha256>.msgpack.zst`. The current OCI
channel layout defines no mapping for either. Reporting those requests as a 404 costs no
network round trip and lets the caller degrade. The match deliberately remains narrow so
other `.msgpack.zst` paths can still be valid bare OCI repository references. Future
sharded-repodata support needs an agreed OCI layout and corresponding mappings in both
`conda-oci-mirror` and rattler.

**Why not report every unrecognised filename as a 404.** That would break bare repository
references. Conda package names may contain dots, and
`repository:channel-mirrors/conda-forge/noarch/ruamel.yaml:pull` is a real repository that
returns 200, so "contains a dot but is not a known artifact type" is not a safe signal for
absence.

**Why not widen `is_missing_sharded_repodata_status` to accept 403 instead.** The gateway is
the wrong layer to reinterpret registry-specific semantics, and it would weaken the signal
for plain HTTP channels where a 403 is a genuine authorization failure.

**Deliberately out of scope.** An anonymous token exchange that returns 401/403 with no
stored credentials is arguably also "not available", which would additionally cover other
registries that report absence as denial. That is a broader behavioural change and risks
masking real permission errors on private registries, so it is left for a separate
discussion.

### How Has This Been Tested?

`cargo test -p rattler_networking --features rustls --lib oci` — 13 passed, 0 failed. This
includes every pre-existing ghcr.io test (`test_oci_middleware`,
`test_oci_middleware_repodata`, `test_oci_middleware_unknown_digest_falls_back_to_manifest`,
`test_oci_middleware_missing_package_is_404`,
`test_oci_middleware_missing_package_with_digest_is_404`), so package and `repodata.json`
pulls are unaffected.

Two new tests:

- `sharded_repodata_is_unavailable_but_other_zstd_artifacts_are_not` — no network. The
  exact `repodata_shards.msgpack.zst` index and a valid `shards/<sha256>.msgpack.zst`
  shard yield `ShardedRepodataUnavailable`; unrelated `.msgpack.zst` repositories and
  `repodata.json.zst` continue to resolve normally.
- `test_oci_middleware_sharded_repodata_is_404` — the same request returns 404 through the
  full middleware path rather than erroring.

End-to-end against the reported failure:

```console
$ cargo run --bin rattler -- create --dry-run -c oci://ghcr.io/channel-mirrors/conda-forge black
{{E2E_RESULT}}
```

### AI Disclosure

Tools: Claude Code

### Checklist:


wolfv (approved):

wolfv:
Rebased onto current `main` to pick up the dependency updates needed by the semver job. Local OCI test suite still passes (13/13).

XhstormR:
@wolfv All CI checks passed.

PR #2738, opened 2026-08-31
feat: add repodata whoneeds for reverse dependency lookup

### LINKED ISSUE #1362: Add repoquery showneeds and depends
### Checklist

### What is the idea?

Now that https://github.com/conda/rattler/issues/518 is complete, the only missing pieces are the `whoneeds` and the `requires` subcommands.

```
$ conda repoquery --help
usage: conda repoquery [-h] {whoneeds,depends,search} ...

Advanced search for repodata.

positional arguments:
  {whoneeds,depends,search}
    whoneeds            Show packages that depend on this package.
    depends             Show dependencies of this package.
    search              Show all available package versions.

options:
  -h, --help            Show this help message and exit.
```

### Why is this needed?

I don't know if these commands are widely used, but they can be useful.

### What should happen?

_No response_

### Additional Context

_No response_

pavelzw:
### Description

closes #1362

This PR adds reverse dependency lookup (`whoneeds`) to rattler: a pure analysis module, a streaming gateway query, py-rattler bindings, and a `rattler whoneeds` CLI command. (The `depends` half of #1362 is intentionally out of scope: forward dependencies are already covered by `rattler solve` and `rattler search`.)

**Key pieces:**

1. **`repoquery` module** (`crates/rattler_repodata_gateway/src/repoquery.rs`, new `repoquery` feature included in `gateway`): pure (no I/O) reverse dependency matching over records. A record matches when one of its `depends`, `constrains`, or run export entries references the target; every match is tagged with its `DependencyKind`, so callers filter instead of toggling options. The target is either a `PackageName` (name-based matching, also for virtual package names), a concrete `PackageRecord` (proper match spec matching via `Matches<PackageRecord>`), or a `GenericVirtualPackage`.

2. **Streaming `Gateway::who_needs`** (`crates/rattler_repodata_gateway/src/gateway/who_needs.rs`): scans the queried sources/platforms package by package instead of materializing a wildcard query. Scanned records are dropped as soon as each package is processed and are never inserted into the gateway's long-lived per-name record cache (pre-cached entries are reused and left in place); only the matching records are retained, `Arc`-shared in `OwnedDependent`s. Platforms are scanned with bounded concurrency (4, `buffer_unordered` + input-index reordering to keep result order deterministic); within a subdir, batches of 500 names run as their own tasks (8 concurrent) so fetching and scanning parallelize. CEP-42 channel relations are followed like in the other gateway queries.

3. **py-rattler**: `rattler.who_needs(records, target)` (pure, over records you already hold) and `await Gateway.who_needs(sources, platforms, target)` (streaming; only matching records cross into Python). Docs page, doctests, and unit tests included, plus `scripts/reproduce_whoneeds.py` reproducing the pixi-browse all-platform usage pattern.

4. **CLI**: `rattler whoneeds <name-or-package-file-or-url>`, built on the streaming query with sharding disabled (reverse lookup needs every record of the channel; one full `repodata.json` fetch beats one shard fetch per package).

### API overview

**Rust — `rattler_repodata_gateway::repoquery`** (new feature `repoquery`, included in `gateway`):

```rust
pub enum WhoNeedsTarget {
    Name(PackageName),                      // any dep naming the package matches
    Record(Box<PackageRecord>),             // dep matches iff its matchspec matches this record
    VirtualPackage(GenericVirtualPackage),  // same, via Matches<GenericVirtualPackage>
}
// From<PackageName> / From<PackageRecord> / From<RepoDataRecord> / From<GenericVirtualPackage>

pub enum DependencyKind { Depends, Constrains, RunExport(RunExportKind) }
pub enum RunExportKind { Weak, Strong, Noarch, WeakConstrains, StrongConstrains }

// borrowed result + pure scan over records the caller already holds
pub struct Dependent<'r> { pub record: &'r RepoDataRecord, pub dependency: &'r str, pub kind: DependencyKind }
pub fn who_needs<'r>(records: impl IntoIterator<Item = &'r RepoDataRecord>, target: impl Into<WhoNeedsTarget>) -> Vec<Dependent<'r>>;

// owned counterparts (records shared via Arc), used by streaming scans
pub struct OwnedDependent { pub record: Arc<RepoDataRecord>, pub dependency: String, pub kind: DependencyKind }
pub fn who_needs_owned(records: &[Arc<RepoDataRecord>], target: &WhoNeedsTarget) -> Vec<OwnedDependent>;
```

**Rust — gateway** (feature `gateway`):

```rust
impl Gateway {
    pub fn who_needs(
        &self,
        sources: impl IntoIterator<Item = impl Into<Source>>,
        platforms: impl IntoIterator<Item = Platform>,
        target: impl Into<WhoNeedsTarget>,
    ) -> WhoNeedsQuery;
}

pub struct WhoNeedsQuery;   // builder; also impl IntoFuture
impl WhoNeedsQuery {
    pub fn channel_relations(self, mode: ChannelRelationsMode) -> Self;
    pub fn channel_relations_max_depth(self, depth: usize) -> Self;
    pub fn with_reporter(self, reporter: impl Reporter + 'static) -> Self;
    pub async fn execute(self) -> Result<WhoNeedsQueryOutput, GatewayError>;
}
pub struct WhoNeedsQueryOutput { pub dependents: Vec<OwnedDependent>, pub warnings: Vec<GatewayWarning> }

impl RepoDataQueryOutput {
    pub fn who_needs(&self, target: impl Into<WhoNeedsTarget>) -> Vec<Dependent<'_>>;
}
```

Supporting plumbing: `SubdirData::fetch_package_records_uncached` (reuses cached entries, never inserts) and a defaulted `SubdirClient::fetch_package_records_without_deps` that skips the unique-dependency extraction the scan never uses (overridden for the sparse-backed clients).

**Python — exported from `rattler` and `rattler.repo_data`:**

```python
def who_needs(
    records: Sequence[RepoDataRecord],
    target: str | PackageName | PackageRecord | GenericVirtualPackage,
) -> list[Dependent]

class Dependent:
    record: RepoDataRecord            # Arc-shared, not copied
    dependency: str                   # the matching dependency string
    kind: Literal["depends", "constrains", "run_export"]
    run_export_kind: Literal["weak", "strong", "noarch", "weak_constrains", "strong_constrains"] | None

class Gateway:
    async def who_needs(
        self,
        sources: Iterable[Channel | str | RepoDataSource],
        platforms: Iterable[Platform | PlatformLiteral],
        target: str | PackageName | PackageRecord | GenericVirtualPackage,
    ) -> list[Dependent]
```

A `str` target is treated as a `PackageName`.

**CLI — new top-level subcommand:**

```
rattler whoneeds <package>
    <package>: a package name (numpy, __cuda) → name matching,
               or a path/URL to a .conda/.tar.bz2 → the full record is extracted
               from its index.json and matched properly
    -c/--channels <CHANNELS>   (default: conda-forge)
    -p/--platform <PLATFORM>   (default: current; noarch always included)
    --limit <N> (default 100) / --all
    --json
```

### Performance and memory

Measured release-built against cached conda-forge, all 12 platforms, target `polars`, fresh process, 4-core machine:

- one `Gateway.who_needs` call: **7.2 s** (12 concurrent single-platform calls at asyncio concurrency 4: 6.9 s; the earlier wildcard-query implementation: ~15 s, and ~100 s before per-platform splitting)
- peak RSS: **2.4 GB** (previously ~9 GB at 12 platforms), of which ~1.8 GB are clean, kernel-evictable mmap pages of the cached repodata held by the pre-existing subdir cache; the per-name record cache stays empty (asserted by test), so gateway-retained heap drops from gigabytes of records to the ~120 MB sparse-index footprint
- a follow-up query while retaining 2.27 M results from a previous `whoneeds python` query adds only ~0.4 GB peak and runs in 6.5 s (`py-rattler/scripts/reproduce_whoneeds.py`)

Worst-case targets like `python` legitimately match a third of the channel (2.27 M records), so the result itself dominates memory there; a summarizing/paged variant is a possible follow-up.

### How Has This Been Tested?

Rust (`repoquery`): name/record/constrains/run-export/virtual-package matching. Rust (`gateway::who_needs`): streaming results identical to the materialized wildcard-query implementation for name and concrete-record targets across multiple and duplicate platforms; the record cache not being populated by a scan while pre-cached entries survive and stay usable; platform result grouping; error propagation. Python: 6 pytest cases incl. a gateway round-trip against the local test channel, plus doctests. Verified end-to-end against conda-forge (CLI and py-rattler); ruff (locked version), mypy (strict), and clippy (`-D warnings`) are clean.

### Checklist:

https://claude.ai/code/session_01RuU28LzKwsxKFT1hHda28k


crates/rattler_repodata_gateway/src/repoquery.rs
```diff
@@ -0,0 +1,429 @@
...
+
+use rattler_conda_types::{MatchSpec, PackageName, ParseMatchSpecOptions, RepoDataRecord, Version};
+
+/// The concrete package to find reverse dependencies for with [`who_needs`].
+///
+/// Unlike a `MatchSpec`, this describes a single concrete package — a name,
+/// optionally pinned to a version and build string — so that the dependency
+/// match specs of candidate records can be unambiguously matched against it.
+#[derive(Debug, Clone)]
+pub struct WhoNeedsTarget {
```
pavelzw: Not sure whether this is the best structure. Theoretically, one could also have something like this in your matchspec: `foo[license=MIT]`. With this name + version,build optional we could not check against things like the licenses. We could also provide a full repodatarecord alternatively, might be a bit tedious to use though (at least I don't want to specify the whole record in `rattler whoneeds`)

crates/rattler_repodata_gateway/src/who_needs.rs
```diff
@@ -0,0 +1,535 @@
...
+                // and move on: the entry cannot be shown to reference the
+                // target, and one bad record must not fail a whole-channel
+                // scan.
+                Err(error) => {
+                    tracing::warn!(
+                        "ignoring the {kind} entry '{dependency}' of '{}', which is not a valid match spec: {error}",
+                        record.identifier,
+                    );
+                    continue;
+                }
```
pavelzw, line 296: happens in 52 entries in conda-forge: https://github.com/conda-forge/admin-requests/pull/2321

crates/rattler_repodata_gateway/src/gateway/mod.rs
```diff
@@ -192,6 +194,31 @@ impl Gateway {
...
+    /// lookup that scans every package of the given sources and platforms
+    /// against `target` without retaining or caching the scanned records —
+    /// only the matching records are kept. See [`WhoNeedsQuery`] for
+    /// details and [`crate::who_needs::WhoNeedsTarget`] for the matching
+    /// semantics of the different target variants.
+    pub fn who_needs<AsSource, SourceIter, PlatformIter>(
+        &self,
+        sources: SourceIter,
+        platforms: PlatformIter,
+        target: impl Into<crate::who_needs::WhoNeedsTarget>,
```
baszalmstra: Please just use this, instead of using FQN.

crates/rattler_repodata_gateway/src/gateway/mod.rs
```diff
@@ -192,6 +194,31 @@ impl Gateway {
         )
     }
 
+    /// Constructs a new [`WhoNeedsQuery`]: a streaming reverse-dependency
+    /// lookup that scans every package of the given sources and platforms
+    /// against `target` without retaining or caching the scanned records —
+    /// only the matching records are kept. See [`WhoNeedsQuery`] for
+    /// details and [`crate::who_needs::WhoNeedsTarget`] for the matching
+    /// semantics of the different target variants.
```
baszalmstra: Please rewrite this explaining the responsibility of this function from a user facing perspective. An example goes a long way.

crates/rattler_repodata_gateway/src/gateway/query.rs
```diff
@@ -1324,21 +1324,44 @@ fn spawn_one_package_fetch(
...
+/// without pinning it themselves.
+#[cfg(not(target_arch = "wasm32"))]
+pub type BoxStream<T> = futures::stream::BoxStream<'static, T>;
+
+#[cfg(not(target_arch = "wasm32"))]
+pub(super) fn box_stream<T, S: futures::Stream<Item = T> + Send + 'static>(
+    stream: S,
+) -> BoxStream<T> {
+    stream.boxed()
+}
```
baszalmstra: Lets lift these in a shared module instead of who_needs_query now depending on types of query.rs.

crates/rattler_repodata_gateway/src/gateway/mod.rs
pavelzw: 🤖: Done, `WhoNeedsTarget` is imported now.

crates/rattler_repodata_gateway/src/gateway/mod.rs
pavelzw: 🤖: Rewritten from the caller's perspective: what it answers, what counts as a reference per target variant, why it reads more repodata than `query` and what that costs, plus a `no_run` example iterating the dependents.

crates/rattler_repodata_gateway/src/gateway/query.rs
pavelzw: 🤖: Moved `BoxFuture`/`BoxStream` and their constructors into a new `gateway::boxed` module; `query.rs` and `who_needs_query.rs` both import them from there now.

crates/rattler_repodata_gateway/src/who_needs.rs
pavelzw: 🤖: Thanks — I put that number and the link into the comment above the warn, so the "this really happens" justification is in the code.

baszalmstra (approved):

PR #2745, opened 2026-08-31
feat(gateway): configure repodata variant fallback

### LINKED ISSUE #2742: feat(gateway): configure repodata filename fallback order
### Problem

Conda-compatible clients need to implement `repodata_fns`, an ordered list such as `current_repodata.json, repodata.json`. Gateway's low-level fetch API exposes `Variant`, but `Gateway`/`SourceConfig` always fetches `Variant::AfterPatches`; callers cannot select a variant or request ordered fallback.

This blocks prefix-dev/microrattler from implementing the setting faithfully.

### Requested API

Allow `SourceConfig` (including per-channel configuration) to specify an ordered non-empty list of repodata variants. For each subdir, try them in order and fall back only when a variant is unavailable. Preserve variant-specific cache keys. Existing behavior must remain `repodata.json` only.

The existing variants cover Conda's standard filenames: `Current`, `AfterPatches`, and `FromPackages`. Unsupported arbitrary filenames can be rejected by downstream validation.

### Acceptance criteria

- `Current, AfterPatches` uses `current_repodata.json` when present.
- A missing current file falls back to `repodata.json`.
- Errors other than absence are not silently hidden by fallback.
- Cache entries for variants do not collide.
- Default gateway behavior is unchanged.

wolfv:
## Description

Closes #2742.

Add an optional ordered `repodata_variants` list to `SourceConfig`. Local, remote Tokio, and remote WASM subdirs try variants in order and fall back only on `NotFound`; transport, parsing, and other errors are returned immediately. Empty and absent lists preserve the existing `repodata.json` behavior.

The low-level fetch layer already uses variant-specific filenames for cache keys, so fallback variants remain isolated in the cache. Sharded repodata remains the first choice when enabled; downstream callers that explicitly select nonstandard repodata filenames can disable shards in the same `SourceConfig`.

This unblocks Conda-compatible `repodata_fns` support using the existing `Current`, `AfterPatches`, and `FromPackages` variants.

## Tests

- A local Gateway configured with `Current, AfterPatches` reads `current_repodata.json` first.
- Removing it and supplying only `repodata.json` exercises ordered fallback.
- `cargo test -p rattler_repodata_gateway --features gateway configured_repodata_variants --lib`
- `cargo clippy -p rattler_repodata_gateway --all-targets --features gateway -- -D warnings`
- `cargo fmt --check`

## AI usage

This contribution was generated with AI assistance and manually reviewed and tested by the contributor.

### DIFF
--- a/crates/rattler_repodata_gateway/src/gateway/channel_config.rs
+++ b/crates/rattler_repodata_gateway/src/gateway/channel_config.rs
@@ -3,7 +3,7 @@ use std::collections::HashMap;
 use rattler_conda_types::ChannelUrl;
 use url::Url;
 
-use crate::fetch::CacheAction;
+use crate::fetch::{CacheAction, Variant};
 
 /// Describes additional properties that influence how the gateway fetches
 /// repodata for a specific channel.
@@ -23,6 +23,10 @@ pub struct SourceConfig {
     /// Describes fetching repodata from a channel should interact with any
     /// caches.
     pub cache_action: CacheAction,
+
+    /// Ordered repodata variants to try. `None` preserves the existing
+    /// `repodata.json`-only behavior.
+    pub repodata_variants: Option<Vec<Variant>>,
 
     /// When the gateway may only read from the cache, report a package whose
     /// shard is absent as having no records instead of failing the query.
@@ -47,6 +51,7 @@ impl Default for SourceConfig {
             bz2_enabled: true,
             sharded_enabled: true,
             cache_action: CacheAction::default(),
+            repodata_variants: None,
             missing_shards_are_empty: false,
         }
     }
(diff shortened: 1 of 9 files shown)


crates/rattler_repodata_gateway/src/gateway/channel_config.rs
```diff
@@ -24,6 +24,10 @@ pub struct SourceConfig {
     /// caches.
     pub cache_action: CacheAction,
 
+    /// Ordered repodata variants to try. `None` preserves the existing
+    /// `repodata.json`-only behavior.
+    pub repodata_variants: Option<Vec<Variant>>,
```
baszalmstra: Maybe an `IndexSet` is a better container here because we technically do not want duplicates.

The comment describes the "existing" repodata.json only behavior. That explains the the "previous" behavior which doesnt make sense if you just read the current state.

baszalmstra (changes requested):

PR #2748, opened 2026-09-02
perf: extract packages on a blocking worker

baszalmstra:
Async package extraction currently runs the tar/zip loop through Tokio filesystem APIs. Every file create, write and metadata update becomes a separate blocking-pool job, which is expensive for packages with many small files.

This PR pumps the async input into a bounded channel of 128 KiB chunks and runs the sync extractor on one `spawn_blocking` worker. The pump also computes the MD5, SHA-256 and size while it forwards chunks, so hashing overlaps with decompression and writes on the worker instead of running in front of them. Local files go directly to the worker. The sync tar loop also validates parent directories once, buffers writes up to 128 KiB and sets mtimes through the open file handle. Download errors and cancellation stop the worker before returning.

The sync `extract_*` functions keep hashing through `process_with_hashing`; the async path calls new `*_without_hashing` variants. Every extractor reads its input to the end, so the result describes the whole stream even when the archive ends before the input does.

## Benchmarks

Old async extractor → this branch, using loopback HTTP and an empty destination for every run. ReFS and NTFS on Windows, ext4 in WSL2 on the same machine:

| Workload | ReFS | NTFS | ext4 |
|---|---:|---:|---:|
| Python 3.13 | 372 → 188 ms | 1.105 → 0.697 s | 534 → 178 ms |
| NumPy | 262 → 106 ms | 747 → 462 ms | 384 → 127 ms |
| Rust, 183 MiB | 1.096 → 0.925 s | 1.630 → 1.551 s | 1.238 → 0.878 s |
| 5-package env, sequential | 716 → 322 ms | 2.412 → 1.488 s | 1.156 → 0.382 s |
| 5-package env, concurrency 8 | 421 → 200 ms | 1.564 → 1.079 s | 601 → 201 ms |

The Rust package is the one dominated by large files: 176 files, 901 MiB extracted, mostly EXE, DLL and PDB. Profiling puts roughly 700 ms in Zstd, 265 ms in MD5/SHA-256 and 200 ms in writes. With hashing still on the worker this package was a small regression on ReFS (1.08 → 1.17 s). Hashing on the pump hides that cost completely: a build with hashing removed altogether is no faster (895 vs 898 ms on ReFS, 1055 vs 1072 ms on NTFS). Larger channel chunks, Zstd buffers and copy buffers did not help. The NTFS Rust row varies by about ±400 ms between runs on my machine, so read it as roughly flat.

## Testing

The numbers above are hyperfine means with one warmup and ten runs.

The `rattler_package_streaming` and `rattler_cache` test suites pass on Windows and Linux. Integration tests compare sync and async output across the test corpus and cover escaping paths and links, old-format directories, lying size headers, truncated streams, reader failures, cancellation, chunk boundaries and the data-descriptor fallback. A new test feeds more trailing bytes than the channel holds after a complete `.conda` and checks that the digests and size cover all of them.

The tar loop replaces `Entry::unpack_in`, so the path sanitization and link handling deserve a close look.

### DIFF
--- a/crates/rattler-bin/src/commands/extract.rs
+++ b/crates/rattler-bin/src/commands/extract.rs
@@ -1,115 +1,194 @@
-use std::path::PathBuf;
+use std::{
+    path::PathBuf,
+    time::{Duration, Instant},
+};
 
+use futures_util::{StreamExt, stream};
 use miette::{Context, IntoDiagnostic};
+use rattler_conda_types::package::{CondaArchiveIdentifier, CondaArchiveType};
+use rattler_package_streaming::ExtractResult;
+use reqwest_middleware::ClientWithMiddleware;
 use url::Url;
 
-/// Extract a local or remote conda package.
+/// Extract one or more local or remote conda packages.
 #[derive(Debug, clap::Parser)]
 pub struct Opt {
-    /// Path or URL to the conda package archive (.tar.bz2 or .conda)
+    /// Paths or URLs to conda package archives (.tar.bz2 or .conda)
     #[clap(required = true)]
-    package: String,
+    packages: Vec<String>,
 
-    /// Destination directory where the package will be extracted
-    /// If not specified, extracts to a directory with the same name as the package
+    /// Destination directory. With a single package this is the directory the
+    /// package is extracted into. With multiple packages each package is
+    /// extracted into a subdirectory of this directory named after the package.
+    /// Defaults to the current directory.
     #[clap(short, long)]
     destination: Option<PathBuf>,
+
+    /// How local files are read. Remote packages always stream.
+    #[clap(long, value_enum, default_value_t = Mode::Sync)]
+    mode: Mode,
+
+    /// Number of packages to extract concurrently.
+    #[clap(long, default_value_t = 1)]
+    concurrency: usize,
 }
 
-/// Strips package extensions (.tar.bz2 or .conda) from a filename
-fn strip_package_extension(filename: &str) -> String {
-    if let Some(stripped) = filename.strip_suffix(".tar.bz2") {
-        stripped.to_string()
-    } else if let Some(stripped) = filename.strip_suffix(".conda") {
-        stripped.to_string()
-    } else {
-        filename.to_string()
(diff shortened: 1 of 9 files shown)


crates/rattler-bin/src/commands/extract.rs
```diff
@@ -27,89 +52,138 @@ fn strip_package_extension(filename: &str) -> String {
...
+    fn parse(package: &str) -> Self {
+        match Url::parse(package) {
+            Ok(url) if url.scheme().len() > 1 => Self::Url(url),
+            _ => Self::Path(PathBuf::from(package)),
+        }
+    }
 
-    Ok((destination, result))
+    /// The directory name to extract into when no explicit destination is given.
+    fn package_name(&self) -> miette::Result<String> {
```
baszalmstra, line 63: We should use ArchiveIdentifier here.

crates/rattler-bin/src/commands/extract.rs
```diff
@@ -27,89 +52,138 @@ fn strip_package_extension(filename: &str) -> String {
...
+async fn extract_one(
+    source: Source,
+    destination: PathBuf,
+    mode: Mode,
+    offline: bool,
+) -> miette::Result<(Source, PathBuf, ExtractResult, Duration)> {
+    let start = Instant::now();
+    let result = match (&source, mode) {
+        (Source::Url(url), _) => {
+            let client = super::client::create_client_with_middleware(offline)?;
```
baszalmstra: This creates a new client per call which is super expensive. We should have one shared client.

crates/rattler-bin/src/commands/extract.rs
baszalmstra: Done, uses CondaArchiveIdentifier now (and errors on names that don't parse).

crates/rattler-bin/src/commands/extract.rs
baszalmstra: Fixed, one client created in extract() and shared across packages.

pavelzw (approved):

PR #2759, opened 2026-09-04
fix(networking): require a path-segment boundary in MirrorMiddleware prefix match

### LINKED ISSUE #2757: MirrorMiddleware prefix match has no path-segment boundary — cross-channel misrouting
### Summary

`MirrorMiddleware`'s request-matching logic uses a raw string prefix check (`url_str.strip_prefix(key)`) with no path-segment boundary. If a mirror is registered for a channel URL that doesn't end in `/` (e.g. `https://host/conda-forge`), a request to a completely unrelated channel that merely shares that string as a prefix (e.g. `https://host/conda-forge2/...`) is silently matched and forwarded to the `conda-forge` mirror, with the extra characters after the shared prefix (`"2/..."`) appended onto the mirror's path — producing a corrupted URL on top of the misrouting.

`MirrorMiddleware::from_config` avoids triggering this in the common case by force-appending a trailing slash to every configured key (`crates/rattler_networking/src/mirror_middleware.rs` `with_trailing_slash`), but `MirrorMiddleware::from_map`, the underlying public constructor, has no such guard, so any other caller of `from_map` (or a future config path) hits this.

### Where

Matching: https://github.com/conda/rattler/blob/main/crates/rattler_networking/src/mirror_middleware.rs#L148-L153

```rust
let url_str = req.url().to_string();

for (key, url) in self.keys() {
    if let Some(url_rest) = url_str.strip_prefix(key) {
        let url_rest = url_rest.trim_start_matches('/');
        // replace the key with the mirror
        let mirrors = self.mirror_map.get(url).unwrap();
        ...
```

### Reproduction

Added a temporary test to `crates/rattler_networking/src/mirror_middleware.rs` registering a mirror for `"https://host/conda-forge"` (no trailing slash, as a direct `from_map` caller might do) with a local mock server, then requesting the *unrelated* channel `https://host/conda-forge2/count`:

```rust
#[tokio::test]
async fn test_mirror_middleware_cross_channel_prefix_bug() {
    async fn echo_path(req: axum::extract::Request) -> String {
        format!("HIT conda-forge mirror at path: {}", req.uri().path())
    }
    let router = Router::new().fallback(echo_path);
    let addr = SocketAddr::new([127, 0, 0, 1].into(), 0);
    let listener = tokio::net::TcpListener::bind(&addr).await.unwrap();
    let addr = listener.local_addr().unwrap();
    tokio::spawn(axum::serve(listener, router.into_make_service()).into_future());
    let mirror_url: Url = format!("http://{}:{}", addr.ip(), addr.port()).parse().unwrap();

    let mut mirror_map = std::collections::HashMap::new();
    mirror_map.insert(
        "https://host/conda-forge".parse().unwrap(),
        vec![mirror_setting(mirror_url)],
    );

    let middleware = MirrorMiddleware::from_map(mirror_map);
    let client = reqwest_middleware::ClientBuilder::new(reqwest::Client::new())
        .with(middleware)
        .build();

    // "conda-forge2" is an unrelated channel with no mirror registered for it.
    let res = client.get("https://host/conda-forge2/count").send().await.unwrap();
    println!("status: {}", res.status());
    let body = res.text().await.unwrap();
    println!("body: {body}");
    assert!(!body.starts_with("HIT conda-forge mirror"), "misrouted: {body}");
}
```

Actual `cargo test -p rattler_networking --features rattler_config -- --nocapture` output:

```
running 1 test
status: 200 OK
body: HIT conda-forge mirror at path: /2/count

thread 'mirror_middleware::test::test_mirror_middleware_cross_channel_prefix_bug' panicked at crates/rattler_networking/src/mirror_middleware.rs:447:9:
request for the unrelated 'conda-forge2' channel was routed to the 'conda-forge' mirror: HIT conda-forge mirror at path: /2/count
test mirror_middleware::test::test_mirror_middleware_cross_channel_prefix_bug ... FAILED
```

The request for `conda-forge2` (which has no mirror configured) was forwarded to the `conda-forge` mirror server, with the path corrupted to `/2/count` (the `"2"` left over from the unmatched prefix, concatenated onto the mirror's base path) instead of `/count`.

### Suggested fix

When matching `key` against `url_str`, require a path-segment boundary right after the matched prefix (i.e. the next character must be `/` or the prefix must already end in `/`), the same way `from_config`'s `with_trailing_slash` works around it — but enforced centrally in the matching logic (or in `from_map` itself) rather than only in one caller.

### Environment

Found while reading `crates/rattler_networking/src/mirror_middleware.rs` on `main` (commit around `8f438dbc`).

doraem-on:
### Description

`MirrorMiddleware` matches request URLs against configured mirror keys with a raw string prefix check (`url_str.strip_prefix(key)`). A key registered without a trailing slash (e.g. `"https://host/conda-forge"`, which any direct `from_map` caller could do) also matches an unrelated channel that merely shares the string as a prefix (e.g. `"https://host/conda-forge2/..."`), silently forwarding that request to the wrong mirror with a corrupted path.

`from_config` avoids the common case by force-appending a trailing slash to every configured key, but `from_map` — the underlying public constructor — has no such guard. This adds a path-segment boundary check directly in the matching loop, so it holds regardless of how the middleware was constructed.

Fixes #2757

### How Has This Been Tested?

Added `test_mirror_middleware_does_not_cross_channel_boundary` in `crates/rattler_networking/src/mirror_middleware.rs`. It registers a mirror for `"https://host/conda-forge"` (no trailing slash) and requests the unrelated `"...conda-forge2/count"`, asserting the request reaches a separate "real upstream" server unmodified rather than the `conda-forge` mirror.

Before the fix, this failed by actually reaching the mirror server with a corrupted path:
```
status: 200 OK
body: HIT conda-forge mirror at path: /2/count

thread '...' panicked: request for the unrelated 'conda-forge2' channel was routed to the 'conda-forge' mirror: HIT conda-forge mirror at path: /2/count
```

After the fix, full test suite for this module:
```
running 6 tests
test mirror_middleware::test::test_mirror_sort ... ok
test mirror_middleware::test::from_config_appends_trailing_slashes ... ok
test mirror_middleware::test::test_mirror_middleware_path_rewrite ... ok
test mirror_middleware::test::test_mirror_middleware_does_not_cross_channel_boundary ... ok
test mirror_middleware::test::test_mirror_middleware ... ok
test mirror_middleware::test::test_mirror_middleware_broken ... ok

test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 80 filtered out; finished in 0.75s
```

`test_mirror_middleware_path_rewrite` (existing test, same-boundary matching) still passes, confirming legitimate matches are unaffected.

Also ran `cargo fmt -p rattler_networking` and `cargo clippy -p rattler_networking --all-targets --all-features -- -D warnings` with no issues.

### AI Disclosure

Tools: Claude Code

### Checklist:

### DIFF
--- a/crates/rattler_networking/src/mirror_middleware.rs
+++ b/crates/rattler_networking/src/mirror_middleware.rs
@@ -149,6 +149,17 @@ impl Middleware for MirrorMiddleware {
 
         for (key, url) in self.keys() {
             if let Some(url_rest) = url_str.strip_prefix(key) {
+                // `key` is used as a prefix, but a plain string prefix match
+                // has no notion of path-segment boundaries: without this
+                // check, a key like "https://host/conda-forge" would also
+                // match "https://host/conda-forge2/...", silently routing
+                // an unrelated channel to this mirror. Require the matched
+                // prefix to already end in '/', or the remainder to be
+                // empty or start with '/'.
+                if !key.ends_with('/') && !url_rest.is_empty() && !url_rest.starts_with('/') {
+                    continue;
+                }
+
                 let url_rest = url_rest.trim_start_matches('/');
                 // replace the key with the mirror
                 let mirrors = self.mirror_map.get(url).unwrap();
@@ -238,6 +249,13 @@ mod test {
 
     async fn broken_return() -> StatusCode {
         StatusCode::INTERNAL_SERVER_ERROR
+    }
+
+    /// Echoes back whatever path it received, prefixed to identify which
+    /// server handled the request. Used to make misrouted requests
+    /// unmistakable in the cross-channel-boundary test below.
+    async fn echo_path_prefixed(prefix: &'static str, req: axum::extract::Request) -> String {
+        format!("HIT {prefix} at path: {}", req.uri().path())
     }
 
     async fn test_server(name: &str, broken: bool) -> Url {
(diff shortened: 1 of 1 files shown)


crates/rattler_networking/src/mirror_middleware.rs
```diff
@@ -149,6 +149,17 @@ impl Middleware for MirrorMiddleware {
 
         for (key, url) in self.keys() {
             if let Some(url_rest) = url_str.strip_prefix(key) {
+                // `key` is used as a prefix, but a plain string prefix match
+                // has no notion of path-segment boundaries: without this
+                // check, a key like "https://host/conda-forge" would also
+                // match "https://host/conda-forge2/...", silently routing
+                // an unrelated channel to this mirror. Require the matched
+                // prefix to already end in '/', or the remainder to be
+                // empty or start with '/'.
```
baszalmstra, line 158: Lets make this comment much much more consise. It doesnt need this much text.

baszalmstra (changes requested):

crates/rattler_networking/src/mirror_middleware.rs
doraem-on: Alright @baszalmstra would see to it , thanks for reviewing

PR #2767, opened 2026-09-05
feat(rattler-bin)!: add output formats to search and whoneeds

pavelzw:
### Description

Adds `--format urls` and `--format json` to both `rattler search` and `rattler whoneeds`. Human-readable output remains the default. URL output prints one package URL per line so results can be piped into tools such as `head` or `xargs`.

```sh
rattler search xtensor --format urls
rattler whoneeds numpy --format urls
rattler search numpy --format json
rattler whoneeds numpy --format json
```

This is a breaking CLI change: replace `--urls-only` with `--format urls` and `--json` with `--format json`. Both formats output all matching records and conflict with explicit `--limit` and `--all` options, plus `--limit-packages` for search.

For `whoneeds`, URL output deduplicates packages reported through multiple dependency kinds and sorts by package name, version (newest first), build number, and URL. It retains only the fields needed for sorting and output while consuming the dependency stream. Both commands handle broken pipes gracefully in URL mode. Help text includes the new syntax.

### How Has This Been Tested?

- `pixi run build -p rattler-bin`
- `pixi run cargo-fmt`
- `pixi run cargo-clippy`
- Ran both commands against a temporary local channel populated from `test-data/channels/conditional-repodata/noarch/repodata.json`, with an empty `linux-64` subdirectory. Verified exact URL output and JSON records for `search package` and `whoneeds linux-dependency`, and confirmed default human-readable output.
- Checked help and CLI validation: the removed flags are rejected, `--json` is absent from help, invalid formats are rejected, and format/limit conflicts are enforced.

### AI Disclosure

  - [ ] I take responsibility for any AI-generated content in my PR.

Tools: Claude, Codex.

User prompts for the format changes:

```plaintext
take a look at this pr.

> How about instead we have something like --format urls?

do this for rattler search and rattler whoneeds

remove the old --json flag.
```

### Checklist:

- [x] I have made corresponding changes to the documentation (CLI help)


crates/rattler-bin/src/commands/search.rs
```diff
@@ -48,13 +50,13 @@ pub struct Opt {
     #[clap(long, default_value = "true", action = clap::ArgAction::Set)]
     sharded: bool,
 
-    /// Output in JSON format
+    /// Output in JSON format (equivalent to --format json)
     #[clap(long, conflicts_with_all = ["limit", "limit_packages", "all"])]
     json: bool,
```
pavelzw: we could also remove or deprecate this. any opinions @baszalmstra?

baszalmstra: Just remove. We dont guarentee anything

PR #2775, opened 2026-09-08
refactor(rattler-bin): share gateway construction and table rendering

pavelzw:
### Description

Two consolidation items from the `rattler-bin` review (follow-up to #2766), branched off latest `main`:

**Shared gateway construction (`commands/gateway.rs`)** — the repodata `Gateway` setup was copy-pasted in `create`, `solve`, `exec`, `search`, and `whoneeds`, and the copies had drifted. All five now go through one `build_gateway(client, config, offline, sharded)` helper (plus a shared `load_config()`). That fixes the drift:

- `solve`, `exec`, `search`, and `whoneeds` now honor the configured download concurrency — only `create` picked that up in #2762.
- `search` and `whoneeds` never set a cache dir, so they fell back to the gateway's built-in default: the cache *root* (`~/.cache/rattler/cache` instead of the `repodata/` subdir every other command uses) and ignored `RATTLER_CACHE_DIR`. They now share the same repodata cache as the rest of the CLI and get the `ensure_cache_dir` treatment (CACHEDIR.TAG).

**Shared table rendering (`commands/table.rs`)** — `solve`, `list`, and `exec --list` each had their own aligned-column renderer. They now share one small `Table`/`Cell` helper (padding computed on plain text so ANSI styling doesn't skew widths, two-space padding, no trailing whitespace). Side effects, all improvements:

- `rattler list` computed widths in *bytes*, so non-ASCII fields misaligned the table — now fixed (widths are counted in characters); its rows also no longer carry trailing whitespace, and its header is bold on a terminal like `solve`'s.
- `exec --list` used a fixed 40-column name field, so longer names broke alignment — the version column now fits the actual content.

### How Has This Been Tested?

- Unit tests for the new table module (alignment, character-vs-byte widths, styled-cell padding, trailing whitespace, indent).
- `cargo test -p rattler-bin`: 33 tests pass, including the CLI output snapshot tests from #2766 (unchanged — `inspect`/`compare-packages` don't use the table).
- `cargo clippy -p rattler-bin --all-targets` and `cargo fmt --check` are clean.

### AI Disclosure

Tools: Claude Code

```plaintext
Give me a full review of rattler-bin. What would you do better?
create a second pr for 6 and 9. remember to take the latest `main` state
```

### Checklist:

https://claude.ai/code/session_013mKoD2V6gJNJfmaEHMif3o

---
_Generated by [Claude Code](https://claude.ai/code/session_013mKoD2V6gJNJfmaEHMif3o)_


crates/rattler-bin/src/commands/create.rs
```diff
@@ -87,22 +74,7 @@ pub async fn create(opt: Opt, offline: bool) -> miette::Result<()> {
...
-        .with_channel_config(rattler_repodata_gateway::ChannelConfig {
-            default: SourceConfig {
-                sharded_enabled: true,
-                cache_action: super::client::repodata_cache_action(offline),
-                ..SourceConfig::default()
-            },
-            per_channel: HashMap::new(),
-        })
-        .finish();
+    let gateway = super::gateway::build_gateway(download_client.clone(), &config, offline, true)?;
```
baszalmstra: Please just import build_gateway or at least the gateway module.

baszalmstra (changes requested):

crates/rattler-bin/src/commands/create.rs
pavelzw: [`6aeeaa0` (this PR)](https://github.com/conda/rattler/pull/2775/changes/6aeeaa0458d2b4226f41aebaf331bb8fa46ac4f9)

baszalmstra (approved):

PR #2781, opened 2026-09-09
feat(conda_lock): read and write CEP-37 lockfiles

baszalmstra:
### Description

We can read old `conda-lock.yml` files today, but only by degrading them: `parse_v3_or_lower` drops `category`, `optional` and everything in `metadata` except channels, and we can't write the format at all. So rattler can't produce a `conda-lock.yml`, and can't round-trip one.

This adds `rattler_conda_lock` for [CEP-37](https://github.com/conda/ceps/blob/main/cep-0037.md) and `rattler_lock::conda_lock` for conversion. The new crate only knows the file format; the lossy judgement calls live in the conversion layer. Reading is lenient about what conda-lock actually writes (omitted `version`, channels as bare strings, explicit `null`, `*` and bare version literals as constraints, a git revision in `hash.sha256`), writing is always canonical v1. We also accept SHA256-only conda packages, which CEP-37 allows but conda-lock's model rejects. `Document` keeps the source and byte spans, so even semantic errors found later during conversion point at the right line.

Conversion is offline and fallible both ways. `import` takes an explicit category set per environment (default `main` -> `default`), since categories are install groups, not environments. `export` handles one pixi `Environment` as non-optional `main` packages. Dropping metadata is fine (`requires_python`, index URLs, repodata extras), changing what installs is not, so source builds, local paths, markers/extras and mismatched noarch modes error out.

`metadata.content_hash` is the awkward part: conda-lock hashes the *input* spec plus virtual packages, which isn't recoverable from a solved pixi lockfile. On export we hash the resolved packages instead, unless the caller supplies the original hashes. Same field name, different number, but better than writing a fake value. The preimage is `json.dumps(packages, sort_keys=True)` via `serde-json-python-formatter`.

Two limitations to check: a pip package with per-platform requirements (`click` needing `colorama` only on win-64) can't be imported, because pixi stores one requirement set per artifact and I'd rather error than invent a marker. Poetry-style `>=4.6.0,<4.7.0 || >4.7.0` is rejected for the same reason.

The 9 fixtures in `test-data/cep37` are ~23.5k of the ~26k line diff; the code is ~2.7k. Happy to trim to 3-4 fixtures. Draft because I'd like agreement on the conversion boundaries first.

### How Has This Been Tested?

`cargo test -p rattler_conda_lock -p rattler_lock`, plus two checks against real conda-lock:

1. Installed conda-lock at `350c4a04` and compared its parse of each fixture with its parse of our canonical output. All 9 match on package identities, versions, URLs, hashes, dependencies, categories and metadata. It also accepts files we exported from pixi lockfiles.
2. Recomputed every generated `content_hash` in Python with `hashlib.sha256(json.dumps(packages, sort_keys=True).encode())`. All 32 platform hashes match, and `generated_hashes_match_python_json_dumps` pins the preimage so a formatter change can't silently move them.

conda-lock rejects our SHA256-only output, which is the documented divergence from CEP-37 rather than a bug here.

My adversarial review subagents both died on provider usage limits, so the review is my own. Worth a close look: span bookkeeping in `parse.rs` and the noarch/subdir reconstruction in `export.rs`.

### AI Disclosure

Tools: Claude Code

### Checklist:


crates/rattler_lock/Cargo.toml
```diff
@@ -14,8 +14,10 @@ readme.workspace = true
 [dependencies]
 ahash = { workspace = true }
 jiff = { workspace = true }
+hex = { workspace = true }
 indexmap = { workspace = true, features = ["serde"] }
 itertools = { workspace = true }
+rattler_conda_lock = { workspace = true }
```
pavelzw, line 20: should we make this an optional feature? i imagine i don't need this in a lot of contexts (for example pixi-pack) and it would otherwise pull in dependencies like `serde-json-python-formatter`

pavelzw (approved):
mainly looked over the API, didn't do a super thorough review. mostly lgtm

PR #2790, opened 2026-09-13
feat(rattler_conda_types): Add indexed_timestamp

pavelzw:
### Description

this intentionally only adds the indexed_timestamp to `rattler_conda_types` only. rattler_index and rattler_solve will come in a follow-up PR

part 1 of https://github.com/conda/rattler/issues/2791

### AI Disclosure

Tools: astra

### Checklist:


crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -581,6 +581,12 @@ pub struct PackageRecord {
     #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub flags: Vec<Flag>,
 
+    /// When this artifact first entered the channel index (CEP-0047).
+    /// Assigned by the channel server or indexer, in Unix milliseconds.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[serde_as(as = "Option<crate::utils::serde::StrictTimestampMs>")]
```
pavelzw, line 587: the CEP requires indexed_timestamp to be `ms` not `s`, so this gets a different non-guessing `{de,}`serialization behavior. we could also use the same behavior we have with `timestamp` which could simplify our code here a bit 🤔 any opinions on this?

crates/rattler_conda_types/src/repo_data/mod.rs
```diff
@@ -785,10 +791,10 @@ impl PackageRecord {
...
     /// Returns the timestamp used by indexing operations.
     ///
-    /// This currently returns the package build timestamp. A future index
-    /// timestamp can change this method without changing its callers.
+    /// Prefers the server-assigned index timestamp, falling back to the
+    /// builder-controlled build timestamp for older channel metadata.
     pub fn timestamp_for_indexing(&self) -> Option<TimestampMs> {
-        self.timestamp
+        self.indexed_timestamp.or(self.timestamp)
     }
```
pavelzw, line 798: is this actually used anywhere? i haven't seen this being used anywhere 🤔

i would prefer an explicit `TimestampPolicy` in exclude-newer over this: either `AllowMissing`, `RequireTimestamp` (indexed preferred over build) and `RequireIndexedTimestamp` (only allow `indexed_timestamp`)

pavelzw: since this pr will be a breaking change anyway, we could also just remove this function

baszalmstra (approved):

